PR #10, opened 2017-04-13
feat(store): Add 'createSelector' and 'createFeatureSelector' utils

MikeRyanDev:
Adds a Reselect-inspired `createSelector` and `createFeatureSelector`. Only implements a minimal set of Reselect's features.


fxck:
Can you explain why is this needed? Given how reselect is a tiny lib anyway

MikeRyanDev:
@fxck Reselect is still going to be a more powerful library. We wanted to build in selector creation to better support other ngrx libraries (without adding a peer dependency on Reselect) and to start thinking about APIs for supporting feature modules (like the included `createFeatureSelector`).

PR #16, opened 2017-04-25
feat(docs): Added migration and API docs for V4

brandonroberts:


MIGRATION.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+`app.module.ts`
+```ts
+import { StoreRouterConnectingModule } from '@ngrx/router-store';
+import { reducers } from './reducers';
+
+@NgModule({
+  imports: [
+    StoreModule.provideStore(reducers),
```
danielgek: this should be forRoot

docs/effects/api.md
```diff
@@ -50,42 +45,13 @@ dispatched by effect streams.
...
-          .map(res => ({ type: 'LOGOUT_SUCCESS', payload: res }))
-          .catch(err => Observable.of({ type: 'LOGOUT_FAILURE', payload: err }))
-      );
-  }
-}
+actions$.ofType('LOGIN', 'LOGOUT');
 ```
 
+
+### Terminating Effects
```
Matmo10: IMO, "Terminating Effects" sounds like a way to kill an effect from running completely rather than meaning it won't automatically dispatch to the store. Maybe my terminology is just incorrect though.

brandonroberts: Agreed. I didn't want to call them "dead end" effects either. Maybe "Non-dispatching Effects"?

Matmo10: That sounds good to me!

docs/store/api.md
```diff
@@ -0,0 +1,68 @@
...
+    StoreModule.forRoot(reducers, { reducerFactory: debugReducerFactory })
+  ]
+})
+export class AppModule {}
+```
+
+## Feature Module State Composition
+
+Store uses fractal state management, which provides state composition through feature modules,
+loaded eagerly or lazily. Feature states are added using the `StoreFeature.forFeature` method. This
```
drew-moore: should be StoreModule.forFeature

MIGRATION.md
brandonroberts: Fixed

docs/effects/api.md
brandonroberts: Got it. Thanks!

docs/store/api.md
brandonroberts: Fixed

PR #23, opened 2017-05-07
Effect classes only run once

elwynelwyn:
As discussed in https://github.com/ngrx/effects/pull/148 and https://github.com/ngrx/effects/issues/123.

Changed the default behaviour of effect classes to only run once.

This means doing `EffectsModule.run` multiple times for an Effect from different modules (e.g. via lazy loading) does not cause duplicate effect subscriptions.

I have removed the config option - this is now just always on. This means people will get No Provider errors for `SingletonEffectsService` if they forget to call `EffectsModule.forRoot()` in their root app module.

I also made a minor tweak to `./build/utilities.ts` to make the build run on Windows.

### DIFF
--- a/build/util.ts
+++ b/build/util.ts
@@ -85,7 +85,7 @@ export async function ignoreErrors<T>(promise: Promise<T>): Promise<T | null> {
 }
 
 export function fromNpm(command: string) {
-  return `./node_modules/.bin/${command}`;
+  return path.normalize(`./node_modules/.bin/${command}`);
 }
 
 export function getPackageFilePath(pkg: string, filename: string) {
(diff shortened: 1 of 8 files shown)


MikeRyanDev:
Thanks @elwynelwyn!

RicardoVaranda:
@MikeRyanDev Has this been pushed into npm by any chance, we came upon this use case today and are looking to implement this fix to our codebase.

brandonroberts:
@RicardoVaranda you can use the nightly builds to start using 4.0 right now: https://github.com/ngrx/platform/tree/master/docs/effects

PR #24, opened 2017-05-10
WIP: Effects v4

MikeRyanDev:
DO NOT MERGE YET!

This introduces @ngrx/effects v4. Highlights:
* Simplified API
* Better error reporting
* More control over the effect stream

## Simplified API
Previously you had to call `EffectsModule.run(source)` for every effect class. This has been changed to accept an array of sources. 

In your root module:
```ts
@NgModule({
  imports: [
    EffectsModule.forRoot([
      SourceA,
      SourceB,
      SourceC,
    ])
  ]
})
export class AppModule { }
```

In your feature modules:
```ts
@NgModule({
  imports: [
    EffectsModule.forFeature([
      FeatureSourceA,
      FeatureSourceB,
      FeatureSourceC,
    ])
  ]
})
export class SomeFeatureModule { }
```

## Better Error Reporting
@ngrx/effects now inspects effects streams for common gotchas:

1. Mistakenly dispatching values that do not implement the Action interface
2. Throwing errors

Either of those will generate console errors giving you enough information to determine which effect is being problematic:
Effect Error Reporting

## More Control Over the Effect Stream
Previously you could use the imperative APIs `mergeEffects` and `EffectsSubscription` to manually control when an effect group starts or stops. Now there is a new lifecycle hook called `OnRunEffects` that allows effect classes to have control over when effects should start or end:

```ts
export class LightManagerEffects implements OnRunEffects {
  @Effect() someEffect$: Observable<Action>;

  ngrxOnRunEffects(effects$: Observable<EffectOutput>): Observable<EffectOutput> {
    return this.actions$.ofType(DevicesPage.ENTER)
      .exhaustMap(() =>
        effects$.takeUntil(this.actions$.ofType(DevicesPage.LEAVE))
      );
  }
}
```


modules/effects/src/actions.ts
```diff
@@ -6,31 +6,25 @@ import { filter } from 'rxjs/operator/filter';
 
 
 @Injectable()
-export class Actions extends Observable<Action> {
-  constructor(@Inject(ScannedActionsSubject) actionsSubject: Observable<Action>) {
+export class Actions<V = Action> extends Observable<V> {
```
fxck: 👍

artaommahe:
> EffectsModule.forFeature()

why not widely used `.forChild()`?

russelltrafford:
With this change, how can one ensure this is trigger later in the APP_INITIALISER chain? This is a major issue for loading app config via APP_INITIALISER, as currently we're finding the Effects are initialised first, before the config, which fails if the effects require config variables on app start.

Appreciate if this is not the place to question in detail, but all here are closest to this code change, so could perhaps offer best insight, and then it can be discussed in the appropriate place.

PR #45, opened 2017-06-22
Modified ofType to return Actions with the associated generic.

vigneshnrfs:
This change allows for better type inference when using @effetcs.


MikeRyanDev:
Thanks! Sorry for the delay in review.

PR #66, opened 2017-07-10
Parallelize cp.exec and add mangle flag to uglify

bfricka:
This MR cuts the build time in half and adds the "mangle" flag to the uglify step, which reduces bundle size (even after gzip).

### DIFF
--- a/build/tasks.ts
+++ b/build/tasks.ts
@@ -20,17 +20,22 @@ export async function removeDistFolder(config: Config) {
  * AOT and Closure compatible JavaScript
  */
 export async function compilePackagesWithNgc(config: Config) {
-  for (let pkg of config.packages) {
-    await util.exec('ngc', [
-      `-p ./modules/${pkg}/tsconfig-build.json`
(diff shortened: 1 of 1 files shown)


bfricka:
Here are the relevant changes.

### Build Times

Each of the build time numbers is the mean average of 3 runs.

**Current (master)**
Build time: 18.873s
CPU usage: 124.3%

**With optimizations**
Build time: 9.66s
CPU usage: 294%

### UglifyJS + gz

**Current (master) -> Optimized w/ mangle**
effects.umd.min.js: 2.62 kB -> 2.02 kB (-23%)
router-store.umd.min.js: 1.1 kB -> 994 B (-10%)
store.umd.min.js: 3.46 kB -> 2.69 kB (-22%)
store-devtools.umd.min.js: 4.32 kB -> 3.71 kB (-14%)

Not huge differences, but it's pretty consistent, and it's basically free.

MikeRyanDev:
Thanks @bfricka!

PR #67, opened 2017-07-10
feat(actions): group of type actions

kurtaqui:
This feature is to be able to group all actions defined in group of type, if all actions are dispatched the effect is executed.

### DIFF
--- a/docs/effects/README.md
+++ b/docs/effects/README.md
@@ -33,7 +33,7 @@ effect to the store.
 
 The `Actions` observable represents an observable of all actions dispatched to the
 store. The `ofType` operator lets you filter for actions of a certain type in which you
-want to use to perform a side effect.
+want to use to perform a side effect. The `whenAll` operator lets you observe on all the actions defined and only when all of the actions have been observed it performs the side effect.
 
 ## Example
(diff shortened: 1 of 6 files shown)


Matmo10:
As a user, "Filter actions by action types and only execute when all actions have been dispatched." does not sufficiently explain what this does for me IMO. 

I think you're trying to say "once all of these action types have been observed, THEN continue down the steam", is that right?  If that's true, that also raises a lot of questions.

If you're waiting on A, B, and C, and you observe A, B, A, B, C. Will it run once or twice?

kurtaqui:
Hi, yes sorry for the brief description,
If you're waiting on A, B, and C, and you observe A, B, A, B, C. it will run only once since C only got dispatched once.
eg: Waiting on A , B and C
dispatched  A, B, A, B, C, A => execute once
dispatched  A, B, A, B, C, A, B, C => execute twice
dispatched A, A, C => do not execute
All these scenarios are covered in the tests

brandonroberts:
Thanks for the PR! We appreciate the work you've put in, but we'd like to keep the API surface small and this functionality is something you can do already with Rx operators.

PR #69, opened 2017-07-11
feat(all): added dispatch thunk

kurtaqui:
Added `store.getSync` this enables the user to get synchronous data (get the current value of the selector) once.
Added thunk actions to `store.dispatch`, this will enable the user to dispatch async actions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,4 @@ output
 *.ngsummary.json
 *.ngfactory.ts
 tmp
+package-lock.json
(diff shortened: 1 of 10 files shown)


brandonroberts:
Getting store synchronously was supported in a previous version of store and is considered an anti-pattern which is why it doesn't exist in the current version. If you want to propose new features or additions, open an issue with `RFC:` in the title so we can discuss it first.

Issue #63, opened 2017-07-06
Issue with Type in Aot Build 


elvirdolic:
Hello,

Versions:
typescript: 2.4.1
ngrx/store: latest nightly build
angular: 4.2.5

On AOT Build im receiving type errors

```
Type 'Store<
Customer[]>' is not assignable to type 'Observable<Customer[]>'.
  Types of property 'subscribe' are incompatible.
    Type '{ (): Subscription; (observer: PartialObserver<Readonly<Customer[]>>): Subscription; (next?: (val...' is not assignable to type '{ (): Subscription; (observer: Pa
rtialObserver<Customer[]>): Subscription; (next?: (value: Custom...'.
      Types of parameters 'next' and 'next' are incompatible.
        Types of parameters 'value' and 'value' are incompatible.
          Type 'Readonly<Customer[]>' is not assignable to type 'Customer[]'.
```

in customers-list.component
```
customers$: Observable<Customer[]>; 
this.customers$ = this.store.select(customers);

```
customers selector

```
export const customers = createSelector(customerState, (state: state) => {
    return state.customers.ids.map(id => state.customers.entities[id]);
});

```

agugan:
#46  Could be due to part explained here.

jinder:
@elvirdolic @agugan Yes, this is a dupe of #46.

MikeRyanDev:
Closes in #72

PR #72, opened 2017-07-12
fix(Store): Removed readonly from type

### LINKED ISSUE #46: Extending T to Readonly<T> on Store is unsuitable for arrays
The ```Store<T>``` definition has been changed to extend ```Observable<Readonly<T>>```.  The problem with this is that this does not cater properly for all read only objects - specifically read only index accessors and read only arrays.

If you select an object that implements ```ReadonlyArray<T>``` and try to use that as an array, you'll receive an error because ```[Symbol.iterator]``` is not included in any mapped type enumerations (see https://github.com/Microsoft/TypeScript/issues/12377).  ```Readonly<[]>``` does not prevent you from adding or removing items in the array, and that's why using ```ReadonlyArray``` is necessary.

I think ```Readonly<T>``` should be changed to ```T``` and the developer should be responsible for applying it to the definition of their state object.

brandonroberts:
Closes #46 #63

### DIFF
--- a/modules/store/src/store.ts
+++ b/modules/store/src/store.ts
@@ -13,7 +13,7 @@ import { isSelector, createSelector } from './selector';
 
 
 @Injectable()
-export class Store<T> extends Observable<Readonly<T>> implements Observer<Action> {
+export class Store<T> extends Observable<T> implements Observer<Action> {
   constructor(
     state$: StateObservable,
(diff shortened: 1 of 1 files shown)


gentlejoseph:
Hi am using version 4. The fix not seems to be working

PR #75, opened 2017-07-12
fix(tslint): fix ts-lint errors & add lint npm script

krjordan:


MikeRyanDev:
Thanks, @krjordan! Now that Prettier has robust TypeScript support can you implement an autoformat command and use Husky to run the autoformat command during commits? The out-of-the-box Prettier style is good though we would prefer single quotes and trailing commas.

PR #77, opened 2017-07-14
fix(Store): Remove parameter destructuring for strict mode (#33)

### LINKED ISSUE #33: Error when compiling with option "strict: true" (nightly build)
node_modules/@ngrx/store/src/state.d.ts (20,68): Type 'StateActionPair<T, V> | undefined' has no property 'state' and no string index signature.
When I add the option `"strictNullChecks": false`, the error goes away.

jinder:
Fixes #33


MikeRyanDev:
Thanks!

PR #78, opened 2017-07-14
chore(docs): Updated migration guide for testing effects

brandonroberts:
Added testing example to effects documentation and removed reducerFactory snippet from example app


brandonroberts:
I removed it from the example because we're looking at changing the API for registering a custom reducer factory

t-p-l-k:
@brandonroberts okay, got it.

PR #81, opened 2017-07-14
Fix devtools compiler warning.

t-p-l-k:
Fixes example-app compiler warning:
```
WARNING in ./modules/store-devtools/src/devtools.ts
"126:562-581 export 'StoreDevtoolsConfig' was not found in './config'
```


MikeRyanDev:
Thanks!

PR #84, opened 2017-07-15
build(platform): Improve consistency (yarn), and async

bfricka:
- Add CircleCI override to use latest yarn for deterministic builds
- Refactor some overly obtuse functions
- Use async throughout
- Add commitizen-friendly badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@ Reactive libraries for Angular
 
 ## Support
 
-[![Join the chat at https://gitter.im/ngrx/store](https://badges.gitter.im/ngrx/store.svg)](https://gitter.im/ngrx/store?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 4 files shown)


bfricka:
Couple things about this PR:

1. Using `yarn@0.28.1` because several earlier versions have a regression that causes issues symlinking to `./node_modules/.bin`. See [this issue](https://github.com/yarnpkg/yarn/issues/3801).

2. The biggest improvement is using `yarn` instead of `npm`. This improves perf a bit, but much more importantly, it gives build determinism. 

3. This doesn't improve build perf like the last one. The async stuff is just for consistency, and likely wouldn't have a demonstrable impact unless the numbers of blocking tasks got significantly higher. That said, it's probably a good idea for io stuff to be async anyways, if only because most of it already was anyways.

I was actually planning for this PR to be unit tests, but with the latest commits (lots of new deps), I ended up running into the aforementioned symlink bug with my version of yarn, and couldn't get the project to build. So I went down the rabbit hole, ended up tweaking a few things, and here we are.

bfricka:
One more thing: We should probably update [contributing](https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md), to refer to `yarn`. Until npm@5 becomes more mainstream, the only way for determinism is something like [`shrinkpack`](https://github.com/JamieMason/shrinkpack), but since this repo is already using yarn, it seems reasonable to consider standardizing on that.

MikeRyanDev:
Thank you! These build-related PRs are very welcome.

MikeRyanDev:
Regarding the contributing doc, I'll gladly accept a PR updating it.

PR #85, opened 2017-07-16
Test/add specs

bfricka:
A few things to note about this PR:

1. `initialState` function for features. This was only implemented for root previously, but I added it to features as well. Unfortunately, the implementation is a bit different b/c features aren't using a token to provide initial state.

2. Removed `OpaqueToken`. The deps for this project are clear, and we're already using `InjectionToken` elsewhere, so it doesn't make sense to continue using a deprecated symbol.

3. Stronger types on core actions `INIT`, `UPDATE`, etc. I don't think this actually affects anything currently, but it does make certain inspections nicer.

4. Watch mode using chokidar. This should make writing development a bit nicer (`yarn run watch:tests`). This also outputs an html coverage report if that's your cup.

Some of these tests are a bit obtuse. I'm still getting used to writing ng2+ tests, so feedback is welcome if you see something that could be improved. I'll write some more tests in the future, but I wanted to get this out there.

*ps: Crossing my fingers that the `xit` test passes. Last time it failed in CI, and I have zero clue why. If it fails I'll xit back again*

### DIFF
--- a/modules/store/spec/store.spec.ts
+++ b/modules/store/spec/store.spec.ts
@@ -1,28 +1,22 @@
 import 'rxjs/add/operator/take';
-import { Observable } from 'rxjs/Observable';
 import { ReflectiveInjector } from '@angular/core';
 import { hot } from 'jasmine-marbles';
 import { createInjector } from './helpers/injector';
-import { Store, Action, combineReducers, StoreModule } from '../';
-import { ActionsSubject } from '../src/private_export';
+import { ActionsSubject, ReducerManager, Store, StoreModule } from '../';
 import {
   counterReducer,
   INCREMENT,
   DECREMENT,
   RESET,
 } from './fixtures/counter';
+import Spy = jasmine.Spy;
+import any = jasmine.any;
 
 interface TestAppSchema {
   counter1: number;
   counter2: number;
   counter3: number;
-}
-
-interface Todo {}
-
-interface TodoAppSchema {
-  visibilityFilter: string;
-  todos: Todo[];
+  counter4?: number;
 }
 
 describe('ngRx Store', () => {
@@ -68,7 +62,7 @@ describe('ngRx Store', () => {
     });
   });
 
-  describe('basic store actions', function() {
+  describe('basic store actions', () => {
     beforeEach(() => setup());
 
     it('should provide an Observable Store', () => {
@@ -84,7 +78,7 @@ describe('ngRx Store', () => {
       e: { type: INCREMENT },
     };
 
-    it('should let you select state with a key name', function() {
+    it('should let you select state with a key name', () => {
       const counterSteps = hot(actionSequence, actionValues);
 
       counterSteps.subscribe(action => store.dispatch(action));
@@ -99,7 +93,7 @@ describe('ngRx Store', () => {
       );
     });
 
-    it('should let you select state with a selector function', function() {
+    it('should let you select state with a selector function', () => {
       const counterSteps = hot(actionSequence, actionValues);
 
       counterSteps.subscribe(action => store.dispatch(action));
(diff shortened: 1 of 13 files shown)


bfricka:
Hey there! First off, congrats on 4.0.0!

Second, I resolved conflicts!

modules/store/spec/store.spec.ts
```diff
@@ -114,13 +108,13 @@ describe('ngRx Store', () => {
       );
     });
 
-    it('should correctly lift itself', function() {
+    it('should correctly lift itself', () => {
       const result = store.select('counter1');
 
-      expect(result instanceof Store).toBe(true);
+      expect(result).toEqual(any(Store));
```
brandonroberts: Using `instanceof` seems like a better check than `any(Store)`

modules/store/spec/store.spec.ts
```diff
@@ -183,7 +177,7 @@ describe('ngRx Store', () => {
     });
 
     // TODO: Investigate why this is no longer working
-    xit('should complete if the dispatcher is destroyed', () => {
```
brandonroberts: Let this test remain excluded

modules/store/spec/utils.spec.ts
```diff
@@ -0,0 +1,81 @@
+/**
```
brandonroberts: Remove header text

modules/store/src/store_module.ts
```diff
@@ -51,7 +51,13 @@ export class StoreFeatureModule implements OnDestroy {
     @Inject(STORE_FEATURES) private features: StoreFeature<any, any>[],
     private reducerManager: ReducerManager
   ) {
-    features.forEach(feature => reducerManager.addFeature(feature));
+    for (let feature of features) {
```
brandonroberts: The functional approach is preferred over the for loop.

modules/store/src/utils.ts
```diff
@@ -45,9 +45,15 @@ export function omit<T extends { [key: string]: any }>(
   object: T,
   keyToRemove: keyof T
 ): Partial<T> {
-  return Object.keys(object)
-    .filter(key => key !== keyToRemove)
-    .reduce((result, key) => Object.assign(result, { [key]: object[key] }), {});
+  const ret: Partial<T> = {};
```
brandonroberts: No need to refactor this.

modules/store/src/state.ts
```diff
@@ -4,20 +4,20 @@ import { Observable } from 'rxjs/Observable';
...
 import { observeOn } from 'rxjs/operator/observeOn';
-import { startWith } from 'rxjs/operator/startWith';
 import { withLatestFrom } from 'rxjs/operator/withLatestFrom';
 import { scan } from 'rxjs/operator/scan';
 import { ActionsSubject } from './actions_subject';
 import { Action, ActionReducer } from './models';
 import { INITIAL_STATE } from './tokens';
 import { ReducerObservable } from './reducer_manager';
 import { ScannedActionsSubject } from './scanned_actions_subject';
+import { INIT } from './';
```
brandonroberts: Add `INIT` to existing imports from `./actions_subject`

modules/store/spec/utils.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+import { omit } from '../src/utils';
+import { combineReducers, compose } from '@ngrx/store';
+
+describe(`Store utils`, () => {
+  describe(`combineReducers()`, () => {
+    const s1 = { x: '' };
+    const s2 = { y: '' };
+    const r1 = (s = s1, a: any): typeof s1 =>
```
brandonroberts: Just a preference, but make the variable names more descriptive than `s` and `a` so they are more readable at glance

modules/store/src/state.ts
bfricka: I'm sorry I don't know what you mean.

brandonroberts: On line 9, add INIT to the existing object

```ts
import { ActionsSubject, INIT } from './actions_subject';
```

bfricka:
Not sure why Circle failed. Everything runs fine here.

modules/store/src/state.ts
bfricka: Haha, got it. It took me an embarrassingly long time to figure out what you were referring to.

modules/store/src/store_module.ts
bfricka: You got it!

brandonroberts:
LGTM

PR #87, opened 2017-07-18
feat(Store): Added simplified API for adding meta-reducers

brandonroberts:


modules/store/src/utils.ts
```diff
@@ -83,3 +83,14 @@ export function compose(...functions: any[]) {
     return rest.reduceRight((composed, fn) => fn(composed), last(arg));
   };
 }
+
+export function createReducerFactory(
+  reducerFactory: ActionReducerFactory<any, any>,
+  metaReducers?: ActionReducer<any, any>[]
+): ActionReducerFactory<any, any> {
+  if (Array.isArray(metaReducers) && metaReducers.length > 0) {
+    return compose.apply(null, [...metaReducers, reducerFactory]);
```
bfricka: No need to `apply`:

```ts
return compose(...metaReducers, reducerFactory)
```

bfricka:
I like the idea here. When writing tests, it took me a minute to realize `reducerFactory` would likely be used for this purpose.

modules/store/src/utils.ts
brandonroberts: TypeScript doesn't agree with signature, which is why I used `apply`:

```
ERROR in ngrx/platform/modules/store/src/utils.ts (92,12): Supplied parameters do not match any signature of call target.
```

PR #98, opened 2017-07-18
fix(effects): allow downleveled annotations

### LINKED ISSUE #93: Implement @Annotation style decorator down-leveling for Closure Compatibility
Angular's metadata decorators (`@Component`etc) aren't really runtime decorators, they're used more like "annotations". Angular tooling has the ability to "downlevel" these decorannoations into static class properties, which are then read by Angular's reflector. 

This is accomplished w/ a JSDoc annotation - see https://github.com/angular/angular/blob/a80ac0a8d3238f483a05252c437d4b321c42f5c7/packages/core/src/metadata/directives.ts#L67-L68

When run through `ngc` with the `annotateForClosureCompiler` option, usages of decorators tagged with this JSDoc are rewritten:
```ts
/**
* @Annotation
**/
export function Effect(args?){ return function(...args?){} }
```
used like
```ts
class Foo {
  @Effect() someProp:whatever
}
```
downlevels to
```ts
class Foo {}
Foo.propDecorators = {
  foo: [{type: Effect, args: [] }]
}
```

For full closure compatibility, we need to implement the same functionality - at runtime, we just need to check those static properties before checking via `Reflect`, as seen in : https://github.com/angular/angular/blob/master/packages/core/src/reflection/reflection_capabilities.ts#L73-L85

internal bug : b/63801869

robwormald:
tldr: applications using Google Closure Compiler with Angular's `ngc` can opt into downleveling specifically marked decorators into static properties. This change enables that usage.

closes #93

### DIFF
--- a/modules/effects/spec/effects_metadata.spec.ts
+++ b/modules/effects/spec/effects_metadata.spec.ts
@@ -23,6 +23,24 @@ describe('Effect Metadata', () => {
       ]);
     });
 
+    it('should get the effects metadata for a downleveled class instance', () => {
+      class Fixture {
+        static get propDecorators() {
+          return {
+            a: [{ type: Effect, args: [{ dispatch: false }] }],
+            b: [{ type: Effect, args: [] }],
+          };
+        }
+      }
+
+      const mock = new Fixture();
+
+      expect(getSourceMetadata(mock)).toEqual([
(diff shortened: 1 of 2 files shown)


MikeRyanDev:
LGTM

PR #108, opened 2017-07-19
Added selector function explanation

btroncone:
Hello,

I thought this document did a great job of explaining *how* to use the selector helpers but there could be some confusion with new users regarding *why* they should use them. I added a brief explanation to try to clear this up. Awesome work as always!


brandonroberts:
Great work as always!

PR #113, opened 2017-07-19
fix(example): remove unused imports

bcabanes:
Hi!

I was checking the example part of the repository and saw that some imports aren't used.
This PR just clean this to avoid confusion.


example-app/app/books/effects/book.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { TestBed, fakeAsync, tick } from '@angular/core/testing';
+import { TestBed, fakeAsync } from '@angular/core/testing';
```
brandonroberts: Would you remove the usage of `fakeAsync` in this file? Its not needed.

bcabanes: Done! 👍

brandonroberts:
Thanks!

Issue #103, opened 2017-07-19
@ngrx/store/init firing before effects are listening


PhillipZada:
Since the upgrade, INIT actions are no longer being pickup up via any effects.

I've also confirm that I'mm using the EffectsModule.forRoot([AppEffects])

Test Setup
![image](https://user-images.githubusercontent.com/12227583/28350066-d0d62dd8-6c89-11e7-8d4c-def7bc295b47.png)

MikeRyanDev:
Ah, sorry! This is intentional but we forgot to mention it in the migration guide. You can refactor that action to look like this for the same behavior:

```ts
import { defer } from 'rxjs/observable/defer';

@Effect()
init$: Observable<Action> = defer(() => {
    console.log('ok');
    const username = '';
    const password = '';

    return of(new auth.LoginAction({ username, password });
  });
```

Will gladly accept a PR that adds a more thorough example to the migration guide!

zygimantas:
Strange, if I use

```@Effect() public readonly initEffect = Observable.defer(() => Observable.of(new Action1()));```

then neither Action1, nor any other effect is executed. Probably I haven't migrated properly.

**Comment by gmickel** [comment]
I have the same issue as @zygimantas 

My project has a Init effect that uses defer(). After upgrading to ngrx v4 all actions after the Init effect are no longer run/working.

**Comment by zygimantas** [comment]
@brandonroberts @MikeRyanDev is that a bug and we should reopen this issue or "defer" is not the best way to do it?

**Comment by maxisam** [comment]
If I use defer and return another Observable<Action> directly, it doesn't work.

```
    @Effect() init$ = Observable.defer(() => {
        return Observable.of(this._RootActions.init());
    });

    @Effect() applicationInit$ = this._Actions$
        .ofType(RootActions.INIT)
        .switchMap((action: any) => this.initial());

```
So I double check how defer really works.  http://reactivex.io/documentation/operators/defer.html

> The Defer operator waits until an observer subscribes to it, and then it generates an Observable

I guess defer generate the Observable before next effect observer.

So if I change the sequence, everything works. (Just move `init$` to the bottom of the file.)

zygimantas:
@maxisam I can confirm, that after moving the ```init$``` with ```defer``` to the bottom, ```applicationInit$```, BUT other actions, dispatched from component, using ```this.store.dispatch(new Action3());``` are not handled by effects.

maxisam:
@zygimantas  I don't have the same problem you have.

[...]

PR #126, opened 2017-07-20
Updated the migration guide address @ngrx-store-init action transition

PhillipZada:
See https://github.com/ngrx/platform/issues/103


MikeRyanDev:
Thanks!

PR #137, opened 2017-07-20
make ofType method type safe

karptonite:
This pull request makes the actions `ofType()` method more type safe, allowing you to optionally specify the type of the action. This is possible because of the default generic parameters added to TypeScript in 2.3. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html

Something hacky like this:
```typescript
@Effect()
add$: Observable<Action> = this.actions$
	.ofType( subscriptions.ADD )
	.map( action => <AddAction>action )
	.let( this.addUserService.withLatestUserFlat )
```
becomes
```typescript
@Effect()
add$: Observable<Action> = this.actions$
	.ofType<AddAction>( subscriptions.ADD )
	.let( this.addUserService.withLatestUserFlat )
```

This may be something of a stopgap--hopefully, eventually typescript will be able to infer the return type from the filtering code itself, but who knows if that feature is ever coming.

Sorry, the test may be a bit more verbose than necessary--I had some trouble figuring out the best way to test it.

### DIFF
--- a/modules/effects/spec/actions.spec.ts
+++ b/modules/effects/spec/actions.spec.ts
@@ -72,4 +72,28 @@ describe('Actions', function() {
     actions.forEach(action => dispatcher.next({ type: action }));
     dispatcher.complete();
   });
+
+  it('should let you specify type with ofType', function() {
+    const ACTIONTYPE = 'ActionWithNumberType';
+    class ActionWithNumber implements Action {
+      readonly type = ACTIONTYPE;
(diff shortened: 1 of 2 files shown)


karptonite:
Looks like @maxime1992 beat me to this by 30 minutes with #135.

Great minds think alike.

MikeRyanDev:
Closed in #135

karptonite:
@MikeRyanDev does the test not add any value?

PR #148, opened 2017-07-21
fix(effects) : handle case if no args is defined in annotations

Salimlou:
Hi all,

this PR solves the problem of effects defined in package library, and compiled following the Angular Package Format , for instance: 

if I have this effect : 
```
@Effect()
  loadModels$: Observable<CMSActions> = this.actions$
    .ofType(CMS_ACTIONS.LOAD_MODELS)
    .map(toPayload)
    .map(models => new LoadModelsSuccessAction(models));

```

Compiled the propDecorators look like this : 

```
CMSEffects.propDecorators = {
    'loadModels$': [{ type: Effect },]
};
```

which causes the following error in this [line](https://github.com/ngrx/platform/blob/master/modules/effects/src/effects_metadata.ts#L31) 

```
ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'length' of undefined
```

Also it seems that the args attribute is marked as optional see this [line](https://github.com/angular/angular/blob/14fd78fd85b3dd230742eaaeab65f75226f37167/packages/core/src/reflection/reflection_capabilities.ts#L245)

Thanks to the ngrx team for their work :)!

### DIFF
--- a/modules/effects/spec/effects_metadata.spec.ts
+++ b/modules/effects/spec/effects_metadata.spec.ts
@@ -29,6 +29,7 @@ describe('Effect Metadata', () => {
           return {
             a: [{ type: Effect, args: [{ dispatch: false }] }],
             b: [{ type: Effect, args: [] }],
+            c: [{ type: Effect }],
           };
         }
       }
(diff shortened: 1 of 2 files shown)


MikeRyanDev:
Thanks!

lbertenasco:
I'm having this same issue.
Has this been released yet?
Is there a workaround?

Issue #147, opened 2017-07-21
reducers are not called when using --prod flag


morhi:
hello :)

I try to migrate my ngrx store to v4 and I have an issue when using the `--prod` flag with `ng serve` or `ng build`.

The store is splitted up in directories of modules and I bundled the store imports into a separate `NgModule`. The `StoreModule` is then imported in the  `AppModule`.

**`./store/index.ts`**
```ts
import { StoreModule as NgrxStoreModule, ActionReducerMap } from '@ngrx/store';
import pet from './pet/reducer';

export const reducers: ActionReducerMap<RootState, actions> = {
  pet: pet
};

@NgModule({
  imports: [
    NgrxStoreModule.forRoot(reducers, RootInitialState),
    environment.devTools ? StoreDevtoolsModule.instrument({maxAge: 50}) : [],
    EffectsModule.forRoot([
      PetEffects
    ])
  ]
})
export class StoreModule {}
```

**`./store/pet/reducer.ts`**
```ts
import { PetState } from './state';
import * as actions from './actions';

export default function reducer(state: PetState, action: actions.Actions): PetState {
  switch (action.type) {
    // ...
  }
}
```

However, the problem is that in production mode the reducers are not called. I can see the actions dispatched but the state is not changed. In development mode everything works fine. The console does not throw any errors or warnings. I hope that one of you has an idea how to fix it :)

I am using `node v8.1.4` and `@ngrx/store v4.0.0`

victornoel:
I'm facing the same problem, actually some reducers are called at the beginning, then after not.
Surprisingly, I use almost the same configuration than you.

```ts
export const reducers: ActionReducerMap<IStore> = {
  ui: UiReducer.reducer,
  users: UsersReducer.reducer,
  workspaces: WorkspacesReducer.reducer,
  buses: BusesReducer.reducer,
  busesInProgress: BusesInProgressReducer.reducer,
  containers: ContainersReducer.reducer,
  components: ComponentsReducer.reducer,
  serviceUnits: ServiceUnitsReducer.reducer,
  serviceAssemblies: ServiceAssembliesReducer.reducer,
  sharedLibraries: SharedLibrariesReducer.reducer,
};

// if environment is != from production
// use storeFreeze to avoid state mutation
const metaReducersDev = [storeFreeze, enableBatching];

const metaReducersProd = [enableBatching];

export const metaReducers = environment.production
  ? metaReducersProd
  : metaReducersDev;
```

And then use them as follows:
```ts
[...]

Issue #141, opened 2017-07-20
StoreModule.forFeature should accept injection token


karptonite:
See issue #116. The problem with AOT and reducers that are created by factories also exists for reducers imported in feature modules. However, the forFeature method lacks the handling for injector tokens that are present in forRoot. It seems like a similar approach should be added to forFeature.

juliekoubova:
hello, is is there any chance this would get released soon? it's kind of blocking me 😢

karptonite:
I believe that this was added in 4.0.2: https://github.com/ngrx/platform/blob/master/modules/store/CHANGELOG.md#402-2017-08-02

juliekoubova:
ok i should be able just "as any" my way until this gets released 😄 
https://github.com/ngrx/platform/commit/38b2f955e980c8f717a184e860fdda2db159018a

PR #153, opened 2017-07-22
feat(Store): Added injection token option for feature modules

### LINKED ISSUE #116: AOT and reducer factories in v4
I'm experimenting with factories for ngrx.
Here is example app.
https://github.com/ValeryVS/angular-storage-experiment

with v2
https://github.com/ValeryVS/angular-storage-experiment/blob/16febfb4d141476dfefe97b5de4b8f90d318d68b/src/app/store/reducer.ts

with v4
https://github.com/ValeryVS/angular-storage-experiment/blob/46c20b987369b0017435f37fada4febdcf1e6937/src/app/store/reducer.ts

However, in v4 version I have AOT error.

`
ERROR in Error encountered resolving symbol values statically. Expression form not supported (position 16:7 in the original .ts file), resolving symbol createReducer in /Users/valery/Projects/storage-experiment/src/app/store/organization/reducers/index.ts, resolving symbol compileReducers in /Users/valery/Projects/storage-experiment/src/app/store/reducer.ts, resolving symbol AppModule in /Users/valery/Projects/storage-experiment/src/app/app.module.ts, resolving symbol AppModule in /Users/valery/Projects/storage-experiment/src/app/app.module.ts, resolving symbol AppModule in /Users/valery/Projects/storage-experiment/src/app/app.module.ts
`

Can you give an advice, what to change?

brandonroberts:
- Fixes AOT issues when providing reducers via an InjectionToken
- Adds a build step before starting the example app
- Adds the example-app to CI testing

Closes #116, #141, #147


victornoel:
@brandonroberts I tried to use this branch to validate it did fix #147 with my code (I build it and replaced the folders in my `node_modules/@ngrx` with those from `dist`), and when I run my application with `--prod`, I get the following error:
```
ERROR in /home/victor/code/petals-cockpit/frontend/src/$$_gendir/node_modules/@ngrx/store/store.ngfactory.ts (51,65): Argument of type '{}' is not assignable to parameter of type 'StoreFeature<any, any>[]'.
  Property 'includes' is missing in type '{}'.
```

I suspect there is something wrong with this PR or is it me doing something wrong?

victornoel:
using typescript 2.4, maybe that's why, I noticed you have both typescript 2.4 and 2.3 in your `node_modules` (2.4 at the root, 2.3 under angular-cli), and angular-cli will use 2.3 so the example app is most certainly built and executed with 2.3.

victornoel:
I tried to execute the example-app with ts 2.4 and it works, so I suppose it is just that now #147 breaks at build time instead of runtime... (and thus this PR does not solve #147 :)

victornoel:
even though it is strange because I don't use features at all in my application (see https://github.com/ngrx/platform/issues/147#issuecomment-316957763)

bfricka:
I have to admit, I tried implementing this the other day, and couldn't figure it out. Nice work.

PR #155, opened 2017-07-22
feat(Effects/Store) Typed version of toPayload

karptonite:
For discussion, this may require some refinement.

But with the newly type safe version of `ofType`, this opens up the option of a type safe version of `toPayload` as well. I wasn't able to retrofit `toPayload` without breaking previous behavior, unfortunately (maybe someone who knows typescript better than my can give it a go?), but I was able to implement a separate `toPayloadTyped`, which infers the type of the payload based on the Action.

Is this useful? Is there an even cleaner way to implement it?


karptonite:
Oops, sent in the pull request before i was ready.

PR #157, opened 2017-07-22
feat(Store/Effects) Add toPayloadTyped function

karptonite:
(Third time's a charm)
For discussion, this may require some refinement.

But with the newly type safe version of `ofType`, this opens up the option of a type safe version of `toPayload` as well. I wasn't able to retrofit `toPayload` without breaking previous behavior, unfortunately (maybe someone who knows typescript better than my can give it a go?), but I was able to implement a separate `toPayloadTyped`, which infers the type of the payload based on the Action.

Is this useful? Is there an even cleaner way to implement it?

Note that this doesn't require that a user declare that their actions implement `ActionWithPayload` (It is probably best to do so, but Typescript can infer that), but it DOES require that they declare the type of their action when they call `ofType`.

### DIFF
--- a/modules/store/src/models.ts
+++ b/modules/store/src/models.ts
@@ -2,6 +2,10 @@ export interface Action {
   type: string;
 }
 
+export interface ActionWithPayload<T> {
+  payload: T;
+}
+
 export type TypeId<T> = () => T;
 
 export type InitialState<T> = Partial<T> | TypeId<Partial<T>> | void;
--- a/modules/store/src/models.ts
+++ b/modules/store/src/models.ts
@@ -2,7 +2,7 @@ export interface Action {
   type: string;
 }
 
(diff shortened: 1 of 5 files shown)


modules/store/src/models.ts
```diff
@@ -2,6 +2,10 @@ export interface Action {
   type: string;
 }
 
+export interface ActionWithPayload<T> {
```
bfricka: This should be:

```ts
export interface ActionWithPayload<T> extends Action {
  payload: T;
}
```

karptonite: You are right! I had that at some point. I'll fix it.

karptonite:
Do not merge this! I think I may have figured out a way to do it without requiring a separate toPayload function.

PR #179, opened 2017-07-25
docs(MIGRATION.md); add store-devtools warning to the top

niklas-dahl:
please don't let this "warning" be the last thing in the migration document.


brandonroberts:
Thanks for the PR, but we don't need a different warning about the devtools and router-store. If you want to add the warning from the devtools section to the router-store section that would be better.

PR #182, opened 2017-07-25
docs(MIGRATION.md): add import order for effects, storeRouter and dev-tools

loiane:
Added message for EffectsModule, StoreRouterConnectingModule and StoreDevtoolsModule import order. Should be imported only after StoreModule, otherwise it does not work.


brandonroberts:
Thanks for the PR! We think this is a bug and we've got a fix in the works so that the import order isn't an issue. If you could open an issue for tracking that would help out alot.

loiane:
Done: #184 :)

brandonroberts:
Thanks!

Issue #104, opened 2017-07-19
Consider removing the RouterStateSnapshot from router-store state


shlomiassaf:
Hi,

Using v4 I saw the the router state that comes in `@ngrx/router-store` has a `state` property that holds a reference to `RouterStateSnapshot` from `@angular/router`.

I'm not sure why the state needs to hold this object but it causes some issues.

  - In #97 we can see a performance issue, this is probably due to the `RouterStateSnapshot` object, pretty sure about that... the redux dev tool try to render it and its a huge object when rendering deep.

  - The store can not be frozen.

The `RouterStateSnapshot` hold reference to a lot of objects... components, router tree's and what not. A lot of them change internally, i.e.  get's mutated which makes the store mutable.

This means using a library like `ngrx-store-freeze` in development mode is not possible since deep freezing the store (state) will result in exceptions all over since deep objects in `RouterStateSnapshot` are expected to be mutable.

In general the store should be light, holding POCOs only.

cc @vsavkin, thoughts?

PR #188, opened 2017-07-26
feat(RouterStore): Added serializer for router state snapshot

### LINKED ISSUE #97: Performance Issues with Dev Tools + Router-Store
I've found that my app becomes quite slow if `StoreRouterConnectingModule` and `StoreDevtoolsModule` are both imported.

```ts
StoreModule.forRoot(reducers),
...(!environment.production ? [StoreDevtoolsModule.instrument({ maxAge: 10 })] : []),
EffectsModule.forRoot([]),
StoreRouterConnectingModule,
```

#### Observations

- CPU usage by Chrome is at least 30% constantly. In the Chrome task manager I've seen the dev tools extension process use 100%+ CPU and 2+ GB of RAM.
- I started with `maxAge: 50`, changing it to as low as `maxAge: 2` didn't help.
- Performance seemed to be worse if I flipped the order of `EffectsModule` and `StoreRouterConnectingModule`. 
- I found that exporting the state from the dev tools yielded a 27 MB JSON file. If the `StoreRouterConnectingModule` was removed, it was 71 KB.
- Most of the 27 MB file has data like this:

```json
 "computedStates": [
    {
      "state": {
        "router": {
          "state": {
            "_root": {
              "value": {
                "url": [],
                "params": {},
                "queryParams": {},
                "data": {},
                "outlet": "primary",
                "_routeConfig": null,
                "_urlSegment": {
                  "segments": [],
                  "children": {
                    "primary": {
                      "segments": [
                        {
                          "path": "app",
                          "parameters": {}
                        },
                        {
                          "path": "users",
                          "parameters": {}
                        }
                      ],
                      "children": {},
                      "parent": {
                        "segments": {
                          "$jsan": "$.computedStates[0].state.router.state._root.value._urlSegment.segments"
                        },
                        "children": {
                          "$jsan": "$.computedStates[0].state.router.state._root.value._urlSegment.children"
                        },
                        "parent": null,
                        "_sourceSegment": {
                          "$jsan": "$.computedStates[0].state.router.state._root.value._urlSegment"
                        },
                        "_segmentIndexShift": 0
                      }
                    }
                  },
                  "parent": null
                },
                "_lastPathIndex": -1,
                "_resolve": {},
                "_routerState": {
                  "$jsan": "$.computedStates[0].state.router.state"
                }
              },
```

#### Profiling Results

Top of call tree:

screen shot 2017-07-18 at 13 16 38

Next highest in tree:

screen shot 2017-07-18 at 13 19 17

#### Versions

Angular: 4.2.6
Angular CLI: 1.2.1
NGRX: 4.0.0

Browser: Chrome 59.0.3071.115
OS: macOS 10.12.5
Node: 6.11.0
Yarn: 0.27.5

brandonroberts:
This adds a serializer that can be customized for returning the router state snapshot. By default, the entire RouterStateSnapshot is returned. Documentation has been updated with example usage.

```ts
import { StoreModule } from '@ngrx/store';
import {
  StoreRouterConnectingModule,
  routerReducer,
  RouterStateSerializer
} from '@ngrx/router-store';

export interface RouterStateUrl {
  url: string;
}

export class CustomSerializer implements RouterStateSerializer<RouterStateUrl> {
  serialize(routerState: RouterStateSnapshot): RouterStateUrl {
    const url = routerState ? routerState.url : '';

    // Only return an object including the URL
    // instead of the entire snapshot
    return { url };
  }
}

@NgModule({
  imports: [
    StoreModule.forRoot({ routerReducer: routerReducer }),
    RouterModule.forRoot([
      // routes
    ]),
    StoreRouterConnectingModule
  ],
  providers: [
    { provide: RouterStateSerializer, useClass: CustomSerializer }
  ]
})
export class AppModule { }
```

Closes #97, #104, #237

### DIFF
--- a/modules/router-store/src/serializer.ts
+++ b/modules/router-store/src/serializer.ts
@@ -1,0 +1,13 @@
+import { InjectionToken } from '@angular/core';
+import { RouterStateSnapshot } from '@angular/router';
+
+export abstract class RouterStateSerializer<T> {
+  abstract serialize(routerState: RouterStateSnapshot): T;
+}
+
+export class DefaultRouterStateSerializer
+  implements RouterStateSerializer<RouterStateSnapshot> {
+  serialize(routerState: RouterStateSnapshot) {
+    return routerState;
+  }
+}
--- a/docs/router-store/README.md
+++ b/docs/router-store/README.md
@@ -15,8 +15,8 @@ Install @ngrx/router-store from npm:
 During the navigation, before any guards or resolvers run, the router will dispatch a ROUTER_NAVIGATION action, which has the following signature:
 
 ```ts
-export type RouterNavigationPayload = {
-  routerState: RouterStateSnapshot,
+export type RouterNavigationPayload<T> = {
+  routerState: T,
   event: RoutesRecognized
 }
 ```
@@ -46,3 +46,6 @@ import { App } from './app.component';
 })
 export class AppModule { }
 ```
+
+## API Documentation
+- [Custom Router State Serializer](./api.md#custom-router-state-serializer)
--- a/docs/router-store/api.md
+++ b/docs/router-store/api.md
@@ -1,0 +1,46 @@
+# API
+
+## Custom Router State Serializer
+
+During each navigation cycle, a `RouterNavigationAction` is dispatched with a snapshot of the state in its payload, the `RouterStateSnapshot`. The `RouterStateSnapshot` is a large complex structure, containing many pieces of information about the current state and what's rendered by the router. This can cause performance
+issues when used with the Store Devtools. In most cases, you may only need a piece of information from the `RouterStateSnapshot`. In order to pair down the `RouterStateSnapshot` provided during navigation, you provide a custom serializer for the snapshot to only return what you need to be added to the payload and store.
+
+To use the time-traveling debugging in the Devtools, you must return an object containing the `url` when using the `routerReducer`.
+
+```ts
+import { StoreModule } from '@ngrx/store';
+import {
+  StoreRouterConnectingModule,
+  routerReducer,
+  RouterStateSerializer,
+  RouterStateSnapshotType
+} from '@ngrx/router-store';
+
+export interface RouterStateUrl {
+  url: string;
+}
+
+export class CustomSerializer implements RouterStateSerializer<RouterStateUrl> {
+  serialize(routerState: RouterStateSnapshot): RouterStateUrl {
+    const { url } = routerState;
+
+    // Only return an object including the URL
+    // instead of the entire snapshot
+    return { url };
+  }
+}
+
+@NgModule({
+  imports: [
+    StoreModule.forRoot({ routerReducer: routerReducer }),
+    RouterModule.forRoot([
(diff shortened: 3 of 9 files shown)


kylecordes:
From a quick read of the code, it looks like this keeps the default behavior of sticking the unserializable router state in the State, but makes it possible for an application to avoid that by providing something custom, for example by copying code from the example application.

My way-out-here-in-userland perspective is that it would be more helpful if the default serializer did the "Right Thing", so that most applications get a serializable State without effort. I dream of a future where Store runs in a "frozen, enforced serializability" mode by default, and requires a `StoreModule.forRoot(mapOfReducers, { footGun: true})` invocation to turn that enforcement off.

brandonroberts:
@kylecordes I agree that the default should be a slimmed down version of the router state snapshot. It would be a breaking change to the current functionality though, so we'll keep it as is, deprecate it and change the default in the next version. Me and @MikeRyanDev have talked about have a config option out of the box to freeze the state, but it was more of an opt-in approach. We'll revisit that soon for the next version, as we'll need a replacement for `ngrx-store-freeze` if its no longer maintained.

modules/router-store/src/serializer.ts
```diff
@@ -0,0 +1,15 @@
+import { RouterStateSnapshot } from '@angular/router';
+
+export type RouterStateSnapshotType =
+  | RouterStateSnapshot
+  | Partial<RouterStateSnapshot>;
+
+export abstract class RouterStateSerializer {
+  abstract serialize(routerState: RouterStateSnapshotType | null): RouterStateSnapshotType | null;
```
vsavkin: Maybe we guard against nulls outside? Cause I think it should never be called with `null`.

modules/router-store/src/serializer.ts
```diff
@@ -0,0 +1,15 @@
+import { RouterStateSnapshot } from '@angular/router';
+
+export type RouterStateSnapshotType =
+  | RouterStateSnapshot
+  | Partial<RouterStateSnapshot>;
```
vsavkin: basically the following will fail:

```
const r: RouterStateSnapshotType;
r.root.params // won't compile
```

vsavkin: Maybe we can parameterize `RouterNavigationAction` instead?

modules/router-store/src/serializer.ts
```diff
@@ -0,0 +1,15 @@
+import { RouterStateSnapshot } from '@angular/router';
+
+export type RouterStateSnapshotType =
```
brandonroberts: Good suggestion

brandonroberts:
@vsavkin this one is ready for another review

vsavkin:
LGTM

strizzaflex:
@brandonroberts Thanks for the workaround. 
1) When can we expect the next major version with the actual fix? (Ballpark)
2) Having store freeze built in would be a big thumbs up from me. Our team is new to redux and one of the biggest pitfalls up front was mutating state. 

Again thanks for the fix we are back up and running with the nightly build.

PR #204, opened 2017-07-28
Added toEqualObservable jasmine matcher for effect testing

PhillipZada:
Hey guys,

Our team was having trouble with deciphering the error messages provided by toBeObservable => jasmine.toBe so I created this matcher which can provide and easier to understand error message when the effect action doesn't match the observable.

### DIFF
--- a/modules/effects/testing/index.ts
+++ b/modules/effects/testing/index.ts
@@ -1,1 +1,2 @@
 export * from './src/testing';
+export * from './src/toEqualObservable';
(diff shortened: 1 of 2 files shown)


brandonroberts:
I think this PR would be better served going in the `jasmine-marbles` repo here: https://github.com/synapse-wireless-labs/jasmine-marbles

PhillipZada:
Thanks @brandonroberts will do

Issue #208, opened 2017-07-28
Subscriptions are not triggered for primitive selectors on subsequent dispatches with same payload


ashkandaie:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
</code></pre>

## What is the current behavior?

When subscribing to a primitive selector such as a boolean, the first dispatched message triggers the subscription but subsequent dispatched messages with the same payload do not trigger the subscription. If you change the selector to the parent selector which contains the primitive data everything works as expected.

## Expected behavior:

All dispatched messages should trigger subscriptions for selectors that return complex objects or primitive selectors. 

## Minimal reproduction of the problem with instructions:

layout state: 
```javascript

export interface LayoutState {
  showSidenav: boolean;
}

```

layout actions:

```javascript
import { Action } from '@ngrx/store';

export const OPEN_SIDENAV = '[Layout] Open Sidenav';
export const CLOSE_SIDENAV = '[Layout] Close Sidenav';

export class OpenSidenavAction implements Action {
  readonly type = OPEN_SIDENAV;
}

export function openSidenav() {
  return new OpenSidenavAction();
}

export class CloseSidenavAction implements Action {
  readonly type = CLOSE_SIDENAV;
}

export function closeSidenav() {
  return new CloseSidenavAction();
}

export type Actions = OpenSidenavAction | CloseSidenavAction;

```

layout reducer:
```javascript
import { createSelector, createFeatureSelector } from '@ngrx/store';
import { LayoutState } from './state';
import * as layout from './actions';

const initialState: LayoutState = {
  showSidenav: false,
};

export function layoutReducer(state = initialState, action: layout.Actions): LayoutState {
  switch (action.type) {
    case layout.CLOSE_SIDENAV:
      return {
        showSidenav: false,
      };

    case layout.OPEN_SIDENAV:
      return {
        showSidenav: true,
      };

    default:
      return state;
  }
}

export const selectLayout = createFeatureSelector<LayoutState>('layout');

export const selectShowSidenav = (state: LayoutState) => state.showSidenav;

export const getShowSidenav = createSelector(selectLayout, selectShowSidenav);
```

subscription:

```javascript
showSidenav$: Observable<boolean>;

this.showSidenav$ = <any>this._store.select(fromLayout.getShowSidenav);   
  this.showSidenav$.subscribe((value) => {
[...]

PR #211, opened 2017-07-29
fix: Store select function

ashkandaie:
Updated call to distinctUntilChanged RXJS operator to mark primitive javascript data types as unique values.

Here is a PR for #208

### DIFF
--- a/modules/store/src/store.ts
+++ b/modules/store/src/store.ts
@@ -78,7 +78,13 @@ export class Store<T> extends Observable<T> implements Observer<Action> {
       );
     }
 
-    return distinctUntilChanged.call(mapped$);
+    return distinctUntilChanged.call(mapped$, (x: any, y: any) => {
+      const isPrimitive = typeof x === 'boolean' || typeof y === 'boolean' ||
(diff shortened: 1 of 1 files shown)


karptonite:
Won't this also trigger subscribers when there has been no change at all, but the selector returns a primitive? That doesn't seem like the desired behavior to me.

One possible solution (but a breaking change) would be to remove that distinctUntilChanged entirely from the `select` entirely, and make calling `distinctUntilChanged` the responsibility of the user.  Another solution would be for you to implement your own `select` with your own desired behavior with respect to primitives; since store is just an observable, you can chain off of it directly or pass it to your own `select` function with `.let`.

wundo:
@brandonroberts looks like in the same day you assigned @MikeRyanDev to this and closed the issue, what is the status here?

fxck:
@karptonite select is just a map + distinctUntilChanged, so removing the latter wouldn't really make sense

karptonite:
@fxck select also handles passing a path rather than a map function, so even without distinctUntilChanged, it would still have use.

fxck:
Oh right, never been fan of that feature. Completely forgot it existed.

PR #225, opened 2017-08-01
chore(Platform): fix type def files search path

juristr:
Hi,

as I mentioned on Gitter I'm currently studying the build process of ngrx which you have set up for another project of mine. During that phase the build process broke a couple of times on the `copyTypeDefinitionFiles` step giving sometimes random errors on files. Some people (like @og-carefree) also mentioned it in the Gitter Channel: 

image

After applying this fix the error didn't happen to me any more, so it should hopefully be ok now :)


juristr:
Strange tho the builds didn't seem to fail before this patch...that said, also locally it was only occasionally, but like every 2nd build failed

MikeRyanDev:
Thanks!

Issue #226, opened 2017-08-01
createSelector() not memoizing correctly


03byron:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
</code></pre>

## What is the current behavior?
The projector fn passed to createSelector() is invoked even when its arguments remain unchanged.

## Expected behavior:
The projector fn should be called only when one of its arguments change.

## Minimal reproduction of the problem with instructions:
http://plnkr.co/edit/GEdUHwk3XEnHHwmPVMSr?p=preview

When you click increase/decrease buttons you can see in the console that the projector function executes but its arguments didnt change.

or in short summary:

```javascript
const mySelector = createSelector(
  (state) => state.substate,
  (substate) => {
    // heavy computing here, should be run only when substate changes,
    // but is running whenever mySelector is called.
    return substate.something;
  }
);
```

interestingly when wrapping the projector fn in createSelector as well, i get the desired behavior:

```javascript
const mySelector = createSelector(
  (state) => state.substate,
  createSelector( s => s, (substate) => {
    // heavy computing here, is called only when substate changes
    return substate.something;
  })
);
```

## Version of affected browser(s),operating system(s), npm, node and ngrx:
Chrome 59
OS X El Capitan
npm 3.10.8
node 6.9.1
ngrx 4.0.0

karptonite:
Can confirm that this is an issue.

I don't know that it is the entire issue, but there clearly seems to be an issue here:
https://github.com/ngrx/platform/blob/master/modules/store/src/selector.ts#L119

line 119, `selectors` should probably be `memoizedSelectors`. I'm seeing if I can write a failing test for your bug.

karptonite:
OK, I can confirm that this is an issue.
If we substitute `createSelector` from reselect, the bug goes away.
http://plnkr.co/edit/ygjOlw9I2WOqqMAo83Wf?p=preview

karptonite:
One other data point: I ran into a similar issue the other day that I didn't report because it was difficult to reproduce. Here is how it manifested: 
[...]

PR #228, opened 2017-08-02
fix(createselector): memoize projector function

03byron:
This is a fix for: https://github.com/ngrx/platform/issues/226

With this fix, the projector function is memoized as well.


MikeRyanDev:
Thanks!

Issue #241, opened 2017-08-06
Start of the e2e test gives an error because of wrong path to type script config file.


sharikovvladislav:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
</code></pre>

## What is the current behavior?

e2e is not working.
## Expected behavior:

e2e is working :)

## Minimal reproduction of the problem with instructions:

1. clone repo
2. yarn
3. run e2e tests

Now you see this:

```
TypeError: Cannot find a tsconfig.json file at the specified directory: e2e
    at resolveSync (/Users/svlad/dev/platform/node_modules/tsconfig/src/tsconfig.ts:69:11)
    at Object.loadSync (/Users/svlad/dev/platform/node_modules/tsconfig/src/tsconfig.ts:140:16)
    at readConfig (/Users/svlad/dev/platform/node_modules/ts-node/src/index.ts:425:18)
    at Object.register (/Users/svlad/dev/platform/node_modules/ts-node/src/index.ts:189:18)
    at beforeLaunch (/Users/svlad/dev/platform/protractor.conf.js:24:24)
    at q_1.Promise (/Users/svlad/dev/platform/node_modules/protractor/lib/util.ts:48:39)
    at Function.promise (/Users/svlad/dev/platform/node_modules/q/q.js:682:9)
    at Object.runFilenameOrFn_ (/Users/svlad/dev/platform/node_modules/protractor/lib/util.ts:39:10)
    at Object.initFn [as init] (/Users/svlad/dev/platform/node_modules/protractor/lib/launcher.ts:111:10)
    at Object.<anonymous> (/Users/svlad/dev/platform/node_modules/protractor/lib/cli.ts:235:10)
```

## Version of affected browser(s),operating system(s), npm, node and ngrx:
```
sharikovvlad:platform svlad$ yarn --version
0.24.6
sharikovvlad:platform svlad$ npm -v
5.0.3
sharikovvlad:platform svlad$ sw_vers -productVersion 
10.12.5
sharikovvlad:platform svlad$ 
```

## Other information:
This can be fixed by changing `project: 'e2e'` to `project: 'e2e/tsconfig.e2e.json'`.

brandonroberts:
Thanks for the PR 😉

sharikovvladislav:
I can submit the PR for this. But I will open another issue right after this :)

[...]

PR #242, opened 2017-08-06
Fix e2e tests: fix path to tsconfig.json and fix spec to pass tests

sharikovvladislav:
Fix for this: https://github.com/ngrx/platform/issues/241


brandonroberts:
Thanks!

PR #261, opened 2017-08-10
docs: fix selection of state using arrow function

jdjuan:
The select method from the Store expects a function, when I put the `this.counter = store.select<number>('counter');` the compiler says: `Argument of type '"counter"' is not assignable to parameter of type '(state: AppState) => string'.`

### DIFF
--- a/docs/store/README.md
+++ b/docs/store/README.md
@@ -96,7 +96,7 @@ export class MyAppComponent {
 	counter: Observable<number>;
 
 	constructor(private store: Store<AppState>) {
-		this.counter = store.select<number>('counter');
+		this.counter = store.select<number>((state: AppState) => state.counter);
 	}
 
 	increment(){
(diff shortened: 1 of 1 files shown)


docs/store/README.md
```diff
@@ -96,7 +96,7 @@ export class MyAppComponent {
 	counter: Observable<number>;
 
 	constructor(private store: Store<AppState>) {
-		this.counter = store.select<number>('counter');
+		this.counter = store.select<number>((state: AppState) => state.counter);
```
brandonroberts: Don't remove the string selector, remove the `<number>` type instead. You can't use the generic type with the string selector, which is the cause for the error.

jdjuan: You were right, I guess it is easier for new comers to select state using the `string` selector instead of the callback.

brandonroberts:
Thanks!

PR #266, opened 2017-08-12
Deprecate toPayload utility function

karptonite:
As discussed here, https://github.com/ngrx/platform/pull/161#issuecomment-321128856, this is a pull request to deprecate toPayload. It also adds a bit of documentation for the type-safe `ofType` method. 

Do you think it needs to be mentioned anywhere else in the docs? I could add something to the Migration notes as well, but I wasn't sure it really needed to be there.

This shouldn't be merged until a 4.1 release.

### DIFF
--- a/docs/effects/api.md
+++ b/docs/effects/api.md
@@ -55,7 +55,7 @@ export class SomeEffectsClass {
 
 ### ofType
 
-Filter actions by action types.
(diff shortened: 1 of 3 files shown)


MikeRyanDev:
Thanks!

Issue #170, opened 2017-07-24
metaReducer typing expects ActionReducer array


luchsamapparat:
According to the documentation, the `metaReducer` property of the store configuration expects an array of functions that accept the reducer and return a reducer function.

```ts
function debug(reducer) {
  return function(state, action) {
    console.log('state', state);
    console.log('action', action);

    return reducer(state, action);
  }
}

const metaReducers = [debug];

@NgModule({
  imports: [
    StoreModule.forRoot(reducers, { metaReducers })
  ]
})
```
https://github.com/ngrx/platform/blob/master/docs/store/api.md#meta-reducers

I tested this using this code...

```ts
        StoreModule.forRoot(REDUCERS_TOKEN, {
            metaReducers: [
                (...args) => console.log(args)
            ]
        }),
```

... and the browser console confirms this:

![image](https://user-images.githubusercontent.com/875017/28520813-5675ae6a-7071-11e7-93ad-8a823711f25d.png)

However, when I want to actually implement the code from the documentation, I get typing issues.

```
[ts]
Argument of type '{ metaReducers: ((reducer: AppState) => (state: any, action: any) => any)[]; }' is not assignable to parameter of type 'StoreConfig<AppState, Action>'.
  Types of property 'metaReducers' are incompatible.
    Type '((reducer: AppState) => (state: any, action: any) => any)[]' is not assignable to type 'ActionReducer<AppState, Action>[]'.
      Type '(reducer: AppState) => (state: any, action: any) => any' is not assignable to type 'ActionReducer<AppState, Action>'.
        Type '(state: any, action: any) => any' is not assignable to type 'AppState'.
```

From what I see, the `StoreConfig` interface incorrectly defines `metaReducers` as `ActionReducer<any, any>[];`

```ts
export interface StoreFeature<T, V extends Action = Action> {  key: string;
  reducers: ActionReducerMap<T, V> | ActionReducer<T, V>;
  reducerFactory: ActionReducerFactory<T, V>;
  initialState?: InitialState<T>;
  metaReducers?: ActionReducer<any, any>[];
}
```

https://github.com/ngrx/platform/blob/d2295c78a29635d40b70b44155d7f0e783310c0b/modules/store/src/models.ts#L29

From what I understand the correct typing should be `ActionReducer<T, V>[]`.

Can anyone confirm this?

**Update**

Here's a repo to reproduce this issue:

https://github.com/luchsamapparat/ngrx-meta-reducer-issue

[...]

PR #270, opened 2017-08-15
fix(Store): Add type signature for metareducer

### LINKED ISSUE #264: Store: is metaReducers signature correct?
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Feature request
[ ] Support request
</code></pre>

In `StoreModule.forRoot()`, in its `StoreConfig<T,V>` parameter, its `metaReducers` property is typed as an (optional) array of `ActionReducer<T,V>`.

But then in [meta reducers docs](https://github.com/ngrx/platform/blob/master/docs/store/api.md#meta-reducers) provided example (which is correct, I think), does not adhere to that type: 

```ts
function debug(reducer) {
  return function(state, action) {
    // ... log things ...
    return reducer(state, action);
  }
}
```

which is like (adding fictional generic types):

```ts
function debug(reducer: ActionReducer<T,V>): 
  (red: ActionReducer<T,V>) => ActionReducer<T,V> {

  return function wrapperReducer(state: T, action: V): ActionReducer<T,V> {
    // ... log things ...
    return reducer(state, action);
  }

}
```

Infact, being a meta-reducer, `debug` is not a reducer itself: is a function taking a reducer and giving another (transformed, wrapped, decorated ...) reducer.

So it seems to me that `metaReducers` cannot be an array of `ActionReducer<T,V>`. As a matter of fact, when I add (non generic) types to my `debug` function, code does not compile because of `metareducers` issue with types: 

> [...]
> Types of property 'metaReducers' are incompatible. [...]
>      Type '(reducer: ActionReducer<PublicAppState, Action>) => ActionReducer<PublicAppState, Action>' is not assignable to type 'ActionReducer<PublicAppState, Action>'.

## Expected behavior:
`metaReducers` property should be typed as an (optional) array of reducer-transformers, each item like: `(red: ActionReducer<T,V>) => ActionReducer<T,V>`

brandonroberts:
Fixes the type signature issue for the array of metareducers. Each function should take a reducer and return a new reducer function.

Closes #264, Related to #170

### DIFF
--- a/docs/store/api.md
+++ b/docs/store/api.md
@@ -49,11 +49,11 @@ export function getInitialState() {
 configuration option to provide an array of meta-reducers that are composed from right to left.
 
 ```ts
-import { StoreModule, combineReducers, compose } from '@ngrx/store';
+import { StoreModule, ActionReducer } from '@ngrx/store';
 import { reducers } from './reducers';
 
 // console.log all actions
-function debug(reducer) {
+export function debug(reducer: ActionReducer<any>): ActionReducer<any> {
   return function(state, action) {
     console.log('state', state);
     console.log('action', action);
@@ -62,7 +62,7 @@ function debug(reducer) {
   }
 }
 
-const metaReducers = [debug];
+export const metaReducers = [debug];
 
 @NgModule({
   imports: [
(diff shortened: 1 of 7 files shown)


MikeRyanDev:
LGTM. Feel free to merge once the build turns green.

PR #288, opened 2017-08-17
fix(Effects): Use factory provide for console

### LINKED ISSUE #276: Cannot read property 'group' of undefined
## I'm submitting a...
<pre><code>
[x] Bug report  
[ ] Feature request
[ ] Support request
</code></pre>

I've found the error `Cannot read property 'group' of undefined` logged in my sentry errors. The error is thrown from this line of code: https://github.com/ngrx/platform/blob/015107f88dc1713a33f7a7596791513985fd8e35/modules/effects/src/error_reporter.ts#L9

That means that `this.console` should be undefined.

The stack trace is coming from https://github.com/ngrx/platform/blob/48a2381c212d5dd3fa2b9435776c1aaa60734235/modules/effects/src/effect_sources.ts#L37 and inside here from https://github.com/ngrx/platform/blob/48a2381c212d5dd3fa2b9435776c1aaa60734235/modules/effects/src/effect_notification.ts#L19 

I'm importing `EffectsModule.forRoot(...)` exactly once in my application, in my main module, so I don't quite understand how `this.console` could be undefined.

## Expected behavior:

Not getting that error, instead seeing the log output

Unfortunately I don't quite understand how to get into that part of code, and therefore couldn't reproduce it.

- Browser: Chrome 60.0.3112
- OS: Windows 10

brandonroberts:
Closes #276

### DIFF
--- a/modules/effects/src/effects_module.ts
+++ b/modules/effects/src/effects_module.ts
@@ -40,7 +40,7 @@ export class EffectsModule {
         },
         {
           provide: CONSOLE,
-          useValue: console,
+          useFactory: getConsole,
         },
       ],
     };
@@ -49,4 +49,8 @@ export class EffectsModule {
 
 export function createSourceInstances(...instances: any[]) {
   return instances;
+}
+
+export function getConsole() {
+  return console;
 }


pgrm:
thx for the quick fix 🎆

PR #291, opened 2017-08-17
feat(createSelector): Ability to test the projectorFn independent from the selectors it is composed of

### LINKED ISSUE #290: Ability to test the projectorFn of createSelector() independent from the selectors it is composed of
## I'm submitting a...

<pre><code>
[ ] Bug report  
[x] Feature request
[ ] Support request
</code></pre>

Currently, when you want to test a selector composed of other selectors you have to mock the state in a way that it satisfies all selectors. 

The following is taken from [here](https://github.com/reactjs/reselect#q-how-do-i-test-a-selector).
```javascript
// tests for the first three selectors...
test("firstSelector unit test", () => { ... })
test("secondSelector unit test", () => { ... })
test("thirdSelector unit test", () => { ... })

// We have already tested the previous
// three selector outputs so we can just call `.resultFunc`
// with the values we want to test directly:
test("myComposedSelector unit test", () => {
  // here instead of calling selector()
  // we just call selector.resultFunc()
  assert(myComposedSelector.resultFunc(1, 2, 3), true)
  assert(myComposedSelector.resultFunc(2, 2, 1), false)
})
```

## Expected behavior:
It should be possible to test the projector function independent from the selectors it is composed of.

03byron:
fixes #290

### DIFF
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -5,7 +5,7 @@ export type AnyFn = (...args: any[]) => any;
 export interface MemoizedSelector<State, Result>
   extends Selector<State, Result> {
   release(): void;
-  resultFunc: AnyFn;
+  projector: AnyFn;
 }
 
 export function memoize(t: AnyFn): { memoized: AnyFn; reset: () => void } {
@@ -135,7 +135,7 @@ export function createSelector(...args: any[]): Selector<any, any> {
 
   return Object.assign(memoizedState.memoized, {
     release,
-    resultFunc: memoizedProjector.memoized,
+    projector: memoizedProjector.memoized,
   });
 }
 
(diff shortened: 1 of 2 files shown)


modules/store/src/selector.ts
```diff
@@ -1,12 +1,13 @@
 import { Selector } from './models';
 
+export type AnyFn = (...args: any[]) => any;
+
 export interface MemoizedSelector<State, Result>
   extends Selector<State, Result> {
   release(): void;
+  resultFunc: AnyFn;
```
MikeRyanDev: Let's call this `projector` instead

03byron: Thanks for reviewing. Just updated it accordingly.

MikeRyanDev:
Thanks!

PR #298, opened 2017-08-18
chore: fix link to collaborators in readme

sharikovvladislav:
Current relative link has */blob/master*.

Any thoughts how to save relative path but get correct link? Now there is extra "/blob/master" in it.


sharikovvladislav:
I removed extra diff. I had problems with merge, but I solved everything.

PR #299, opened 2017-08-18
chore: fix contributors link in readme

sharikovvladislav:
Current link does not work. It contains */blob/master* at the moment (I think if another branch will be active link will be changed too).

Any ideas how to save relativeness of link but make it working?


brandonroberts:
Thanks!

Issue #128, opened 2017-07-20
Interface Action should be generic.


ovangle:
This issue is a resubmission (and improvement)  of the proposal `https://github.com/ngrx/store/issues/297`: _`Action` to be made generic_.

----- 

## New `Action<T>` Interfaces

`ActionType` represents the interface that is statically implemented by an action. 

It guarantees that `Action` must be constructable using a single parameter, and be declared as part of an `NgModule`.  

    interface ActionType<T> {
      /**
       * The `NgModule` that this action is declared as part of.
       */
      ngModule: Type<T>;
      /**
       * The action is carrying information of type `payload`
       */
      new (payload: T): Action<T>;
    }

     /**
      * Represents an instruction that the reducer will apply to the module/global state.
      */
     interface Action<T> {
        readonly payload: T;
     }

The following declarations would all be a conforming Actions:

```
class VoidAction implements Action<null> {
   static ngModule: AppModule;
   constructor(public readonly payload: null) {}
}

class IncrementAction implements Action<number> {
   static ngModule: MathModule;
   constructor(public readonly payload: number) {}
}

class ComplexAction implements Action<ComplexInterface> {
    constructor (public readonly payload: ComplexInterface) {}
```

## Some utility functions

It needs to be possible to obtain a value which corresponds to the old `action.type`. 

    /** 
     * This function can be used wherever there was previously a call to `action.type`, 
     * which is expected to return a string.
     */
    export function actionTypeToString(action: Action<T>): string {
        let type = actionType(action);
        function stringify(type: Type<T>) { return type.constructor.name; }
        return `${stringify(actionType.module)} (${stringify(type)})`;
    }

    export function actionType(action: Action<T>): ActionType<T> {
        return Object.getPrototypeOf(action).constructor;
    }

See [this attached gist](https://gist.github.com/ovangle/699f6a45423b8a67360be1b98a77da76) for an example of using the improved interface for `Action`, and how the example (despite being a stupid example) feels more "angular"y, since it uses the static type of an object as a token...
        
----

## Simple inline actions

By making the interface for `Action` rely upon the runtime type of the action for use in reducers, it removes the possibility for calls like the following:

   ```
   store.dispatch({type: 'some-quick-and-dirty-string', payload: Math.PI});
   ```

However, we can workaround this by enforcing a simple rule throughout the code:

    ```
    action has a property 'type' and the value is a string? Use it. Else, try use the runtime type.
    ```

Thus, for example, the function `actionTypeToString` would actually be

[...]

PR #305, opened 2017-08-20
Issue 303 PR: Recommended migration doc change.

TheBolmanator:
TL;DR  I conclude removal of the Before and After example removes a source of confusion.  
Payload property has been removed from the Action interface.  I really like the "Before" and "After" concept of this migration guide however in this case, the change seems to me to be only observable via intellisense or IDE.  I may well be wrong, but I just don't see how the change is reflected in the Before and After Case provided.  I think do think that users need to understand that when a payload is appropriate it may be added, but that the Action interface will no longer enforce inclusion of a payload property.

In the future I will be more proactive with my pull requests as I hope to be an increasingly active member this and other Repos that I frequent.  (Newbie).  

Many thanks for the great work 
Steve

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -43,7 +43,7 @@ import { compose } from '@ngrx/store';
 
 ### Action interface
 
-The `payload` property has been removed from the `Action` interface. As a result `payload` will no longer be included in the typing for `Actions` in your IDE.  Of course `Actions` may still recieve a `payload` but as an optional parameter.
+The optional property `payload` has been removed from the `Action` interface. As a result `payload` is no longer be included in the typing for `Actions` (i.e. will not show as a property in your IDE).  `Actions` may still recieve a `payload`, just don't be supprised when your IDE does not hint that you provide one.
 
 ### Registering Reducers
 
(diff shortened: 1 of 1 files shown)


brandonroberts:
I don't think the Before and After example should be completely removed since we're providing guidance on migration. If anything, the explanation from here https://github.com/ngrx/platform/issues/128#issuecomment-316680916 could be incorporated into the guide.

karptonite:
Also, if you add type information, with `ofType<FooAction>`, the IDE and compiler should correctly infer whichever properties you include on the action.

TheBolmanator:
I will take a crack at incorporating the comment from #128 into the PR.
@karptonite I do not follow your point, and would need some additional detail to attempt to capture it.

TheBolmanator:
Now that I have read issue #128 I am afraid to say anything on the subject.  To be perfectly honest this whole thing is over my head.   I see that even removal of an optional parameter can be a source of disquiet in the community.  A possible rewrite may be as proposed below however I am not able to translate how this fits into an "AFTER" example.  For what it is worth I just find the current "AFTER" example to be very confusing.  

PR to be resubmitted, but I am equally happy to close this with no action taken.

brandonroberts:
Don't remove the migration example. Users still need sample code to reference. If you want to add an interface for an untyped payload, then add it as additional information.

TheBolmanator:
No Action will be taken.

PR #344, opened 2017-09-04
docs(RouterStore): type signature of ROUTER_NAVIGATION action

dherges:
Readme is a little bit mis-leading here. It tells that the signature of the action is `RouterNavigationPayload`. In fact, it is a `RouterNavigationAction` with a payload of the above type.


brandonroberts:
Thanks!

PR #355, opened 2017-09-06
fix(RouterStore): Fix cancelled navigation with async guard (fixes #354)

### LINKED ISSUE #354: Navigation is cancelled by any action during an async guard checking
## I'm submitting a...
<pre><code>
[x] Bug report
[ ] Feature request
[ ] Support request
</code></pre>

The navigation is cancelled when an action dispatched during a guard checking.

A simple timeline:

1. Navigate to a page which has an async guard
2. Before the guard is resolved, emit an action
3. The navigation will be cancelled and relaunched even if the guard accepts the navigation

## Expected behavior:

Simple timeline:

1. Navigate to a page which has an async guard
2. Before the guard is resolved, emit an action
3. The navigation should not be cancelled when the guard accepts the navigation

(For a simple reproduction see my unit test: https://github.com/kondi/platform/blob/13feb6c12118955473f351e6fc7e3575699d93f3/modules/router-store/spec/integration.spec.ts#L377.)

Detailed plunker example:
https://plnkr.co/edit/VVe26uhP3qoNARyhPmi1?p=preview

After clicking on button, you can see in the console:

<pre><code>[Router] /page NavigationStart
[Router] /page NavigationCancel
[Router] /page NavigationStart
[Router] /page NavigationEnd
</code></pre>

However you should see:

<pre><code>[Router] /page NavigationStart
[Router] /page NavigationEnd
</code></pre>

Explanation of the attached plunker example:

1. Click on the button, 'GO_TO_PAGE' action is dispatched
2. `AppEffects.goToPage$` initiates a navigation to `/page` (using `navigateByUrl`)
3. `AppEffects.whatever$` starts to wait half second
3. Async `AppGuard` is activated which starts to wait 1 second
4. `AppEffects.whatever$` has finished waiting, emits any action (without any other side-effect)
5. `AppGuard` has finished waiting, returns true
6. Navigation is cancelled and started again for `/page`
7. Async `AppGuard` is activated again which starts to wait 1 second
8. `AppGuard` has finished waiting, returns true
9. Navigation is finished for `/page`

The expected behavior is to skip step 6, 7 and 8.

@ngrx/router-store: 4.0.4

## Other information:

You can "fix" this on the user side by not registering the routerReducer, or registering it, but with different name.
I have a simple fix for the library too, PR is on the way.

kondi:
Fix for #354, see more details there.
Contains unit test which fails without the fix.

### DIFF
--- a/modules/router-store/spec/integration.spec.ts
+++ b/modules/router-store/spec/integration.spec.ts
@@ -14,7 +14,10 @@ import {
 } from '../src/index';
 import 'rxjs/add/operator/filter';
 import 'rxjs/add/operator/first';
+import 'rxjs/add/operator/mapTo';
+import 'rxjs/add/operator/take';
 import 'rxjs/add/operator/toPromise';
+import { of } from 'rxjs/observable/of';
 
 describe('integration spec', () => {
   it('should work', done => {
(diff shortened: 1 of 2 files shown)


brandonroberts:
Thanks!

PR #375, opened 2017-09-11
fix(StoreModule) Fix typing to accept InjectionToken

Szpadel:
forFeature method accepts also InjectionToken as reducer, but this was disallowed by typescript definition.


brandonroberts:
Thanks!

PR #376, opened 2017-09-11
Another change to keep closure compiler happy

ukrukarg:
According to closure compiler, you cannot have non-optional arguments after optional ones.


MikeRyanDev:
Any chance we could get Closure integration tests to verify this? cc @robwormald

robwormald:
We should probably revert this, as Dan points out, this is unsafe - a state of `0` or `false` would reset back to the initial state...

PR #381, opened 2017-09-12
Broken unit tests on master

Manduro:
I ran into these failing tests while debugging a problem I'm having where my custom RouterStateSnapshot serializer is not receiving url fragment parameters in the snapshot.

The guards and resolve events were missing from the expected results. Also, one test was nested inside another. (no whitespace diff: https://github.com/ngrx/platform/pull/381/files?w=1)

The serializer test still fails, with params not getting passed. So I guess that's a bug which was missed because of the other failing tests?

CI seems to be running old code, as the tests I fixed still fail there?


brandonroberts:
These tests are valid if you are running Angular 4.3 or later. Our dependencies are locked to 4.2.6 using yarn.

PR #390, opened 2017-09-14
Minor fixes to entity/interfaces.md

buge:
- Fixes the type for the selectedUserId in the example.
- Fixes the link to creating an entity adapter.


brandonroberts:
Thanks!

Issue #378, opened 2017-09-11
Default selectId function for EntityAdapter


code2nguyen:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?

I have several entities, and I use always id property for the primary key.  So, to create the EntityAdater I must repeat the code like:

```JS
export const myEntityAdapter: EntityAdapter<MyEntity> = createEntityAdapter<MyEntity>({
  selectId: (myEntity: MyEntity) => myEntity.id
});
```
## Expected behavior:

This could be greate if createEntityAdapter uses a default selectId method which will take <code>id</code> property of object as a primary key. So I just need call this function
 ```JS
export const myEntityAdapter: EntityAdapter<MyEntity> = createEntityAdapter<MyEntity>();
```

brandonroberts:
Fixed via #405

PR #405, opened 2017-09-19
Default selectId function for EntityAdapter

PhillipZada:
See #378 

**What is the current behavior?**

I have several entities, and I use always id property for the primary key. So, to create the EntityAdater I must repeat the code like:

```
export const myEntityAdapter: EntityAdapter<MyEntity> = createEntityAdapter<MyEntity>({
  selectId: (myEntity: MyEntity) => myEntity.id
});
```

**Expected behavior:**

This could be greate if createEntityAdapter uses a default selectId method which will take id property of object as a primary key. So I just need call this function

```
export const myEntityAdapter: EntityAdapter<MyEntity> = createEntityAdapter<MyEntity>();
```


modules/entity/src/create_adapter.ts
```diff
@@ -10,10 +10,12 @@ import { createSelectorsFactory } from './state_selectors';
 import { createSortedStateAdapter } from './sorted_state_adapter';
 import { createUnsortedStateAdapter } from './unsorted_state_adapter';
 
-export function createEntityAdapter<T>(options: {
+export function createEntityAdapter<T>(options?: {
```
brandonroberts: Instead of making `options` optional, make `selectId` optional and provide a default below

modules/entity/src/create_adapter.ts
```diff
@@ -10,10 +10,12 @@ import { createSelectorsFactory } from './state_selectors';
...
   selectId: IdSelector<T>;
   sortComparer?: false | Comparer<T>;
 }): EntityAdapter<T> {
+  options = options || { selectId: (instance: any) => instance.id };
+
   const { selectId, sortComparer }: EntityDefinition<T> = {
     sortComparer: false,
```
brandonroberts: Added a default `selectId` function here under the `sortComparer`

modules/store/src/store_module.ts
```diff
@@ -149,12 +149,16 @@ export class StoreModule {
   ): ModuleWithProviders;
```
brandonroberts: Do these changes in a separate commit

PhillipZada: Hey Brandon, these changes are happening automatically as per the precommit script. Any suggestions on how to stop it.

modules/entity/src/create_adapter.ts
PhillipZada: Done

PhillipZada:
Hey @brandonroberts,

The original request was to make it possible to create an adaptor using this 

```
export const myEntityAdapter: EntityAdapter<MyEntity> = createEntityAdapter<MyEntity>();
```

I removed the optional options flag and replaced it with a default value if not provided. Hope this meets requirements.

modules/store/src/store_module.ts
brandonroberts: Don't worry about it then. Must have merged another change that didn't get formatted.

brandonroberts:
Thanks!

PR #407, opened 2017-09-19
docs(effects): Remove usage of deprecated `toPayload` util in example

jayphelps:
`toPayload` is [listed as deprecated](https://github.com/ngrx/platform/blob/master/docs/effects/api.md#topayload-deprecated), but the effects docs still use them in the example.


MikeRyanDev:
Thanks!

PR #409, opened 2017-09-20
added missing import for `rxjs/add/operator/catch` in example doc

shcarrico:
just a minor fix, does not compile for me otherwise .. 

typescript: `2.5.2`
rxjs: `5.4.2`
angular: `4.2.4`


brandonroberts:
Thanks!

Issue #410, opened 2017-09-20
Replacing 'routerReducer' as an injectable token rather than static text


phillipzada:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
Currently in router_store_module.ts routerReducer is a static text key. 

## Expected behavior:
Allow for the key name to be injected via a provider and token but defaulting to 'routerReducer' of not provided

## Other information:
Happy to put together a PR if approved.

brandonroberts:
Seems reasonable to me. I would suggest adding a `forRoot` method that takes a config object with a key for the router reducer name. This way you don't have to register a new provider to override it.

PhillipZada:
Cool, so the end result will be something like this

```
StoreRouterConnectingModule.forRoot( { stateName: 'routerReducer' } );
```

brandonroberts:
Yep. What about `stateKey` for the name?

edmundo096:
This catched me by surprise. I used another name for the router state key and time traveling didn't worked on my first try.
After checking the example-app, I noticed the persistence on using the key 'routerReducer', so I checked the src and found that key. I gave it a try and it worked.

At least the setup code on the docs makes mention it in the `StoreModule.forRoot({ routerReducer: routerReducer }),` line, but maybe a note would be useful either if this is changed or not.

pauldubois777:
@edmundo096 - Thanks for your post!  This was driving me crazy trying to debug why time-traveling wasn't working!  I tried a lot of things, and did a lot of searching for fixes until I found your post.  I renamed the property to `routerReducer` and now it works!

I hope this can be fixed so that specific naming of the property isn't required, or at least update the documentation.  I can help if needed.

PR #417, opened 2017-09-21
Replacing 'routerReducer' with a configurable option

phillipzada:
See https://github.com/ngrx/platform/issues/410

- Creation of a forRoot method to allow for configuration of StoreRouterConnectingModule
- Replacement of 'routerReducer' static state key with a configurable value


modules/router-store/spec/integration.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { StoreRouterConfig } from '../src/router_store_module';
```
brandonroberts: This can be added to the imports below from `../src/index`

modules/router-store/src/router_store_module.ts
```diff
@@ -155,18 +165,31 @@ export function routerReducer<T = RouterStateSnapshot>(
   ],
 })
 export class StoreRouterConnectingModule {
```
brandonroberts: This would be a breaking change for existing users unless you add default providers to to NgModule metadata

modules/router-store/src/router_store_module.ts
```diff
@@ -155,18 +165,31 @@ export function routerReducer<T = RouterStateSnapshot>(
...
+  static forRoot(config?: StoreRouterConfig): ModuleWithProviders;
+  static forRoot(config: StoreRouterConfig = {}): ModuleWithProviders {
+    return {
+      ngModule: StoreRouterConnectingModule,
+      providers: [{ provide: _ROUTER_CONFIG, useValue: config }],
```
brandonroberts: This won't work in AoT because the actual value for `useValue` property won't get resolved. You'll need to have two tokens, one for internal use that you register using `useValue` and another you register with `useFactory` that uses the first token as a dependency. Then you'll inject the public token in the constructor.

modules/router-store/src/router_store_module.ts
```diff
@@ -155,18 +165,31 @@ export function routerReducer<T = RouterStateSnapshot>(
...
+  }
+
   private routerState: RouterStateSnapshot;
   private storeState: any;
   private lastRoutesRecognized: RoutesRecognized;
 
   private dispatchTriggeredByRouter: boolean = false; // used only in dev mode in combination with routerReducer
   private navigationTriggeredByDispatch: boolean = false; // used only in dev mode in combination with routerReducer
 
+  private stateKey: string = 'routerReducer';
```
brandonroberts: Initialize this value in the factory function used to register the token

phillipzada:
Hey @brandonroberts just wanted to get your thoughts my my previous comments.

brandonroberts:
Hey @phillipzada. If you're talking about jasmine-marbles, I just responded to your comments.

phillipzada:
@brandonroberts sorry mate, I was referring the comments to your comments to the code changes above.

Hey Brandon,

Would this work, you do you still prefer an internal and external token?

```

export function createDefaultRouterConfig(config: StoreRouterConfig = {}): StoreRouterConfig {
  return {
    stateKey: 'routerReducer',
    ...config
  }
}



@NgModule({
  providers: [
    { provide: RouterStateSerializer, useClass: DefaultRouterStateSerializer },
    { provide: ROUTER_CONFIG, useFactory: createDefaultRouterConfig }
  ],
})
export class StoreRouterConnectingModule {
  static forRoot(config?: StoreRouterConfig): ModuleWithProviders;
  static forRoot(config: StoreRouterConfig = {}): ModuleWithProviders {
    return {
      ngModule: StoreRouterConnectingModule,
      providers: [{ provide: ROUTER_CONFIG, useFactory: createDefaultRouterConfig, deps: [config] }],
    };
  }

  private routerState: RouterStateSnapshot;
  private storeState: any;
  private lastRoutesRecognized: RoutesRecognized;

  private dispatchTriggeredByRouter: boolean = false; // used only in dev mode in combination with routerReducer
  private navigationTriggeredByDispatch: boolean = false; // used only in dev mode in combination with routerReducer

  private stateKey: string;

  constructor(
    private store: Store<any>,
    private router: Router,
    private serializer: RouterStateSerializer<RouterStateSnapshot>,
    @Inject(ROUTER_CONFIG) private config: StoreRouterConfig
  ) {
    const routerConfig = createDefaultRouterConfig(config);
    this.stateKey = (routerConfig.stateKey as string);
```

brandonroberts:
@phillipzada changes look good for the most part. You can't use `config` as a dependency without it being a token though. And then you won't need to call `createDefaultRouterConfig` in the constructor.

phillipzada:
Thanks @brandonroberts, appreciate the assist, will make the changes this week.

phillipzada:
Hey @brandonroberts, Changes have been made.

brandonroberts:
Thanks @phillipzada. One small thing. Change `@ngrx/router` to `@ngrx/router-store` in the InjectionToken strings. Everything else looks good.

brandonroberts:
The providers in the forRoot method should be in the NgModule metadata also to prevent a breaking change for existing users.

brandonroberts:
@phillipzada Can you fix the merge conflicts and failing tests? I'm ready to land this one.

brandonroberts:
@phillipzada can you remove the `yarn.lock` file from the changes? There shouldn't be any dependency changes.

brandonroberts:
@phillipzada sorry, I wasn't clear. Can you remove the `yarn.lock` changes? The `yarn.lock` file in master should remain unchanged

phillipzada:
So make a copy of the current version and put it back in?

PR #426, opened 2017-09-25
Add information to docs on usage of custom router state serializer wi…

zakhenry:
…th store freeze

also s/pair/pare


brandonroberts:
Thanks!

PR #440, opened 2017-09-29
Improve performance for sorted entity adapter

dinvlad:
Prior to this fix, collection methods in sorted_state_adapter
ran in quadratic time, because they iterated over the collection
with single-item mutators, each of which also running in linear time.

Now, we first collect all modifications (each in constant time),
then sort them in linearithmic time, and finally merge
them with the existing sorted array linearly.

As a result, we can improve performance of collection methods
to N + M log M, where N is the number of existing items
and M is the number of modifications.

Single-item methods are now expressed through those
for collections, still running linearly.

### DIFF
--- a/modules/entity/src/sorted_state_adapter.ts
+++ b/modules/entity/src/sorted_state_adapter.ts
@@ -20,102 +20,89 @@ export function createSortedStateAdapter<T>(
   );
 
   function addOneMutably(entity: T, state: R): boolean {
-    const key = selectId(entity);
-
-    if (key in state.entities) {
-      return false;
-    }
-
-    const insertAt = findTargetIndex(state, entity);
-    state.ids.splice(insertAt, 0, key);
-    state.entities[key] = entity;
-
-    return true;
+    return addManyMutably([entity], state);
   }
 
   function addManyMutably(newModels: T[], state: R): boolean {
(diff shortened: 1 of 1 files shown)


MikeRyanDev:
Thanks @dinvlad!

dinvlad:
Glad to help!

Something I haven't addressed is if an entity is updated with a new ID (as determined by `selectId()`), then the sorted code doesn't check if this ID already exists (in other words, another entity may already correspond to this ID). In that case, the code may put 2 identical IDs in the new sequence of state IDs. Additionally, they may actually violate the order, e.g. if the other entity is larger than a particular element, but the new entity is smaller. The new entity will then overwrite the other one, but its ID will appear twice in the sequence:

`(new entity) ... < (element) ... > (new entity) <= ...`

To prevent that, we could `delete state.entities[selectId(updated)];` in addition to https://github.com/ngrx/platform/blob/7fa0cf2559bab9b1090148ae6042aff82881a4c4/modules/entity/src/sorted_state_adapter.ts#L55

However, this case prompts the question: should such updates be considered illegal in the first place? After all, they modify multiple entities with a 'single' update and it may not be clear which update should take precedence (imagine doing multiple of these at the same time).

PR #441, opened 2017-09-29
feat(Entity): Add support for string or number type for entity ID

### LINKED ISSUE #398: Incompatible type error when following docs
## I'm submitting a...

<pre><code>
[ ] Bug report  
[ ] Feature request
[X] Documentation issue or request
</code></pre>

When following the docs for the [Entity Adapter/Adapter Collection Methods](https://github.com/ngrx/platform/blob/c31573f1c4d9cc02a90cd2033f68d83b17b74969/docs/entity/adapter.md#adapter-collection-methods), I get a `"Types of property 'id' are incompatible."` error at several points:

```
export const adapter: EntityAdapter<User> = createEntityAdapter<User>({
  selectId: (user: User) => user.id
});
```

and

```
[...]
case UserActions.UPDATE_USER: {
  return adapter.updateOne(action.payload.user, state);
}

case UserActions.UPDATE_USERS: {
  return adapter.updateMany(action.payload.users, state);
}
[...]
```

This seems to happen because the docs declare `user.id` as a `number`, but the code behind declares `id` everywhere as `string`.

## Expected behavior:
Types should be compatible when following the docs.

Follow the docs  for the [Entity Adapter/Adapter Collection Methods](https://github.com/ngrx/platform/blob/c31573f1c4d9cc02a90cd2033f68d83b17b74969/docs/entity/adapter.md#adapter-collection-methods).

## Other information
**EDIT:** The interface `State` that extends `EntityState<User>` declares the `selectedUserId` property as a type of `number`, too. It may should be also changed to type `string` to prevent further issues.
```
export interface State extends EntityState<User> {
  // additional entities state properties
  selectedUserId: number | null;
}
```
____________________
Sorry for the sparse issue description, I don't have much time at the moment.

brandonroberts:
Closes #398

### DIFF
--- a/modules/entity/src/entity_state.ts
+++ b/modules/entity/src/entity_state.ts
@@ -1,4 +1,4 @@
-import { EntityState } from './models';
+import { EntityState, EntityStateStr, EntityStateNum } from './models';
 
 export function getInitialEntityState<V>(): EntityState<V> {
   return {
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -1,24 +1,56 @@
-export type Comparer<T> = {
+export type ComparerStr<T> = {
+  (a: T, b: T): string;
+};
+
+export type ComparerNum<T> = {
   (a: T, b: T): number;
 };
 
-export type IdSelector<T> = {
+export type Comparer<T> = ComparerNum<T> | ComparerStr<T>;
+
+export type IdSelectorStr<T> = {
   (model: T): string;
 };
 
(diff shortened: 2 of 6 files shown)


Lycidas0815:
Stupid question, since my project broke due to that change: how to I specify the type argument to use properly? My `selectId` is of type number, but I get a `the type argument for type parameter Result cannot be inferred from the usage` error in my reducers, using the ngrx-example app as template.

I also had a look into the entity docs but didn't find anything new there.


Many thanks,
Ly

PR #454, opened 2017-10-05
fix(Entity): Change type for EntityState to interface

brandonroberts:
closes #458

### DIFF
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -40,17 +40,10 @@ export type UpdateNum<T> = {
 
 export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
-export interface EntityStateStr<T> {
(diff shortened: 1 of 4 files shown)


modules/entity/src/models.ts
```diff
@@ -40,18 +38,11 @@ export type UpdateNum<T> = {
 
 export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
-export interface EntityStateStr<T> {
-  ids: string[];
+export interface EntityState<T> {
+  ids: any[];
```
ksaifullah: If the possible values of `ids` are either string or number, it better to be explicit about that `ids: string[] | number[];`.

brandonroberts: This won't work with TypeScript

PR #457, opened 2017-10-06
EntityState is changed to interface from type and book model is changed to class from interface.

sanbornsen:
- These are couple of problems I saw while trying the example out, thought of fixing them.


brandonroberts:
Thanks for the PR! Please rebase off latest master

PR #462, opened 2017-10-08
Example App: interface can extend only interface or class

ksaifullah:
## The problem

An interface may only extend a class or another interface.
screen shot 2017-10-08 at 12 28 37 pm

## The solution

Instead of splitting the interface `EntityState` into two separate interfaces we can have a unified one.

### DIFF
--- a/modules/entity/src/entity_state.ts
+++ b/modules/entity/src/entity_state.ts
@@ -1,4 +1,4 @@
-import { EntityState, EntityStateStr, EntityStateNum } from './models';
+import { EntityState } from './models';
 
 export function getInitialEntityState<V>(): EntityState<V> {
   return {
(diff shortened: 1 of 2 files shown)


TheRedD0t:
Duplicate of #457

brandonroberts:
TypeScript doesn't allow union types with a number and string.

TheRedD0t:
> TypeScript doesn't allow union types with a number and string.

Sorry, but Typescript allow union types with number and string, even the [official documentation](https://www.typescriptlang.org/docs/handbook/advanced-types.html) includes a sample:

> function padLeft(value: string, padding: string | number) {
>     // ...
> }

brandonroberts:
Yes, but that is a function and not a generic interface. It won't compile if you try to build the example app after making that change to the interface.

PR #473, opened 2017-10-11
feat(Effects): Add root effects init action

### LINKED ISSUE #246: The way to fire an effect after effects are settled up
## I'm submitting a...

<pre><code>
[ ] Bug report  
[ ] Feature request
[ ] Support request
</code></pre>

There is no action, no function which fired when all effects are settled up. Where we can dispatch actions to fire other effects.

## Expected behavior:

```typescript
    this.onInit$ = this.actions$
      .ofType('@ngrx/effects/up')
      .switchMap((isAuthorized) => {
        return this.isAuthorized$;
      })
      .map((isAuthorized) => {
        if (isAuthorized) {
          this.store.dispatch({ type: '[posts] getAll' })
        }
      });

    @Effect()
    this.onGetPosts$ = this.actions$
      .ofType('[posts] getAll')
      .map(() => ... )
```
or, may be
```typescript
  public ngrxOnRunEffects(resolvedEffects$: Observable<EffectNotification>): Observable<EffectNotification> {
    this.isAuthorized$
      .take(1)
      .subscribe({
        next: (isAuthorized) => {
          if (isAuthorized) {
            this.store.dispatch({ type: '[posts] getAll' })
          }
        },
      });
    return resolvedEffects$;
  }
```

There a few issues, of `@ngrx/store/init` action. Seems, that this action is not useful for described purpose anymore. So, we probably need another.
#103 
#152 

ngrx v4

## Other information:

Possible workaround:

dispatch action, somewhere after application start.
In the `AppComponent` for example.
```typescript
this.store.dispatch({ type: '[app] init' });
```
And use it to dispatch other actions, that must be dispatched after all effects are settled up.

tdeschryver:
Closes #246

### DIFF
--- a/modules/effects/src/effects_root_module.ts
+++ b/modules/effects/src/effects_root_module.ts
@@ -16,11 +16,12 @@ export class EffectsRootModule {
     @Optional() storeModule: StoreModule
   ) {
     runner.start();
-    store.dispatch({ type: ROOT_EFFECTS_INIT });
 
     rootEffects.forEach(effectSourceInstance =>
       sources.addEffects(effectSourceInstance)
     );
+
+    store.dispatch({ type: ROOT_EFFECTS_INIT });
   }
 
   addEffects(effectSourceInstance: any) {
--- a/modules/effects/src/effects_root_module.ts
+++ b/modules/effects/src/effects_root_module.ts
@@ -1,18 +1,22 @@
 import { NgModule, Inject, Optional } from '@angular/core';
-import { StoreModule } from '@ngrx/store';
+import { StoreModule, Store } from '@ngrx/store';
 import { EffectsRunner } from './effects_runner';
 import { EffectSources } from './effect_sources';
 import { ROOT_EFFECTS } from './tokens';
+
+export const ROOT_EFFECTS_INIT = '@ngrx/store/root effects init' as '@ngrx/store/root effects init';
 
 @NgModule({})
 export class EffectsRootModule {
   constructor(
     private sources: EffectSources,
     runner: EffectsRunner,
+    store: Store<any>,
     @Inject(ROOT_EFFECTS) rootEffects: any[],
     @Optional() storeModule: StoreModule
   ) {
     runner.start();
+    store.dispatch({ type: ROOT_EFFECTS_INIT });
 
     rootEffects.forEach(effectSourceInstance =>
       sources.addEffects(effectSourceInstance)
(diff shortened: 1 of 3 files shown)


modules/effects/src/effects_root_module.ts
```diff
@@ -1,18 +1,22 @@
...
 export class EffectsRootModule {
   constructor(
     private sources: EffectSources,
     runner: EffectsRunner,
+    store: Store<any>,
     @Inject(ROOT_EFFECTS) rootEffects: any[],
     @Optional() storeModule: StoreModule
   ) {
     runner.start();
+    store.dispatch({ type: ROOT_EFFECTS_INIT });
```
MikeRyanDev: I think this needs to be beneath the lines where you add the effects.

modules/effects/src/effects_root_module.ts
```diff
@@ -1,18 +1,22 @@
 import { NgModule, Inject, Optional } from '@angular/core';
-import { StoreModule } from '@ngrx/store';
+import { StoreModule, Store } from '@ngrx/store';
 import { EffectsRunner } from './effects_runner';
 import { EffectSources } from './effect_sources';
 import { ROOT_EFFECTS } from './tokens';
 
+export const ROOT_EFFECTS_INIT = '@ngrx/store/root effects init' as '@ngrx/store/root effects init';
```
brandonroberts: I would rename the type string to `@ngrx/effects/init`

modules/effects/src/effects_root_module.ts
tdeschryver: Changed.
Why exactly :smile: ?

modules/effects/src/effects_root_module.ts
tdeschryver: Yep you're right, don't know what I was thinking!

modules/effects/src/effects_root_module.ts
MikeRyanDev: Dispatching and registering the effects happens synchronously. Moving the dispatch to after registering the effects will give the effects a chance to listen for this new action type.

MikeRyanDev:
Thanks @tdeschryver!

felikf:
I am just wondering ... is this supposed to work in feature effects (`EffectsModule.forFeature([Effects]`)? Because it does not...

If I use:

    @Effect({ dispatch: false }) init$= this.actions$
        .ofType(ROOT_EFFECTS_INIT)
        .do(console.log.bind(console, 'HIT!!!!'));

in root module, it works.

When I use it in feature module, it does not.

I would need this for feature module though... :)

PR #481, opened 2017-10-14
Getting tests to pass under Angular 4.3+

mattmutt:
Using Angular 4.3 and running the tests via `yarn test` would result in several failures due to the newer lifecycle hooks. 

http://angularjs.blogspot.com/2017/07/angular-43-now-available.html

This PR makes the necessary patches to ensuring tests pass. Verified basic linting passed as well via Angular CLI


brandonroberts:
Included with 63ab98a30f54f5eb2a385fabfa5bee7a1ef083b0

PR #482, opened 2017-10-14
feat(material): Upgrade @angular/material to v 2.0.0-beta.12

### LINKED ISSUE #448: Support for Angular 4.4.3+, deprecated MaterialModule and changes to md-input-container
## I'm submitting a...

<pre><code>
[] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
</code></pre>

MaterialModule has been deprecated, causing a compiling error.
Consider creating a "Material" module within their application (e.g., FooMaterialModule) that imports only the set of components actually used in the application.

## Expected behavior:

No compiling error caused in:
- core.module.ts
- auth.module.ts,
- app/books/components/index.ts
- app/books/books.module.ts

> To replace MaterialModule, users can create their own "Material" module within their application (e.g., GmailMaterialModule) that imports only the set of components actually used in the application.
[https://github.com/angular/material2/blob/master/CHANGELOG.md#materialmodule](url)

## Other information:
Updates on md-input-container:
- md-input-container renamed to md-form-field 
- input: Inputs have a width of 200px by default (similar to native input elements). The width can be overridden by via the mat-form-field css class. 

**In book-search.ts**
```
md-form-field {
    width: 300px;
}
```
**instead of**
```
input {
    width: 300px;
}
```

izifortune:
Angular material removed the global MaterialModule all the
components needs to be imported individually.

Fixes #448


example-app/app/books/components/book-search.ts
```diff
@@ -6,42 +6,39 @@ import { Component, Output, Input, EventEmitter } from '@angular/core';
...
-        <md-spinner [class.show]="searching"></md-spinner>
-      </md-card-content>
-    </md-card>
+    <mat-card>
+      <mat-card-title>Find a Book</mat-card-title>
+      <mat-card-content>
+        <mat-input-container>
+          <input matInput placeholder="Search for a book" [value]="query" (keyup)="search.emit($event.target.value)">
+        </mat-input-container>
+        <mat-progress-spinner mode="indeterminate" [class.show]="searching"></mat-progress-spinner>
```
krjordan: Instead of using `mat-progress-spinner` you can use the `mat-spinner`. The `<mat-spinner>` component is an alias for `<mat-progress-spinner mode="indeterminate">`. You should also add the `diameter` and `strokeWidth` inputs. This will allow us to not have to use `!important` when setting the `width` and `height`. It should look something like:
```
<mat-spinner [class.show]="searching" [diameter]="30" [strokeWidth]="3"></mat-spinner>
```

example-app/app/books/components/book-search.ts
```diff
@@ -6,42 +6,39 @@ import { Component, Output, Input, EventEmitter } from '@angular/core';
...
-    md-card-spinner {
-      padding-left: 60px; // Make room for the spinner
-    }
-
-    md-spinner {
-      width: 30px;
-      height: 30px;
+    .mat-progress-spinner {
+      width: 30px !important;
+      height: 30px !important;
```
krjordan: Remove the `width` and `height`

example-app/styles.css
```diff
@@ -10,3 +10,8 @@ html {
   -ms-overflow-style: none;
   overflow: auto;
 }
+
+.mat-progress-spinner svg {
```
krjordan: This is no longer needed after fixing the `book-search` component.

package.json
```diff
@@ -63,7 +63,7 @@
     "@angular/core": "^4.2.0",
     "@angular/forms": "^4.2.0",
     "@angular/http": "^4.2.0",
-    "@angular/material": "^2.0.0-beta.7",
+    "@angular/material": "^2.0.0-beta.12",
```
krjordan: We should lock this to version `2.0.0-beta.12` so it won't break if a new release is published with breaking changes.

package.json
```diff
@@ -123,7 +123,7 @@
     "zone.js": "^0.8.12"
   },
   "dependencies": {
-    "@angular/cdk": "^2.0.0-beta.8",
+    "@angular/cdk": "^2.0.0-beta.12",
```
krjordan: We should lock this to version `2.0.0-beta.12` so it won't break if a new release is published with breaking changes. This can probably also be moved to the `devDependencies`.

setup-jest.ts
```diff
@@ -1,6 +1,6 @@
 import 'jest-preset-angular';
-import { MdCommonModule } from '@angular/material';
-
+// import { MdCommonModule } from '@angular/material';
```
krjordan: These should be deleted instead of commented out.

example-app/app/books/components/book-search.ts
izifortune: great i was struggling to set the correct values for this :)

izifortune:
All comments addressed

example-app/styles.css
krjordan: Can you remove this block? It's no longer needed.

package.json
krjordan: We should lock this to version 2.0.0-beta.12 so it won't break if a new release is published with breaking changes. This can probably also be moved to the devDependencies.

Issue #491, opened 2017-10-16
Integration testing of @Effect() observables


dinvlad:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?

One is able to unit-test individual effect observables.

## Expected behavior:

We should be able to test that these observables are properly registered with the store (i.e. provide an integration test). For example, an observable may be provided and behave correctly in isolation, but a developer can forget or accidentally remove a corresponding `@Effect` decorator.

We should also be able to test that the decorator has been configured correctly. E.g. to test that "dispatch: false" was set.

Although we can indeed instantiate the "real" store and subscribe to its Actions, actions returned from effects in this way don't get "immediately" dispatched in a test. As a result, it is unclear how to "wait" until a particular effect "settled down" before testing whether this effect resulted in a dispatch of a new action, or whether no dispatch happened (if "dispatch: false").

If that's possible to achieve with the current implementation, could you provide an example how to?

## Minimal reproduction of the problem with instructions:
https://github.com/ngrx/platform/blob/master/example-app/app/auth/effects/auth.effects.ts

## Version of affected browser(s),operating system(s), npm, node and ngrx:
any

dinvlad:
As a possible solution, we could expose https://github.com/ngrx/platform/blob/100a8ef9c83d4eff4f1eb74800bf49d90773f5ab/modules/effects/src/effects_metadata.ts#L42-L45
`getSourceMetadata()` returns all of the necessary information to enable these tests.

Moreover, we can then get by with simple unit tests, and the integration test will consist of a single
`expect(TestBed.get(EffectsClass)).toBeDefined()` when `TestBed` imports an entire store module.

Currently, we can
```typescript
import { getSourceMetadata } from '@ngrx/effects/src/effects_metadata';
```
but I assume that is not reliable as it's not exposed through the API. I can submit a PR to export it.

PR #493, opened 2017-10-17
Expose @Effect() metadata for testing

dinvlad:
This PR exposes `getEffectsMetadata()` method
to access metadata supplied through `@Effect()` decorators.

This method is useful for when we want to test
that effects have been properly decorated/registered
with the store.

Without this method, it can be hard to ensure that
the decorators were supplied and properly configured
(this could be done through integration tests,
but that would be much more involved,
especially for "negative" cases with `dispatch: false`).

Relevant commits are 8efba70, f06e7444 and 0673d12
(earlier commits have already been merged in https://github.com/ngrx/platform/pull/440).

Relevant discussion: https://github.com/ngrx/platform/issues/491

### DIFF
--- a/modules/effects/src/effects_metadata.ts
+++ b/modules/effects/src/effects_metadata.ts
@@ -25,9 +25,6 @@ function setEffectMetadataEntries(sourceProto: any, entries: EffectMetadata[]) {
   Array.prototype.push.apply(meta, entries);
 }
 
-/**
- * @ExportDecoratedItems
- */
 export function Effect({ dispatch } = { dispatch: true }): PropertyDecorator {
   return function(target: any, propertyName: string) {
     const metadata: EffectMetadata = { propertyName, dispatch };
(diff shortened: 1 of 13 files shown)


modules/effects/src/effects_metadata.ts
```diff
@@ -40,3 +40,19 @@ export const getSourceMetadata = compose(
   getEffectMetadataEntries,
   getSourceForInstance
 );
+
+export interface EffectsMetadata {
```
MikeRyanDev: What do you think about changing this to accept a type parameter:

```ts
interface EffectsMetadata<T> {
  [keyof T]: undefined | { dispatch: boolean };
}
```

modules/effects/src/effects_metadata.ts
```diff
@@ -40,3 +40,19 @@ export const getSourceMetadata = compose(
...
+  [propertyName: string]: {
+    dispatch: boolean;
+  };
+}
+
+export function getEffectsMetadata(instance: any): EffectsMetadata {
```
MikeRyanDev: And then the signature of this function would become:

```ts
function getEffectsMetadata<T>(instance: T): EffectsMetadata<T>
```

dinvlad: Makes sense, additional type checking is a plus. Not sure it works with the interfaces though, so submitting a change that defines EffectsMetadata as a mapped type.

MikeRyanDev:
Thanks!

PR #504, opened 2017-10-19
chore(docs): made example code of using a selector w/ store more incl…

BobbieBarker:
I had to use these docs today and it seemed to me that the example of using a store with a selector function would be better if it was presented as working for both the `createSelector` and `createFeatureSelector` methods. In it's current presentation in my opinion it is too closely associated with the `createSelector` method and this might throw some people off. I also added a missing import to the demo code.


brandonroberts:
Thanks!

BobbieBarker:
You are very welcome. Thank you for providing this tool.

PR #509, opened 2017-10-20
Add TodoMVC-App in examples

nicobytes:
Example and simple TodoApp utilizing @ngrx/store with UI/UX and full functionalities by TodoMVC project.

https://github.com/nicobytes/todo-app-ngrx/blob/master/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Reactive libraries for Angular
 
 ## Examples
 - [example-app](./example-app/README.md) - Example application utilizing @ngrx libraries, showcasing common patterns and best practices.
+- [todoMVC-app](https://github.com/nicobytes/todo-app-ngrx/blob/master/README.md) - Example and simple TodoApp  utilizing @ngrx/store.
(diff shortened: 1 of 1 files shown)


brandonroberts:
Thanks for the PR @nicobytes and nice work on the repo. We are looking at a better way to provide examples from the community like this one so I'd rather hold off on adding more examples. I will reopen then.

nicobytes:
@brandonroberts ok I get it! but tell me, What are the features you should have? (add ngrx/entity for example)... I do not want the PR to be forgotten

bbaia:
I've updated the TODO sample for a meetup @ GDG in Toulouse, France.
You can find the sample code here: https://github.com/bbaia/gdgtoulouse-ngrx

I've updated components to have only presentational components and an unique container component that only select state and dispatch actions.
You can browse the commits to see the progressive integration of ngrx/effects, ngrx/entity and ngrx/router-store packages.

Issue #463, opened 2017-10-09
Add Redux DevTools `name` option.


krismeister:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ X ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
There is no support for the `__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` `name` option, so you can not see a friendly name in the Redux DevTools `AutoSelect instances` dropdown.

## Expected behavior:
 It would be convenient to pass  a `name` variable with the `StoreDevtoolsOptions` when calling `StoreDevtoolsModule.instrument`

roopkt:
Hi @brandonroberts I would like to work on this an create PR.

brandonroberts:
Hey @roopkt, we'll accept a PR for this

roopkt:
Hi @brandonroberts I checked in the fix in my forked branch.

roopkt:
Hi @brandonroberts  I am not able to rebase or fetch from upstream it says access denied. 
```
git fetch upstream
Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists
```

roopkt:
@brandonroberts  I finished fixing this issue.

PR #517, opened 2017-10-21
Added support to give name for store dev tool

roopkt:
This is the changes to put name on store dev tool configuration object for the  issue #463


roopkt:
Finished passing name to devtools fixes #463

roopkt:
Hi @brandonroberts  I updated this pull request with my latest changes.

roopkt:
Hi @brandonroberts ,
I fixed code merge issue as well. Please check if the implementation or the fix for the issue #463 is correct.

brandonroberts:
@roopkt the changelog shouldn't be included in your changes. Please verify that your changes change the name in the extension also.

roopkt:
Hi @brandonroberts  thanks for giving me the direction to pass options to send method. I am planning to inject then ```STORE_DEVTOOLS_CONFIG``` into ```DevtoolsExtension``` so that we can always pass the entire option in send method. This is one potential solution that I think of.

I even wanted to validate by running example app however unfortunately when I run 
``` yarn run eample:star``` I always get error ```NONEMPTY dir``` related error. I tried deleting '../platform/dist' folder manually still when I run the start script I get error at some point while "Copying type definition files" step. I event tried opening new command terminal to execute still same issue. Have you encountered this issue is there any wise workaround of this.

**Comment by roopkt** [comment]
Hey @brandonroberts  
I pushed one more change-set this time I verified the ReduxDevToolExtension name is indeed changing. 👍

File: CHANGELOG.md [hunk 1]
```diff
@@ -1,6 +1,204 @@
-# Changelogs
```
brandonroberts: This file should not be included in the changes

modules/store-devtools/src/instrument.ts
```diff
@@ -64,12 +64,16 @@ export function noMonitor(): null {
   return null;
 }
 
+export const DEFAULT_NAME = 'ngrx-store-devtool-instance';
```
brandonroberts: Change to `NgRx Store DevTools`

modules/store-devtools/src/instrument.ts
```diff
@@ -64,12 +64,16 @@ export function noMonitor(): null {
...
+
 export function createConfig(
   _options: StoreDevtoolsOptions
 ): StoreDevtoolsConfig {
-  const DEFAULT_OPTIONS: StoreDevtoolsConfig = {
+  const DEFAULT_OPTIONS: Partial<StoreDevtoolsConfig> = {
```
brandonroberts: Why is this a partial?

CHANGELOG.md
brandonroberts: The changelog should not be included in these changes. We'll generate a changelog in the next release

roopkt: sure

roopkt: Thanks for explaination

modules/store-devtools/src/instrument.ts
roopkt: I changed the name to NgRx Store DevTools

modules/store-devtools/src/instrument.ts
roopkt: I removed the partial from here.

brandonroberts:
Thanks!

PR #520, opened 2017-10-22
refactor(http): replace HttpModule with HttpClientModule

krjordan:


example-app/app/core/services/google-books.ts
```diff
@@ -1,22 +1,22 @@
...
   private API_PATH = 'https://www.googleapis.com/books/v1/volumes';
 
-  constructor(private http: Http) {}
+  constructor(private http: HttpClient) {}
 
   searchBooks(queryTitle: string): Observable<Book[]> {
     return this.http
-      .get(`${this.API_PATH}?q=${queryTitle}`)
-      .map(res => res.json().items || []);
+      .get<Book[]>(`${this.API_PATH}?q=${queryTitle}`)
```
brandonroberts: The response is an object that contains items, which is an array of books. Pass that as the type

```
.get<{ items: Book[] }>
```

Then you won't have to type `books` as `any`, because `items` will be inferred as `Book[]` and just use `res => res.items || []`

PR #534, opened 2017-10-30
feat(Codegen): Add base code and build for @ngrx/codegen

MikeRyanDev:
Includes basic implementation for the [@ngrx/codegen proposal](https://paper.dropbox.com/doc/ngrxcodegen-Proposal-DhD934mmHfqTljpntnqJ3)


bmayen:
Why not devkit schematics for this?

MikeRyanDev:
@bmayen Why would you suggest schematics? Genuinely curious.

bmayen:
Primarily because a) it already exists to handle code generation, and b) would give us a single workflow when using something like Angular CLI. 

In the same way that Angular provides a set of schematics which CLI consumes, so could ngrx. Then organizations, which may have customization needs, have an easy way to extend the defaults.

Also, over time, schematics looks like it will become more powerful; supporting user prompting, for instance. Would be nice to be able to leverage this without having to develop and maintain similar functionality in parallel.

mko:
I agree that having an `@ngrx/cli` which provides similar functionality to the schematics of `@angular/cli` would likely be a better long-term strategy than codegen, but this might be a good starting point in the interim.

MikeRyanDev:
We'll explore how to integrate this tool with Schematics. Merging early to start producing nightly builds.

bmayen:
Sounds good! Let me know if I can help with any Schematics effort.

PR #546, opened 2017-11-04
Boost the provideMockActions example for @ngrx/effects testing

getDanArias:
* Add link to article that describes the mechanics and syntax used for marble testing and refer to it within the body of the example.
* Add imports that increase the clarity of the example flow: Observable and MyActions.
* Add 'imports' property to Testing module configuration for further completeness.
* Showcase an example of how Actions are created for the test and how they are used - the semantics of the Action names help paint a flow of the timeline and what is to be tested.


brandonroberts:
Thanks!

Issue #421, opened 2017-09-22
Entity: adapter.addOrUpdate


jobvanstiphout:
## I'm submitting a...
<pre><code>
[ ] Regression
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
Sometimes when receiving a full entity from the server, you don't know if it already exists. At the moment you need to check for that manually and choose to update instead of add.
```ts
    case user.ADD_USER: {
      const changes = action.payload.user;
      const id = changes.id;

      if (state.ids.indexOf(id) > -1) {
        return adapter.updateOne({ id, changes }, state);
      }
      
      return adapter.addOne(changes, state);
    }
```

## Expected behavior:
The adapter provides a method that adds the entity when it does not exists, but merges (or replaces as it's not a partial entity) the entity into the existing one when it does exist.
```ts
    case user.ADD_USER: {
      return adapter.addOrUpdateOne(action.payload.user, state);
    }

    case user.ADD_USERS: {
      return adapter.addOrUpdateMany(action.payload.users, state);
    }
```

Silthus:
This should be added for many entities as well: `addOrUpdateMany`.

karptonite:
This is called "Upsert" in some databases. There could also be call for "Replace", which would Add or replace rather than than Add or merge; this would be important if, for example, an entity with an optional field has that field unset (and not returned by the server).

dinvlad:
+1, though another possible workaround is to simply compose the reductions:
```typescript
case user.ADD_USER: {
  const changes = action.payload.user;
  const id = changes.id;
  state = adapter.addOne({ id, changes }, state); // safe
  return adapter.updateOne(changes, state);
}
```

brandonroberts:
Adding `upsertOne` and `upsertMany` seems reasonable.

dinvlad:
Btw, in https://github.com/ngrx/platform/blob/aa7172a7e65faca530c6d4fd1c5f14b3805d3deb/modules/entity/src/unsorted_state_adapter.ts#L71-L90 why do we call computationally expensive `indexOf` when `index` only `has` to be used if the id has changed (and the check for `index === -1` can be easily replaced with `!(update.id in state.entities)`)? It may be wise to move it inside `if (newKey !== update.id) {` to avoid possible quadratic time penalty when calling `updateOneMutably` from `updateManyMutably`.

[...]

PR #550, opened 2017-11-06
 Implement upsertOne() and upsertMany() for entity adapters

dinvlad:
These methods are used to update one ore more entities
if they already exist in the state, or to add them
if they don't.

The signature of the methods is identical to that of
the corresponding update...() methods.

The commit for this PR is 9826793, others have been merged previously.

Relevant discussion: #421

EDIT: please don't merge until we agree on the signature.


modules/entity/src/unsorted_state_adapter.ts
```diff
@@ -112,13 +112,41 @@ export function createUnsortedStateAdapter<T>(selectId: IdSelector<T>): any {
...
+    const updated: Update<T>[] = [];
+
+    for (let index in updates) {
+      const update = updates[index];
+      if (update.id in state.entities) {
+        updated.push(update);
+      } else {
+        added.push({
+          ...update.changes,
+          id: update.id,
```
dinvlad: This line is problematic: if `selectId != entity => entity.id`, this insert will fail to create an entity with the same effective id.

MikeRyanDev: Let's make this change.

brandonroberts:
@dinvlad this one will be ready to land once that changed is made and the conflicts are resolved

dinvlad:
Sorry, didn't notice @MikeRyanDev's comment from earlier. Will submit soon.

bbaia:
Need some help ?

dinvlad:
Apologies, will make it happen this weekend.

acurrieclark:
Anything I can do to assist with this?

Silthus:
What is the state of this PR? Will it come soon?

jchaganti:
Any updates on this PR?

icepeng:
Any news?

mcblum:
I certainly know everyone is busy and is working hard, but if you happen to have a rough estimate of when this may be rolled out that would be amazing. If it's coming soon I'll wait to work on my next thing, if not I'll write it by hand.

Thanks for Entities -- it's amazing!

shahafal:
another workaround until the guys here gets upsert in master:
```
function upsert(state: State, entities: Post[]): State {
    entities.forEach(entity => {
        let key = entity.id;
        if (key in state.entities) {
            state = adapter.updateOne({ id: key, changes: entity}, state);
        }
        else {
            state = adapter.addOne(entity, state);
        }
    });
    return state;
}
```
enjoy.

peterbsmith2:
ngRx is MIT license. @dinvlad may take more time to finish this. Open source is open source. So I can make a new PR for this with the requested changes.

mcblum:
@peterbsmith2 I actually forked the pull request in order to try and fix it but I couldn't figure out two things: 1) how do I push the changes back (dumb, I know) or do I push changes to my repo and then open a new pull request and 2) I was (indirectly) in dialog with @MikeRyanDev who explained that the `upsert*` methods needed to handle changes like the `update*` methods do to allow for circumstances where the app wrote a dummy `id` to the entity until save and now the `id` is actually changing after a successful save. 

I understand the idea but couldn't actually figure out what I was supposed to do :) In my branch is also a new `Upsert` type that accepts `T` rather than `Partial<t>`.

mcblum:
@shahafal thanks for that function. one thing I did that not only assists with this issue but any others that might come up is I made a `DataFacade` abstract class that can be extended to give access to a bunch of common methods that retrieve state. It gives me a way to use that facade anywhere to do things like `if exists($entity)` and then the base class method knows what to do with that. Even when `upsert` is full implemented it does seem like you can gain a lot by wrapping the entity adapter.

dinvlad:
Hi folks, sorry for being silent on this and taking a long time to re-visit it. I've actually moved away from Angular to Vue (and possibly Vuex), but if everyone is happy with the signature `(updates: { id: string | number; changes: T}[], state: R)` then I can get to it shortly.

bbaia:
There's a rebase to do first!

sandangel:
@dinvlad let's finish this off

shahafal:
@dinvlad i went over your pull request and i have a question:
in upsert we pass list of entities to be added if missing or updates if exists.
in your implementation - if an entity already exists will it get updated even if it is the same as the original element?
if the answer is yes - this raises a performance question, as entities that have no changes would get new references and reload in components with OnPush change detection strategy.

i am writing this as a comment on the pull request code too so it wont get missed.

modules/entity/spec/sorted_state_adapter.spec.ts
```diff
@@ -231,4 +231,72 @@ describe('Sorted State Adapter', () => {
...
+          ...firstChange,
+        },
+        [AClockworkOrange.id]: {
+          ...AClockworkOrange,
+          ...secondChange,
+        },
+      },
+    });
+  });
 });
```
shahafal: in upsert we pass list of entities to be added if missing or updates if exists.
in your implementation - if an entity already exists will it get updated even if it is the same as the original element?
if the answer is yes - this raises a performance question, as entities that have no changes would get new references and reload in components with OnPush change detection strategy.

anyway whatever the answer is - perhaps a test that clarifies it would be a good idea:
so if an existing entity gets an update with no changes -
a) if the functionality is to recreate it the test will show that a new entity was created.
b) if the functionality is to ignore it the test will show that no new entity was created.

MikeRyanDev:
I'm going to go ahead and close this. @dinvlad or anyone else is welcome to open up a new PR!

dinvlad:
Sounds good @MikeRyanDev, @sandangel is more motivated and his PR LGTM.

Issue #487, opened 2017-10-16
Store devtools show initialstate of a lazy loaded slice of state instead of actual state.


diegomaninetti:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[*] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
When a new reducer get loaded with StoreModule.forFeature(), checking the initial state (corresponding to @ngrx/store/init action), the state contains all the initialized states after first loading of the feature reducer.

## Expected behavior:
Devtools should show the correct state relative to the selected action in devtools app.

## Minimal reproduction of the problem with instructions:
https://github.com/ngrx/platform/tree/master/example-app

- Activate devtools in browser
- Reload page at http://localhost:4200/#/asdasd url (BooksModule not loaded).
- Click on @ngrx/store/init in devtools and see the state (no "books" property).
- Navigate to home (click on "Take me home")
- Click on @ngrx/store/init in devtools and see the state ("books" state populated).

(see attached images)

## Version of affected browser(s),operating system(s), npm, node and ngrx:
Chrome Version 61.0.3163.100, windows 10, ngrx/* 4.0.x, devtools 2.15.1

![before](https://user-images.githubusercontent.com/2275525/31607359-507ac6d4-b26c-11e7-8360-e4e89140a599.png)

![after](https://user-images.githubusercontent.com/2275525/31607388-628e31da-b26c-11e7-8f51-242a232948cf.png)

phillipzada:
Hi @diegomaninetti, 
The reason the BooksModule/State is loading is if you look at the history, when you click Take me home it is navigating to the books module/feature and being immediately redirected back to the login module/feature by the Guard which checks if your logged in. Also its worth noting with Feature States they are not unloaded when you go to another section. Does this solve your issue?

diegomaninetti:
Thank's @phillipzada 

My issue regards the "regression" to the state history BEFORE the module loading. 
The "books" slice is correctly loaded, but if I select an action dispatched before the module was loaded (e.g. @ngrx/store/init) I would like to see the state of the store as it were **when the action was dispatched** and not as when the first initial state was created.

The problem is the difference between the state shown in the first image and the one shown in the second image: they should be identical regardless of the actions issued in the meantime.

phillipzada:
[...]

PR #570, opened 2017-11-16
fix(StoreDevtools): Only recompute current state when reducers are updated

### LINKED ISSUE #229: Devtools seems to cause issues with actions dispatching additional times
## I'm submitting a...

<pre><code>
[X] Bug report  
[ ] Feature request
[ ] Support request
</code></pre>

I'm running into an issue with the devtools where sometimes (but not always), an action might be dispatched once, but hit the corresponding reducer many (8-10) times.
No minimal implementation, but I can say that it happens fairly often (at least 1 time in 3). By putting debugging into the ngrx store code, I've confirmed that these actions aren't coming into the stores's dispatch or next methods, nor are they showing up in the devtools lists of actions. But this is having real effects--not only are the reducer functions run, but they are causing changes to the store that then trigger additional (real) actions (because of subscriptions).

## Expected behavior:
with devtools off, I don't see this problem. I expect that devtools will not cause additional actions to be dispatched.

This I can't provide. I can't even guarantee it will happen every time in my application, which has quite a few moving parts. I do, however, have a "working" example of the issue locally, and am more than happy to work with anyone who has a better understanding of devtools to try to figure this out. Find me on Gitter!

ngrx 4.02. , Chrome 59 with the latest redux-devtools. MacOS 10.12.2.

brandonroberts:
Closes #229, #487

### DIFF
--- a/modules/store-devtools/spec/store.spec.ts
+++ b/modules/store-devtools/spec/store.spec.ts
@@ -259,7 +259,7 @@ describe('Store Devtools', () => {
       expect(getState()).toBe(2);
     });
 
-    it('should replace the reducer', () => {
+    it('should replace the reducer and preserve previous states', () => {
       store.dispatch({ type: 'INCREMENT' });
       store.dispatch({ type: 'DECREMENT' });
       store.dispatch({ type: 'INCREMENT' });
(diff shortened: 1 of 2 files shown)


phillipzada:
Hey gents, do you have a release timeline for this fix please?

brandonroberts:
We're going to try to wait for the 2 pending PRs for `@ngrx/entity` to land before doing the next release.

Issue #573, opened 2017-11-16
StoreRouterConnectingModule.forRoot does not support config object


fredgate:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
When I add StoreRouterConnectingModule of ngrx/router-store in my module using forRoot method instead of direct type, the options passed to the function as an object are ignored. If I pass the options as a function factory, the options are used correctly.
The interest of using the forRoot method instead of the class directly, is to change the name of the router state in the store. But actually it does not work, and so the time travel debugging in developper tools does not change URL and components are not replaced in the router outlet.

## Expected behavior:
The state key passed in the config options should be used, so the time travel debugging should work in developer tools.

## Minimal reproduction of the problem with instructions:

```
import * as fromRouter from '@ngrx/router-store';
export const reducers = {
    routes: fromRouter.routerReducer,
    ...
};
```
```
@NgModule({
  ...
  imports: [
    ...
    StoreRouterConnectingModule.forRoot({stateKey: 'routes'})    // KO
    StoreRouterConnectingModule.forRoot(() => <StoreRouterConfig>{stateKey: 'routes'})    // OK
  ]
  ...
})
```

## Version of affected browser(s),operating system(s), npm, node and ngrx:
ngrx 4.1.1

## Other information:
I located the problem, and I am preparing a pull request.

brandonroberts:
Closed via #575

PR #575, opened 2017-11-17
fix(RouterStore): support objects for options of StoreRouterConnectingModule

fredgate:
Fix issue #573

### DIFF
--- a/modules/router-store/src/router_store_module.ts
+++ b/modules/router-store/src/router_store_module.ts
@@ -124,11 +124,15 @@ export const ROUTER_CONFIG = new InjectionToken(
 );
 export const DEFAULT_ROUTER_FEATURENAME = 'routerReducer';
 
-export function _createDefaultRouterConfig(config: any): StoreRouterConfig {
-  let _config = {};
+export function _createDefaultRouterConfig(
+  config: StoreRouterConfig | StoreRouterConfigFunction
+): StoreRouterConfig {
+  let _config: StoreRouterConfig;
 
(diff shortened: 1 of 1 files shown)


modules/router-store/src/router_store_module.ts
```diff
@@ -125,10 +125,12 @@ export const ROUTER_CONFIG = new InjectionToken(
 export const DEFAULT_ROUTER_FEATURENAME = 'routerReducer';
 
 export function _createDefaultRouterConfig(config: any): StoreRouterConfig {
```
brandonroberts: Change `config` argument type to `StoreRouterConfig | StoreRouterConfigFunction`

modules/router-store/src/router_store_module.ts
```diff
@@ -125,10 +125,12 @@ export const ROUTER_CONFIG = new InjectionToken(
...
-  let _config = {};
+  let _config: any;
```
brandonroberts: Change `_config` to `StoreRouterConfig`

brandonroberts:
Thanks @fredgate for the PR! I left you some feedback.

modules/router-store/src/router_store_module.ts
fredgate: I agree

modules/router-store/src/router_store_module.ts
fredgate: It makes sense

brandonroberts:
Thanks!

Issue #382, opened 2017-09-12
Router Store does not save url parameters


jobvanstiphout:
## I'm submitting a...
<pre><code>
[ ] Regression
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
URL parameters are not stored in the router state snapshot.

## Expected behavior:
They should be.

## Minimal reproduction of the problem with instructions:
This demo has a link 'Test parameter', which navigates to a route using a url parameter.
It also shows the current router state, which is created using a custom serializer.

It should display the url parameter, but only shows the url itself.

http://plnkr.co/edit/rgNb1U72etGOZmTQP8ey?p=preview

## Version of affected browser(s),operating system(s), npm, node and ngrx:
Latest ngrx.

brandonroberts:
Your example is incorrect. Query params are bound as an input to a `routerLink`. If you want route specific parameters, you have to traverse the route tree to get the params from the current `ActivatedRoute`.

Here is a working example: http://plnkr.co/edit/l0O9AE?p=preview

jobvanstiphout:
@brandonroberts Ahh thanks so much for clearing that up! I was indeed talking about route specific parameters and it really wasn't clear I'd need to traverse the tree for those. Might be nice to add your solution to the serializer doc!

EdoB:
Just copying Brandon solution for future reference:
```javascript
export class CustomRouterStateSerializer implements RouterStateSerializer<RouterStateUrl> {
  serialize(routerState: RouterStateSnapshot): RouterStateUrl {
    const { url } = routerState;
    const queryParams = routerState.root.queryParams;

    let route = routerState.root;
    while (route.firstChild) {
      route = route.firstChild;
    }

    const params = route.params;
    return { url, params, queryParams };
  }
}
```

PR #589, opened 2017-11-21
chore(RouterStore): add documentation on custom RouterStateSerializer

bcabanes:
Hi!

I was facing the same issue as describe [here](https://github.com/ngrx/platform/issues/382)(getting `params` object from the url) and spent some time finding this thread. I found that it could be useful to add the information to the documentation.

Have a great day!


brandonroberts:
Thanks!

PR #596, opened 2017-11-23
Fix of EntityState ids definition

### LINKED ISSUE #595: ngrx/entity - wrong definition of ids in EntityState interface
## I'm submitting a...

<pre><code>
[ ] Bug report  
[ ] Feature request
</code></pre>

indexOf() of EntityState ids array cause an error:

> error TS2349: Cannot invoke an expression whose type lacks a call signature. Type '((searchElement: number, fromIndex?: number) => number) | ((searchElement: string, fromIndex?: nu...' has no compatible call signatures.
On indexOf ids array in EntityState.

## Expected behavior:

Should not throw an error.

```typescript
interface ProcessListEntityState extends EntityState<IInstitutionProcessList> {
  selectedId: number;
  selectedProcess: IInstitutionProcess;
}

export interface FeatureState {
  processList: ProcessListEntityState;
}

export function reducers(state: FeatureState = initialState, action: InstitutionActions.All) {
  case InstitutionActions.STORE_PROCESS_LIST:
    if (state.processList.ids.indexOf(processList.id) !== -1) {
        ....
      }
}
```

Statement `state.processList.ids.indexOf(processList.id)` cause error.

ngrx/entity: 4.1.1, TypeScript: 2.4.1, node: 8.9.1, yarn: 1.3.2, Linux Mint 18.2 Sonya.

yp2:
EntityState ids definition changed from `string[] | number[]` to `Array<string|number>` - fixes #595

### DIFF
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -39,7 +39,7 @@ export type UpdateNum<T> = {
 export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
 export interface EntityState<T> {
-  ids: string[] | number[];
+  ids: Array<string|number>;
   entities: Dictionary<T>;
 }
 
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -66,7 +66,7 @@ export interface EntityStateAdapter<T> {
 }
 
 export type EntitySelectors<T, V> = {
-  selectIds: (state: V) => string[] | number[];
+  selectIds: (state: V) => Array<string|number>;
   selectEntities: (state: V) => Dictionary<T>;
   selectAll: (state: V) => T[];
   selectTotal: (state: V) => number;


brandonroberts:
Thanks for the PR @yp2 ! Will you add some tests to verify the selector types are correct?

yp2:
I will try to to do this ASAP.

MattiJarvinen-BA:
Doesn't this change allow arrays containing numbers and strings instead of arrays containing only strings or numbers?

yp2:
@MattiJarvinen-BA Yes. But on prevoiuse one you have to do something like this:
```
...
(Array<number>state.processList.ids).indexOf(processList.id) !== -1
...
```
App was failed to compile without this casting.
Maybe this is a bigger problem then only a types definition? Should ids of entities be only one type eg. string or number? 

I would prefer strings.

yp2:
Maybe correct type should be only string[] from `ids` in `EntityState` and `selectedIds` in `EntitySelectors`?

brandonroberts:
When entity was initially implemented, the only type for ids were strings but it was requested that it be more flexible to allow numbers also. Casting might be the option for now.

brandonroberts:
Closing this one for now. If it becomes a larger issue, we'll revisit.

kyranjamie:
Would it be possible to address this issue with an additional type argument for `EntityState`?

```ts
export interface EntityState<T, I = string | number> {
  ids: I[];
  entities: Dictionary<T>;
}
```

TeoTN:
@brandonroberts Why did that get closed? It practially renders adapters unusable

msegmx:
Would you guys please reevaluate your decision and reopen this issue ?

@kyranjamie 's solution seems perfect, no ?

MattiJarvinen-BA:
@msegmx just read the actual implementation.

Ids is either a string array or number array.
```ts
export interface EntityState<T> {
  ids: string[] | number[];
  entities: Dictionary<T>;
}
```
https://github.com/ngrx/platform/blob/master/modules/entity/src/models.ts#L35-L38

msegmx:
I get a "Cannot invoke an expression whose type lacks a call signature" error. To solve this I have to typecast the ids array every time.
That wouldn't be the case if I could tell typescript what type my ids are : 

`    (state: UsersState) => (<number[]>state.ids).filter(id => id === ...)`

vs

`    (state: UsersState) => state.ids.filter(id => id === ...)`

Issue #597, opened 2017-11-24
example-app: Update actions types to use enum instead of const


BeastCode:
## I'm submitting a...
<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
Actions types are currently declared as const.

## Expected behavior:
Suggestion, update example app to use enum instead of const for actions types

## Minimal reproduction of the problem with instructions:

Current:
export const OPEN_SIDENAV = '[Layout] Open Sidenav';
export const CLOSE_SIDENAV = '[Layout] Close Sidenav';

Suggestion:
export enum LayoutActionTypes {
  OpenSidenav = '[Layout] Open Sidenav',
  CloseSidenav = '[Layout] Close Sidenav',
}

## Other information:

I would like to help document best practices and conventions for NgRX code and turn the standards into code snippets for Angular.

jotatoledo:
I remember that this was the implemented approach on the old ngrx/store example and afterwards got refactored. Maybe check the reason for it in thw repo

BeastCode:
One difference between the old commit and my pr is that I’m using String Enums, which became available in TS2.4: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html

brandonroberts:
Fixed via #598

PR #598, opened 2017-11-24
example-app: Update actions types to use enum instead of const

BeastCode:
Suggestion, update actions types to use enum instead of const in example-app. See Issue #597 .

I've only updated two files as an example, but can update the rest if approved. Let me know what you think.

### DIFF
--- a/example-app/app/auth/actions/auth.ts
+++ b/example-app/app/auth/actions/auth.ts
@@ -1,39 +1,41 @@
 import { Action } from '@ngrx/store';
 import { User, Authenticate } from '../models/user';
 
-export const LOGIN = '[Auth] Login';
-export const LOGOUT = '[Auth] Logout';
-export const LOGIN_SUCCESS = '[Auth] Login Success';
-export const LOGIN_FAILURE = '[Auth] Login Failure';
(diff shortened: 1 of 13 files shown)


brandonroberts:
@BeastCode seems reasonable

sharikovvladislav:
Notice https://github.com/ngrx/platform/issues/597#issuecomment-347787793.

It was changed in that commit: https://github.com/ngrx/example-app/commit/d7547f282cd3f22a1ec9e03f07e27365d5242bdb#diff-e5d3a5a3504ff9ea5eaaf397b67fbd90

BeastCode:
@sharikovvladislav One difference between the old commit and my pr is that I’m using String Enums, which became available in TS2.4: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html

brandonroberts:
@BeastCode are you still going to work on this PR or can this be closed?

BeastCode:
Been away from the computer (vacation), PR is coming soon.

brandonroberts:
:+1:

BeastCode:
Updated PR with the rest of the files. Ready for review.

brandonroberts:
Thanks!

PR #602, opened 2017-11-25
chore(Example): Add auth effects marble tests

bcabanes:
Hi!

This PR adds some marble tests for _authEffects_ in the example application.

Feel free to ask any change, available to do so!

### DIFF
--- a/example-app/app/auth/effects/auth.effects.spec.ts
+++ b/example-app/app/auth/effects/auth.effects.spec.ts
@@ -7,7 +7,13 @@ import { Router } from '@angular/router';
 
 import { AuthEffects } from './auth.effects';
 import { AuthService } from '../services/auth.service';
-import { Login, LoginFailure, LoginSuccess } from '../actions/auth';
+import {
+  Login,
+  LoginFailure,
+  LoginRedirect,
+  LoginSuccess,
+  Logout,
+} from '../actions/auth';
 import { Authenticate, User } from '../models/user';
 
 export class TestActions extends Actions {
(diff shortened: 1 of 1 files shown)


example-app/app/auth/effects/auth.effects.spec.ts
```diff
@@ -0,0 +1,82 @@
+import { TestBed } from '@angular/core/testing';
+import { Actions } from '@ngrx/effects';
+import { RouterTestingModule } from '@angular/router/testing';
+import { cold, hot } from 'jasmine-marbles';
+import { Observable } from 'rxjs/Observable';
+import 'rxjs/add/observable/empty';
```
brandonroberts: Use `import { empty } from 'rxjs/observable/empty';` instead of patching the Observable

bcabanes:
@brandonroberts I added the tests and fix the comments!

example-app/app/auth/effects/auth.effects.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      actions$.stream = hot('-a---', { a: action });
+      const response = cold('-#|', {}, error);
+      const expected = cold('--b', { b: completion });
+      authService.login = jest.fn(() => response);
+
+      expect(effects.login$).toBeObservable(expected);
+    });
+  });
+
```
brandonroberts: The two test below don't seem correct. You're not dispatching any actions to trigger login success and `router.navigate` should be called with the string wrapped in an array (e.g. `['/']`)

bcabanes: Thank you for the comment!
I dispatch now the action and have verified that `router.navigate` method was indeed called with the right arguments (using `router.navigate.calls.allArgs()`). So no _false positive_ anymore 🙂

brandonroberts:
Thanks!

PR #607, opened 2017-11-27
fix(prettier) ignore node_modules if glob path begins with dot

Fost:
Its depends from 
https://github.com/isaacs/node-glob/issues/309

If we use glob path begin ./ , then we include 'node_modules' folders (exists in modules packages).
This affects only OS - such as Linux, MacOS.


brandonroberts:
Thanks!

PR #617, opened 2017-11-29
chore(docs): remove `--` from yarn run

phra:
the `--` to escape arguments is not needed anymore. ref -> https://yarnpkg.com/lang/en/docs/cli/run/#toc-yarn-run-script


brandonroberts:
This is for the latest version of yarn, so the minimum version in the `engines` section of the `package.json` needs to be updated also along with the scripts.

phra:
@brandonroberts it will be deprecated, so it's breaking anyway in the end.

brandonroberts:
@phra yes, but only changing the instructions is going to lead to an error unless you have the latest version. To prevent that, the requirements need to be updated so we have a baseline.

phra:
@brandonroberts this is not valid just for the latest version of yarn.. the deprecation message states:

```
yarn run v1.3.2
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
```

btw i've updated the mininum version required for yarn from 1.1.0 -> 1.3.2 and removed the deprecation warning from the circleci build.

brandonroberts:
@phra thanks! Will you update the scripts in the `package.json` that have the dashes? Then this will be ready to land

phra:
done!

brandonroberts:
Thanks!

Issue #184, opened 2017-07-25
Bug: EffectsModule, StoreRouterConnectingModule and StoreDevtoolsModule need to be imported only after StoreModule


loiane:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
</code></pre>

## What is the current behavior?

EffectsModule, StoreRouterConnectingModule and StoreDevtoolsModule need to be imported only after StoreModule, otherwise it does not work.

## Expected behavior:

The order of the imports describe above in the app.module or feature module should not matter.

## Minimal reproduction of the problem with instructions:

In the app.module or feature module, import EffectsModule or StoreRouterConnectingModule or StoreDevtoolsModule before StoreModule. Effects, storeRouter and devTools will not work.

## Version of affected browser(s),operating system(s), npm, node and ngrx:

ngrx v4

## Other information:

Issue opened due PR #182 

:heartpulse: ngrx? Please consider supporting our collective: 👉  [donate](https://opencollective.com/ngrx/donate)

lucastheisen:
@loiane , what was your symptoms for this problem, and how did you figure out it was related to import order?  After I updated, i just get a `modules[moduleId] is undefined` message in my console, but am having trouble determining if this is the same problem...  My `app.module` actually imports something that imports `StoreModule` so i am having trouble determining how to change the order, and whether or not it would be effective as a temporary workaround.

loiane:
@lucastheisen the first one I tried was StoreDevtoolsModule. It did not do anything and I did not get any errors. So I tried to change the order and it worked. Then, I tried EffectsModule. I got the no provider for Actions error, but it was because I forgot to add the Effects forRoot import. After fixing this error, it was also not working, so I decided to try again to change the import order and it worked. 
The current import order I'm using is the following:

```js
StoreModule.forRoot(reducers),
EffectsModule.forRoot([]),
StoreRouterConnectingModule,
!environment.production ? StoreDevtoolsModule.instrument({ maxAge: 50 }) : []
```

And I'm doing the same in my feature modules for the store and effects modules.

schmkr:
Maybe related: I was just setting up my first Effect and ran into the error ` No provider for Actions`. For me this was caused by a wrong path in the `import` statement for `EffectsModule` that was automatically generated by WebStorm. 
[...]

Issue #428, opened 2017-09-26
ngrx/effects breaks uglifyjs in Angular CLI with the use of es2015


remmeier:
## I'm submitting a...
<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
Angular CLI production build fails:

```
ERROR in vendor.8ab1d7bd4e428012688c.bundle.js from UglifyJs
Invalid assignment [C:/projects/asap/asap-resource-browser/asap-resource-browser-demo-frontend/node_modules/@ngrx/effects/src/effects_metadata.js:17,0][vendor.8ab1d7bd4e428012688c.bundle.js:19127,57]
```

The issue lies in 

https://github.com/ngrx/platform/blob/master/modules/effects/tsconfig-build.json

making use of target=es2015

See https://github.com/angular/angular-cli/issues/5730 for more information.

## Expected behavior:

make use of target=es5 for compilation

## Version of affected browser(s),operating system(s), npm, node and ngrx:

windows, @ngrx/effects@4.0.5, angular cli 1.4.3, angular  4.4.3.

MikeRyanDev:
We will need a reproduction for this. @ngrx/effects is published in a variety of formats, including ES2015, ES5 with ES2015 modules, and an ES5 UMD build. You can explore the build artifacts in our nightly repositories and verify that the published `package.json` points to the correct distribution formats. https://github.com/ngrx/effects-builds

RobotnickIsrael:
I have it too with ngrx4. Works on ng serve but fails on ng build.
I think it started after I switched to angular cli 1.3.2 from 1.1.
My Angular version is 4.2.2 and @ngrx is 4.0.0 (and @ngrx/effects 4.0.1)

Swtiching to ES6 on my tsconfig.app.json doesn't help, just multiply the issue.

Any ideas on this one? I researched the web about it and I didn't find any real solution
(seems like it's webpack2 issue with vendors using ES6 while I target ES5.
I recently installed angular-materials as well)

> ERROR in vendor.58a358ad5ad796dccd3b.bundle.js from UglifyJs
> Unexpected token operator «=», expected punc «,» [C:/Projects/Torch10/torch10-web/node_modules/@ngrx/store/src/utils.js:6,0][vendor.58a358
> ad5ad796dccd3b.bundle.js:112548,48]

remmeier:
[...]

PR #627, opened 2017-12-05
fix(*): package code is bundled twice from `src` and `es5` file

alan-agius4:
At the moment both the `src` and `FESM` are being bundled, at least when using Angular 4.

Remove `JavaScript` from `src` folder as this is not required by the consumers and in some cases with `Angular 4.x.x` and `@ngtools/webpack` it causes the `src` to be bundled togather with the `FESM`.

With Angular 4 and Webpack the javascript in `src` is being used rather than the FESM.

Relates: https://github.com/ngrx/platform/issues/184#issuecomment-319649909

Somewhat this is an urgent fix as currently we are blocked by this.

### DIFF
--- a/build/builder.ts
+++ b/build/builder.ts
@@ -9,6 +9,7 @@ export default createBuilder([
   ['Creating UMD Bundles', tasks.createUmdBundles],
   ['Renaming package entry files', tasks.renamePackageEntryFiles],
   ['Cleaning TypeScript files', tasks.cleanTypeScriptFiles],
+  ['Cleaning JavaScript files', tasks.cleanJavaScriptFiles],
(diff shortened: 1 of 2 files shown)


brandonroberts:
We publish multiple formats as [stated here ](https://github.com/ngrx/platform/issues/428#issuecomment-332200889). The bundler is responsible for consuming the correct bundle.

alan-agius4:
But what’s the point of bundling files which shouldn’t be used?

Also, this is causing me problems and aparently even some other folks have
it.

brandonroberts:
Open an issue with a reproduction first so we can see if a fix is indeed needed on our end

Issue #629, opened 2017-12-05
Packages are being bundled twice


alan-agius4:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
When using @angular v4, @ngrx v4 and have the following config in tsconfig.json:
```json
"paths": {
  "*": [
    "./node_modules/*"
  ]
}
```
webpack will generate incorrect output because it bundles @ngrx/store twice in the dist folder (correct ES5 bundle and the ES6 code from src folder).

Open the generated app.js and search for class Store and Store.prototype. You will see that the same code is bundled twice.

Then the app will fail with this error message:
```
Unhandled Promise rejection: No provider for Store! ; Zone: ; Task: Promise.then ; Value: Error: No provider for Store!
````
Reason: the class will be registered to the DI but EffectsModule will try to get the Store by using the ES5 function as the token.

The issue does not happen when using Angular 5 even if paths config is specified in tsconfig.

## Expected behavior:
The FESM5 contents are bundled.

## Minimal reproduction of the problem with instructions:
https://github.com/peterbakonyi05/angular4-ngrx4-webpack-bundling-issue

**Read the readme for further explanation how to get the problem**

## Other information:
Angular 4.XX
NGRX 4.X.X

alan-agius4:
@brandonroberts here's the issue mentioned earlier.

PR #630, opened 2017-12-05
fix(*): packages are being bundled twice

alan-agius4:
Remove `JavaScript` from `src` folder as this is not required by the consumers with `Angular 4.x.x`, `@ngtools/webpack` and `paths` in `tsconfig.json` it causes the `src` to be bundled togather with the `FESM`.

Closes: #629

### DIFF
--- a/build/builder.ts
+++ b/build/builder.ts
@@ -9,6 +9,7 @@ export default createBuilder([
   ['Creating UMD Bundles', tasks.createUmdBundles],
   ['Renaming package entry files', tasks.renamePackageEntryFiles],
   ['Cleaning TypeScript files', tasks.cleanTypeScriptFiles],
+  ['Cleaning JavaScript files', tasks.cleanJavaScriptFiles],
   ['Removing remaining sourcemap files', tasks.removeRemainingSourceMapFiles],
   ['Copying type definition files', tasks.copyTypeDefinitionFiles],
   ['Minifying UMD bundles', tasks.minifyUmdBundles],
--- a/build/tasks.ts
+++ b/build/tasks.ts
@@ -123,6 +123,21 @@ export async function cleanTypeScriptFiles(config: Config) {
 }
 
 /**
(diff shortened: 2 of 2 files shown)


brandonroberts:
I'm still not inclined to merge this as seems like a configuration issue with the bundler. I've used NGRX V4 with Angular 4 and webpack wth no issues. Maybe @TheLarkInn can weigh in?

alan-agius4:
The problem is when you use NGRX 4, Angular 4 and Paths options in TSConfig.

NGC 4 reaolved the paths of the source files next to the .d.ts files
instead of using the package.json if paths options is specified in the ts
config and for some reason.

This is no longer a problem in NG5 because Angular change how files are
resolved, as they now rely on TypeScript to do the resolution.

alan-agius4:
The problem is when you use NGRX 4, Angular 4 and Paths options in TSConfig.

NGC 4 reaolved the paths of the source files next to the .d.ts files instead of using the package.json if paths options is specified in the ts config.

This is no longer a problem in NG5 because Angular change how files are resolved, as they now rely on TypeScript to do the resolution.

It’s not the bundler that is resolving the files incorrectly but rather NGC, that said you are still shipping js files which in theory shouldn’t be present in the distributable as consumers shouldn’t be using, so why not delete them really? Will even make the final package smaller when downloading in npm.

So really this PR, fixes a bug, removes uneeded files from the distributable, and align your final package format with Google’s ‘Angular Package Format V 4.0’.

brandonroberts:
Good points, but I'm not going to introduce what could be a breaking change for many existing users to handle this edge case. We'll look at changing this in the next major release.

brandonroberts:
It _could_ be a breaking change, mainly for consumers who are reaching into the source files but you shouldn't be doing that anyway. If it affects someone using the nightlies, they will chime in. Thanks for your patience.

alan-agius4:
Thank you

PR #635, opened 2017-12-07
feat(ErrorReporter): expose ErrorReporter as a ErrorHandler for ngrx

### LINKED ISSUE #626: Effect error logging should be customizable or use ErrorHandler
## I'm submitting a...
<pre><code>
[ ] Bug report  
[X] Feature request
</code></pre>

Errors are sent to the internal `ErrorReporter` which logs to the console

## Expected behavior:
Errors should be logged to `ErrorHandler` (from `@angular/core`) or `ErrorReporter` should be public and customizable/overridable to allow doing so (or both).

4+ ?

## Other information:
Personally I'd think just sending all errors to `ErrorHandler` would be best and the `ErrorReporter` can  be removed, but maybe others depend on the console logging from `ErrorReporter`? Maybe just making `ErrorReporter` public/overridable is a good start without any breaking changes?

jbedard:
Fixes #626 

Could potentially change:
* What (if any) extra error information is exposed. Right now this information is quite specific to dumping it to the console. I think for any non-console error reporting this information (such as the instance of `Observable` and `Notification`) wouldn't be very useful.
* How that extra information is exposed. Right now it is direct properties on the `Error`, typed using interfaces extending `Error`.

In the `EffectError` case it would be nice to expose the real `Notification.error` better so custom `ErrorHandler`s will (by default) be handling the real underling error that was thrown.  Again... if we weren't building the custom `EffectError` for the console logging purposes I think just invoking `ErrorHandler.handleError(Notification.error)` would be best.

Or as an alternative (which I keep mentioning, sorry :P), avoid expanding the public API and just delete this custom error handler: https://github.com/jbedard/platform/commit/41bdf03f216fb9b55690f5fb1498d0c9429dcad6

### DIFF
--- a/modules/effects/spec/effect_sources.spec.ts
+++ b/modules/effects/spec/effect_sources.spec.ts
@@ -32,7 +32,7 @@ describe('EffectSources', () => {
     mockErrorReporter = TestBed.get(ErrorReporter);
     effectSources = TestBed.get(EffectSources);
 
-    spyOn(mockErrorReporter, 'report');
+    spyOn(mockErrorReporter, 'handleError');
   });
 
   it('should have an "addEffects" method to push new source instances', () => {
@@ -103,7 +103,7 @@ describe('EffectSources', () => {
 
       toActions(sources$).subscribe();
 
-      expect(mockErrorReporter.report).toHaveBeenCalled();
+      expect(mockErrorReporter.handleError).toHaveBeenCalled();
     });
 
     it('should not complete the group if just one effect completes', () => {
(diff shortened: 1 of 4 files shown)


brandonroberts:
LGTM @MikeRyanDev

MikeRyanDev:
By the way, for NgRx 5 I'm totally cool just using `ErrorHandler` instead of our own reporter.

jbedard:
Looks like master is now for NgRx 5. Shall I open a PR replacing `ErrorReporter` with `ErrorHandler`? Is there going to be another v4 patch that will include this PR?

brandonroberts:
Yes, open a PR to replace the `ErrorReporter`. We may do another 4.x release if necessary. There aren't going to be many breaking changes aside from bumping the minimum RxJS version and removing deprecated code.

jbedard:
Will do, thanks for the info 👍

PR #648, opened 2017-12-15
chore(Example): use rxjs lettable aperators

bcabanes:
Hi!

Starting in version 5.5 `rxjs` has shipped "lettable operators", which can be accessed in `rxjs/operators`. These are meant to be a better approach for pulling in just the operators you need than the "patch" operators found in `rxjs/add/operator/*`.

This PR convert the **example app** to "lettable operators".

Have a great day!


brandonroberts:
Fixed via #655

PR #666, opened 2017-12-22
Update instrument initialization options

NightBrownie:
As `StoreDevtoolsModule.instrument()` parameter type has changed in latest version of `@ngrx/store-devtools` - `name` option is no more applicable.


brandonroberts:
The `name` option in `@ngrx/store-devtools` has not been removed. You my be referring to an older version before this feature was introduced.

PR #667, opened 2017-12-23
feat(ErrorHandler): use the angular ErrorHandler for reporting errors

### LINKED ISSUE #626: Effect error logging should be customizable or use ErrorHandler

jbedard:
Fixes #626

This is a v5 alternative to #635. Can potentially try to keep `InvalidActionError` and `EffectError` but I'm not sure if that is a standard thing to do with `ErrorHandler`?

BREAKING CHANGE:

Errors are now reported to the @angular/core ErrorHandler.
Errors are no longer reported to the console.


brandonroberts:
The ErrorHandler reports to the console by default, so it wouldn't be a breaking change.

jbedard:
It is a breaking change if #635 gets released in v4. Shall I remove the breaking change from the commit message though?

brandonroberts:
@jbedard The breaking change should remain, but the format should be altered so it shows in the changelog correctly..

```
BREAKING CHANGE:

BEFORE:

AFTER:
```

brandonroberts:
@jbedard this one is ready to land once the commit message is updated.

jbedard:
I've updated the commit message, and rebased. Let me know if the commit message is ok...

brandonroberts:
LGTM

PR #685, opened 2018-01-04
feat(schematics): alias for container

peterbsmith2:


brandonroberts:
Go ahead and add one for an `action` (a) and the `store` (st)

peterbsmith2:
done

PR #696, opened 2018-01-06
chore(docs): make store README more friendly by uniqueifying variable…

### LINKED ISSUE #694: @ngrx/store README improvement
https://github.com/ngrx/platform/blob/master/docs/store/README.md

Issue: overuse of the name 'counter' in different context.

When the reducer is registered in the NgModule, it is given the label '__counter__'.

In the sample code of the component, the name __counter__ is used inside interface AppState, then in the constructor this is passed to the Store.

So counter is now tired to a variable and as a key to the reducer?

```js
constructor(private store: Store<AppState>)
```

__counter__ is used again as a class variable

```js
counter: Observable<number>;
```

Same name, differing type, different context, it is confusing what is going on for a new reader, how things are related.

Then there is a call to select, more counter.

```js
this.counter = store.pipe(select('counter'));
```

Please consider changing the name and also adding improved description of what is taking place. It may be super obvious to you, but not new readers.

peterbsmith2:
… names (resolves #694)

### DIFF
--- a/docs/store/README.md
+++ b/docs/store/README.md
@@ -64,7 +64,7 @@ import { counterReducer } from './counter';
 @NgModule({
   imports: [
     BrowserModule,
-    StoreModule.forRoot({ counter: counterReducer })
+    StoreModule.forRoot({ count: counterReducer })
   ]
 })
 export class AppModule {}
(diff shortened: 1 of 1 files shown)


docs/store/README.md
```diff
@@ -79,24 +79,24 @@ import { Observable } from 'rxjs/Observable';
...
     <button (click)="reset()">Reset Counter</button>
   `
 })
 export class MyAppComponent {
-  counter: Observable<number>;
+  displayCount: Observable<number>;
 
   constructor(private store: Store<AppState>) {
-    this.counter = store.pipe(select('counter'));
+    this.displayCount = store.pipe(select('count'));
```
brandonroberts: Let's denote this as an Observable by convention also by changing it to `count$`

peterbsmith2:
done

brandonroberts:
Thanks!

PR #738, opened 2018-01-22
fix(Effects): Remove toPayload utility function

brandonroberts:
BREAKING CHANGE: toPayload function is no longer exported

BEFORE:
import { toPayload } from '@ngrx/effects';

actions$.ofType('SOME_ACTION').map(toPayload);

AFTER:

actions$.ofType('SOME_ACTION').map((action: SomeActionWithPayload) => action.payload)


MikeRyanDev:
@karptonite Major release is imminent. Other than bumping the minimum version of RxJS and removing `toPayload` it should be a very straightforward upgrade.

PR #768, opened 2018-01-29
Add missing comma in example in README.md

igghera:
If you copy/pasted that code it would raise a static error for that missing comma


brandonroberts:
Thanks!

PR #780, opened 2018-02-04
feat(entity): introduce upsert methods

sandangel:
This PR continue some works of @dinvlad , include some fix:

- change function return type from boolean to DidMutate
- return correct DidMutate based on update or add

Refs: #550
Closes: #421

### DIFF
--- a/modules/entity/spec/sorted_state_adapter.spec.ts
+++ b/modules/entity/spec/sorted_state_adapter.spec.ts
@@ -1,4 +1,4 @@
-import { EntityStateAdapter, EntityState } from '../src/models';
+import { EntityStateAdapter, EntityState, Update } from '../src/models';
 import { createEntityAdapter } from '../src/create_adapter';
 import {
   BookModel,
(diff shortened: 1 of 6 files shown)


dinvlad:
Thanks @sandangel for picking up the baton - apologies again for not finding time for it. Your PR looks great!

shahafal:
@sandangel 
hi i'll write here what i wrote in dinvlad pr, as the pr was closed before anyone could address it.

as opposed to the simple update case, where perhaps we pass an object id and a change and can assume there is a change,
in upsert we'll porobably pass a list of full-blown objects to either be added if missing or updated if exists.
in such case, we might add objects that exists and do not need any update because they are exactly the same as they were before.

i went over your code, and just like in dinvlad's pr -
in the current implementation what would happen is that the objects will be updated, thus assigned a new reference. this would cause them, given angular's change detection on-push strategy, to reload into the page.

do you think this can be addressed?

sandangel:
hi @shahafal 
What you want to address does not relate to my PR, it relates to update methods. upsert methods is just detect whether id in entities and call addMany, updateMany accordingly.

shahafal:
@sandangel 
you are right that this is a performance issue that actually effects the update method,
however,
when you do a regular update you specify a change you want to make, which means you are more likely to pass a real change and not the same data.

and this brings me to why i think it is related to your pr too -
due to the fact that you'd expect upsert to get list of full blown entities that might not necessarily contain changes in them

Issue #781, opened 2018-02-04
null value is sometimes added to entity ids by @ngrx/entity


livthomas:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
When I add entities to the store using `adapter.addAll()` or `adapter.addMany` methods, `null` value is sometimes stored in `ids` array. If I then try to get entities using some selector, I get a function among those entities. This causes a lot of bugs in my application unless I always filter the returned entities.

## Expected behavior:
There should not be any unwanted values inserted into the store.

## Minimal reproduction of the problem with instructions:
I have not been able to reproduce this problem by unit tests in this project. However, I can always reproduce it in my application. I have debugged it for a while and discovered that [this loop](https://github.com/ngrx/platform/blob/master/modules/entity/src/unsorted_state_adapter.ts#L28) not only goes through entity keys but also iterates over `Array` prototype method names and as the result `null` key is pushed to `ids` array [on this line](https://github.com/ngrx/platform/blob/master/modules/entity/src/unsorted_state_adapter.ts#L18).

## Version of affected browser(s),operating system(s), npm, node and ngrx:
Chrome 64, Fedora 27, NPM 5.6.0, Node 8.9.4, NgRx 5.0.1

DanielKucal:
Was it solved? Because I'm still experiencing the issue:
![image](https://user-images.githubusercontent.com/8938562/67741901-c902ac00-fa1a-11e9-89cf-b94ae9af9526.png)

timdeschryver:
@DanielKucal feel free to open an issue with a reproduction if you're experiencing issues.
AFAIK, this is resolved.

PR #782, opened 2018-02-04
fix(Entity): Do not add Array.prototype properties to store

livthomas:
I have tried to fix issue #781 and verified the problem is really no longer there by building my application (where I can always reproduce the problem) with a version of `@ngrx/entity` containing these changes.

### DIFF
--- a/modules/entity/spec/unsorted_state_adapter.spec.ts
+++ b/modules/entity/spec/unsorted_state_adapter.spec.ts
@@ -11,6 +11,20 @@ describe('Unsorted State Adapter', () => {
   let adapter: EntityStateAdapter<BookModel>;
   let state: EntityState<BookModel>;
 
+  beforeAll(() => {
+    Object.defineProperty(Array.prototype, 'unwantedField', {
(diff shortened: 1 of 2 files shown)


MikeRyanDev:
It sounds like this is hard to reproduce. Any chance you can isolate it into a unit test that fails on current version but passes with this fix?

livthomas:
I simply cannot reproduce it in the unit tests. But I think I have found the root cause. My project uses `core-js` library which adds some additional `Array.prototype` methods and the `for..in` loop in `@ngrx/entity` code then iterates over them (as well as over the indices).

In general, iterating over an array with `for..in` loop is a very bad idea and should be avoided. See the relevant sections of [Speaking JavaScript](http://speakingjs.com/es5/ch13.html#for-in) book or [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/statements/for...in) documentation.

livthomas:
@MikeRyanDev, I have added a reproducer here as you suggested.

Issue #604, opened 2017-11-26
action-/stateSanitizer configuration does not make it into the extension.


timblakely:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?

`actionSanitizer` and `stateSanitizer`, if specified in the config, are not called.

## Other information:

#544 added support for `actionSanitizer` and `stateSanitizer`, but unfortunately they never make it into the extension. `StoreDevtools` uses `notify` to send updates to the extension, which in turn calls `devtoolsExtension.send` with `null, state, this.config, this.instanceId`. The config object does have the state-/actionSanitizers properties, but the `sendMessage` function in the extension [does not reference them nor passes the config into the extension](https://github.com/zalmoxisus/redux-devtools-extension/blob/58f6bd3c0d2fcb0ed831d7761d3db3c6030f415e/src/app/api/index.js#L84), and are subsequently ignored.

timblakely:
Note: This also blocks the ability to dispatch an `Action` whose payload is/contains an `ActivatedRoute` while using `@ngrx/store-devtools` (or any other payload with circular references).

roopkt:
Hi @timblakely 
I will start working on this issue.

roopkt:
Hi @brandonroberts  I just committed the fix and tested it. Action/State sanitizers are getting called now.

kgkma:
Hey @timblakely! I've also created #795 to address this.

> Note: This also blocks the ability to dispatch an Action whose payload is/contains an ActivatedRoute while using @ngrx/store-devtools (or any other payload with circular references).

I'm not sure how to test that this problem is resolved, but please feel free to try it out.

roopkt:
Hi @kgkma I think you are correct. The way you did the fix makes more sense.

brandonroberts:
Fixed via #795

Issue #494, opened 2017-10-17
Add support for actionSanitiser in @ngrx/store-devtools


aboyton:
## I'm submitting a...

<pre><code>
[ ] Regression (a behaviour that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behaviour?
`redux-devtools-extension` has support for having an `actionSanitizer` (see https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#actionsanitizer--statesanitizer).

## Expected behaviour:
I'd like be able to use this with @ngrx/store-devtools if possible, but I can't seem to work out how this can be done.

roopkt:
@brandonroberts  I checked in the support for both state and action sanitizer. Not sure how to create pull request.

PR #786, opened 2018-02-05
added action action and state sanitizer #494

roopkt:
Hi @brandonroberts 
I have checked in the fix for #604 and tested action and state sanitizers are getting called please review the code.


roopkt:
I will update my branch code and check in again therefore closing this pull request.

Issue #753, opened 2018-01-25
ofType pipable operator does not work with tsconfig "strict": true


seritools:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
Typescript throws the following error using ofType in a pipe when working in strict mode:

```
Argument of type '(source$: Actions<Action>) => Actions<Action>' is not assignable to parameter of type 'UnaryFunction<Observable<Action>, Observable<Action>>'. Types of parameters 'source$' and 'source' are incompatible. Type 'Observable<Action>' is not assignable to type 'Actions<Action>'. Property 'ofType' is missing in type 'Observable<Action>'.
```

Since typescript still compiles with these errors, the resulting javascript still works fine.

## Expected behavior:
No error.

## Minimal reproduction of the problem with instructions:

The [sample in the docs](https://github.com/ngrx/platform/blob/master/docs/router-store/api.md#effects) exhibits this problem.

## Version of affected browser(s),operating system(s), npm, node and ngrx:

```
Angular CLI: 1.6.5
Node: 9.3.0
OS: win32 x64
Angular: 5.2.2
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cdk: 5.1.0
@angular/cli: 1.6.5
@angular/material: 5.1.0
@angular-devkit/build-optimizer: 0.0.42
@angular-devkit/core: 0.0.29
@angular-devkit/schematics: 0.0.52
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.5
@schematics/angular: 0.1.17
typescript: 2.6.2
webpack: 3.10.0
```

ngrx v5.0.0, v5.0.1

abdulhaq-e:
I confirm. I'm not sure but I thing it's because of TS 2.6.2, the library needs to adapt to the strict checking introudiced in the new version.

sandangel:
you can implement a custom ofType operator as described in this: https://github.com/ngrx/platform/issues/709#issuecomment-362912807

PR #789, opened 2018-02-05
fix(Effects): make ofType strict

tdeschryver:
Closes: #753 

Based on [https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md#build-your-own-operators-easily](https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md#build-your-own-operators-easily).


MikeRyanDev:
Thanks!

PR #790, opened 2018-02-06
fix(StoreDevtools): Do not send full liftedState on simple actions.

timblakely:
A number of users have reported significant slowdown when using store-devtools. Slowdowns were more noticeable with larger states and larger values of maxAge. 
- ngrx/platform#97
- zalmoxisus/redux-devtools-extension#314
- ngrx/store-devtools#45
- ngrx/store-log-monitor#18

It appears that @ngrx/store-devtools currently sends the full lifted state to the extension, which requires serializing not just the current state but all previous states that have not been committed. With this change, only the current action and state snapshots are serialized during normal app actions. When app state is recomputed, a full liftedState update is sent. 

Note: Also requires optionally propagating maxAge into ReduxDevtoolsExtension.connect.


MikeRyanDev:
🔥 

Thank you for the PR! I'll get this reviewed and tested quickly.

PR #794, opened 2018-02-06
fix(RouterStore): type mismatch

### LINKED ISSUE #709: Type mismatch with angular 5.2 and typescript 2.6.2
## I'm submitting a...

<pre><code>
[ ] Bug report  
[ ] Feature request
</code></pre>

consider the following snippet

```ts
export interface RouterStateUrl {
  url: string;
  queryParams: Params;
  params: Params;
}

export interface RootState {
  router: RouterReducerState<RouterStateUrl>;
}

export const reducers: ActionReducerMap<RootState> = {
  router: routerReducer
};
```

tsconfig.json:

```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "outDir": "./dist/out-tsc",
    "sourceMap": true,
    "declaration": false,
    "moduleResolution": "node",
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "skipLibCheck": true,
    "noUnusedParameters": true,
    "noUnusedLocals": true,
    "strict": true,
    "target": "es5",
    "typeRoots": [
      "node_modules/@types"
    ],
    "lib": [
      "es2016",
      "dom"
    ]
  }
}
```
> Do note that `strict: true`

It worked with typescript 2.5.3, but after updating to typescript 2.6.2 (which is now supported in angular 5.2.0), compiler has show an error and project can not be successfully built anymore

```sh
Error:(22, 14) TS2322: Type '{ router: <T = RouterStateSnapshot>(state: RouterReducerState<T>, action: RouterAction<any, Route...' is not assignable to type 'ActionReducerMap<RootState, Action>'.
  Types of property 'router' are incompatible.
    Type '<T = RouterStateSnapshot>(state: RouterReducerState<T>, action: RouterAction<any, RouterStateSnap...' is not assignable to type 'ActionReducer<RouterReducerState<RouterStateUrl>, Action>'.
      Types of parameters 'state' and 'state' are incompatible.
        Type 'RouterReducerState<RouterStateUrl> | undefined' is not assignable to type 'RouterReducerState<RouterStateUrl>'.
          Type 'undefined' is not assignable to type 'RouterReducerState<RouterStateUrl>'.
```

## Expected behavior:

App continue to work with typescript 2.6.2

clone this repo [ngrx-bug-report](https://github.com/sandangel/ngrx-bug-report) and serve --aot

`git clone https://github.com/sandangel/ngrx-bug-report.git`

`ng serve --aot`

Chrome 63, Ubuntu 16.04, npm 5.6, node 8.9.4, ngrx 4.1.1

tdeschryver:
Had to create `Action`s instead of types to make `ActionReducerMap` happy.

Closes #709

### DIFF
--- a/example-app/app/core/reducers/layout.ts
+++ b/example-app/app/core/reducers/layout.ts
@@ -8,9 +8,9 @@ const initialState: State = {
   showSidenav: false,
 };
 
-export function reducer(
+export function reducer<T extends LayoutActions>(
   state: State = initialState,
-  action: LayoutActions
+  action: T
 ): State {
   switch (action.type) {
     case LayoutActionTypes.CloseSidenav:
--- a/example-app/app/reducers/index.ts
+++ b/example-app/app/reducers/index.ts
@@ -4,6 +4,7 @@ import {
   createFeatureSelector,
   ActionReducer,
   MetaReducer,
+  Action,
 } from '@ngrx/store';
 import { environment } from '../../environments/environment';
 import { RouterStateUrl } from '../shared/utils';
@@ -39,14 +40,14 @@ export interface State {
  * These reducer functions are called with each dispatched action
  * and the current or initial state and return a new immutable state.
  */
-export const reducers: ActionReducerMap<State> = {
+export const reducers: ActionReducerMap<State, any> = {
   layout: fromLayout.reducer,
   router: fromRouter.routerReducer,
 };
 
 // console.log all actions
 export function logger(reducer: ActionReducer<State>): ActionReducer<State> {
-  return function(state: State, action: any): State {
+  return function(state: State | undefined, action: any): State {
     console.log('state', state);
     console.log('action', action);
 
--- a/example-app/app/reducers/index.ts
+++ b/example-app/app/reducers/index.ts
@@ -40,7 +40,7 @@ export interface State {
  * These reducer functions are called with each dispatched action
  * and the current or initial state and return a new immutable state.
  */
-export const reducers: ActionReducerMap<State, any> = {
+export const reducers: ActionReducerMap<State> = {
   layout: fromLayout.reducer,
   router: fromRouter.routerReducer,
 };
--- a/modules/router-store/src/router_store_module.ts
+++ b/modules/router-store/src/router_store_module.ts
@@ -34,7 +34,8 @@ export type RouterNavigationPayload<T> = {
  * An action dispatched when the router navigates.
  */
 
-export class RouterNavigationAction<T = RouterStateSnapshot> implements Action {
+export abstract class RouterNavigationAction<T = RouterStateSnapshot>
+  implements Action {
   readonly type = ROUTER_NAVIGATION;
   constructor(public payload: RouterNavigationPayload<T>) {}
 }
@@ -55,7 +56,8 @@ export type RouterCancelPayload<T, V> = {
 /**
  * An action dispatched when the router cancel navigation.
  */
-export class RouterCancelAction<T, V = RouterStateSnapshot> implements Action {
+export abstract class RouterCancelAction<T, V = RouterStateSnapshot>
+  implements Action {
   readonly type = ROUTER_CANCEL;
   constructor(public payload: RouterCancelPayload<T, V>) {}
 }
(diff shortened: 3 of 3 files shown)


modules/router-store/src/router_store_module.ts
```diff
@@ -96,7 +96,7 @@ export type RouterReducerState<T = RouterStateSnapshot> = {
 };
 
 export function routerReducer<T = RouterStateSnapshot>(
-  state: RouterReducerState<T>,
+  state: RouterReducerState<T> | undefined,
```
tdeschryver: Had to use `undefined` because the signature for an `ActionReducer`  is `(state: T | undefined, action: V)`

example-app/app/reducers/index.ts
```diff
@@ -39,14 +40,14 @@ export interface State {
  * These reducer functions are called with each dispatched action
  * and the current or initial state and return a new immutable state.
  */
-export const reducers: ActionReducerMap<State> = {
+export const reducers: ActionReducerMap<State, any> = {
```
tdeschryver: Not quite happy with `any` but I think this is the only way to do it.
Feedback welcome!

modules/router-store/src/router_store_module.ts
```diff
@@ -33,11 +33,11 @@ export type RouterNavigationPayload<T> = {
...
  */
-export type RouterNavigationAction<T = RouterStateSnapshot> = {
-  type: typeof ROUTER_NAVIGATION;
-  payload: RouterNavigationPayload<T>;
-};
 
+export class RouterNavigationAction<T = RouterStateSnapshot> implements Action {
+  readonly type = ROUTER_NAVIGATION;
+  constructor(public payload: RouterNavigationPayload<T>) {}
+}
```
sandangel: after typescript 2.8 landed which support conditional type, I think ngrx team will stick with interface based Action

example-app/app/reducers/index.ts
sandangel: Although this will fix router store but users still have to manually add ‘any’ by themself to by pass typescript check in their store. we need a better approach

tdeschryver: What about the last commit? I think we're getting closer...

modules/router-store/src/router_store_module.ts
tdeschryver: Oh I did not know about this!
But since we're still in 2.7 this is the only way to do it.

example-app/app/core/reducers/layout.ts
```diff
@@ -8,9 +8,9 @@ const initialState: State = {
   showSidenav: false,
 };
 
-export function reducer(
```
brandonroberts: Why is this change needed for router-store?

modules/router-store/src/router_store_module.ts
brandonroberts: If we need to use classes here, make them abstract

brandonroberts:
Does this PR only cover support for router-store with TS 2.6 and Angular 5.2?

example-app/app/core/reducers/layout.ts
tdeschryver: Otherwise the example app will have the following error when we create the `ActionReducerMap`:
`
Type '{ layout: (state: State | undefined, action: LayoutActions) => State; router: <V extends RouterAc...' is not assignable to type 'ActionReducerMap<State, Action>'.
`

tdeschryver:
I'm thinking TS 2.4.2 and up but  I'm not 100% sure.
I'll run some tests tomorrow.

tdeschryver:
So, if I'm not mistaken it should work with TS 2.4.2 and up.
But I do get compile errors on another part (even on the master branch) when I upgraded TS to 2.6.1 and 2.7.1.

````
TSError: ⨯ Unable to compile TypeScript
modules\store\src\reducer_manager.ts (68,5): Type 'Partial<ActionReducerMap<any, any>>' is not assignable to type 'ActionReducerMap<any, any>'.
  Index signatures are incompatible.
    Type 'ActionReducer<any, any> | undefined' is not assignable to type 'ActionReducer<any, any>'.
      Type 'undefined' is not assignable to type 'ActionReducer<any, any>'.
```

PR #544, opened 2017-11-03
Added support for actionSanitiser and stateSanitiser in @ngrx/store-devtools

roopkt:
This is the fix to pass state and action sanitizer to redux dev tool extensions for #494.

PR #795, opened 2018-02-07
feat(StoreDevtools): Add support for ActionSanitizer and StateSanitizer

kgkma:
Continuing the work from #544 and requested by #604, this PR actually adds the sanitization functionality to @ngrx/store-devtools.

Unfortunately, simply passing `actionSanitizer` and `stateSanitizer` to the Redux Devtools Extension does not work, since the `send(action, state)` function exposed by its API does not sanitize anything if the `action` passed in is `null`, which is the current behaviour. As a result, it is necessary to sanitize within ngrx Devtools itself, and then pass that into the extension.

This implementation deliberately avoids mutating the lifted state's `actionsByIds` and `computedStates` properties, but instead creates a sanitized copy of them. This prevents the possibility of a sanitized action from changing the store state if, for example, its sanitized type coincided with a real action's type within the application.

### Testing
In `app.module.ts`:
```typescript
  actionSanitizer: (action: Action, id: number) => {
    return { type: 'SANITIZED_ACTION' };
  },
  stateSanitizer: (state: any, index: number) => {
    return { ...state, extra: 'SANITIZED_STATE' };
  }
})
```
Open Chrome, confirm that it appears, while app and time travel seem to work fine.

### DIFF
--- a/modules/store-devtools/src/devtools.ts
+++ b/modules/store-devtools/src/devtools.ts
@@ -21,7 +21,12 @@ import { queue } from 'rxjs/scheduler/queue';
 
 import { DevtoolsExtension } from './extension';
 import { liftAction, unliftAction, unliftState, applyOperators } from './utils';
-import { liftReducerWith, liftInitialState, LiftedState } from './reducer';
+import {
+  liftReducerWith,
+  liftInitialState,
+  LiftedState,
+  ComputedState,
+} from './reducer';
 import * as Actions from './actions';
 import { StoreDevtoolsConfig, STORE_DEVTOOLS_CONFIG } from './config';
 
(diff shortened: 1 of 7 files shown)


modules/store-devtools/src/devtools.ts
```diff
@@ -97,6 +106,26 @@ export class StoreDevtools implements Observer<any> {
     this.state = state$;
   }
 
+  /**
+   * Restructures the lifted state passed in to prepare for sending to the
+   * Redux Devtools Extension
+   */
+  getSanitizedState(state: LiftedState, config: Partial<StoreDevtoolsConfig>) {
```
brandonroberts: Pass the `stateSanitizer` as an argument instead of the config.

kgkma:
@brandonroberts Thank you! Fixed as requested.

brandonroberts:
Thanks!

aboyton:
Any chance of a patch release so I can use this?

PR #798, opened 2018-02-08
test(Entity): Add Array.prototype problem reproducer

livthomas:
I have added a reproducer for bug #781. Once the tests fail, you can clearly see where the problem is.

I have not added it to [my previous pull request](https://github.com/ngrx/platform/pull/782) with the fix because I am not sure if you want to have something like this in the unit tests.


MikeRyanDev:
Thanks! Now if you'll roll this up into the other MR I'll get it merged in!

PR #805, opened 2018-02-11
fix(Entity): Avoid for..in iteration in sorted state adapter

livthomas:
I have just noticed that the changes in commit f871540f67191195a347297f1208e8f821d511cc introduced to the sorted state adapter the same problem I have already fixed in the unsorted state adapter. So I'm adding the reproducer to those tests as well.

Just one question: Where can we expect these fixes to be released? Is there going to be version `5.0.2` anytime soon?


MikeRyanDev:
There will probably be a 5.1.0 at some point. Not sure when, but if you need the fixes soon then I would recommend using the nightlies for now.

Thanks for these fixes!

PR #824, opened 2018-02-15
build: Introduce Bazel

alexeagle:
Can now `bazel build modules/...`


kylecordes:
For the usual reasons around eventually running into a case insensitive file system, would it be wise to name the other BUILD file BUILD.bazel?  Or can we be confident that there would never be any possible reason for a build directory to be needed in that spot?

(Scars over here from non-Bazel cases where a project started on Linux, ended up on Mac or Windows, with great misery around cased name conflicts...)

alexeagle:
@kylecordes I did have to name the root one `BUILD.bazel` to avoid conflict. I was playing with the idea of the shorter filenames here since it's green-field, but you're probably right that it's just not worth the hassle

MikeRyanDev:
@kylecordes I'm going to merge as-is. I don't foresee a conflict ever arising.

PR #826, opened 2018-02-16
chore: Remove Codegen from Platform

MikeRyanDev:


kuncevic:
Hi guys can you rovide some backgorund on that PR?
Does that means that all the functionality that Codegen supposed to provide now will be handled by https://github.com/ngrx/platform/tree/next/modules/schematics ?

PR #836, opened 2018-02-17
build: Add Store unit tests to Bazel

MikeRyanDev:


BUILD.bazel
```diff
@@ -1,17 +1,25 @@
...
-genrule(
-    name = "true",
-    outs = ["true.sh"],
-    cmd = "echo true > $@",
-)
-
-sh_test(
-    name = "tautology_test",
-    srcs = [":true.sh"],
+filegroup(
```
MikeRyanDev: It isn't clear to me if I should be using the dependencies installed from running Yarn at the root for tests or if I should do something similar to the `ngrx_compile_time_deps` for tests.

modules/store/spec/edge.spec.ts
```diff
@@ -21,11 +21,17 @@ describe('ngRx Store', () => {
     let store: Store<TodoAppSchema>;
 
     beforeEach(() => {
-      const injector = createInjector(
-        StoreModule.forRoot<TodoAppSchema>({ todos, todoCount } as any)
-      );
+      try {
+        TestBed.resetTestingModule();
```
MikeRyanDev: When running these tests via `jasmine_node_test` the `TestBed` is not automatically reset.

tools/defaults.bzl
```diff
@@ -1,9 +1,27 @@
...
+load("@build_bazel_rules_nodejs//:defs.bzl", _jasmine_node_test = "jasmine_node_test")
 
 def ts_library(tsconfig = None, node_modules = None, **kwargs):
   if not tsconfig:
     tsconfig = "//:tsconfig.json"
   if not node_modules:
     node_modules = "@ngrx_compiletime_deps//:node_modules"
   _ts_library(tsconfig = tsconfig, node_modules = node_modules, **kwargs)
+
+def ts_test_library(node_modules = None, **kwargs):
```
MikeRyanDev: Is this an OK pattern of defining my own rules that just set defaults for existing rules similar to what you've done with `ts_library`? Or should I be more explicit in `BUILD` files?

tools/defaults.bzl
```diff
@@ -1,9 +1,27 @@
...
+  if not node_modules:
+    node_modules = "//:ngrx_test_dependencies"
+  ts_library(node_modules = node_modules, testonly = 1, **kwargs)
+
+def jasmine_node_test(node_modules = None, bootstrap = None, deps = [], **kwargs):
+  if not node_modules:
+    node_modules = "//:ngrx_test_dependencies"
+  if not bootstrap:
+    bootstrap = ["ngrx/tools/testing/bootstrap_node_tests.js"]
```
MikeRyanDev: It didn't seem like I could pass a label here.

BUILD.bazel
alexeagle: You have both options.
In the case of tests, your users probably don't expect to build/run these from source. At the same time, your users *would* download and install the `ngrx_compiletime_deps` so it's not nice to add test-only dependencies there.
I think I'd probably re-use the `node_modules` directory installed by yarn (note you can use `bazel run @yarn//:yarn` to do that with Bazel's version of node and yarn).

modules/store/spec/edge.spec.ts
alexeagle: I don't know much about TestBed, but I would expect `jasmine_node_test` just evaluates all scripts, and wouldn't do anything special to reset test state - that would be the job of whatever framework you are testing?

alexeagle: Can you help me understand what layer was doing this job before?

tools/defaults.bzl
alexeagle: It ought to allow `"$(location //my:label)"` but doesn't yet :(
https://github.com/bazelbuild/rules_nodejs/issues/32

tools/defaults.bzl
alexeagle: This is a macro (you can tell because you didn't call the `rule` factory function) and macros can be the gateway to a bad time. In Google, there are a lot of macros that produce surprising combinations of targets, and make the BUILD files look very different from the result of `bazel query --output=build`.
This usage is okay though, matching the criteria:
- it should only wrap one rule, not multiple
- it should just set defaults

This looks clever and reasonable to me :)

tools/testing/bootstrap_node_tests.ts
```diff
@@ -0,0 +1,32 @@
...
+// otherwise the patch will not work.
+//    This is needed since Bazel creates a new instance of jasmine and it's ENV and we want to make
+//    sure it gets the same one.
+jasmineCore.boot = function() {
+  return patchedJasmine;
+};
+
+require('zone.js/dist/jasmine-patch.js');
+
+TestBed.initTestEnvironment(ServerTestingModule, platformServerTesting());
```
alexeagle: maybe in this spot you could install jasmine lifecycle hooks so that you don't need to worry about resetting the TestBed and whatnot?
I've used https://jasmine.github.io/2.1/custom_reporter.html before (but it's smelly to have a reporter messing with test state... maybe there's another way?)

modules/store/spec/edge.spec.ts
MikeRyanDev: Currently Angular's core testing library owns resetting the testing module: https://github.com/angular/angular/blob/master/packages/core/testing/src/before_each.ts#L25

I'm assuming that the `beforeEach` hook it is calling belongs to a different Jasmine environment.

BUILD.bazel
MikeRyanDev: Good deal. This PR follows your suggestion and re-uses the top level `node_modules` directory.

tools/testing/bootstrap_node_tests.ts
MikeRyanDev: I got around the issue for now by overriding Angular's `TestBed.configureTestingModule` to always reset the testing module before reconfiguring it.

modules/store/spec/edge.spec.ts
MikeRyanDev: I was able to confirm that the Jasmine environment I'm working with in the bootstrap script is different from the one being used when running my tests. I wasn't able to add a custom reporter in bootstrap, for instance.

brandonroberts:
LGTM

alexeagle:
Mike, any action items for me as follow-ups to this PR? Does this all work
the way you'd expect, or should we fix something around the
bootstrap/jasmine/beforeEach ergonomics?

MikeRyanDev:
@alexeagle I do think the issue with Angular's testing package and Jasmine is a real bug. We have a temporary fix working for us right now.

alexeagle:
Okay, could you file that somewhere? I'll probably need to enlist some help
on the team to understand how this is meant to work.
/cc @vikerman

PR #838, opened 2018-02-18
chore(docs): remove typo in store README

peterbsmith2:


docs/store/README.md
```diff
@@ -96,7 +96,7 @@ export class MyAppComponent {
   count$: Observable<number>;
 
   constructor(private store: Store<AppState>) {
-    this.count$ = store.pipe(select('count'));
+    this.count$ = this.store.pipe(select('count'));
```
brandonroberts: Why the change? Either way is valid since its in the constructor.

peterbsmith2: @brandonroberts I was running examples last night and encountered an error. I may have been at less than my best. So you're right and I think this is unnecessary now.

PR #844, opened 2018-02-20
fix(StoreDevtools): Recompute state history when reducers are updated

brandonroberts:
This fixes the issue with time-traveling to previous states before the reducers were updated. Each state history item will be recomputed with the latest reducer.


MikeRyanDev:
This is the same behavior as Redux's devtools, right?

brandonroberts:
Mostly. The difference is that the history is recomputed with the `@ngrx/store/update-reducers` action instead of the original one.

PR #845, opened 2018-02-20
build: Add remaining unit tests to Bazel

MikeRyanDev:


MikeRyanDev:
This also verifies that all packages except for Schematics are now building with Bazel

PR #851, opened 2018-02-22
docs(effects): fix document issues

SangKa:
Here should be a mistake in writing. It should be `EffectsModule` rather than `NgModule`.


brandonroberts:
Thanks!

Issue #853, opened 2018-02-23
RFC: Dev Tools support for persist, lock, pause


dummdidumm:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
The redux dev tools extension has the following features which are not usable with ngrx at the moment:
- persist: If enabled, the state of the redux dev tools survive a full reload of the page. The persisted actions are fired against the initial state. This makes developing applications faster. Imagine you don't use HMR and have to fix something from the example application - every reload throws you back to the login page. With the persist feature, this would no longer be the case.
- lock/unlock: If locked, no new actions will get through to the reducer, the app is basically frozen. In combination with persist / HMR, this is a great developer experience when you want to focus on a specific feature with a specific number/order of actions.
- pause/unpause: If paused, the redux dev tools will not record the actions fired. The app will continue to work. Useful if you don't want to get spammed with lots of actions. When you unpause, the current state will be committed and recording is resumed.

Locking and pausing is explained in more details [here](https://medium.com/@zalmoxis/improve-your-development-workflow-with-redux-devtools-extension-f0379227ff83).

## Expected behavior:
Features also work with ngrx dev tools.

## Other information:
I'm willing to help with PRs.

MikeRyanDev:
I'm down with these changes for NgRx 6. I don't know enough about Redux's devtools to help implement these. Would you be able to make these changes @dummdidumm?

dummdidumm:
Sure, I'll look into it. The referenced PR is a first step to lay the groundwork for persist.

PR #855, opened 2018-02-24
State / Action sanitizer refactoring/bugfix + options fix

dummdidumm:
The new state / action sanitization feature added by #544 and #795 has a bug:

1. Start up example app and open redux dev tools
2. Open side drawer
3. Export state (file A) -> you can already see that sanitizedActionsById and actionsById doesn't line up
4. Import state (file A)
5. Export again (file B)
6. Import state (file B)
7. Error

This may now be an edge case, but, [if the persist feature would get implemented](https://github.com/ngrx/platform/issues/853) (which under the hood is basically export/import), it may occur very often.

This is because the redux devtools extension does not know about any sanitizedActionsById object and will not update that. But the ngrx devtools reducer relied on this getting updated. This PR moves the state / action sanitization out of the reducer into the DevtoolsExtension service. Now, the sanitization is done directly before passing the state updates to the redux devtools extension. This makes the code more clear and fixes that bug.

During this I noticed that the options object which is handed to the devtools extension on startup is different to the options object which is sent on a full state replacement. Fixed that, too.

No worries on the big added/removed lines count in the store.spec.ts btw, I only removed the now obsolete tests for sanitizedActionsById, but for some reason git made that into a "replaced the whole file"-commit.


modules/store-devtools/src/extension.ts
```diff
@@ -158,4 +175,65 @@ export class DevtoolsExtension {
...
+          return sanitizedActions;
+        },
+        <LiftedActions>{}
+      );
+    } else {
+      return actions;
+    }
+  }
+
+  private sanitizeAction(action: LiftedAction, actionIdx: number) {
```
brandonroberts: Pass the sanitizer as an argument instead of checking its presence from the config inside the function.

modules/store-devtools/src/extension.ts
```diff
@@ -158,4 +175,65 @@ export class DevtoolsExtension {
...
+    return this.config.actionSanitizer
+      ? {
+          ...action,
+          action: this.config.actionSanitizer(action.action, actionIdx),
+        }
+      : action;
+  }
+
+  private sanitizeStates(states: ComputedState[]) {
```
brandonroberts: Pass the sanitizer as an argument instead of checking its presence from the config inside the function.

modules/store-devtools/src/extension.ts
```diff
@@ -158,4 +175,65 @@ export class DevtoolsExtension {
...
+        state: this.sanitizeState(computedState.state, idx),
+        error: computedState.error,
+      }));
+    } else {
+      return states;
+    }
+  }
+
+  private sanitizeState(state: any, stateIdx: number) {
+    return this.config.stateSanitizer
```
brandonroberts: Pass the sanitizer as an argument instead of checking its presence from the config inside the function.

modules/store-devtools/src/extension.ts
```diff
@@ -158,4 +175,65 @@ export class DevtoolsExtension {
...
+      ? this.config.stateSanitizer(state, stateIdx)
+      : state;
+  }
+
+  private getExtensionConfig() {
+    const extensionOptions: ReduxDevtoolsExtensionConfig = {
```
brandonroberts: Pass the config as an argument instead of using it inside the function

brandonroberts:
These changes look reasonable to me. I left some comments on minor tweaks. I would also consider moving the sanitization functions to the `utils.ts` file

dummdidumm:
Thanks for the feedback! I added a commit implementing your suggestions.

brandonroberts:
Thanks!

PR #863, opened 2018-02-25
feat(entity): add entity selection option

yairtawil:
# entity selection
resovle #869 
## Ability to select entities.

`example.reducer.ts`:

```typescript
    case ActionTypes.SELECT_ONE: {
      return adapter.selectOne((<SelectOne>action).payload, state);
    }

    case ActionTypes.SELECT_MANY: {
      return adapter.selectMany((<SelectMany>action).payload, state);
    }

    case ActionTypes.SELECT_ALL: {
      return adapter.selectAll(state);
    }

    case ActionTypes.SELECT_ONLY: {
      return adapter.selectOnly((<SelectOnly>action).payload, state);
    }

    case ActionTypes.UNSELECT_ONE: {
      return adapter.unSelectOne((<UnSelectOne>action).payload, state);
    }

    case ActionTypes.UNSELECT_MANY: {
      return adapter.unSelectMany((<UnSelectMany>action).payload, state);
    }

    case ActionTypes.UNSELECT_ALL: {
      return adapter.unSelectAll(state);
    }
```

`example.actions.ts`:
```typescript

import { SelectedId, SelectedIds} from '@ngrx/entity';

export class SelectOnly implements Action {
  readonly type = ActionTypes.SELECT_ONLY;

  constructor(public payload: SelectedIds) {
  }
}

export class SelectOne implements Action {
  readonly type = ActionTypes.SELECT_ONE;

  constructor(public payload: SelectedId) {
  }
}

export class SelectMany implements Action {
  readonly type = ActionTypes.SELECT_MANY;

  constructor(public payload: SelectedIds) {
  }
}

export class SelectAll implements Action {
  readonly type = ActionTypes.SELECT_ALL;
}

export class UnSelectAll implements Action {
  readonly type = ActionTypes.UNSELECT_ALL;
}


export class UnSelectOne implements Action {
  readonly type = ActionTypes.UNSELECT_ONE;

  constructor(public payload: SelectedId) {
  }
}

export class UnSelectMany implements Action {
  readonly type = ActionTypes.UNSELECT_MANY;

  constructor(public payload: SelectedIds) {
  }
}

```

models: 
```typescript
export type SelectedId = string | number;
export type SelectedIds = SelectedId[];
```

selector:
```typescript
selectedIds$: Observable<SelectedIds> = this.store.select(selectSelectedIds)
```
example: https://github.com/yairtawil/ngrx-entity-selection
website: https://yairtawil.github.io/ngrx-entity-selection

### DIFF
--- a/example-app/app/books/reducers/__snapshots__/books.spec.ts.snap
+++ b/example-app/app/books/reducers/__snapshots__/books.spec.ts.snap
@@ -26,7 +26,7 @@ Object {
   "ids": Array [
     "1",
   ],
-  "selectedBookId": null,
+  "selectedIds": Set {},
 }
 `;
 
@@ -56,7 +56,7 @@ Object {
   "ids": Array [
     "1",
   ],
-  "selectedBookId": null,
+  "selectedIds": Set {},
 }
 `;
 
@@ -106,7 +106,7 @@ Object {
     "1",
     "222",
   ],
-  "selectedBookId": null,
+  "selectedIds": Set {},
 }
 `;
 
@@ -156,7 +156,7 @@ Object {
     "1",
     "222",
   ],
-  "selectedBookId": null,
+  "selectedIds": Set {},
 }
 `;
 
@@ -226,7 +226,7 @@ Object {
     "222",
     "333",
   ],
-  "selectedBookId": null,
+  "selectedIds": Set {},
 }
 `;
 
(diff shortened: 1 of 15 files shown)


brandonroberts:
We are not looking to add these methods to the API as entity is focused on managing model collections. In general, we don't use `Set` because its not serializable, which is one of our "core principles" of data in the store. It also causes issues with the Devtools. This would be better as a separate package of functions to manage selections. 

Thanks for the PR though!

PR #867, opened 2018-02-26
refactor(Example): Unify the form of the actions import name

SangKa:
Hi @brandonroberts 
In the example app, there are two forms for the actions import name, one is pascal and another one is lowercase, as show below:

```ts
// app.ts
import * as layout from '../actions/layout';
import * as Auth from '../../auth/actions/auth';
```
If I was a beginner of ngrx, I couldn't figure out which one was the best practices. It was a little confusing for me.

I suggest to unify the form of the import name, meanwhile I don't know which one is better.

In the [Typed Actions](https://github.com/ngrx/platform/blob/master/docs/store/actions.md#typed-actions) section of the store document, It used the pascal, as show below:

```ts
import { Store, select } from '@ngrx/store';
import { Observable } from 'rxjs/Observable';
import * as Counter from './counter.actions';

interface AppState {
  counter: number;
}

@Component({
  selector: 'my-app',
  template: `
    <button (click)="increment()">Increment</button>
    <button (click)="decrement()">Decrement</button>
    <button (click)="reset()">Reset Counter</button>
    
    <div>Current Count: {{ counter | async }}</div>
  `
})
export class MyAppComponent {
  counter: Observable<number>;

    this.counter = store.pipe(select('counter'));
  }

  increment(){
    this.store.dispatch(new Counter.Increment());
  }

  decrement(){
    this.store.dispatch(new Counter.Decrement());
  }

  reset(){
    this.store.dispatch(new Counter.Reset(3));
  }
}
```

And if we import actions and model for the same entity both, we can just use the lowercase:

```ts
// find-book-page.ts
import * as book from '../actions/book';
import { Book } from '../models/book';
```

So, I need the answer from you guys and then complete this PR that in progress.

### DIFF
--- a/example-app/app/auth/actions/auth.ts
+++ b/example-app/app/auth/actions/auth.ts
@@ -35,7 +35,7 @@ export class Logout implements Action {
   readonly type = AuthActionTypes.Logout;
 }
 
-export type AuthActions =
+export type AuthActionsUnion =
   | Login
   | LoginSuccess
   | LoginFailure
--- a/example-app/app/auth/containers/login-page.component.spec.ts
+++ b/example-app/app/auth/containers/login-page.component.spec.ts
@@ -5,7 +5,7 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { StoreModule, Store, combineReducers } from '@ngrx/store';
 import { LoginPageComponent } from './login-page.component';
 import { LoginFormComponent } from '../components/login-form.component';
-import * as Auth from '../actions/auth';
+import * as AuthActions from '../actions/auth';
 import * as fromAuth from '../reducers';
 
 describe('Login Page', () => {
@@ -57,7 +57,7 @@ describe('Login Page', () => {
 
   it('should dispatch a login event on submit', () => {
     const $event: any = {};
-    const action = new Auth.Login($event);
+    const action = new AuthActions.Login($event);
 
     instance.onSubmit($event);
 
--- a/example-app/app/auth/containers/login-page.component.ts
+++ b/example-app/app/auth/containers/login-page.component.ts
@@ -2,7 +2,7 @@ import { Component, OnInit } from '@angular/core';
 import { Store, select } from '@ngrx/store';
 import { Authenticate } from '../models/user';
 import * as fromAuth from '../reducers';
-import * as Auth from '../actions/auth';
+import * as AuthActions from '../actions/auth';
 
 @Component({
   selector: 'bc-login-page',
@@ -24,6 +24,6 @@ export class LoginPageComponent implements OnInit {
   ngOnInit() {}
 
   onSubmit($event: Authenticate) {
-    this.store.dispatch(new Auth.Login($event));
+    this.store.dispatch(new AuthActions.Login($event));
   }
 }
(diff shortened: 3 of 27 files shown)


SangKa:
Hi @brandonroberts again.
Although this is a [WIP] PR, but I need your suggestion first.

brandonroberts:
"Best practice" for naming conventions is subjective, but in general I prefer consistency with either choice. For the actions, I'd go with Title/Pascal case for the imports and suffix them with `Actions` (e.g. import * as BookActions from 'somewhere';`. In addition, I'd rename the book model to `BookModel` to also guard against wildcard imports collisions like the one above.

SangKa:
@brandonroberts 
Thanks for your answer that benefits me a lot.
I will make the consistency in this PR and I prefer the way that you handle it.

SangKa:
@brandonroberts Done

brandonroberts:
Changes LGTM, but they highlight another point of confusion. The union exported from https://github.com/ngrx/platform/blob/master/example-app/app/core/actions/layout.ts#L16 is `LayoutActions` also. What about refactoring the union exports to `All`?

SangKa:
How about this?
If need to import the union, take another import  statement and use the alias to avoid the confusion.

```ts
import * as LayoutActions from '../actions/layout';
import { LayoutActions as LayoutActionTypes } from '../actions/layout';
```

brandonroberts:
Layout actions already includes `LayoutActionTypes`

SangKa:
Generally speaking, the union only be imported by the reducer to specify the typed actions, and the import all object (eg `import * as LayoutActions`) is imported by the container component.  They are actually separate.

brandonroberts:
Yes but it's still leads to confusion. I'd rather the imports be distinct

SangKa:
I agree. Now we need a appropriate alias name for union type `LayoutActions`.
How about the `LayoutActionUnion`?

brandonroberts:
I like `LayoutActionsUnion`

SangKa:
That is better. :+1:

SangKa:
hi @brandonroberts 
Does this PR ready to be merged?

brandonroberts:
@SangKa I didn't see where you updated the Unions with the naming scheme we discussed.

SangKa:
All we discussed is for the possibility of the import confusion that doesn't appear in the example for now, and there are nothing to update.

brandonroberts:
Yes, there are no collisions currently but we if want people to use that pattern the example should reflect it

SangKa:
Ok, I got it. I will find a appropriate place to add comment for this.

SangKa:
Already add the comment for explaining, pleases review the description of the comment. Thank you. @brandonroberts

brandonroberts:
@SangKa I would like the union names to actually be changed instead of adding a note about collisions. That way there are no need for the notes.

SangKa:
You means that change [https://github.com/ngrx/platform/blob/master/example-app/app/core/actions/layout.ts#L16](https://github.com/ngrx/platform/blob/master/example-app/app/core/actions/layout.ts#L16) to like this:
`export type LayoutActionsUnion = OpenSidenav | CloseSidenav;`

I hope I got your point.

brandonroberts:
Exactly

SangKa:
Done.

brandonroberts:
Thanks!

SangKa:
My pleasure

PR #868, opened 2018-02-26
Provide ActionReducerMap<T> with every reducer map in the example-app

jtcrowson:
In the example-app, provide the ActionReducerMap<T> in the books and auth reducer maps for [added type checking](https://github.com/ngrx/platform/blob/master/docs/store/actions.md).  This makes the example-app a better example of best practices.


brandonroberts:
Thanks!

PR #879, opened 2018-03-01
docs(schematics): add alternative setup command

kiwikern:
Instead of editing the `.angular-cli.json` manually, you can run `ng set`. This is also used when setting up the `@angular/service-worker`.

### DIFF
--- a/docs/schematics/README.md
+++ b/docs/schematics/README.md
@@ -43,6 +43,11 @@ add the following to the `defaults` section in your `.angular-cli.json`.
     }
 ```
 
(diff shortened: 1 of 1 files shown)


brandonroberts:
I like this method instead. Will you update the PR to replace manually editing the `.angular-cli.json` file?

kiwikern:
Editing the `.angular-cli.json` file manually has proven to be rather difficult to me... 🤦‍♂️  Now, this should be foolproof.

brandonroberts:
Awesome. Thanks!

Issue #880, opened 2018-03-02
store should support falsey state


jbedard:
## I'm submitting a...
<pre><code>[X] Bug report  
</code></pre>

## What is the current behavior?
Feature store/reducers with a falsey state fallback to the initial state. This might only be for feature and not root reducers?

See https://github.com/ngrx/platform/blob/7f49b9f413471306d614cd5eae8b4e8d7e41f611/modules/store/src/reducer_manager.ts#L46 where `state || initialState` is used.

## Expected behavior:
Any non-undefined state value should be considered "set" and not be reset to the `initialState`.

https://github.com/ngrx/platform/blob/7f49b9f413471306d614cd5eae8b4e8d7e41f611/modules/store/src/reducer_manager.ts#L46 should be changed to `(state !== undefined) ? state : initialState`

## Minimal reproduction of the problem with instructions:
Can create one if necessary...?

## Version of affected browser(s),operating system(s), npm, node and ngrx:
I'm using 5.x

## Other information:
As an example... in my case this feature-store represents the user session. An empty store value (`undefined`) means the session state is unknown (not fetched yet), a `null` value means no session exists and a `Session` object means the user is logged in. In this case the `null` value is being converted to `undefined`.

jbedard:
Actually... should `initialState` be the initial state, or the default state? Is it correct that an `undefined` store value should always be converted to the `initialState` like the feature store does (in one case) today?

brandonroberts:
Fixed via #886

PR #886, opened 2018-03-02
fix(Store): only default to initialValue when store value is undefined

jbedard:
This assumes the store will only have an `undefined` state on initialization and sets the `initialSate` at that time. If the store state ever becomes `undefined` again in the future it will be reset to the `initialState` again.  This basically means `initialState` is more of a "default state" then "initial", but this at least improves it to be an undefined-check instead of falsey-check.  If we wanted to change this to be more strictly an "initial" state then this change would grow quite a bit.

For example if you add the following diff to the tests weird things occur:
```
@@ -74,8 +74,8 @@ describe('ngRx Store', () => {
       store = TestBed.get(Store);
       dispatcher = TestBed.get(ActionsSubject);
 
-      const actionSequence = '--a--b--c--d--e--f--g';
+      const actionSequence = '--a--b--c--d--e--f--g--h';
       const actionValues = {
         a: { type: INCREMENT },
         b: { type: 'OTHER' },
@@ -84,6 +84,7 @@ describe('ngRx Store', () => {
         e: { type: INCREMENT },
         f: { type: INCREMENT },
+        g: { type: CLEAR },   //set state to undefined
+        h: { type: 'OTHER' }, //reproduces https://github.com/ngrx/platform/issues/880
       };
       const counterSteps = hot(actionSequence, actionValues);
       counterSteps.subscribe(action => store.dispatch(action));
```

### DIFF
--- a/modules/store/spec/store.spec.ts
+++ b/modules/store/spec/store.spec.ts
@@ -68,6 +68,76 @@ describe('ngRx Store', () => {
         complete: done,
       });
     });
+
+    function testInitialState(feature?: string) {
+      store = TestBed.get(Store);
+      dispatcher = TestBed.get(ActionsSubject);
+
+      const actionSequence = '--a--b--c--d--e--f--g';
+      const stateSequence = 'i-w-----x-----y--z---';
+      const actionValues = {
+        a: { type: INCREMENT },
+        b: { type: 'OTHER' },
+        c: { type: RESET },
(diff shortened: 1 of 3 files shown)


MikeRyanDev:
Looks great! Thanks!

jbedard:
Thanks for the quick merge.

Sorry I forgot to put "fixes #880" in the commit message though. Should #880 be closed? Or leave it open regarding the initial vs default issue?

PR #887, opened 2018-03-03
refactor(Store): avoid createFeatureReducer being invoked per-action

jbedard:
I noticed this while doing #886. `createFeatureReducer` was being invoked for each action. In some cases (when `metaReducers` exist) this would invoke `compose` and create a new composed method for each action.

This also changes `createFeatureReducer` to `createFeatureReducerFactory` more like `createReducerFactory`. This could potentially be undone to simplify this PR, but I thought the end result (the two methods being more alike) was better.

However... I think the better solution would be changing the `ActionReducerFactory` `reducerMap` argument from `ActionReducerMap<T, V>` to `ActionReducerMap<T, V> | ActionReducer<T, V>`. Then this `createFeatureReducer` method can be dropped and the reducer vs map of reducers code paths would be combined into one. Things like the `initialValue` logic would no longer be in two places ([a](https://github.com/ngrx/platform/commit/8deffc3dd8dc0932efb8626fc1926ebfd0fac95b#diff-46ac3744db99c69c50de39d2dc253e37R46), [b](https://github.com/ngrx/platform/commit/8deffc3dd8dc0932efb8626fc1926ebfd0fac95b#diff-b1befd0d816408efc33272e96ede4a4aR30)). But I think such a change to `ActionReducerFactory` would be a breaking API change...?

### DIFF
--- a/modules/store/spec/utils.spec.ts
+++ b/modules/store/spec/utils.spec.ts
@@ -1,4 +1,4 @@
-import { omit, createFeatureReducer } from '../src/utils';
+import { omit, createFeatureReducerFactory } from '../src/utils';
 import { combineReducers, compose } from '@ngrx/store';
 
 describe(`Store utils`, () => {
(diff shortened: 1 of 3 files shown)


brandonroberts:
LGTM. I think keeping them separate is fine for now. The root store only supports an action reducer map

PR #890, opened 2018-03-04
feat(Effects): Add support for effects inheritance from parent classes.

mrgusev:
I was trying to implement some abstraction for effects components, but faced some confusing behaviour: inherited class could pick up parent's class `@Effect` decorators only in case of you do not have any in the child one. Also the `setEffectMetadataEntries` method is performing properly and assigns metadata to all the parent and child class properties. So, I think it's worth to add the same behaviour for `getEffectMetadataEntries`.

The solution is quite simple and straightforward and does not change existing API.

### DIFF
--- a/modules/effects/spec/effects_metadata.spec.ts
+++ b/modules/effects/spec/effects_metadata.spec.ts
@@ -35,6 +35,46 @@ describe('Effect Metadata', () => {
 
       expect(getSourceMetadata(mock)).toEqual([]);
     });
+
+    it('should get the effects metadata for class instance including metadata from parent classes', () => {
+      class AbstractFixture {
+        @Effect() a: any;
+        @Effect() b: any;
+        @Effect({ dispatch: false })
+        c: any;
+      }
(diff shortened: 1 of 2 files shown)


MikeRyanDev:
Hi @mrgusev,

I appreciate the PR however I don't think this is a change we want to make to @ngrx/effects. NgRx tends to promote composition over inheritance and we want the API to push developers towards that.

Thanks again!

robwormald:
I'll expand on this: I definitely see a whole bunch of super useful decorators emerging, *once the decorator spec* has finalized - at the moment, the semantics are so up in the air, its very difficult to guarantee compatibility on the semantics of decorators (as really, Angular has basically "invented" our own) for the moment. 

Once the spec lands, and the behavior is properly documented, then we should no longer need anything like this PR - it'll either work as the spec mandates, or not!

igor-pavlichenko:
The inheritable effects classes work as expected if one replace all decorated effects with the new createEffect function available since 8.0.0-beta.0

PR #893, opened 2018-03-06
generate named umd bundles

thelgevold:
Submitting a PR to generate named umd bundles for ngrx.

The motivation is to make ngrx bundles compatible with the Bazel dev server used in the sample here: https://github.com/alexeagle/angular-bazel-example

cc: @alexeagle


brandonroberts:
Thanks @thelgevold!

Issue #672, opened 2017-12-27
RFC: Entity: update\remove by predicate


aigosh:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?

Current behaviour does not let us update or remove many entities by predicate.

## Expected behavior:

There is often a need to update or remove by the way described.
So it would be great to have an ability to update\remove many entities this way:
```typescript
adapter.updateMany([
        {
            predicate: entity => entity.someField === someValue,
            changes: {
                someAnotherField: someAnotherValue
            }
        }])

adapter.removeMany(entity => entity.someField === someValue, predicate1, predicate2, ...)
```
Also some there are some cases that requires attention:
- there could be several conditions, depending on which could be required different changes
- there should be no breaking changes

## Other information:
Please look at the issue #671. It is tightly tied to this one, because it would be great to provide predicate and changes function at the same time.

brandonroberts:
If you want to filter based on some predicate, why not find the ids based on the predicate and use them to remove items?

aigosh:
Yes, I can do as you say, but I find this useful, to have this feature "out of the box".

tdeschryver:
I'll take this one :smile:

tdeschryver:
While implementing the `updateMany`, I came up with an edge case and would like some feedback.

Would it be clear if we let the option open to add multiple predicates at once?
If so, what is the desired behaviour when a model satisfies multiple predicates?

If it should be possible I would opt to accumulate all changes, but maybe this would raise some questions with users.
If it shouldn't be possible (which has my favor because then the intention is very clear, and I also can't think of a use case where I want to be able to update entities by using multiple predicates), we don't have this 'problem'.


For example:
```
Given state: [
  {id: 1, title 'Title 1', description: 'Description 1'}, 
  {id: 2, title 'Title 2', description: 'Description 2'}
]

adapter.updateMany([
  { predicate: book => book.id === 1, changes: { title : 'Changed title 1', description: 'Awesome description' }},
  { predicate: book => book.id < 5, changes: { title : 'Awesome title' }}
])

Output: [
  {id: 1, title 'Awesome title', description: 'Awesome description'}, 
  {id: 2, title 'Awesome title', description: 'Description 2'}
]
```

cburnicki:
[...]

PR #900, opened 2018-03-09
feat(Entity): add removeMany by predicate

timdeschryver:
This PR allows to remove entities based on a predicate.
See #672

### DIFF
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -28,6 +28,10 @@ export interface UpdateNum<T> {
 
 export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
+export type Predicate<T> = {
+  (entity: T): boolean;
+};
+
 export interface EntityState<T> {
   ids: string[] | number[];
   entities: Dictionary<T>;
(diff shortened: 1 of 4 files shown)


MikeRyanDev:
This looks great to me!

I am going to label this at v6.x. We are in the middle of adding new infrastructure to support multiple versions in the same repo. Please don't be surprised if this doesn't get merged immediately.

tdeschryver:
@MikeRyanDev Should this change (and the update many from #907) also be documented in the current docs? 
If so, I'll edit the docs with an extra `Action`.

brandonroberts:
@timdeschryver will you rebase this so it can land?

timdeschryver:
✔️

modules/entity/src/models.ts
```diff
@@ -38,6 +38,10 @@ export type UpdateNum<T> = {
 
 export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
+export type Predicate<T> = {
```
brandonroberts: Make this the same as `export type ComparerStr<T> = (a: T, b: T) => string;`

PR #901, opened 2018-03-09
fix(RouterStore): change the default serializer to work around cycles in RouterStateSnapshot

vsavkin:
RouterStateSnapshot contains cycles, which means that the default serializer doesn't work with the devtools.

This serializer implements a work around for the issue.


MikeRyanDev:
@vsavkin Does this introduce a breaking change?

brandonroberts:
Fixes included in 7917a276fa199b19d6d8f2e3d2fa1708e3481c2b

PR #903, opened 2018-03-11
fix: type mismatch when upgrade to ts2.6

sandangel:
Refs: #709

### DIFF
--- a/modules/router-store/src/router_store_module.ts
+++ b/modules/router-store/src/router_store_module.ts
@@ -96,7 +96,7 @@ export type RouterReducerState<T = RouterStateSnapshot> = {
 };
 
 export function routerReducer<T = RouterStateSnapshot>(
-  state: RouterReducerState<T>,
+  state: RouterReducerState<T> | undefined,
   action: RouterAction<any, T>
 ): RouterReducerState<T> {
   switch (action.type) {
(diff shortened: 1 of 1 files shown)


brandonroberts:
This is to support `strict: true` correct? We're already building against 2.6 in the repo. We're not going to change `Action` to `any` as we lose the type safety.

sandangel:
yes, this is for supporting strict: true.
could you suggest a better fix for map type between reducer and action type when there are multiple feature in ActionReducerMap ?
currently this won't work:

```typescript
import * as fromAuth from './auth.reducer';
import * as fromToken from './token.reducer';

export interface AuthFeatureState {
  authState: fromAuth.AuthState;
  tokenState: fromToken.TokenState;
}

export const getAuthFeatureState = createFeatureSelector<AuthFeatureState>('authFeature');

export const reducers: ActionReducerMap<AuthFeatureState, AuthActions | TokenActions> = {
  authState: fromAuth.authReducer,
  tokenState: fromToken.tokenReducer
};
```

user have manually cast ActionReducerMap<AuthFeatureState, any> for ts to compile.

sandangel:
if you not agree to change the default type variable in ActionReducerMap, I can revert it, but do you consider the fix for routerStore signature ?

brandonroberts:
The routerStore signature fix is ok. I'd rather use `ActionReducerMap<AuthFeatureState, any>` for now but those changes don't impact this repo as `strict: true` isn't enabled for the example-app

sandangel:
hi @brandonroberts , I have revert the ActionReducerMap default type, please check.

PR #907, opened 2018-03-13
feat(Entity): add updateMany by predicate

timdeschryver:
This PR allows to edit entities based on a predicate.
See #672, in the issue it was mentioned to update the entities via multiple predicates - see my comment as why I don't think thats a good idea. If needed I can implement it that way tho.

I do got one question about the implementation of the sorted state adapter. 
Why does this have an own implementation? 
In my head the sorted adapter should call the unsorted adapter and do the sorting afterwards, but I must be missing someting :sweat_smile:

### DIFF
--- a/modules/entity/src/index.ts
+++ b/modules/entity/src/index.ts
@@ -1,2 +1,9 @@
 export { createEntityAdapter } from './create_adapter';
-export { Dictionary, EntityState, EntityAdapter, Update } from './models';
+export {
+  Dictionary,
+  EntityState,
+  EntityAdapter,
+  Update,
+  EntityMap,
+  Predicate,
+} from './models';
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -30,11 +30,7 @@ export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
 export type Predicate<T> = (entity: T) => boolean;
 
-export type UpdateMap<T> = {
-  (entity: T): Partial<T> | false;
-};
-
-export type Map<T> = (entity: T) => T;
+export type EntityMap<T> = (entity: T) => T;
 
 export interface EntityState<T> {
   ids: string[] | number[];
@@ -62,10 +58,11 @@ export interface EntityStateAdapter<T> {
 
   updateOne<S extends EntityState<T>>(update: Update<T>, state: S): S;
   updateMany<S extends EntityState<T>>(updates: Update<T>[], state: S): S;
-  updateMany<S extends EntityState<T>>(map: UpdateMap<T>, state: S): S;
 
   upsertOne<S extends EntityState<T>>(entity: T, state: S): S;
   upsertMany<S extends EntityState<T>>(entities: T[], state: S): S;
+
+  map<S extends EntityState<T>>(map: EntityMap<T>, state: S): S;
 }
 
 export interface EntitySelectors<T, V> {
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -35,9 +35,7 @@ export type UpdatePredicate<T> = {
   changes: Partial<T>;
 };
 
-export type Change<T> = {
-  (entity: T): Partial<T>;
-};
+export type Map<T> = (entity: T) => T;
 
 export interface EntityState<T> {
   ids: string[] | number[];
@@ -65,7 +63,7 @@ export interface EntityStateAdapter<T> {
 
   updateOne<S extends EntityState<T>>(update: Update<T>, state: S): S;
   updateMany<S extends EntityState<T>>(updates: Update<T>[], state: S): S;
-  updateMany<S extends EntityState<T>>(change: Change<T>, state: S): S;
+  updateMany<S extends EntityState<T>>(map: Map<T>, state: S): S;
 
   upsertOne<S extends EntityState<T>>(entity: T, state: S): S;
   upsertMany<S extends EntityState<T>>(entities: T[], state: S): S;
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -30,6 +30,11 @@ export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
 export type Predicate<T> = (entity: T) => boolean;
 
+export type UpdatePredicate<T> = {
+  predicate: Predicate<T>;
+  changes: Partial<T>;
+};
+
 export interface EntityState<T> {
   ids: string[] | number[];
   entities: Dictionary<T>;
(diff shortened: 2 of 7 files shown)


MikeRyanDev:
Implementation seems reasonable. What are your opinions on this:

Instead of a predicate with a list of changes the signature is updated to use a filter/map function:

```ts
 // Name is bad but it illustrated the interface
interface Updater<T> {
  (entity: T): false | Change<T>;
}
function updateMany(updater: Updater<T>, state: EntityState<T>);
```

tdeschryver:
I don't think I get the idea.
Would this be a `map` implementation where the user implements `map`?
For example: 
```
adapter.updateMany(
      entity => entity.id.startsWith('a') ? {title: 'New title'} : false, // maybe null or an empty object instead of false?
      withMany
);
```

MikeRyanDev:
Yes, that's right.

tdeschryver:
I must say your proposal feels a bit weird at first but after seeing it in action, it is better in my opinion.
I did change the signature to return a `Partial<T>`, this brings the flexibility to do a second change.
See the unsorted state test: `book => (book.id.startsWith('a') ? firstChange : secondChange),`
Thoughts?

If someone got a better name than `Change` please feel free to speak up :smile:

modules/entity/src/models.ts
```diff
@@ -38,6 +38,10 @@ export type UpdateNum<T> = {
 
 export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
+export type Map<T> = {
```
MikeRyanDev: Recommend renaming this to something more specific. Maybe `UpdateMap`?

modules/entity/src/sorted_state_adapter.ts
```diff
@@ -72,8 +73,16 @@ export function createSortedStateAdapter<T>(selectId: any, sort: any): any {
...
 
   function updateManyMutably(updates: Update<T>[], state: R): DidMutate;
-  function updateManyMutably(updates: any[], state: any): DidMutate {
+  function updateManyMutably(map: Map<T>, state: R): DidMutate;
+  function updateManyMutably(updatesOrMap: any[] | any, state: any): DidMutate {
     const models: T[] = [];
+    const updates: Update<T>[] =
+      updatesOrMap instanceof Array
+        ? updatesOrMap
+        : state.ids.map((id: string | number) => ({
```
MikeRyanDev: I think that creating an `Update` object for each entity means even entities that shouldn't change will lose referential equality. I think this `map` should be a `reduce` instead that shrinks the `ids` array into a smaller array of `Update`s for only changed entities.

modules/entity/src/models.ts
tdeschryver: LGTM!

modules/entity/src/sorted_state_adapter.ts
```diff
@@ -72,8 +73,25 @@ export function createSortedStateAdapter<T>(selectId: any, sort: any): any {
...
+  function updateManyMutably(map: UpdateMap<T>, state: R): DidMutate;
+  function updateManyMutably(updatesOrMap: any[] | any, state: any): DidMutate {
     const models: T[] = [];
+    const updates: Update<T>[] =
+      updatesOrMap instanceof Array
+        ? updatesOrMap
+        : state.ids.reduce((changes: Update<T>[], id: string | number) => {
+            const change = updatesOrMap(state.entities[id]);
+            if (change && Object.keys(change).length > 0) {
+              return [
```
MikeRyanDev: Recommend changing this line to:
```ts
changes.push({ id, changes: change });

return changes;
```

That way we don't allocate and deallocate an array for every item we are upserting.

MikeRyanDev:
One minor change. Otherwise this is looking good to me!

modules/entity/src/sorted_state_adapter.ts
tdeschryver: Done, thanks for the feedback/reviews!

MikeRyanDev: @tdeschryver Thanks for all of the hard work! We really appreciate it! 😄

sandangel:
Hi, how about adding this feature to upsert too?

brandonroberts:
@timdeschryver will you rebase this so it can land?

timdeschryver:
✔️

MikeRyanDev:
I wonder if we could just have `entityAdapter.map(...)` instead.

timdeschryver:
That would make sense, since this is what we're actually doing.
But I think having a `updateMany` would have its value if we would change the API to the following ?

```ts
{
  predicate: entity => boolean,
  changes: Partial<Entity>
}
```

timdeschryver:
Ha, now I see we started with this a API  😄 -  https://github.com/ngrx/platform/pull/907/commits/149042b47dc55d62a0e28c092079899e23de6824

timdeschryver:
I've updated the PR to be a `map` function.
While doing this I also changed the signature to actually return an entity object.

Let me know your thoughts about this one and if I have to change something.
I could also submit a PR if you agree with https://github.com/ngrx/platform/pull/907#issuecomment-421244609.

modules/entity/src/models.ts
```diff
@@ -38,6 +38,10 @@ export type UpdateNum<T> = {
 
 export type Update<T> = UpdateStr<T> | UpdateNum<T>;
 
+export type EntityMap<T> = {
```
brandonroberts: Update type to match `export type ComparerStr<T> = (a: T, b: T) => string;`

brandonroberts:
Minor nit, otherwise changes LGTM

timdeschryver:
There we go 😄

brandonroberts:
@timdeschryver rebase pls

modules/entity/src/index.ts
```diff
@@ -1,2 +1,9 @@
 export { createEntityAdapter } from './create_adapter';
-export { Dictionary, EntityState, EntityAdapter, Update } from './models';
+export {
+  Dictionary,
+  EntityState,
+  EntityAdapter,
+  Update,
+  EntityMap,
```
timdeschryver: I did also `EntityMap` and `Predicate` here.

modules/store/src/store.ts
```diff
@@ -19,9 +19,6 @@ export class Store<T> extends Observable<T> implements Observer<Action> {
     this.source = state$;
   }
 
-  /**
```
brandonroberts: Revert this change because its in a separate PR

timdeschryver: Doh, I was using 2 VSCode instances and was wondering what was happening.
Sorry... 😅

PR #909, opened 2018-03-14
fix(Store): Handling of @ngrx/store/update-reducers Action on Startup leads to strange behaviour for initialStates

### LINKED ISSUE #906: Store: Handling of @ngrx/store/update-reducers Action on Startup leads to strange behaviour for initialStates
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Feature request
</code></pre>

I configure a object as an initialState for the store. Unfortunately some values are "overridden" on start-up of the store/app.

## Expected behavior:
All properties from the initial-state should remain untouched by any internal ngrx-action.

I think my explanation was i bit poor sofar (sorry it's hard to explain this). I think an example will show the problem:

https://stackblitz.com/edit/angular-6wbjq4

In this example everything is fine. The initial state is produced by the initState() function in app/app.state.ts. 
screenshot 2018-03-12 18 02 05

*BUT*:

Remove the key 'featureB' from this init state will "break" this behaviour. 

screenshot 2018-03-12 18 03 42

It seems to be caused by the combination function: https://github.com/ngrx/platform/blob/9b02f79f1fc3b5e4d85bf0516108f93eed655126/modules/store/src/utils.ts#L29-L44

The function expects all possible state properties to be setted. If a property is not (in initialState) set the 'hasChanged' turns true and returns the nextState. This is some kind of an intermediated state (at least while the  ' @ngrx/store/update-reducers'-Action is fired).

@ngrx/store v 5.1 + 5.2

## Other information:

This is not directly linked to the problem but another finding that confused me.
After '@ngrx/store/update-reducers' is fired all reducers are applied, even if they were already applied previously:

screenshot 2018-03-12 18 09 37

berthertogen:
- assign the new state to the overall initial state instead of returning the new state only
 - Add test to simulate the following situation:
 	Module MA and MB with initial state IA and IB in module.	
 	Overall initial state RootIAB { IB } with only state of MB being
	initiated.
	When action '@ngrx/store/update-reducers' is fired for MA the RootIAB should be merged with IA so we don't lose the IB initialization in RootIAB.
- Fixed failing test to expect 0 instead of undefined.

Closes #906

### DIFF
--- a/modules/store/spec/store.spec.ts
+++ b/modules/store/spec/store.spec.ts
@@ -296,7 +296,7 @@ describe('ngRx Store', () => {
       store.removeReducer(key);
       store.dispatch({ type: INCREMENT });
       store.take(1).subscribe(val => {
-        expect(val.counter4).toBeUndefined();
+        expect(val.counter4).toBe(0);
       });
     });
   });
--- a/modules/store/spec/utils.spec.ts
+++ b/modules/store/spec/utils.spec.ts
@@ -41,7 +41,7 @@ describe(`Store utils`, () => {
     it(`should return full initial state of the root when action '@ngrx/store/update-reducers' is running with no initial state`, () => {
       const initialStateForReducer2Only = { reducer2: { y: 'bar' } };
       const combinationForReducer1 = combineReducers(
-        { reducer1 } as any,
+        { reducer1 },
         initialStateForReducer2Only as any
       );
       const updateAction1 = { type: 'state1', payload: 'foo' };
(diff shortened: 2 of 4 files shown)


MikeRyanDev:
If I'm understanding this change correctly it won't be possible to remove state keys by removing reducers, right? I'm not sure if I want to merge in the fix if that's the reprecussion. 

The harder fix would be to change the way the `Store` bootstraps itself. Technically it could load all `forFeature()` calls that exist in the same injector and resolve them right away as part of the root module. That way you don't get `update-reducers` calls on bootstrap, only for lazy loaded modules.

Bretto:
What's the status on this ? Should we worry about all the ```@ngrx/store/update-reducers``` in the console ? Once the feature modules are loaded we don't get those actions fired multiple times only the first time. Thanks

PR #912, opened 2018-03-14
fix(Entity): change upsert signature to have entity property

### LINKED ISSUE #817: (entity) Upsert Add overrides the id property
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Feature request
</code></pre>

The `upsert` methods of the `EntityStateAdapter` override the `id` property, even if this is not the primary entity key. This only happens when adding with upsert.

## Expected behavior:

ngrx/entity should not make any changes to the added entity. It should leave the `id` property as is OR should change the property we assigned with the `selectId` function.

## Explanation / reproduction

**DEMO:** https://stackblitz.com/edit/ngrx-upsert-bug-r96kbx
(thanks to @miniplus for the StackBlitz skeleton)

Imagine an entity like a *book* whose "primary key" is not named `id` but `isbn`.
However, this entity also has an `id` property which is not the primary key.

```
export interface Book {
  isbn: string,
  title: string,
  id: string;
}
```

(We might discuss why there is an `id` property which is not the ID, but that's not the point here at all 😉 )

To make ngrx/entity use the `isbn` property as key, we set the `selectId` selector function accordingly:

```
export const adapter: EntityAdapter<Book> = createEntityAdapter<Book>({
  selectId: book => book.isbn
})
```

Now, in the reducer, we're using the new `upsertOne`/`upsertMany` methods from the StateAdapter to **add** a new entity which is not yet in the entity list:

```
const book: Book = {
    isbn: '123',
    name: 'My Book',
    id: 'book-abc'
}
const change: Update<Book> = { id: book.isbn, changes: book }
return adapter.upsertOne(change, state);
```

(Note: The `Update` object seems to be necessary here, but shouldn't, see https://github.com/ngrx/platform/issues/818 )

This leads to the following entities object in the state:

```
'123': {
    isbn: '123',
    name: 'My Book',
    id: '123' // <-- THIS CHANGED 💥
}
```

...where it actually should be

```
'123': {
    isbn: '123',
    name: 'My Book',
    id: 'book-abc'  // <-- ✅
}
```

Notice, when doing an upsert as update (i.e. for an entity that already exists), the `id` property is set correctly.

## Where's the error?

The issue might be in this line:
https://github.com/ngrx/platform/blob/master/modules/entity/src/sorted_state_adapter.ts#L125

```
for (const update of updates) {
    if (update.id in state.entities) {
        updated.push(update);
    } else {
        added.push({
            ...update.changes,
            id: update.id, // <-- overrides the id property
        });
    }
}
```

This sets/overrides the `id` property, regardless of whether this is the actual ID or not.

## Possible solution

Remove the line completely.
`update.changes` contains the complete object to be added and there is no need to set the ID.
If we want to make sure that `update.id` and the ID from the entity are the same we should check this here using the `selectId()` function.

**I'd like to submit a PR but wanted to discuss this first.**

Thanks!

## Affected versions

@ngrx/entity 5.1.0

tdeschryver:
Fixes #817.

If this is approved, I can edit the documentation if needed.

### DIFF
--- a/modules/entity/src/sorted_state_adapter.ts
+++ b/modules/entity/src/sorted_state_adapter.ts
@@ -5,6 +5,7 @@ import {
   Dictionary,
   EntityStateAdapter,
   Update,
+  Upsert,
 } from './models';
 import { createStateOperator, DidMutate } from './state_adapter';
 import { createUnsortedStateAdapter } from './unsorted_state_adapter';
@@ -106,24 +107,21 @@ export function createSortedStateAdapter<T>(selectId: any, sort: any): any {
     }
   }
 
-  function upsertOneMutably(update: Update<T>, state: R): DidMutate;
-  function upsertOneMutably(update: any, state: any): DidMutate {
-    return upsertManyMutably([update], state);
+  function upsertOneMutably(upsert: Upsert<T>, state: R): DidMutate;
+  function upsertOneMutably(upsert: any, state: any): DidMutate {
+    return upsertManyMutably([upsert], state);
   }
 
-  function upsertManyMutably(updates: Update<T>[], state: R): DidMutate;
-  function upsertManyMutably(updates: any[], state: any): DidMutate {
+  function upsertManyMutably(upserts: Upsert<T>[], state: R): DidMutate;
+  function upsertManyMutably(upserts: any[], state: any): DidMutate {
     const added: T[] = [];
     const updated: Update<T>[] = [];
 
-    for (const update of updates) {
-      if (update.id in state.entities) {
-        updated.push(update);
+    for (const { id, entity } of upserts) {
+      if (id in state.entities) {
+        updated.push({ id, changes: entity });
       } else {
-        added.push({
-          ...update.changes,
-          id: update.id,
-        });
+        added.push(entity);
       }
     }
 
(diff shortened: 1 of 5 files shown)


MikeRyanDev:
Change looks good to me.

I sort of wonder if changing the ID of entity with an upsert is going to be prevalent. The API would be a little more ergonomic if it just accepted the entity like `addOne`/`addMany`.

tdeschryver:
The API would indeed make more sense with your recommendation. 
I'll try to change it later today.

modules/entity/src/sorted_state_adapter.ts
```diff
@@ -106,24 +106,21 @@ export function createSortedStateAdapter<T>(selectId: any, sort: any): any {
...
+  function upsertManyMutably(entities: T[], state: R): DidMutate;
+  function upsertManyMutably(entities: any[], state: any): DidMutate {
     const added: T[] = [];
     const updated: Update<T>[] = [];
 
-    for (const update of updates) {
-      if (update.id in state.entities) {
-        updated.push(update);
+    for (const entity of entities) {
+      if (entity.id in state.entities) {
```
MikeRyanDev: You can't safely access `entity.id`. You will need to use the adapter's `selectId` function to get it.

tdeschryver: Right, totally forgot about it, sorry :sweat_smile: 
I think some tests on the state adapters with a different `selectId` implementation would prevent these mistakes. But I don't know if these are worth the effort? Or maybe rename `id` on the book fixture?

Issue #643, opened 2017-12-11
Using store-devtools in "logOnly" mode in production


akrueger:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Feature request
[ X ] Documentation issue or request
</code></pre>

My goal is to send error reports with a redux state history to my server as part of our logging service in production. I have no problem sending the current state at the time of the error, but what I'd really like is something that looks like the `state.json` object that the `redux-devtools-extension` allows you to download. With an object like that with a max age (say 5 to keep performance in a good spot), I could take the data from my server log and upload the state into the `redux-devtools-extension`; then replay the state back to see how users have managed to get into their error state. 

I don't need to use any of the features that can manipulate the state--JUMP, SKIP, PAUSE, COMMIT, etc--I'd want to avoid that completely in production. I have read about the extension's [production modes](https://github.com/zalmoxisus/redux-devtools-extension#14-using-in-production) that have `logOnly` as well as [this article](https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f).

I am not sure how this is best achieved in my Angular ngRx app. Is there a way for me to achieve this using `store-devtools`. Should I be using some other type of meta-reducer similar to the `enhancer`?

I saw the options for sanitizing actions and serializing state, but I'm more interested in a setting for `logOnly` and accessing the data in my error handling that could be read back into `redux-devtools-extension` for error analysis from our logs.

https://github.com/ngrx/store-devtools/issues/53#issuecomment-279043936 addresses use in production and seems to indicate that there is no `logOnly` option, but I didn't want to give up just on that assumption: 

> The devtools allow external users to look at and potentially manipulate the state of your application.

Thanks!

akrueger:
Thanks a lot!

As a follow-up to this:

Is it possible to access the `liftedState` of @ngrx/store-devtools from the client application? This lifted state seems akin to the `state.json` object that is imported/exported by Redux Dev Tools. Since I want to log that at the point of an error, I'm trying to figure out the best way of accessing that at runtime when there is an error.

brandonroberts:
@akrueger you can inject the `StoreDevtools` to access the `lifedState` property that is an Observable of the current lifted state.

akrueger:
@brandonroberts Thanks, worked perfectly!

akrueger:
I'm having some trouble figuring out the exact effects of including `logOnly: true`.

[...]

PR #927, opened 2018-03-22
fix(StoreDevtools): report errors to ErrorHandler instead of console

sylvaindumont:
store-devtools catch errors happening in reducers and log them to console

## Expected behavior:
store-devtools catch errors happening in reducers and pass them to angular ErrorHandler

As stated in #643 store-devtools should  be usable in production, but the current behavior hide reducers errors in the console instead of reporting them to ErrorHandler as #667 do

### DIFF
--- a/modules/store-devtools/src/devtools.ts
+++ b/modules/store-devtools/src/devtools.ts
@@ -1,4 +1,4 @@
-import { Injectable, Inject, OnDestroy } from '@angular/core';
+import { Injectable, Inject, OnDestroy, ErrorHandler } from '@angular/core';
 import {
   State,
   Action,
@@ -51,6 +51,7 @@ export class StoreDevtools implements Observer<any> {
     reducers$: ReducerObservable,
     extension: DevtoolsExtension,
     scannedActions: ScannedActionsSubject,
+    errorHandler: ErrorHandler,
     @Inject(INITIAL_STATE) initialState: any,
     @Inject(STORE_DEVTOOLS_CONFIG) config: StoreDevtoolsConfig
   ) {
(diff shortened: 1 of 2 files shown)


brandonroberts:
Thanks for the PR @sylvaindumont ! Will you update the commit message for the breaking change?

```
Commit message here

BREAKING CHANGE:

Description

BEFORE:

Usage before change

AFTER:

Usage after change
```

sylvaindumont:
In #667 you said
> The ErrorHandler reports to the console by default, so it wouldn't be a breaking change.

so I thought it wasn't a breaking change

If it is a breaking change is the following description accurate? 
```
BREAKING CHANGE:

Errors in reducers are no longer hidden from ErrorHandler by StoreDevtools

BEFORE:

Errors in reducers are caught by StoreDevtools and logged to the console

AFTER:

Errors in reducers are reported to ErrorHandler

```

dummdidumm:
Maybe keep the errorHandler outside of the reducer by changing the signature like this?

```
export function liftInitialState(
   initialCommittedState?: any,
   monitorReducer?: any,
   errorHandler?: any // <- new
 )
```

This way the errorHandler can be kept outside of the lifted state.

brandonroberts:
That's true but since we're not merely logging the error anymore and passing it the error handler. Implementing custom error handlers are pretty common so I'd rather be on the safe side so users are aware of the change.

sylvaindumont:
Breaking change added to the commit, even if basically the commit just restore the error handling behavior that we have without using StoreDevtools

brandonroberts:
Fixed via 32df3f0fc92c5724562a2cb60c7ad0abb3e27f3f

Issue #920, opened 2018-03-20
RFC: Schematics `ng-add` implementation


vitaliy-bobrov:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x ] Feature request
[ ] Documentation issue or request
</code></pre>

There is the new command introduced for `@angular/cli` [v6.0.0-beta.5](https://github.com/angular/angular-cli/releases/tag/v6.0.0-beta.5) (and will be part of v6 final release) called `ng add`. What does it do? It fetches schematic collection from npm registry according to passed collection name and executes `ng-add` schematic from that collection. It should be used for initial automatic setup of the library inside users project. We could implement such `ng-add` schematic for `@ngrx/schematics` package to produce store and feature setup like it described [here](https://github.com/ngrx/platform/blob/master/docs/schematics/README.md) by calling store and feature schematics. 

Flow example:
1. User runs `ng add @ngrx/schematics` + additional arguments needed.
2. Angular CLI runs `ng-add` schematics from the ngrx collection
3. Root store, effects, devtools setup in users project automatically

Great [article](https://www.softwarearchitekt.at/post/2018/03/20/custom-schematics-part-iv-frictionless-library-setup-with-the-angular-cli-and-schematics.aspx) by @manfredsteyer for the reference.

robwormald:
Yep, all for this.

vitaliy-bobrov:
Great, I'll prepare PR

MikeRyanDev:
@robwormald, @brandonroberts: Does it make more sense for the `:ng-add` schematic to live in each package? That way running `ng add @ngrx/store` would add Store, `ng add @ngrx/effects` would add Effects, etc.

brandonroberts:
Good question. I think it might be a little confusing if you add store to an application and it gets set up, but you can't generate anything else unless you then install schematics, which would run its initial setup also.

vitaliy-bobrov:
It is possible to use any platform package without the schematic, so IMHO only schematic package should implement `ng-add`. We can add arguments to specify optional parts generation (StoreDevtoolsModule, Effects).

Another point here that `@ngrx/schematics` should have some packages as `peerDependencies`, isn't it?

robwormald:
+1 to Mike - that’s the right mental model for ng add. Let’s go with this.

vitaliy-bobrov:
[...]

PR #929, opened 2018-03-23
feat(Schematics): implement "ng-add" schematic

vitaliy-bobrov:
Add implementation for `ng-add` schematic that initializes root store and optionally root effect.

See https://github.com/ngrx/platform/issues/920

### DIFF
--- a/modules/schematics/collection.json
+++ b/modules/schematics/collection.json
@@ -1,97 +1,103 @@
 {
   "schematics": {
     "action": {
-      "aliases": [ "a" ],
+      "aliases": ["a"],
       "factory": "./src/action",
       "schema": "./src/action/schema.json",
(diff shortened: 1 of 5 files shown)


brandonroberts:
Thanks for the PR @vitaliy-bobrov! We're going to implement an `ng-add` schematic per package instead of adding it to `@ngrx/schematics`. Do you want to update this PR to support that?

vitaliy-bobrov:
@brandonroberts , Yes, I would like to prepare PR for each package. And in this case we won't to add `ng-add` to ngrx/schematics ?

brandonroberts:
Correct

vitaliy-bobrov:
Closing as not relevant. Will create the new one.

PR #928, opened 2018-03-23
refactor: upgrade to rxjs 6 and ts 2.7

sandangel:
Closes #926 , #940 , #941

PR #941, opened 2018-03-28
Fix type annotations for select operator overloads

### LINKED ISSUE #940: select operator provides incorrect type information when using string paths
## I'm submitting a...

<pre><code>[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request</code></pre>

Currently, when piping a `select` operator that receives either one string argument or more than two string arguments, the type information is incorrect, resulting in the operator function's return value always being computed as `Store<{}>`. For example:

screen shot 2018-03-28 at 12 50 32 pm

## Expected behavior:
The correct behavior would be for the operator function's return value to be `Store<T>`, where `T` would be replaced by the type of the value at the selected state path. For example:

screen shot 2018-03-28 at 12 51 21 pm

For reference, I've created a [plunker example](https://plnkr.co/edit/zZ9nQRWJ5mfr4j0Z8tw4?p=preview) (check `src/app.ts`), although I don't think you'll be able to see the relevant type information in the browser. You'll likely need to download the files and view them in an editor.

A similar minimal reproduction can be achieved in any existing ngrx project by piping a `select` operator that uses either one or more than two string path segments as its selection method.

Browser: N/A
OS: macOS 10.12.6
npm: 5.6.0
node: 8.9.3
ngrx: 5.2.0

## Other information:
I have a fix prepared, I'm creating this issue for reference purposes.

disquisition:
Fixes #940

### DIFF
--- a/modules/store/spec/edge.spec.ts
+++ b/modules/store/spec/edge.spec.ts
@@ -38,7 +38,7 @@ describe('ngRx Store', () => {
       let todosNextCount = 0;
       let todosCountNextCount = 0;
 
-      store.pipe(select<TodoAppSchema, Todo[]>('todos')).subscribe(todos => {
+      store.pipe(select('todos')).subscribe(todos => {
         todosNextCount++;
         store.dispatch({ type: 'SET_COUNT', payload: todos.length });
       });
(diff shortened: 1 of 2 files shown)


brandonroberts:
Thanks for the PR! These changes are already included in #928

PR #954, opened 2018-04-03
build: Add BUILD files for Schematics

MikeRyanDev:
This does not get the tests running with Bazel. I believe we need bazelbuild/rules_nodejs#84 to be fixed first. cc @alexeagle


alexeagle:
That issue is just about the load paths being inconsistent at runtime - it shouldn't be blocking you?

BUILD.bazel
```diff
@@ -7,16 +7,17 @@ exports_files([
 ])
 
 filegroup(
```
alexeagle: note, you might want to convert this to a `node_modules_filegroup` see https://github.com/bazelbuild/rules_nodejs/blob/master/defs.bzl#L41

modules/schematics/BUILD
```diff
@@ -0,0 +1,29 @@
...
+filegroup(
+    name = "schematics_templates",
+    srcs = glob([
+        "src/*/files/**/*.ts",
+        "src/*/schema.json",
+    ]),
+)
+
+npm_package(
+    name = "npm_package",
```
alexeagle: is this published to npm separately from other modules?
just FYI if you need it - there is a `packages` attribute on `ng_package` and `npm_package` that allows you to nest them if needed.

MikeRyanDev:
@alexeagle Each schematic needs to be able to load its own `schema.json` and any template files at run time. I can get the package artifact to put these JSON and template files in the correct locations: https://github.com/MikeRyanDev/schematics-bazel-builds/tree/master/src/action

What I can't seem to get working is getting these runtime config files to be in the same place at test time. It seemed like that issue was related but I may have been misconfiguring the `jasmine_node_test` rule.

BUILD.bazel
MikeRyanDev: Changed. Thanks for the tip. 👍

modules/schematics/BUILD
MikeRyanDev: Yes, it is published separately.

MikeRyanDev:
Looks like this is also captured in bazelbuild/rules_typescript#154

@brandonroberts Since the tests are being run with the old runner I don't think this blocks the PR.

brandonroberts:
@MikeRyanDev ok, will you run buildifier so linting passes?

alexeagle:
@MikeRyanDev I found an easy workaround with @robwormald - tell Bazel to copy the json file into bazel-bin folder so it appears in the place the `require` tries to load from.

```
genrule(
  name = "copy_json",
  srcs = ["my_input_copy.json"],
  outs = ["my_input.json"],
  cmd = "cp $< $@",
  output_to_bindir = True,
)
```

Note that Bazel frowns on having inputs at the same filesystem path as outputs, so the .json file has a different name in the source tree than it has in the bazel-bin dir.

Issue #919, opened 2018-03-20
Bug(Devtools): Import state buggy when there are lazy loaded feature reducers


dummdidumm:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
Importing a state which has lazy loaded feature modules along the way leads to wrong state.

## Expected behavior:
Importing a state with lazy loaded feature modules works.

## Minimal reproduction of the problem with instructions:
1. Go to https://stackblitz.com/github/ngrx/platform/ , open in new window, open dev tools
2. Wait till you are redirected to login, then reload on that page to make sure the book feature module is not loaded
3. Login, add a book to your collection, go back to book search (you should NOT stay on the "my collection" page)
4. Export state
5. Go back to login page, reload to make sure book feature module is not loaded
6. Import state
7. Jump to action "[Collection] Load Success". You will see that your added book does not appear. If you jump to other states before that, you will notice more buggy behavior.

Reason: Since #570 older states are not recomputed anymore when a feature reducer update happens. This means that all actions of the import are replayed with the initial reducer which does not yet contain the books reducer, therefore all actions having to do with books have no effect.

## Version of affected browser(s),operating system(s), npm, node and ngrx:
Independent of browser/system/node, latest ngrx version (5.2.0)

## Other information:

Here is what came to my mind so far for fixing this:

### Auto-Commit state when "update reducers" action is fired
Get rid of the whole "what state should be available in which form when and how"-problem and just commit the state up until the update reducers action.
Then the whole state is consistently built with one reducer:
- No possibility of bugs when recomputation of old states has side effects like #229
- Importing state works cleanly without additional code

This behavior could be made flaggable (default: true), so users can turn it off but know of the consequences because of the docs (that would need to be added).

### Add a ROUTER_REQUEST action that is fired before "update reducers" action

Implement something along the lines of #1010 so that a router action is fired BEFORE the navigation kicks in and the "update reducers" action is fired. Doing an import then works like this:
1. the devtools see that ROUTER_REQUEST-action and fires it, then waits
2. the StoreRouterModule does the redirect accordingly
3. the devtools waits until a NavigationEnd event and then continues importing. The way lazy loading works, we can guarantee that any possibly lazy loaded feature module is ready by then.

This would work in most cases but there could be some corner cases (errors during redirect etc) where it could crash.

<br><br>

[...]

PR #955, opened 2018-04-03
feat(Devtools): Support for persist, lock, pause

### LINKED ISSUE #853: RFC: Dev Tools support for persist, lock, pause

dummdidumm:
Closes #853:

Devtools now support:
- persist: When devtools option is clicked, state survives full page reload
- lock: When activated, all actions are ignored
- pause: When activated, recording of actions is paused

Fixes part of #919:

When importing a state, there is now a one second delay to give lazy loaded reducers time to instantiate.
Think of this situation: User is on a page where a reducer was lazy loaded. He wants to use the persist feature. The page is reloaded. Now the "import state" action is fired from the devtools, but the actions are replayed too soon because the lazy loaded reducer is not instantiated yet. That means, the actions are replayed with incomplete reducers. To fix this, a small delay is added before importing a state.


modules/router-store/src/router_store_module.ts
```diff
@@ -279,7 +279,10 @@ export class StoreRouterConnectingModule {
 
     if (this.router.url !== this.storeState[this.stateKey].state.url) {
       this.navigationTriggeredByDispatch = true;
-      this.router.navigateByUrl(this.storeState[this.stateKey].state.url);
```
brandonroberts: This should be a separate PR. I don't think this will work for SSR apps also because you can't guarantee this will be finished before the next navigation starts.

dummdidumm:
I removed the router-store-commit from the PR and instead opened #960 .

What do you think about the other changes?

dummdidumm:
@brandonroberts any plans if/when this will be merged?

brandonroberts:
I'll take a look this week. I'd like something a little more predictable for waiting for the reducers to load when importing, but we'll see.

dummdidumm:
Thanks for the update!

Regarding the import and "wait on lazy loaded reducers": Since lazy loaded features/reducers are directly tied to lazy loaded modules - what about injecting the `Router` with `@Optional()` and then on import
- check if the router exists
- if yes, if it did navigate yet
- if it did not navigate yet, it means the import is done straight after reload, so we wait for the initial router navigation to end (reducers will be loaded by then) and then import the state. 
- if it did already navigate, we can be sure that every module/reducer we know of at this point is already loaded, so we import the state straight away.
- if no router dependency exists, we fall back to waiting for 1 second. Maybe users use a different router implementation or something like that. 

I could add a commit implementing this behavior. What do you think?

Side note: This would not solve all import issues with lazy loading. What about a state history which does some navigation action and only at this point a new reducer is loaded (#919)? But I think this is far more complex to think through / implement and does not belong into this PR.

dummdidumm:
@brandonroberts any updates on this? What do you think about the more predictable "wait for reducers by using Router"-proposal?

dummdidumm:
Sorry to ask again but is there any timeline when (or if at all) this will be merged? Any feedback would be appreciated.

brandonroberts:
Apologies for getting behind on this one. Will you rebase against master?

dummdidumm:
Thanks for the update, I will do it next week.

dummdidumm:
I resolved some new merge conflicts. The build fails but this seems to have nothing to do with the commits (bazel ran out of memory and crashed).

modules/store-devtools/src/extension.ts
```diff
@@ -163,7 +169,31 @@ export class DevtoolsExtension {
     // Listen for lifted actions
     const liftedActions$ = changes$.pipe(
       filter(change => change.type === ExtensionActionTypes.DISPATCH),
-      map(change => this.unwrapAction(change.payload))
+      map(change => this.unwrapAction(change.payload)),
+      concatMap((action: any) => {
+        if (action.type === IMPORT_STATE) {
+          if (this.router) {
```
brandonroberts: I don't think the router should be integrated here. The route may need the state we are importing, but this prevents that from happening until the navigation is over.

modules/store-devtools/src/extension.ts
```diff
@@ -185,9 +215,8 @@ export class DevtoolsExtension {
     return typeof action === 'string' ? eval(`(${action})`) : action;
   }
 
-  private getExtensionConfig(instanceId: string, config: StoreDevtoolsConfig) {
+  private getExtensionConfig(config: StoreDevtoolsConfig) {
     const extensionOptions: ReduxDevtoolsExtensionConfig = {
-      instanceId: instanceId,
```
brandonroberts: Why is this removed?

modules/store-devtools/src/extension.ts
brandonroberts: Maybe we can listen for any `@ngrx/store/update-reducers` actions and delay the import until none have been received within a span of time.

brandonroberts:
@dummdidumm we are still working through the Bazel memory issues, but at long as the tests pass, its fine for now. I left you some feedback on the PR. Let me know what you think.

modules/store-devtools/src/extension.ts
dummdidumm: `liftedActions$` retrieves the IMPORT_STATE-action before it is passed to the reducer. So as long as IMPORT_STATE is not emitted by `liftedActions$`, the import will not be triggered. `update-reducers` etc will be fired before the route is loaded because that happens on a different stream of actions (`actions$`), so there is no block. So in my opinion this is fine.

dummdidumm:
I added you some feedback, let me know what you think.

modules/store-devtools/src/extension.ts
dummdidumm: I think it's not quite clear what I mean, so I'll expand on it:

### Situation 1: Router is used by user. User clicks the "import"-button of the redux devtools and manually imports a state
The added logic will not matter because  router.navigated will be true (the first navigation already happened) , therefore nothing to wait on, therefore the IMPORT action is fired straight away.

### Situation 2: Router is used by user. User turned on the "persist"-feature and reloads.
The added logic now matters. router.navigated will be false because the router is in the process of navigating when the action comes in. What that means is that the modules connected to that route are still getting instantiated, so for example components/reducers/effects for that route are not there yet.  So if we just let the IMPORT action pass through, the reducer would import the state too soon, some actions would get ignored because the reducers handling it do not exist yet. That's why we wait until the NavigationEnd-event of the router is fired. At this point we know all the initialization logic is done.<br>
There will be no blocking of state needed for the router because initial reducers/effects/state will be loaded during navigation. These things do not come through the `liftedActions$` stream, so blocking further events there until we are finished importing is not an issue. Also something like "oh you are not logged in, I will redirect you" is fine because, well, we just import the state after that and the application will get back to the correct route.

### Situation 3: Router is not used by user.
The whole router logic does not matter now, we wait 1 second just to make sure.

brandonroberts: I understand what's going on here, but I'd still rather support this in a consistent way without the router if possible.

Idea: After the IMPORT_STATE action comes in, wait some amount of time, then switch to listening for update-actions and debounce the IMPORT_STATE action. We can inject and listen to the store actions being dispatched.

dummdidumm: Like this?

```
concatMap((action: any) => {
  if(action.type === IMPORT_STATE) {
    return actions$.pipe(
      ofType(UPDATE), // '@ngrx/store/update-reducers'
      debounceTime(1000),
      map(() => action)
    );
  }
...
```

dummdidumm: We cannot know whether the IMPORT_STATE action is fired because of the persist option turned on or because of a manual import. But that's no harm, the user just has a minimal delay (of 1 second) until the manual import starts.

dummdidumm: Good point, a race condition / timeout in the likes of "there is no UPDATE action at all" that fires after 1 second is needed.
I will implement it and update the PR.

modules/store-devtools/src/extension.ts
```diff
@@ -17,6 +28,8 @@ import {
   sanitizeStates,
   unliftState,
 } from './utils';
+import { UPDATE } from '../../store/src';
```
brandonroberts: This should be `@ngrx/store` instead of a relative import

PR #958, opened 2018-04-04
WIP: chore(Example): Upgrade to Angular CLI 6.0.0-rc.0

brandonroberts:


package.json
```diff
@@ -63,29 +63,30 @@
...
-    "@ngrx/db": "github:brandonroberts/ngrx-db-builds",
+    "@ngrx/db": "^2.2.0-beta.0",
     "core-js": "^2.5.4",
     "hammerjs": "^2.0.8",
     "ngrx-store-freeze": "^0.2.1",
     "opencollective": "^1.0.3",
-    "rxjs": "6.0.0-beta.4",
-    "rxjs-compat": "6.0.0-beta.1",
+    "rxjs": "6.0.0-rc.0",
+    "rxjs-compat": "6.0.0-rc.0",
```
tdeschryver: Is this still needed?

brandonroberts: For now, yes, but only for `husky`. We are using `lint-staged` which has a dependency on `listr` and its still using RxJS 5.x imports.

https://github.com/SamVerschueren/listr/blob/master/lib/task.js#L4

Issue #938, opened 2018-03-28
[Store-DevTools] actionsBlacklist / actionsWhitelist


Wykks:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
There is no way to filter out actions in Store-DevTools.

## Expected behavior:
Implement `actionsBlacklist` / `actionsWhitelist`
https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#actionsblacklist--actionswhitelist

I can try to implement this one, if you're ok with this.

TobyBackstrom:
Yes, I'd like to see this, too. I've got some actions that come with a very high frequency and I need to filter them somehow to be able to use the devtools. I would guess this is a pretty common scenario in any app that displays "realtime data".

nasreddineskandrani:
the `predicate` just below `actionsBlacklist` / `actionsWhitelist` is more useful as feature in my opinion (that can do the blacklisting or whitelisting).

@Wykks what do you think?

Wykks:
Yup, you're right. I added `predicate` too

PR #970, opened 2018-04-14
feat(StoreDevtools): implement actionsBlacklist/Whitelist & predicate

Wykks:
Ref: #938 

Note: the unit tests doesn't test jump / slider (I'm not sure how to test that).

There an issue when using the slider (and the action filter feature).
The app doesn't get the right state (whereas the extension show the correct state).
To reproduce in the example app, add `actionsBlacklist`:
```
      name: 'NgRx Book Store DevTools',
      logOnly: environment.production,
      actionsBlacklist: ['\\[Book\\] Search$']
    }),
```
Do two different search, then, jump or slide back to the previous `Search Complete`.

Sorry for leaving this issue here, I spend more than half a day figuring out how this works... This is as far as I'm able to do, for now... I'll try to investigate more If needed, but you probably already know what's going on :sweat_smile:

Edit: Ok I see what's going on. Since I filter the `liftedState` only for the `devtoolsExtension.send`, the `currentStateIndex` send by the extension is desync with the internal `liftedState`.

I could actually filter-out the state outside `extension.notify` but according to: https://github.com/zalmoxisus/redux-devtools-extension/issues/316#issuecomment-283914716 , the internal `liftedState` should not be filtered.

Edit 2 : Welp, I guess I should filter the state outside `extension.notify` after all. It works fine. But I'm a bit puzzled about how to "properly" code this. I'll update the PR next time (maybe tomorow).


brandonroberts:
Thanks for the PR @Wykks! As an alternative, I'm looking into sending all actions/states through the extension since it already has support for blacklist/whitelist/predicate and I don't think we need to duplicate all that on our end.

kevinahuber:
Any word/work on this? Happy to help get this through.

Wykks:
Hi!

Sorry for this very long silence... I totally got demoralized doing this PR.
I can't help thinking that something is really wrong.

Just like Brandon said, this should be handled by the extension itself. But it doesn't work when sending the full state (with `this.devtoolsExtension.send`); because the extension doesn't filter anything in this case. That's why I started to handle both cases.

But, I really don't get why it's implemented like that in the extension. This should just be a visual thing, with some skip when moving through states. There's no actual need to filter the state like that...

brandonroberts:
@Wykks apologies for getting behind on this one. Will you rebase on master? Handling it here may be the option for now. We may need to open an issue on the extension to at least get some clarification on filtering when full state is provided.

brandonroberts:
@Wykks will you rebase this one?

Wykks:
ops didn't saw the last message. I can do the rebase this week yep!

brandonroberts:
👍

Wykks:
Rebased and updated!
Same thought on this stuff since the last time I spent time on it; it doesn't feel correct but it works.

brandonroberts:
@Wykks will you rebase again? Thanks

Wykks:
Done, sorry for the delay..

brandonroberts:
🎉

wendelstam:
Really nice addition. In what version will this be included in the NPM package?

timdeschryver:
It will become available in the next version (7).
You can use it now by using the nightly builds.

Issue #825, opened 2018-02-16
StoreDevtools: window postMessage, object Position cannot be cloned


montella1507:
## I'm submitting a...
[x ] Bug report

## What is the current behavior?
Just migrated to the newest typescript and we got new error from StoreDevTools.
"window postMessage, object Position cannot be cloned"

montella1507:
Fixed. But i would let open this issue, because it is interesting problem. Maybe it relates to some problem..

If i create an observable from window.geolocation.getPosition() and pass Position object from the callback directly to observer. it will crash in navigation in StoreDevTools.

However. if i map that position to my own new object, it works..

`
ERROR in StoreDevTools
return Observable.create(obs =>  {
   window...getPosition((pos)=> {
    obs.next(pos);
    obs.complete();
   });
});

OK
return Observable.create(obs =>  {
   window...getPosition((pos)=> {
    obs.next({
    coords: {
       ...
    }
});
    obs.complete();
   });
});
`

it looks like  StoreDevTools is trying to copy the object when it is already freed:? 

Feel free to close the issue

warapitiya:
What is the fix here? I'm getting an error like this too.

montella1507:
WELL... do not use Position object from geolocation callback directly in your code.. Just copy Longtitude, lattitude values to new object. 

`obs.next({
coords: {
   long : original.coords.long,
   lat : original.coords.lat,
}`

BTW-... this wasnt the only one problem we had.. We had to remove ALL "Position" type in our code, else we got "Position is undefined" error in runtime - lol..

I think there may be problem in Typescript lib.dom.ts where are Position 2times: as a interface and as a value.

warapitiya:
ok. looks like this is not the same issue for me! But when i disable the `Redux extension` everything works normally.

**Below is the error message I get:**
```
ERROR Error: Uncaught (in promise): DataCloneError: Failed to execute 'postMessage' on 'Window': function DashboardComponent(title, activatedRoute, dashboardApiService, userService, rootStore, store, route, m...<omitted>... } could not be cloned.
```

montella1507:
Well.. this looks littlebit like your own mistake, are you pushiung to observer a component?

Overdozed:
same issue if redux extension is on after updating from
"@ngrx/effects": "^5.0.1",
   "@ngrx/store": "^5.0.0",
    "@ngrx/store-devtools": "^5.0.1",

to 5.1.0

warapitiya:
[...]

PR #971, opened 2018-04-15
DevTools Sanitizer Service and allow Devtool Options to be injectable

phillipzada:
Hi team,

Created a DevTools sanitizer service that allows for multiple (or single) sanitizers to be injectable. This provides an avenue to address #825 

I've also exposed the DevTools Options to be injectable via a provider.

Thanks
Phill


dummdidumm:
In what way does it adress the issue? DevTools already have a action/state sanitizer. What is the advantage of having them injectable and provided via tokens?

phillipzada:
Hey @dummdidumm, We've had the use case with lazy loaded modules (eg. Nrwl NX Framework) where we need sanitizers added to handle specific cases. whereas the current sanitizer is for a single action/state santizier this option gives the ability to register multiple.

cbeall:
This has been sitting for a while and appears it has a few conflicts at this point, but I agree that including it would be valuable. Using DI would allow for clean separation of concerns, and the current implementation really presents complications with regard to lazy loading. 

Are there an updates here?

brandonroberts:
@cbeall at a minimum I'd like to see the configuration tokens be exposed. I think the custom sanitizer service should be left out for now. Or added as a recipe in the docs. Would you want to put in a PR for that part?

phillipzada:
@brandonroberts happy to put a PR and start it again from a newer codebase. Let me know if there is anything else you would like to see other than exposed DI tokens.

brandonroberts:
@phillipzada thanks Phillip. If you want to add a recipe to the docs about using a custom sanitizer that would be great.

brandonroberts:
Going to close this one. Open a new PR for the changes we talked about.

PR #979, opened 2018-04-18
feat(effects): Improve performance of ofType

### LINKED ISSUE #978: Performance of ofType?
## I'm submitting a...

<pre><code>
[ ] Bug report  
[x] Feature request
</code></pre>

Right now, ofType is implemented via `filter`. As I understand it, this creates O(n) performance when you have multiple effects.

As an application grows, the performance is going to continually degrade, unless we re-write "ofType" to work without filter, and instead use something that has a performance closer to O(1).

## Possible options:
I think that the best option would be doing a dictionary lookup on a Map or Set. We can always fall back to a POJO in the case that Map or Set are not supported.

The tricky bit is, when you have multiple `ofType`s being called, each creating its own observable, how do you make that work in a performant fashion?

I can work on a PR for this, I just need to know whether or not it's a direction the project wants to go in.

bryanrideshark:
Instead of naively applying `filter`, ofType now uses a Map to only send actions along to the exact places where they will be used.

Closes #978

### DIFF
--- a/modules/effects/spec/of_type.spec.ts
+++ b/modules/effects/spec/of_type.spec.ts
@@ -1,0 +1,111 @@
+import { ofType, getOfTypeMetadata } from '../';
+import { forceBasicMode } from '../src/of_type';
+import { hot, cold } from 'jasmine-marbles';
+
+describe('ofType', function() {
+  const ADD = 'ADD';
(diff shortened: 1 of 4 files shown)


brandonroberts:
See https://github.com/ngrx/platform/issues/978#issuecomment-383005194

PR #980, opened 2018-04-19
fix(Entity): Fix type annotations for getInitialState methods

rob3c:
This adds intellisense support for extended state objects within the `getInitialState` method if the state type is passed as a generic parameter.
For example, here's an excerpt from the added test for this (given the constraints of type annotation testing):

```typescript
    interface StateWithAdditionalProps extends EntityState<BookModel> {
      isHydrated: boolean;
    }

    // initialState is type StateWithAdditionalProps
    const initialState = adapter.getInitialState<StateWithAdditionalProps>({
        // intellisense here for StateWithAdditionalProps
        isHydrated: true,
        // nonStateProp: 'not allowed', // red squigglies
    });
```

This commit also preserves the previous overload, which allows adding arbitrary objects without complaining (valid for some use cases, I suppose, so didn't want to break it):

```typescript
    const additionalProperties = { isHydrated: true };

    // initialState is type EntityState<BookModel> & { isHydrated: boolean; }
    const initialState = adapter.getInitialState(additionalProperties);
```

### DIFF
--- a/modules/entity/src/entity_state.ts
+++ b/modules/entity/src/entity_state.ts
@@ -9,11 +9,14 @@ export function getInitialEntityState<V>(): EntityState<V> {
 
 export function createInitialStateFactory<V>() {
   function getInitialState(): EntityState<V>;
+  function getInitialState<S extends EntityState<V>>(
+    additionalState: Partial<S>
+  ): S;
   function getInitialState<S extends object>(
     additionalState: S
   ): EntityState<V> & S;
   function getInitialState(additionalState: any = {}): any {
-    return Object.assign(getInitialEntityState(), additionalState);
+    return Object.assign(getInitialEntityState<V>(), additionalState);
   }
 
   return { getInitialState };
--- a/modules/entity/spec/entity_state.spec.ts
+++ b/modules/entity/spec/entity_state.spec.ts
@@ -1,4 +1,4 @@
-import { createEntityAdapter, EntityAdapter } from '../src';
+import { createEntityAdapter, EntityAdapter, EntityState } from '../src';
 import { BookModel } from './fixtures/book';
 
 describe('Entity State', () => {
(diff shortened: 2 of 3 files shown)


modules/entity/src/entity_state.ts
```diff
@@ -9,11 +9,14 @@ export function getInitialEntityState<V>(): EntityState<V> {
 
 export function createInitialStateFactory<V>() {
   function getInitialState(): EntityState<V>;
+  function getInitialState<S extends EntityState<V>>(
+    additionalState: Partial<S>
```
sandangel: is this good when make ids and entities optional too?

sandangel: why do you use Partial here?

sandangel: I mean we should not allow user declare ids or entities, so we should use conditional type here to ban those props.

rob3c: My personal preference is to not any properties that aren't on the extended state, and that includes not allowing ids or entities. That's how my original fix worked. However, since the original typings allowed anything, and the single test confirmed that was the desired behavior, I changed it to a partial without removing the `EntityState<T>` properties. If you'd prefer that we only strictly allow the additional properties explicitly declared on a custom state object, then great - I'd love to restrict it to that as I originally had done! However, that would require deleting the previous test and breaking current functionality.

I can do either, so which do you prefer?

rob3c: This is what I had before, but the existing test suggested it was too restrictive:
```typescript
    getInitialState<S extends EntityState<T>>(
        state: Pick<S, Exclude<keyof S, keyof EntityState<T>>>
    ): S {
        return Object.assign(this.getInitialEntityState(), state) as S;
    }
```
which gives:
```typescript
export const initialState = adapter.getInitialState<StateWithAdditionalProps>({
    // only intellisense for additional properties beyond those in EntityState<T>:
    isHydrated: false,

    // and anything else gives red squigglies, including ids and entities:
    // ids: [],
    // entities: {},
    // foos: [],
});
```

rob3c: That `Exclude` is new in TypeScript 2.8. If we need to use something earlier, then this `Diff` implementation works instead:
```typescript
export type Diff<T extends string, U extends string> = (
    & { [P in T]: P }
    & { [P in U]: never }
    & { [x: string]: never }
)[T];
```
So we just replace `Exclude` with `Diff`.

sandangel: Sorry let me think again. there is a use case that user would like to provide initial entities different from {}.

brandonroberts:
Are these changes still valid or is there a different way to handle this the latest version of TypeScript?

brandonroberts:
Closing for now. Will reopen if needed.

PR #991, opened 2018-04-24
Fix 'oftype after'

tdeschryver:
There was a question on the gitter channel what the change was at the `oftType` before and after example.
Seems like there was none :sweat_smile:


brandonroberts:
👍

Issue #245, opened 2017-08-07
Redux devtools not being connected to the store.


mnemanja:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[x] Support request
</code></pre>

## What is the current behavior?
I've followed these 2 guides:
https://github.com/ngrx/platform/blob/master/docs/store/README.md
https://github.com/ngrx/platform/blob/master/docs/store-devtools/README.md
After the implementation the Redux plugin returns this message: `No store found. Make sure to follow the instructions.`

## Expected behavior:
The expected behavior is for the plugin and the store to be connected through the `@ngrx/store-devtools` module.

## Minimal reproduction of the problem with instructions:
I have two examples where it doesn't work. One is the app that I'm building, and the second one is the clean slate app generated just for purpose of testing this.
This is `my app.module.ts`
```
import {BrowserModule} from '@angular/platform-browser';
import {NgModule} from '@angular/core';
import {HttpModule} from '@angular/http';
import {FormsModule, ReactiveFormsModule} from '@angular/forms';

import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
import {MdButtonModule, MdDialogModule, MdIconModule, MdInputModule} from '@angular/material';

import 'hammerjs';

import { StoreDevtoolsModule } from '@ngrx/store-devtools';
import { StoreModule } from '@ngrx/store';

import {AppComponent} from './app.component';

import {ForecastComponent} from './forecast/forecast.component';
import {ForecastService} from './forecast/forecast.service';
import {HandleLocationDialogComponent} from './handle-location-dialog/handle-location-dialog.component';

import { forecast } from './forecast/forecast.reducer';


@NgModule({
  declarations: [
    AppComponent,
    ForecastComponent,
    HandleLocationDialogComponent,
  ],
  imports: [
    StoreModule.forRoot({forecast}),
    StoreDevtoolsModule.instrument({
      maxAge: 25 //  Retains last 25 states
    }),

    BrowserModule,
    HttpModule,
    FormsModule,
    ReactiveFormsModule,
    BrowserAnimationsModule,
    MdButtonModule,
    MdIconModule,
    MdDialogModule,
    MdInputModule,
  ],
  providers: [ForecastService],
  bootstrap: [AppComponent],
  entryComponents: [HandleLocationDialogComponent]
})
export class AppModule { }
```
And my reducer
[...]

PR #992, opened 2018-04-24
docs(devtools): Specify that you need to inject the store for devtools to work

mtaran-google:
I spent some time trying to debug things until I found https://github.com/ngrx/platform/issues/245 which mentions that the store needs to be injected before the devtools start working! Hopefully the next person that runs into this will be able to find the answer here.


brandonroberts:
Thanks!

PR #1008, opened 2018-05-01
docs: describe composition of selectors in pipeable operators

### LINKED ISSUE #1002: Selectors: filter undefined values
## I'm submitting a...

<pre><code>
[ ] Bug report  
[x] Feature request
</code></pre>

I created a repro / example:
https://stackblitz.com/edit/angular-ngrx-effects-1rj88y

Let's look at the following selectors:

```ts

export const selectFoo = createFeatureSelector<State>('foo');

export const selectFooBar = createSelector(selectFoo, state => state.bar);

export const selectFooFoobar = createSelector(selectFoo, state => state.foobar);

export const selectProjectedValues = createSelector(
  selectFooBar,
  selectFooFoobar,
  (foo, foobar) => {
    if (foo && foobar) {
      return { foo, foobar };
    }

    // XX: Is it possible to filter this undefined value in the selector?
    return undefined;
  }
);
```

`selectProjectedValues` will emit an `undefined` value if `foo` or `foobar` have an undefined value.

Filtering the projected `undefined` value would need to happen in the subscribor, e.g.

```ts
store
  .select(selectProjectedValues)
  .pipe(filter(value => !!value))
```

## Expected behavior:

A projected value from  `foo` and `foobar` should be returned, if both states have a defined value. Let's pretend we have a `list` and an `activeItemId`, an active item will be emitted if both the list and the identifier are set. If one of them is undefined, there's no active item.

The question or feature request is: is it possible to filter the undefined value in terms of the `selectProjectedValues` API?

Meaning that `store.select(selectProjectedValues)` will never emit an undefined value.

See above.

ngrx@5.2.0

dherges:
Ready for review.

Closes #1002

### DIFF
--- a/docs/store/selectors.md
+++ b/docs/store/selectors.md
@@ -220,7 +220,7 @@ Selectors empower you to compose a [read model for your application state](https
 In terms of the CQRS architectural pattern, NgRx separates the read model (selectors) from the write model (reducers).
 An advanced technique is to combine selectors with [RxJS pipeable operators](https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md).
 
-This section covers some basics of how selectors compare to pipeable operators and demonstrates 
+This section covers some basics of how selectors compare to pipeable operators and demonstrates how `createSelector` and `scan` are utilized to display a history of state transitions.
 
 ### Breaking Down the Basics
 
--- a/docs/store/selectors.md
+++ b/docs/store/selectors.md
@@ -270,7 +270,7 @@ import { pipe } from 'rxjs';
 import { filter } from 'rxjs/operators';
 
 export const selectFilteredValues = pipe(
-  select(selectProjectedValues),
+  select(selectValues),
   filter(val => val !== undefined)
 );
 
@@ -306,7 +306,6 @@ We are not only interested in the current state but rather like to display the l
 Meaning that we will map a stream of state values (`1`, `2`, `3`)  to an array of state values (`[1, 2, 3]`).
 
 ```ts
-
 // Compose a pipeable operators that combines the last `count` state values in an array
 export const selectLastStateTransitions = (count: number) => {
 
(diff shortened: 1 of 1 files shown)


docs/store/selectors.md
```diff
@@ -213,3 +213,61 @@ class MyAppComponent {
...
+```
+
+Refactoring to leverage the `select()` utility function:
+
+```ts
+import { select } from '@ngrx/store';
+import { map, filter } from 'rxjs/operators';
+
+store.pipe(
+  select(selectProjectedValues(state),
```
MattiJarvinen-BA: Missing one right parenthesis.

docs/store/selectors.md
```diff
@@ -213,3 +213,61 @@ class MyAppComponent {
...
+
+export const selectProjectedValuesFiltered = pipe(
+  select(selectProjectedValues),
+  filter(val => val !== undefined)
+)
+
+store.pipe(selectProjectedValuesFiltered)
+     .subcribe(/* .. */);
+```
+
```
MattiJarvinen-BA: Hmm... this is an interesting one although I might add a new pipeable operator that just leaves out undefined values. It would give more reusability throughout the application.

dherges:
@brandonroberts Here's my write-up on combining selectors and pipeable operators. I also added a stackblitz to demosntrate the advanced example of such a read model. Maybe the comments in the code lkisting could be better? Wdyt?

/FYI @dummdidumm @MattiJarvinen-BA

brandonroberts:
@dherges I think the writeup is great! I also feel like it should be a separate blog post in itself and it should be under an `Advanced Usage` section in the store docs. Will you add a header above your section for `Advanced Usage`?

dherges:
I updated, removed some typos, and added the "Advanced Usage" headline. I think the basics how the selector compares to the rxjs pipeable operators should be defo in the docs - that's what I was looking for initially. I'd either take the advanced example for a blog post or blog something on the ngrx/cqrs analogies. The server-side/client-side filtering is a classic example and worth a good blog post.

brandonroberts:
Thanks!

PR #1012, opened 2018-05-03
Fix for Schematics Documentation for Creating Stores

firrae:
As the title says, this should update the docs to the correct command for creating a new store in a module other than the root app module.


firrae:
Not a problem 😃

PR #1014, opened 2018-05-04
remove unnecessary code

metasong:
no need to slice the last function to use array.reduceRight function

### DIFF
--- a/modules/store/src/utils.ts
+++ b/modules/store/src/utils.ts
@@ -76,14 +76,7 @@ export function compose<A, B, C, D, E, F>(
 export function compose<A = any, F = any>(...functions: any[]): (i: A) => F;
 export function compose(...functions: any[]) {
   return function(arg: any) {
-    if (functions.length === 0) {
-      return arg;
-    }
-
-    const last = functions[functions.length - 1];
-    const rest = functions.slice(0, -1);
-
-    return rest.reduceRight((composed, fn) => fn(composed), last(arg));
+    return functions.reduceRight((composed, fn) => fn(composed), arg);
   };
 }
 


MattiJarvinen-BA:
I haven't checked if your changes work (tests seem to work). 

Here's the react/Redux compose for reference.

https://github.com/reactjs/redux/blob/master/src/compose.js#L12-L22

Current source
```javascript
/**
 * Composes single-argument functions from right to left. The rightmost
 * function can take multiple arguments as it provides the signature for
 * the resulting composite function.
 *
 * @param {...Function} funcs The functions to compose.
 * @returns {Function} A function obtained by composing the argument functions
 * from right to left. For example, compose(f, g, h) is identical to doing
 * (...args) => f(g(h(...args))).
 */

export default function compose(...funcs) {
  if (funcs.length === 0) {
    return arg => arg
  }

  if (funcs.length === 1) {
    return funcs[0]
  }

  return funcs.reduce((a, b) => (...args) => a(b(...args)))
}
```

ngrx compose tests
https://github.com/ngrx/platform/blob/9b02f79f1fc3b5e4d85bf0516108f93eed655126/modules/store/spec/utils.spec.ts#L58-L75

redux compose tests
https://github.com/reactjs/redux/blob/3a17aef0468e279cba048411f8f91dce07c55a7e/test/compose.spec.js#L4-L52

```javascript
  describe('compose', () => {
    it('composes from right to left', () => {
      const double = x => x * 2
      const square = x => x * x
      expect(compose(square)(5)).toBe(25)
      expect(compose(square, double)(5)).toBe(100)
      expect(compose(double, square, double)(5)).toBe(200)
    })

    it('composes functions from right to left', () => {
      const a = next => x => next(x + 'a')
      const b = next => x => next(x + 'b')
      const c = next => x => next(x + 'c')
      const final = x => x

      expect(compose(a, b, c)(final)('')).toBe('abc')
      expect(compose(b, c, a)(final)('')).toBe('bca')
      expect(compose(c, a, b)(final)('')).toBe('cab')
    })

    it('throws at runtime if argument is not a function', () => {
      const square = x => x * x
      const add = (x, y) => x + y

      expect(() => compose(square, add, false)(1, 2)).toThrow()
      expect(() => compose(square, add, undefined)(1, 2)).toThrow()
      expect(() => compose(square, add, true)(1, 2)).toThrow()
      expect(() => compose(square, add, NaN)(1, 2)).toThrow()
      expect(() => compose(square, add, '42')(1, 2)).toThrow()
    })

    it('can be seeded with multiple arguments', () => {
      const square = x => x * x
      const add = (x, y) => x + y
      expect(compose(square, add)(1, 2)).toBe(9)
    })

    it('returns the first given argument if given no functions', () => {
      expect(compose()(1, 2)).toBe(1)
      expect(compose()(3)).toBe(3)
      expect(compose()()).toBe(undefined)
    })

    it('returns the first function if given only one', () => {
      const fn = () => {}

      expect(compose(fn)).toBe(fn)
    })
  })
```

brandonroberts:
I'm inclined to leave this as is ...

metasong:
i've looked at the redux implementation posted above, it uses reduce to implement reduceright, it has to do argement numbser check because of no initial value to the reduce function, as describe in reduce if the array is empty there would an exception. so the mdn document suggest to always provide an initial value. but for our case, we use reduceright and provide the args as the initial value, it's safe.

the test has passed, as a professional programer, there should be a cause to not use it.

Issue #915, opened 2018-03-15
RFC: Mock Store


TeoTN:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?
There's no simple way to mock a Store for testing purposes. 

## Expected behavior:
When testing a component or an effect, I'd like to be able to inject a fake Store to the tested entity. It became increasingly non-trivial after the pipable operators were introduced.

For instance, one may want to test behavior component with a given state that is a result of many various actions. Similarly, one may want to test an effect with various fixtures for state. 
Across the web there are multiple snippets with custom implementation of `MockStore` that extends either `Store` or `BehaviorSubject` but all of them are outdated. The presence of such snippets suggests though that there's a real need for some testing utils.
I propose introduction of `@ngrx/store/testing` module which would contain `provideMockStore` for providing `MockStore` in place of `Store` and an implementation of `MockStore` that provides `nextMock` method that replaces current state just like `BehaviorSubject.next` does on Subject.

## Version of affected browser(s),operating system(s), npm, node and ngrx:
Newest ngrx, reasonably new npm, node, Chrome and Windows.

## Other information:
I'll make a contribution with `MockStore` class and a provider, if you feel it's a valid issue.

MikeRyanDev:
Great suggestion! @brandonroberts and I have a mock Store implementation we use internally that works similarly to how you propose. It should also setup `Store#dispatch` to be a spy.

vitaliy-bobrov:
Great proposal! IMHO will be nice to have a separate package inside platform with all testing helper tools.

hoangdovan:
Yeah, We really need it!

alex-okrushko:
This has been raised a number of times already and I'd like to get the consensus whether mock Store is needed or not :)

E.g. in https://github.com/ngrx/store/issues/128 @robwormald says: 

> _If you want to implement a mock version (**which, IMHO, is entirely unnecessary, but feel free..**.)_

The same recommendation is at current testing guide for the store https://github.com/ngrx/platform/blob/master/docs/store/testing.md#providing-store-for-testing

> Reducing state is synchronous, so mocking out the Store isn't required.

[...]

PR #1027, opened 2018-05-07
feature: Testing module for @ngrx/store

TeoTN:
Addresses #915: This PR introduces a simple `MockStore` that allows setting state shape for testing purposes.
Alongside with `MockStore` provider function is declared, and both are structured into `@ngrx/store/testing`.

Please review changes and help me ensure that the `testing` module is setup properly, or feel free to commit appropriate changes.


eppsilon:
How could this be used to mock lazy-loaded feature state? The `nextMock(nextState: T)` method prevents providing state that isn't in the root/global state interface.

TeoTN:
The purpose of this solution is to mock current state of the app, it doesn't care about the way the state was build. So it doesn't matter whether some part of state was built with `forRoot` or `forFeature`, it all goes to a single object.

TeoTN:
Hey @MikeRyanDev  I've added a little bit more code and tests :) Can we work towards merging it?

marcelnem:
Additionally, to nextMock and spyOnDispatch, I think, it would be good to add an ability to mock selectors to the proposed MockStore implementation. Being able to mock selectors is mentioned quite often in the thread #915, so I think people like to use it.

SerkanSipahi:
@TeoTN Thank you so much for that :) Is there any documentation/examples/practices for that?

brandonroberts:
@TeoTN posting here because it will get lost in the issue. First, thanks for your work and patience! I've played around with this in the example app and I have some suggestions. 

StoreModule.forRoot() should not be required in addition to this.
Mocking the dispatch depends on the developers test suite of choice, so a method to patch it should not be included
I think `setState` as the method name instead of `nextMock` is clearer.
We shouldn't focus on mocking selectors. If you want to test selectors, provide a state that the selectors can use or use StoreModule.forRoot() with reducers for integration.

Here is a proposed update to the testing module.

```typescript
import { Injectable, Inject } from '@angular/core';
import {
  StateObservable,
  Store,
  ReducerManager,
  ActionsSubject,
  ActionReducer,
  Action,
  ScannedActionsSubject,
  INITIAL_STATE,
} from '@ngrx/store';
import { BehaviorSubject } from 'rxjs';

@Injectable()
export class MockState<T> extends BehaviorSubject<T> {
  constructor(@Inject(INITIAL_STATE) initialState: any) {
    super(initialState as T);
  }
}

@Injectable()
export class MockReducerManager extends BehaviorSubject<
  ActionReducer<any, any>
> {
  constructor() {
    super(() => undefined);
  }
}

@Injectable()
export class MockStore<T> extends Store<T> {
  private state$ = new BehaviorSubject<T>({} as T);

  constructor(
    state$: StateObservable,
    actionsObserver: ActionsSubject,
    reducerManager: ReducerManager,
    public scannedActions$: ScannedActionsSubject
  ) {
    super(state$, actionsObserver, reducerManager);
    this.source = this.state$.asObservable();
  }

  setState(state: T): void {
    this.state$.next(state);
  }

  dispatch<V extends Action = Action>(action: V) {
    super.dispatch(action);
    this.scannedActions$.next(action);
  }

  addReducer() {
    // noop
  }

  removeReducer() {
    // noop
  }
}

export function provideMockStore<T = any>(config: { initialState?: T } = {}) {
  return [
    { provide: INITIAL_STATE, useValue: config.initialState },
    ActionsSubject,
    ScannedActionsSubject,
    { provide: StateObservable, useClass: MockState },
    { provide: ReducerManager, useClass: MockReducerManager },
    {
      provide: Store,
      useClass: MockStore,
    },
  ];
}
```

Thoughts?

ef32:
@brandonroberts  your proposal for the testing module above seems like a perfect fit, the selectors do not need to be mocked; by providing state initially, the required selectors should be able to work with the store. That said, here are some suggestions to slightly enhance your proposal. 

- the observable source field is now deprecated and in my opinion,  i dont think it needs to be populated in this scenario.

- StateObservable being an observable, can be substituted using MockState. No need to provide with  useClass. 

- MockState can be generic, requiring the same generic Type as MockStore  hence providing strong type checking. 

- Initial state can be initialised by calling the next method with the initialstate token  in the MockStore constructor.

- Test suites could require additional data to state, so setState can cater for such scenarios

please find below slight modification:

`import { Injectable, Inject } from '@angular/core';
import {
  StateObservable,
  Store,
  ReducerManager,
  ActionsSubject,
  ActionReducer,
  Action,
  ScannedActionsSubject,
  INITIAL_STATE,
} from '@ngrx/store';
import { BehaviorSubject } from 'rxjs';

@Injectable()
export class MockState<T> extends BehaviorSubject<T> {
  constructor() {
    super({} as T);
  }
}

@Injectable()
export class MockReducerManager extends BehaviorSubject<
  ActionReducer<any, any>
> {
  constructor() {
    super(() => undefined);
  }
}

@Injectable()
export class MockStore<T> extends Store<T> {

// not needed
 // private state$ = new BehaviorSubject<T>(this.initialState as T);

  constructor(
    state$: MockState<T>,
    actionsObserver: ActionsSubject,
    reducerManager: ReducerManager,
    public scannedActions$: ScannedActionsSubject,
    @Inject(INITIAL_STATE) private initialState: T
  ) {
    super(state$, actionsObserver, reducerManager);
    
 //deprecated
    // this.source = this.state$.asObservable();

   this.state$.next(this.initialState);
  }

  setState(state: T): void {
   const newState = {
                              ...this.initialState,
                              ...state
                            };
    this.state$.next(newState);
  }

  dispatch<V extends Action = Action>(action: V) {
    super.dispatch(action);
    this.scannedActions$.next(action);
  }

  addReducer() {
    // noop
  }

  removeReducer() {
    // noop
  }
}

export function provideMockStore<T>(config: { initialState?: T } = {} as T) {
  return [
    { provide: INITIAL_STATE, useValue: config.initialState },
    ActionsSubject,
    ScannedActionsSubject,
    MockState,
    { provide: ReducerManager, useClass: MockReducerManager },
    {
      provide: Store,
      useClass: MockStore,
    },
  ];
}`

just a suggestion. Your thoughts?

ef32:
@brandonroberts  your proposal for the testing module above seems like a perfect fit, the selectors do not need to be mocked; by providing state initially, the required selectors should be able to work with the store. That said, here are some suggestions to slightly enhance your proposal.

the observable source field is now deprecated and in my opinion, i dont think it needs to be populated in this scenario.

- StateObservable being an observable, can be substituted using MockState. No need to provide with useClass.

- MockState can be generic, requiring the same generic Type as MockStore hence providing strong type checking.

- Initial state can be initialised by calling the next method with the initialstate token in the MockStore constructor.

- Test suites could require additional data to state, so setState can cater for such scenarios.

please find below slight modification:

```typescript

import { Injectable, Inject } from '@angular/core';
import {
StateObservable,
Store,
ReducerManager,
ActionsSubject,
ActionReducer,
Action,
ScannedActionsSubject,
INITIAL_STATE,
} from '@ngrx/store';
import { BehaviorSubject } from 'rxjs';

@Injectable()
export class MockState<T> extends BehaviorSubject<T> {
constructor() {
super({} as T);
}
}

@Injectable()
export class MockReducerManager extends BehaviorSubject<
ActionReducer<any, any>

{
constructor() {
super(() => undefined);
}
}

@Injectable()
export class MockStore<T> extends Store<T> {

// not needed
// private state$ = new BehaviorSubject(this.initialState as T);

constructor(
state$: MockState<T>,
actionsObserver: ActionsSubject,
reducerManager: ReducerManager,
public scannedActions$: ScannedActionsSubject,
@Inject(INITIAL_STATE) private initialState: T
) {
super(state$, actionsObserver, reducerManager);

//deprecated
// this.source = this.state$.asObservable();

this.state$.next(this.initialState);
}

setState(state: T): void {
const newState = {
...this.initialState,
...state
};
this.state$.next(newState);
}

dispatch(action: V) {
super.dispatch(action);
this.scannedActions$.next(action);
}

addReducer() {
// noop
}

removeReducer() {
// noop
}
}

export function provideMockStore(config: { initialState?: T } = {} as T) {
return [
{ provide: INITIAL_STATE, useValue: config.initialState },
ActionsSubject,
ScannedActionsSubject,
MockState,
{ provide: ReducerManager, useClass: MockReducerManager },
{
provide: Store,
useClass: MockStore,
deps:[MockState]
},
];
}
```

just a suggestion. Tested with my test suites and it works. Your thoughts?

brandonroberts:
@ef32 for the most part it looks good to me.

I'd rather leave it up the developer on the `state` they want to set beyond the initial state.
The `deps` isn't needed for the provider object.

nasreddineskandrani:
@brandonroberts
`
We shouldn't focus on mocking selectors. If you want to test selectors, provide a state that the selectors can use or use StoreModule.forRoot() with reducers for integration.
`

I totally agree with you testing component this way will work and it's better that trigger actions :+1: 
so `setState` solve the original main issue

But from your statement, i wanted to say that in my case i don't want to test selectors when i want to unit test components. Becasue integration tests are slower and should be fewer than unit tests to fail faster.

Also there is a way `projector` built-in in ngrx to unit tests selectors to be consitent i saw a unit test ability in the lib for components as a nice to have.
https://blog.angularindepth.com/how-i-test-my-ngrx-selectors-c50b1dc556bc
conclusion section:
> A selector can be unit tested, you don’t need a (mocked) store in order to test your selectors.

By your statement you don't provide a way to unit test components (doc update or code update) and encourage integration tests with real selectors to test components. 
I am insisting again about the fact it's slower and it may be a problem for some.

thank you for the answer and i am happy 'setState` is going in to stop triggering actions to be able to tests a components

maxime1992:
Hey, joining the discussion because there is one point where I have a different opinion: 

> your proposal for the testing module above seems like a perfect fit, the selectors do not need to be mocked; by providing state initially, the required selectors should be able to work with the store

and I think it might be worth to have a solution for that :)

I've brought a [piece code](https://github.com/ngrx/platform/issues/915#issuecomment-414601899) where we can manage and mock selectors really easily.
Initially, I thought like you that it would not be a great idea and that providing a state was a better idea. Except that in some cases it's definitely not handy.

Example:

Let say you have a store which contains:
- Cities
- Companies
- Product categories
- Products

Based on some filters given by the user, you want to display for a given city, company and product category, the total sum of the products they have. In that case you'll have to provide a substantial amount of code/mocked data to have the right state. Where all you really want for this integration test is a number.

nasreddineskandrani:
nice example @maxime1992! i ll keep it in my hat

TeoTN:
Hey @brandonroberts, I like your ideas, especially getting isolated from any reducers. (I don't remember now why I've done it that way, maybe I missed some knowledge). Commit is on its way. :)

TeoTN:
I'd appreciate some help with configuring Bazel properly... :)

modules/store/testing/src/mock_reducer_manager.ts
```diff
@@ -0,0 +1,12 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
```
brandonroberts: The import here should just be `rxjs`

modules/store/testing/src/mock_state.ts
```diff
@@ -0,0 +1,9 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
```
brandonroberts: The import here should just be `rxjs`

modules/store/testing/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './src/testing';
```
brandonroberts: Use explicit exports instead of wildcard exports

modules/store/testing/src/mock_store.ts
```diff
@@ -0,0 +1,39 @@
...
+  ActionsSubject,
+  INITIAL_STATE,
+  ReducerManager,
+  StateObservable,
+  Store,
+} from '@ngrx/store';
+
+@Injectable()
+export class MockStore<T> extends Store<T> {
+  private stateSubject = new BehaviorSubject<T>({} as T);
```
brandonroberts: Remove stateSubject as it isn't needed?

modules/store/testing/src/mock_store.ts
```diff
@@ -0,0 +1,39 @@
...
+  public scannedActions$: Observable<Action>;
+
+  constructor(
+    private state$: StateObservable,
```
brandonroberts: Type with `MockState<T>` instead of `StateObservable` and remove the additional type casting below

modules/store/spec/store.spec.ts
```diff
@@ -432,4 +433,46 @@ describe('ngRx Store', () => {
...
+
+    it('should allow tracing dispatched actions', () => {
+      const action = { type: INCREMENT };
+      mockStore.scannedActions$
+        .pipe(skip(1))
+        .subscribe(scannedAction => expect(scannedAction).toEqual(action));
+      mockStore.dispatch(action);
+    });
+
+    it('should allow setting a spy on dispatch method', () => {
```
brandonroberts: Remove this test

brandonroberts:
Thanks for all your work on this @TeoTN! We will revisit mocking selectors since we are going to keep `store.select` for now.

nasreddineskandrani:
thank's for all this work
@brandonroberts the testing doc is updated with this?

brandonroberts:
@nasreddineskandrani the testing in the example app has not been updated yet.

nasreddineskandrani:
hi @brandonroberts,

--- Doc-update
I saw that you moved the documentation to: https://ngrx.io/
For example if i navigate into the [doc](https://github.com/ngrx/platform/tree/master/docs/effects) folder of effects the message:
`
The documentation below is for NgRx versions 6.x and older.
Visit https://ngrx.io for the latest documentation.
` 
is present.

This means the documentation can only be updated by you (the members of ngrx team)?

--- Example-app update
you said:
> the testing in the example app has not been updated yet.
do we want to change all tests to this way? or just have one isolated example of it?

brandonroberts:
@nasreddineskandrani no, it does not. Anyone can still edit the docs. They are just accessed from the site now. You can still edit any of the content for the docs in the `projects/ngrx.io/content` folder. You can also directly edit any page from the site itself by clicking on the pencil in the top right corner of any page.

All the test documentation will be updated to use the testing module. The testing guide will be updated with better documentation on testing.

nasreddineskandrani:
@brandonroberts i am checking some major issues in https://github.com/jest-community/vscode-jest
I'll come back to this update on example app if no one do it in 2 weeks. Can we keep track of it with a dedicated github issue? possibly doing it step by step not one big PR. What do you think?

brandonroberts:
Sure. The guide rewrites are being tracked here https://github.com/ngrx/platform/issues/1383

PR #1032, opened 2018-05-08
feat(Store): add ng update

### LINKED ISSUE #1028: Add ng update support
## I'm submitting a...

<pre><code>
[ ] Bug report  
[x] Feature request
</code></pre>

When running `ng update` the `ngrx/**` library is ignored.

## Expected behavior:
To have `ngrx/**` v6 installed when running `ng update` on a project that has a dependency to ngrx v5.2.

I've got a proposition, see [my commit here](https://github.com/tdeschryver/platform/commit/d472795535a5296633ffaf6dd3bda6468e0ea0a1).
To achieve this I've taken a look at rxjs and material because I didn't find any docs, but be aware that I haven't tried it out (yet) tho :sweat_smile:

tdeschryver:
Closes #1028

### DIFF
--- a/modules/store/migrations/6_0_0/index.ts
+++ b/modules/store/migrations/6_0_0/index.ts
@@ -30,7 +30,7 @@ export default function(): Rule {
       const suffix = match(firstChar, '^') || match(firstChar, '~');
 
       // TODO: remove beta
-      pkg.dependencies['@ngrx/store'] = `${suffix}6.0.0@beta.1`;
+      pkg.dependencies['@ngrx/store'] = `${suffix}6.0.0-beta.2`;
       tree.overwrite(pkgPath, JSON.stringify(pkg, null, 2));
       context.addTask(new NodePackageInstallTask());
     }
--- a/modules/store/migrations/6_0_0/index.ts
+++ b/modules/store/migrations/6_0_0/index.ts
@@ -1,0 +1,44 @@
+import {
+  Rule,
+  SchematicContext,
+  Tree,
+  SchematicsException,
+  chain,
+} from '@angular-devkit/schematics';
+import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
+
+export default function(): Rule {
+  return (tree: Tree, context: SchematicContext) => {
+    const pkgPath = '/package.json';
+    const buffer = tree.read(pkgPath);
+    if (buffer == null) {
+      throw new SchematicsException('Could not read package.json');
+    }
+    const content = buffer.toString();
+    const pkg = JSON.parse(content);
+
+    if (pkg === null || typeof pkg !== 'object' || Array.isArray(pkg)) {
+      throw new SchematicsException('Error reading package.json');
+    }
+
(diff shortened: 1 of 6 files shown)


modules/store/migrations/6_0_0/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (!pkg.dependencies) {
+      pkg.dependencies = {};
+    }
+
+    if (pkg.dependencies['@ngrx/store']) {
+      const firstChar = pkg.dependencies['@ngrx/store'][0];
+      const suffix = match(firstChar, '^') || match(firstChar, '~');
+
+      // TODO: remove beta
+      pkg.dependencies['@ngrx/store'] = `${suffix}6.0.0@beta.1`;
```
brandonroberts: Why wouldn't this be `6.0.0-beta.1` since you're just modifying the `package.json`? Also go ahead and change to to `beta.2` since that will be the next release.

modules/store/package.json
```diff
@@ -19,10 +19,13 @@
   "homepage": "https://github.com/ngrx/platform#readme",
   "peerDependencies": {
     "@angular/core": "NG_VERSION",
-    "rxjs": "RXJS_VERSION"
+    "rxjs": "RXJS_VERSION",
+    "@angular-devkit/core": "NG_DEVKIT_VERSION",
```
brandonroberts: Remove the devkit peer dependencies. If you're using the Angular CLI or the Schematics CLI, we can rely on those dependencies implicitly.

brandonroberts: Also, add the `NG_DEVKIT_VERSION` replaces to the `modules/schematics/package.json` and bump the version to `^0.6.0`

modules/store/migrations/6_0_0/index.ts
tdeschryver: Yep you're totally right!

modules/store/package.json
tdeschryver: Good to know, thanks!

brandonroberts:
Is this PR ready to land? Or are you going to add it to the other packages separately?

tdeschryver:
Any idea of this is testable locally? Because I wanted to ng update, a new project to see if it would update the ngrx store version.

About the other packages, what would you prefer. A PR per package or should I add them all in this PR?

brandonroberts:
Testing locally requires a local npm server. There is an article here http://www.softwarearchitekt.at/post/2018/04/17/seamlessly-updating-your-angular-libraries-with-ng-update.aspx

tdeschryver:
OK, I think everything is fine.
Fyi, when installing the following error is thrown, but can be ignore when using `--force`.
`Package "@ngrx/router-store" has an incompatible peer dependency to "@ngrx/store" (requires "v6.0.0-beta.1", would install "6.0.0-beta.2").`

brandonroberts:
Cool. I'll merge in a bit. I have a follow-up branch ready so we can reuse it across the other packages pretty easily.

PR #1057, opened 2018-05-12
fix(build): Use default build for schematics

brandonroberts:
Fixes the replacements when building @ngrx/schematics
Moves `@ngrx/store` to first in the `packageGroup`, which enables use of `ng update @ngrx/store` to update all installed @ngrx packages


tdeschryver:
Oh cool I didn't know this was possible! :smile:

PR #1067, opened 2018-05-14
feat: implement ng add for store and effects packages

vitaliy-bobrov:
Implementation of #920 WIP.

This is the implementation of @ngrx/store ng-add module to get feedback, cause other `ng-add`s will be done in a similar way.

Done:
- created schematics "submodule" inside @ngrx/store with needed build configuration
- implemented optional addition `@ngrx/store` as `dependency` to `package.json` (idea from https://github.com/angular/material2/tree/master/src/lib/schematics)
- store `ng-add` doesn't import and initialize `StoreDevtoolsModule` (imho should be done ny `store-devtools` ng-add)

Additional code share optimizations could be performed later, after initial implementation.

Will work on other modules after gathering a feedback.


modules/store/schematics/tsconfig-build.json
```diff
@@ -0,0 +1,18 @@
+{
```
brandonroberts: Remove this file. The `npm_package` already has a default tsconfig.json` to use

modules/store/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  addImportToModule,
+  buildRelativePath,
+  findModuleFromOptions,
+  getProjectPath,
+  insertImport,
+  stringUtils,
+  addPackageToPackageJson,
+  platformVersion,
+  parseName,
+} from '../../schematics-core';
```
brandonroberts: Use the `@ngrx/schematics/schematics-core` import here

modules/store/schematics/ng-add/index.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const projectPath = getTestProjectPath();
+
+  let appTree: UnitTestTree;
+
+  beforeEach(() => {
+    appTree = createWorkspace(schematicRunner, appTree);
+  });
+
+  it('should update package.json', () => {
```
brandonroberts: Add a test for the `skipPackageJson` option

modules/store/package.json
```diff
@@ -19,8 +22,11 @@
   "homepage": "https://github.com/ngrx/platform#readme",
   "peerDependencies": {
     "@angular/core": "NG_VERSION",
+    "@angular-devkit/core": "NG_DEVKIT_VERSION",
```
brandonroberts: Remove these from the `package.json` as we are treating them as implicit dependencies instead of adding additional dependencies to `@ngrx/store`

.prettierignore
```diff
@@ -1,5 +1,7 @@
 /dist
 /modules/schematics/src/*/files/*
+/modules/schematics-core/templates/*
```
brandonroberts: Change `templates` to `files` so its consistent with schematics. If we want to change `files` to `templates` across the board, we'll do that in a separate PR.

modules/store/schematics/BUILD
```diff
@@ -0,0 +1,34 @@
+package(default_visibility = ["//visibility:public"])
```
brandonroberts: This should be added to the `packages` section in the `modules/store/BUILD` file in the `npm_package`. That way its output gets copied into the distribution.

brandonroberts:
I left some feedback. I'd rather have 2 copies of the store template files, then put them into the `schematics-core` and have 6 copies in places where they will never be used.

vitaliy-bobrov:
Thanks for the review, will work on changes. Regarding templates, I also not sure about keeping them in core, just thought that it could be good to have the only source.

brandonroberts:
Since its only one file, I think it will be ok to have 2 copies.

modules/store/package.json
```diff
@@ -21,6 +24,7 @@
     "@angular/core": "NG_VERSION",
     "rxjs": "RXJS_VERSION"
   },
+  "schematics": "./schematics/collection.json",
```
brandonroberts: Add this replacement to the `/tools/defaults.bzl` file

modules/store/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,144 @@
...
+      sourceText,
+      ts.ScriptTarget.Latest,
+      true
+    );
+
+    const statePath = `${options.path}/${options.statePath}`;
+    const relativePath = buildRelativePath(modulePath, statePath);
+    const environmentsPath = buildRelativePath(
+      statePath,
+      `/${options.path}/environments/environment`
```
brandonroberts: Update this to match https://github.com/ngrx/platform/blob/master/modules/schematics/src/store/index.ts#L57

modules/store/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  addImportToModule,
+  buildRelativePath,
+  findModuleFromOptions,
+  getProjectPath,
+  insertImport,
+  stringUtils,
+  addPackageToPackageJson,
+  platformVersion,
+  parseName,
+} from '@ngrx/schematics/schematics-core';
```
brandonroberts: This should be `@ngrx/store/schematics-core`

modules/store/package.json
brandonroberts: Add this replacement to `/tools/default.bzl`

brandonroberts:
Added a couple more changes, otherwise I think this is ready to implement in the other packages.

Issue #1036, opened 2018-05-09
After add @ngrx/schematics to ng6 project, cli generate component with css instead of scss


kklimczak:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

## What is the current behavior?

When we add schematics from ngrx and change default schematics collection, cli stop generate components with scss files.
When we revert the default collection, cli start again generate components with scss files.
 
Of course, we set 
```json
"schematics": {
  "@schematics/angular:component": {
    "styleext": "scss"
  }
}
```
in angular.json file, also set with `@ngrx/schematics/angular:component`.

## Expected behavior:

When we use ngrx schematics as a default collection, cli should still generate scss, when it set.

## Minimal reproduction of the problem with instructions:
1. Generate project with `@angular/cli@6.0.0`
2. Follow this [instruction](https://github.com/ngrx/platform/tree/master/docs/schematics) to install ngrx with schematics and set it as default schematics collection.
3. Set SCSS as default css preprocessor in `angular.json` file using schematics: 
```json
"schematics": {
  "@schematics/angular:component": {
    "styleext": "scss"
  }
}
```
**Optional** try with `@ngrx/schematics/angular:component` instead of `@schematics/angular:component`
4. Generate component via CLI.

## Version of affected browser(s),operating system(s), npm, node and ngrx:
node: 8.11.1
npm: 6.0.0
dependencies:
```json
  "dependencies": {
    "@angular/animations": "^6.0.0",
    "@angular/common": "^6.0.0",
    "@angular/compiler": "^6.0.0",
    "@angular/core": "^6.0.0",
    "@angular/forms": "^6.0.0",
    "@angular/http": "^6.0.0",
    "@angular/platform-browser": "^6.0.0",
    "@angular/platform-browser-dynamic": "^6.0.0",
    "@angular/router": "^6.0.0",
    "@ngrx/core": "^1.2.0",
    "@ngrx/effects": "^5.2.0",
    "@ngrx/entity": "^5.2.0",
    "@ngrx/store": "^5.2.0",
    "@ngrx/store-devtools": "^5.2.0",
    "core-js": "^2.5.4",
    "rxjs": "^6.0.0",
    "rxjs-compat": "^6.1.0",
    "zone.js": "^0.8.26"
  },
  "devDependencies": {
[...]

PR #1074, opened 2018-05-16
docs(Schematics): add information about schematics settings (#1036)

kklimczak:
PR for #1036 
I added information of additional configuration when default collection for schematics changed to `@ngrx/schematics` and example of that.


brandonroberts:
Thanks!

PR #1080, opened 2018-05-19
Refactor(schematics-core): remove npm install task during ng update

tdeschryver:
I did some digging into the Angular CLI and noticed we don't need to add the install task `NodePackageInstallTask` during the `ng update` command.
Basically the CLI installs the package automatically except when the user uses the `--migrate-only` flag.
If we don't remove the install task in ngrx it would cause problems if a user uses the `--migrate-only` flag (it will try to update but it can't find the new files in order to update).

While I was there I also:
* removed the `TODO: remove beta` comment
* use the `version` param in `create-package`, instead of hard coding version 5.2.0
* removed `configPath` and `getConfig` inside `config.ts` because this was core for the old `.angular-cli` (and it isn't used anymore)
* removed `getAppFromConfig`  inside `config.ts` because it isn't used
* removed `CliConfig`  inside `config.ts` because it isn't used
* removed the file `get-file-content` because it isn't used - should this be used in favor of `tree.readContent()` ?


tdeschryver:
I maybe should have asked first if this was OK tho... :sweat_smile:

brandonroberts:
Changes LGTM. Just so we can track changes, if you have some unrelated changes, use a separate commit with each set of changes. We can squash them together when merging if needed.

Issue #1079, opened 2018-05-19
RFC: StackBlitz ngrx seed


tdeschryver:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Feature request
[x] Documentation issue or request
</code></pre>

## What is the current behavior?
The issue template currently has a link to an old plunkr template.

## Expected behavior:
If we create a StackBlitz ngxr seed, this would maybe lower the effort to create a repro for an issue.
Currently I've ported the plunkr template to StackBlitz which you can find at [https://stackblitz.com/edit/ngrx-seed](https://stackblitz.com/edit/ngrx-seed).

Thoughts?
Would this be enough, should something be added/removed?

vincecampanale:
I think this is a great idea. Some thoughts:

I'm of two minds when it comes to the seed application structure. When I initially went to [https://stackblitz.com/edit/ngrx-seed](https://stackblitz.com/edit/ngrx-seed), I was looking for actions, effects, reducers, etc. in their own folders, as one would find them in most apps with NgRx. So my first feedback was that maybe we should divide them up like that to make it easier to navigate. That being said, the point of the seed is to be minimal and small, so having all the store stuff in one file makes sense. Maybe we call the directory something other than `reducers`? Would `store` work?

Also, we may want to include a `fakeAsync` util with timeouts and such to mock an HTTP request or something. Pairing that with a simple effect and some `Fail` and `Success` actions would make it easier to reproduce issues that involve asynchronous code, which I would imagine are fairly common.

tdeschryver:
Seems like I forgot to save after I moved the actions to the actions folder :sweat_smile: - now it's there.

tdeschryver:
Closing, because it's implemented in #1100 .

PR #1100, opened 2018-05-27
docs: Add issue templates

brandonroberts:
Add some custom issue templates to help with filing issues and requests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,27 @@
+---
+name: Bug Report
+about: Report a bug or regression in functionality
+---
+
+<!-- ❤️ ngrx? Please consider supporting our collective: 👉  [donate](https://opencollective.com/ngrx/donate) -->
+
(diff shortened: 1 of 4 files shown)


tdeschryver:
Awesome! 
Would the [StackBlitz ngrx-seed](https://stackblitz.com/edit/ngrx-seed) from #1079 be useful to add?
If it would, I was thinking since there wasn't much response, we could ask if the seed was helpful or if there was something they were missing in it and update the seed accordingly until it is 'perfect'.

tdeschryver:
I've added the store devtools integration. I thought it wouldn't work in StackBlitz, but it does ... StackBlitz never disappoints ✨ 

I also added the logger meta reducer.

This may seem stupid but what do you mean with `counter actions need to be fixed` 😅?
I did rename the action types if this is what you meant.

brandonroberts:
That's what I mean. Couple more things. Remove the "Start edting" line, fix the spacing in `AppModule` imports and Store Devtools name (NgRg) and then I think its set for now.

tdeschryver:
Done, done and done. 
Thanks for the review.

Issue #250, opened 2017-08-08
Error: Cannot convert undefined or null to object


ValeryVS:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
</code></pre>

## What is the current behavior?

With AOT build Error `Cannot convert undefined or null to object` is thrown at this line
https://github.com/ngrx/platform/blob/48a2381c212d5dd3fa2b9435776c1aaa60734235/modules/store/src/utils.ts#L16
because `reducers` is `null`.
```typescript
  const reducerKeys = Object.keys(reducers);
```

## Expected behavior:

I'm injecting reducers as described here
https://github.com/ngrx/platform/blob/master/docs/store/api.md#injecting-reducers

## Minimal reproduction of the problem with instructions:

https://github.com/ValeryVS/angular-storage-experiment/blob/02fd926e17547abfdb213b3d8dd3af3392218238/src/app/store/reducer.ts
Same error has been reproduced there.
I can update repo to match instruction above.

## Version of affected browser(s),operating system(s), npm, node and ngrx:
$ node -v
v6.11.1
$ npm -v
3.10.10

@ngrx/store@4.0.2

ValeryVS:
Possible relative #175

Christian24:
I have the same issue. I am on Typescript 2.4.2. Disabling AOT fixes the issue.  My getReducers function is not executed (I added a console.log).

ValeryVS:
Right, I also tried to debug this a while, and found that reducers factory is not executed.

Christian24:
Wouldn't that mean it is more a CLI issue though? AOT is only active in production mode right? Because everything works as expected in development mode.

ValeryVS:
@Christian24 
Well... it can be issue with ngrx/store or documentation or with angular itself. Don't think that CLI is relative here.
You can `ng build --prod --aot=false` to make production build without though.

Christian24:
@ValeryVS true. I have tried that. When disabling AOT it works, but I would hate to give up AOT.

brandonroberts:
I have a fix for this that will work for JiT/AoT. It should land soon once I submit the PR

Christian24:
While the error goes away, my reducer still is not called. I detailed this in #189

tschuege:
@ValeryVS 
I have the same issue. 
- Angular 4.3.5
[...]

PR #1101, opened 2018-05-28
fix(store): fix Object.keys being called on null or undefined

David-Mueller:
This should mitigate #250

### DIFF
--- a/modules/store/src/utils.ts
+++ b/modules/store/src/utils.ts
@@ -14,7 +14,7 @@ export function combineReducers(
   reducers: any,
   initialState: any = {}
 ): ActionReducer<any, Action> {
-  const reducerKeys = Object.keys(reducers);
+  const reducerKeys = reducers ? Object.keys(reducers) : [];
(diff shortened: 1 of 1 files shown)


tdeschryver:
While this might resolve the errors, I don't think this is desired?
Because I expect you would always need reducers, I could be wrong tho.
Maybe an error/warning would be more suitable?

Just my 2 cents 😅

David-Mueller:
Hi @tdeschryver,

Thanks for taking a look onto this!

While investigating further, this seems to be related to using lazy reducers within an Angular 6 AOT-enabled build. Somehow, the feature reducers seem not to be injected properly (they are resolved as `null`) in my case... The proposed patch should probably not be applied but the real cause of calling `reducerManager.addFeature(null)` be eliminated...

tdeschryver:
I didn't meant to be the cause to close this PR, because this might actually be desired.
Let's wait what @brandonroberts or @MikeRyanDev have to say.

Issue #674, opened 2017-12-28
NgRx Schematics: Testing and Feedback


brandonroberts:
This post is to collect feedback on the implementation of schematics. This is not the final documentation, but allows those who want to test the schematics before official release. Subscribe to this issue if you want to be notified of updates.

## Installation Instructions

```shell
yarn add github:ngrx/schematics-builds --dev
```

## Set as Default Collection

In your `.angular-cli.json` file, add the following to the `defaults` section. If you it as set the default, you don't need to add `--collection @ngrx/schematics` to each blueprint command

```json
    "schematics": {
      "collection": "@ngrx/schematics"
    }
```

## Blueprints

Generate the root state. Specifying the module will update the file with the imports for the root State, and the Store Devtools.

```shell
ng generate store State --root --module app.module.ts --collection @ngrx/schematics
```

Generate a feature state. Specifying the module will update the file with the import for the feature State.

```shell
ng generate store path/to/Feature --module feature.module.ts --collection @ngrx/schematics
```

Generate an actions file.

```shell
ng generate action User --collection @ngrx/schematics
```

Generate a reducer and register the reducer in a module with `StoreModule.forFeature`.

```shell
ng generate reducer User --module path/to/to/feature.module.ts --collection @ngrx/schematics
```

Generate a reducer and register the reducer in a reducers file. The reducer state interface will be added to the first defined interface in the file and the reducer added to the map of reducers defined with a type of `ActionReducerMap`.

```shell
ng generate reducer User --reducers path/to/reducers/index.ts --collection @ngrx/schematics
```

Generate an effect and register the reducer in a module with `StoreModule.forFeature`.

```shell
ng generate effect User --module path/to/to/feature.module.ts --collection @ngrx/schematics
```

Generate a _feature_ (actions file, reducer, effect) and register the reducer in a reducers file. The reducer state interface will be added to the first defined interface in the file and the reducer added to the map of reducers defined with a type of `ActionReducerMap`.

```shell
ng generate feature User --reducers path/to/reducers/index.ts --collection @ngrx/schematics
```

Generate a _feature_ (actions file, reducer, effect) and register the reducer in a module with `StoreModule.forFeature`.

```shell
ng generate feature User --module path/to/to/feature.module.ts --collection @ngrx/schematics
```

Generate an entity and register the reducer in a module with `StoreModule.forFeature`.

```shell
ng generate entity User --module path/to/to/feature.module.ts --collection @ngrx/schematics
```

[...]

PR #1111, opened 2018-06-06
Fix command for generate an `Admin` feature state

voidek-work:
https://github.com/ngrx/platform/issues/674
> @crimsonskyfalling this path is implicit. The path you provide for the feature will be the same path provided by the module.

### DIFF
--- a/docs/schematics/store.md
+++ b/docs/schematics/store.md
@@ -56,7 +56,7 @@ Generate an `Admin` feature state within the `admin` folder and register it with
 
 ```sh
 ng generate module admin --flat false
-ng generate store admin/Admin -m admin/admin.module.ts
+ng generate store admin/Admin -m admin.module.ts
 ```
 
 Generate the initial state management files within a `store` folder and register it within the `app.module.ts`.


tdeschryver:
Which version of the CLI are you using? 
Because I still have to do `admin/Admin`.

I'm on v6.0.7

voidek-work:
Angular CLI: 6.0.5
Node: 9.11.1
OS: win32 x64
Angular: 6.0.3

ng generate store **admin/Admin** -m _admin.module.ts_

voidek-work:
I wanted to say the second path is relative to the first path, but the documentation says relative src

tdeschryver:
Oops, I clearly wasn't paying attention 😅 
You're totally right!

tdeschryver:
LGTM.

@brandonroberts just to be sure, it's OK if I 'accept' this PR since you've assigned me to it?

brandonroberts:
@tdeschryver yep

tdeschryver:
Thanks!

PR #488, opened 2017-10-16
Updated store-devtools section of MIGRATION.md

glenngr:
Updated note in @ngrx/store-devtools with updated information after fix of issue #97

PR #1117, opened 2018-06-09
Update note on issue #97

oboukli:
- Update note on issue #97 in MIGRATION.md
- Update examples to RxJS 6
- Edit and clean up

See also #488.


brandonroberts:
I don't see a reason to upgrade these to pipeable operators. Its a migration guide to V4 and RxJS 5.4, which still had support for chainable operators.

brandonroberts:
We can keep this migration doc as-is.

PR #1121, opened 2018-06-11
Allow custom serializer

alex-okrushko:
`serialize` can be boolean or object. One of the teams needs to provide custom replacer.

Adjusting the `serialize` according to the docs for extension:

https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize

### DIFF
--- a/modules/store-devtools/src/config.ts
+++ b/modules/store-devtools/src/config.ts
@@ -3,7 +3,7 @@ import { InjectionToken, Type } from '@angular/core';
 
 export type ActionSanitizer = (action: Action, id: number) => Action;
 export type StateSanitizer = (state: any, index: number) => any;
-export type SerializeOptions = {
+export type SerializationOptions = {
   options?: boolean | any;
   replacer?: (key: any, value: any) => {};
   reviver?: (key: any, value: any) => {};
@@ -17,7 +17,7 @@ export class StoreDevtoolsConfig {
   actionSanitizer?: ActionSanitizer;
   stateSanitizer?: StateSanitizer;
   name?: string;
-  serialize?: boolean | SerializeOptions;
+  serialize?: boolean | SerializationOptions;
   logOnly?: boolean;
   features?: any;
 }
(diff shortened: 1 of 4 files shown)


modules/store-devtools/src/instrument.ts
```diff
@@ -60,7 +60,7 @@ export function createConfig(
     actionSanitizer: undefined,
     stateSanitizer: undefined,
     name: DEFAULT_NAME,
-    serialize: false,
+    serialize: undefined,
```
brandonroberts: This potentially would be a breaking change, since the existing behavior is to handle circular references. We should have changed with the default serializer for router-store.

modules/store-devtools/src/config.ts
```diff
@@ -3,14 +3,21 @@ import { InjectionToken, Type } from '@angular/core';
 
 export type ActionSanitizer = (action: Action, id: number) => Action;
 export type StateSanitizer = (state: any, index: number) => any;
+export type SerializeOptions = {
```
brandonroberts: I'd prefer `SerializationOptions` here

modules/store-devtools/src/instrument.ts
alex-okrushko: As far as I can tell, `undefined` is [recommended default](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize) (will use regular JSON.stringify to send data, and it's the fasted mode).
There are concerns that attempting to resolve circular references with [JSPB](https://developers.google.com/protocol-buffers/docs/reference/javascript-generated) might crash the extension.

brandonroberts: 👍

brandonroberts:
I think we should leave it as `false` for now. You can configure it to be `undefined`. We can update the docs for now to recommend it and change the default in V7.

alex-okrushko:
Done

brandonroberts:
Thanks Alex!

PR #1137, opened 2018-06-18
feat(Store): feature name should be a key of the state

tdeschryver:
Right now you can type everything you want when calling the `createFeatureSelector` method.
This may result in vague errors because the state is undefined at runtime.
This PR should prevent these mistakes.

Result:
snippet


tdeschryver:
Woops, you can ignore this :sweat_smile: 
This is isn't compatible with `combineReducers`, my bad ...

PR #1140, opened 2018-06-21
fix(schematics): correct a type of action class generated

kouMatsumoto:
I used `ng generate action` command and found the generated file are invalid.
Please check the result I generated, thank you.

* Result I generated
```
import { Action } from '@ngrx/store';

export enum FooActionTypes {
  LoadFoos = '[Foo] Load Foos'
}

export class Foo implements Action {
  readonly type = FooActionTypes.LoadFoos;
}

export type FooActions = LoadFoos;
```

* Expected
```
import { Action } from '@ngrx/store';

export enum FooActionTypes {
  LoadFoos = '[Foo] Load Foos'
}

export class Foo implements Action {
  readonly type = FooActionTypes.LoadFoos;
}

export type FooActions = Foo; // <<< this should be Action class name!
```

### DIFF
--- a/modules/schematics/src/action/index.spec.ts
+++ b/modules/schematics/src/action/index.spec.ts
@@ -69,7 +69,7 @@ describe('Action Schematic', () => {
     expect(fileContent).toMatch(/export enum FooActionTypes/);
   });
 
-  it('should create a class named "LoadFoos"', () => {
+  it('should create a class based on the provided name', () => {
     const tree = schematicRunner.runSchematic(
       'action',
       defaultOptions,
@@ -82,7 +82,7 @@ describe('Action Schematic', () => {
     expect(fileContent).toMatch(/export class LoadFoos implements Action/);
   });
 
-  it('should create a type named "FooActions"', () => {
+  it('should create the union type based on the provided name', () => {
     const tree = schematicRunner.runSchematic(
       'action',
       defaultOptions,
--- a/modules/schematics/src/action/index.spec.ts
+++ b/modules/schematics/src/action/index.spec.ts
@@ -69,7 +69,7 @@ describe('Action Schematic', () => {
     expect(fileContent).toMatch(/export enum FooActionTypes/);
   });
 
-  it('should create an type named "FooActions"', () => {
+  it('should create a class named "LoadFoos"', () => {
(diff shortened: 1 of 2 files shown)


MatsumotoKou:
In advance, I suppose the generated action class name `Foo` should be `LoadFoos`.

* Expected
```
import { Action } from '@ngrx/store';

export enum FooActionTypes {
  LoadFoos = '[Foo] Load Foos'
}

export class LoadFoos implements Action {
  readonly type = FooActionTypes.LoadFoos;
}

export type FooActions = LoadFoos;
```

tdeschryver:
@MatsumotoKou Indeed, it should be generated as `LoadFoos`:

```
import { Action } from '@ngrx/store';

export enum FooActionTypes {
  LoadFoos = '[Foo] Load Foos'
}

export class LoadFoos implements Action {
  readonly type = FooActionTypes.LoadFoos;
}

export type FooActions = LoadFoos;
```

MatsumotoKou:
I will commit additionally if needed, thank you.

modules/schematics/src/action/index.spec.ts
```diff
@@ -69,6 +69,32 @@ describe('Action Schematic', () => {
     expect(fileContent).toMatch(/export enum FooActionTypes/);
   });
 
+  it('should create a class named "LoadFoos"', () => {
```
brandonroberts: This should be more like `should create a class based on the provided name`

modules/schematics/src/action/index.spec.ts
```diff
@@ -69,6 +69,32 @@ describe('Action Schematic', () => {
...
+      appTree
+    );
+    const fileContent = tree.readContent(
+      `${projectPath}/src/app/foo.actions.ts`
+    );
+
+    expect(fileContent).toMatch(/export class LoadFoos implements Action/);
+  });
+
+  it('should create a type named "FooActions"', () => {
```
brandonroberts: This should be more like `should create the union type based on the provided name`

kouMatsumoto:
@brandonroberts 
Thank you for pointing that out.
I revised it based on your comments.
Please continue the review, thank you!

brandonroberts:
Thanks!

PR #1154, opened 2018-06-26
feat(Schematics): use ofType operator function instead of Actions#ofType

kouMatsumoto:
I suppose that we should use ofType operator function instead of Actions#ofType in generated files by schematics.
Please consider this change, thank you!

### DIFF
--- a/modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
+++ b/modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
@@ -1,12 +1,12 @@
 import { Injectable } from '@angular/core';
-import { Actions, Effect } from '@ngrx/effects';
-<% if(feature) { %>import { <%= classify(name) %>Actions, <%= classify(name) %>ActionTypes } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
+import { Actions, Effect<% if(feature) { %>, ofType<% } %> } from '@ngrx/effects';
+<% if (feature) { %>import { Load<%= classify(name) %>s, <%= classify(name) %>ActionTypes } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
 
 @Injectable()
 export class <%= classify(name) %>Effects {
-<% if(feature) { %>
+<% if (feature) { %>
   @Effect()
-  effect$ = this.actions$.ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s);
+  effect$ = this.actions$.pipe(ofType<Load<%= classify(name) %>s>(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s));
 <% } %>
   constructor(private actions$: Actions) {}
 }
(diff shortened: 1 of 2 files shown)


modules/schematics/src/effect/index.spec.ts
```diff
@@ -50,6 +50,44 @@ describe('Effect Schematic', () => {
...
+    );
+    expect(content).toMatch(
+      /import { Actions, Effect, ofType } from '@ngrx\/effects';/
+    );
+    expect(content).toMatch(
+      /import { LoadFoos, FooActionTypes } from '\.\/foo.actions';/
+    );
+    expect(content).toMatch(/export class FooEffects/);
+    expect(content).toMatch(
+      /effect\$ = this\.actions\$.pipe\(ofType<LoadFoos>\(FooActionTypes\.LoadFoos\)\);/
```
tdeschryver: Maybe change `effects$` to be `loadFoos$`?

kouMatsumoto: Thank you for your comment!
I renamed "effect$" to "loadFoos$".
It's more suitable in this case.

modules/schematics/src/effect/index.spec.ts
```diff
@@ -37,7 +37,7 @@ describe('Effect Schematic', () => {
     appTree = createWorkspace(schematicRunner, appTree);
   });
 
-  it('should create an effect', () => {
+  it('should create an effect file and its spec file', () => {
```
tdeschryver: This should be `should create an effect with a spec file`.

modules/schematics/src/effect/index.spec.ts
```diff
@@ -50,6 +50,44 @@ describe('Effect Schematic', () => {
     ).toBeGreaterThanOrEqual(0);
   });
 
+  it('should create an effect class which has a "loadFoos$" property if feature is set', () => {
```
tdeschryver: This should be `should create an effect that describes a source of actions within a feature`.

modules/schematics/src/effect/index.spec.ts
```diff
@@ -50,6 +50,44 @@ describe('Effect Schematic', () => {
...
+    expect(content).toMatch(
+      /import { LoadFoos, FooActionTypes } from '\.\/foo.actions';/
+    );
+    expect(content).toMatch(/export class FooEffects/);
+    expect(content).toMatch(
+      /loadFoos\$ = this\.actions\$.pipe\(ofType<LoadFoos>\(FooActionTypes\.LoadFoos\)\);/
+    );
+  });
+
+  it('should create an effect class which does not have a "loadFoos$" property if root is set', () => {
```
tdeschryver: This should be `should create an effect that does not define a source of actions within the root`.
And maybe move the test a bit down? (after the default tests)

MatsumotoKou:
@tdeschryver 
Thank you for review and explaining!
I updated it, please continue your checks.

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
```diff
@@ -1,12 +1,12 @@
...
+<% if (feature) { %>import { Load<%= classify(name) %>s, <%= classify(name) %>ActionTypes } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
 
 @Injectable()
 export class <%= classify(name) %>Effects {
-<% if(feature) { %>
+<% if (feature) { %>
   @Effect()
-  effect$ = this.actions$.ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s);
+  loadFoos$ = this.actions$.pipe(ofType<Load<%= classify(name) %>s>(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s));
 <% } %>
```
brandonroberts: Remove the `<Load<%= classify(name) %>s>`

kouMatsumoto:
@brandonroberts 
Thank you for the pointing out!
I update it, please continue your review.

PR #1158, opened 2018-06-27
Fix typo in CHANGELOG.md.

tplk:
Schematcis > Schematics


brandonroberts:
Thanks for the PR @tplk. The changelog is generated from commit messages, so since its already in master it will just get overwritten next time we generate one.

tplk:
Okay, cool.

Issue #1160, opened 2018-06-28
Confusing Destructuring in Entity Adapter Example


SeymoreThrottle:
This example would probably be a lot more clear if the following code was refactored to not use destructuring. I wasn't used to seeing this syntax, and this particular usage threw me for a loop for a bit. It would be handy to include the snippet as an optimization somewhere in the doc, perhaps after all of the example code.

```
export const {
  // select the array of user ids
  selectIds: selectUserIds,

  // select the dictionary of user entities
  selectEntities: selectUserEntities,

  // select the array of users
  selectAll: selectAllUsers,

  // select the total user count
  selectTotal: selectUserTotal,
} = adapter.getSelectors(); 
```

tdeschryver:
Would you like to make a pull request with your suggestion in order to make it clear?

SeymoreThrottle:
#1161 submitted

brandonroberts:
Closed via #1161

PR #1161, opened 2018-06-29
#1160 - Update Entity Adapter documentation for clarity

SeymoreThrottle:
Destructuring syntax in this example adds a layer of complexity that may trip up developers who aren't used to using it. Declaring the Entity Adapter selector methods more explicitly should help to keep the example focused on its original intent.

### DIFF
--- a/docs/entity/adapter.md
+++ b/docs/entity/adapter.md
@@ -261,19 +261,17 @@ export function reducer(state = initialState, action: UserActionsUnion): State {
 
 export const getSelectedUserId = (state: State) => state.selectedUserId;
 
-export const {
-  // select the array of user ids
-  selectIds: selectUserIds,
+// select the array of user ids
+export const selectUserIds = adapter.getSelectors().selectIds;
 
-  // select the dictionary of user entities
-  selectEntities: selectUserEntities,
+// select the dictionary of user entities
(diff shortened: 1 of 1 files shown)


brandonroberts:
I'd rather do the destructuring differently than call `getSelectors()` multiple times. What about this:

```
// get the selectors
const adapterSelectors = adapter.getSelectors();

// select the array of user ids
export const selectUserIds = adapterSelectors.selectIds;

// select the dictionary of user entities
export const selectUserEntities = adapterSelectors.selectEntities;

// select the array of users
export const selectAllUsers = adapterSelectors.selectAll;

// select the total user count
export const selectUserTotal = adapterSelectors.selectTotal;
```

or

```
// get the selectors
const { selectIds, selectEntities, selectAll, selectTotal } = adapter.getSelectors();

// select the array of user ids
export const selectUserIds = selectIds;

// select the dictionary of user entities
export const selectUserEntities = selectEntities;

// select the array of users
export const selectAllUsers = selectAll;

// select the total user count
export const selectUserTotal = selectTotal;
```

SeymoreThrottle:
That makes sense. I feel like either of those options work well. The second option seems a bit more clean. I Can change my PR if that works.

brandonroberts:
Yea, let's go with the second option.

brandonroberts:
Thanks!

PR #1162, opened 2018-06-30
style(Schematics): revise Effects class name in description of spec file

kouMatsumoto:
I think "FooEffects" is more suitable than "FooService" for description of effects spec file.
please review this change, thank you!


tdeschryver:
Thanks!

PR #1163, opened 2018-06-30
fix(Store): bootstrap store with partial initial state

### LINKED ISSUE #906: Store: Handling of @ngrx/store/update-reducers Action on Startup leads to strange behaviour for initialStates

tdeschryver:
This changes the way features are added and removed.
Instead of adding/removing feature by feature, this adds/removes them in all together in batch.

With consequence that will only dispatch `@ngrx/store/update-reducers` once, if this isn't what we want we can still dispatch every feature one by one.

The "original" `addFeature`, `removeFeature`, `addReducer`, `removeReducer` methods will loop through the corresponding `addFeatures`, `removeFeatures`, `addReducers`, `removeReducers` methods.

Will it be beneficial to also add `addReducers` and `removeReducers` in `store.ts`?

This closes #906 and closes #909


modules/store/src/utils.ts
```diff
@@ -1,117 +1,117 @@
-import {
```
tdeschryver: I don't know what happened here? 😅

brandonroberts: Revert these changes

modules/store/src/reducer_manager.ts
```diff
@@ -1,89 +1,118 @@
...
+    featureKeys.forEach(key => {
+      this.reducers = omit(this.reducers, key) /*TODO(#823)*/ as any;
+    });
+    this.updateReducers(featureKeys);
+  }
+
+  private updateReducers(keys: string[]) {
+    this.next(this.reducerFactory(this.reducers, this.initialState));
+
+    if (keys.length === 1) {
```
brandonroberts: This would be a breaking change. Currently there is one dispatch per reducer update.

brandonroberts:
Don't know what's going on with the whitespace changes. Update the commit message also

modules/store/src/reducer_manager.ts
tdeschryver: To be sure, we want to keep dispatching every update?

brandonroberts: Yes, dispatch individual updates for each key added. In V7, we'll change it dispatch one action.

PR #1165, opened 2018-07-02
chore(Example): remove tslint configuration conflicting with prettier

kouMatsumoto:
Now, tslint fails as described below.

```
$ ng lint

ERROR: /platform/example-app/app/shared/pipes/ellipsis.pipe.spec.ts[42, 6]: " should be '
```

We should use single quotes with escaping.

*ellipsis.pipe.spec.ts#L42
```
// current code
it("should return the string if it's length is less than 250", () => {
// expected
it('should return the string if it\'s length is less than 250', () => {
```

But this file is formatted by prettier as described below.

```
$ yarn run pritter

- it('should return the string if it\'s length is less than 250', () => {
+  it("should return the string if it's length is less than 250", () => {
```

So  I add tslint the "avoid-escape" option.
And ng lint command passes as expected.

```
$ ng lint

All files pass linting.
```

Also, we can use back-quotes instead double quotes.
In this way, we need not update tslint.json.

```
it(`should return the string if it's length is less than 250`, () => {
```

Please consider my change, thank you!

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -1,35 +1,11 @@
 {
   "rules": {
     "class-name": true,
-    "comment-format": [true, "check-space"],
-    "indent": [true, "spaces"],
     "no-duplicate-variable": true,
     "no-eval": false,
     "no-internal-module": true,
-    "no-trailing-whitespace": true,
     "no-var-keyword": true,
-    "one-line": [true, "check-open-brace", "check-whitespace"],
-    "quotemark": [true, "single", "avoid-escape"],
-    "semicolon": true,
     "triple-equals": [false, "allow-null-check"],
-    "typedef-whitespace": [
-      true,
-      {
-        "call-signature": "nospace",
-        "index-signature": "nospace",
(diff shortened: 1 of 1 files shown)


tdeschryver:
I would even suggest to remove the tslint configuration because prettier "is the boss" and will modify the code accordingly.

Based on [Ultimate Angular + Prettier Cheatsheet](https://medium.com/ngconf/ultimate-prettier-angular-cheatsheet-777c9515f4fb), these are the tslint configs that can be removed:
```
- comment-format
- curly
- eofline
- import-spacing
- indent
- max-line-length
- no-trailing-whitespace
- one-line
- quotemark
- semicolon
- typedef-whitespace
- angular-whitespace
```

kouMatsumoto:
That's right.
The above options are unnecessary and should be removed.

kouMatsumoto:
@tdeschryver 
I additionaly updated tslint.config and revise title of this pull-request.
Please consider this change, thank you.

PR #1169, opened 2018-07-03
feat(Entity): log a message when selectId returns undefined in dev mode

### LINKED ISSUE #1133: RFC: Provide a message when `selectId` is undefined
I think if we would log (or even throw) a message when `selectId` returns an `undefined` when using Entity module, it would increase the dev experience. 
Note that this would only be done when the application is in dev mode.

This message could be something like "ngrx/entity: the key 'id' does not exist in {...}, you can provide your own key with `selectId`." where `id` is the return value of `selectId`. The second part would only be needed if the developer doesn't override the default function.

Now the developer has to figure out why the ids in the store are all `undefined`.

## Other information:
This idea came after a question and discussion with @sandangel on the gitter channel.

[ ] No

timdeschryver:
Closes #1133 

Example of the message:

image

### DIFF
--- a/modules/entity/spec/utils.spec.ts
+++ b/modules/entity/spec/utils.spec.ts
@@ -1,14 +1,13 @@
 import * as ngCore from '@angular/core';
 import { selectIdValue } from '../src/utils';
-import { BookModel } from './fixtures/book';
+import { BookModel, AClockworkOrange } from './fixtures/book';
 
 describe('Entity utils', () => {
   describe(`selectIdValue()`, () => {
     it('should not warn when key does exist', () => {
       const spy = spyOn(console, 'warn');
 
-      const book: BookModel = { id: 'PC', title: 'Purple Cow' };
-      const key = selectIdValue(book, b => b.id);
+      const key = selectIdValue(AClockworkOrange, book => book.id);
 
       expect(spy).not.toHaveBeenCalled();
     });
@@ -16,8 +15,7 @@ describe('Entity utils', () => {
     it('should warn when key does not exist in dev mode', () => {
       const spy = spyOn(console, 'warn');
 
-      const book: BookModel = { id: 'PC', title: 'Purple Cow' };
-      const key = selectIdValue(book, (b: any) => b.foo);
+      const key = selectIdValue(AClockworkOrange, (book: any) => book.foo);
 
       expect(spy).toHaveBeenCalled();
     });
@@ -26,8 +24,7 @@ describe('Entity utils', () => {
       spyOn(ngCore, 'isDevMode').and.returnValue(false);
       const spy = spyOn(console, 'warn');
 
-      const book: BookModel = { id: 'PC', title: 'Purple Cow' };
-      const key = selectIdValue(book, (b: any) => b.foo);
+      const key = selectIdValue(AClockworkOrange, (book: any) => book.foo);
 
       expect(spy).not.toHaveBeenCalled();
     });
--- a/modules/entity/spec/utils.spec.ts
+++ b/modules/entity/spec/utils.spec.ts
@@ -20,11 +20,36 @@ describe('Entity utils', () => {
       expect(spy).toHaveBeenCalled();
     });
 
+    it('should warn when key is undefined in dev mode', () => {
+      const spy = spyOn(console, 'warn');
+
+      const undefinedAClockworkOrange = { ...AClockworkOrange, id: undefined };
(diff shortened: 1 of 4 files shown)


modules/entity/spec/utils.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    it('should warn when key does not exist in dev mode', () => {
+      const spy = spyOn(console, 'warn');
+
+      const key = selectIdValue(AClockworkOrange, (book: any) => book.foo);
+
+      expect(spy).toHaveBeenCalled();
+    });
+
+    it('should not warn when key does not exist in prod mode', () => {
+      spyOn(ngCore, 'isDevMode').and.returnValue(false);
```
brandonroberts: You think it would be better to just set prod mode here and reset it after the `selectIdValue` function call?

modules/entity/src/utils.ts
```diff
@@ -0,0 +1,18 @@
+import { isDevMode } from '@angular/core';
+import { IdSelector } from './models';
+
+export function selectIdValue<T>(entity: T, selectId: IdSelector<T>) {
+  const key = selectId(entity);
+
+  if (isDevMode() && key === undefined) {
+    console.warn(
+      '@ngrx/entity: The entity passed to the `selectId` implementation returned undefined. You should probably provide your own `selectId` implementation.',
+      'The entity that was passed:',
```
brandonroberts: Split up this sentence more.

modules/entity/src/utils.ts
```diff
@@ -0,0 +1,18 @@
+import { isDevMode } from '@angular/core';
+import { IdSelector } from './models';
+
+export function selectIdValue<T>(entity: T, selectId: IdSelector<T>) {
+  const key = selectId(entity);
+
+  if (isDevMode() && key === undefined) {
+    console.warn(
```
brandonroberts: What about `false`?

modules/entity/spec/utils.spec.ts
```diff
@@ -0,0 +1,32 @@
+import * as ngCore from '@angular/core';
+import { selectIdValue } from '../src/utils';
+import { BookModel, AClockworkOrange } from './fixtures/book';
+
+describe('Entity utils', () => {
+  describe(`selectIdValue()`, () => {
```
brandonroberts: Need an additional test with a valid key that's undefined.

modules/entity/src/utils.ts
tdeschryver: What do you mean by that? You would also log this when the `key` would be `null`, `false`?

modules/entity/spec/utils.spec.ts
tdeschryver: You mean with `enableProdMode()`? Because this threw an error because the platform was already set.

modules/entity/src/utils.ts
brandonroberts: Yes. We also want to warn if the selected value is `false`, correct?

tdeschryver: My initial thinking was to only check on `undefined` because the id property wouldn't exist on the entity, meaning you probably did something wrong. I'm not sure how this is implemented by other devs, because if one would define an entity as `null`, this is "hes choice" and he should know what he is doing.
Therefor I'm only checking on `undefined`.

For me it would also make more sense to check if the id value is false, but as I said I'm not sure how this is used.

modules/entity/spec/utils.spec.ts
brandonroberts: Ok. Just didn't want to use the wildcard import if we didn't have to.

modules/entity/src/utils.ts
brandonroberts: Ok, that makes sense. My rationale was that we're checking for values that aren't valid ids.

modules/entity/spec/utils.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      spyOn(ngCore, 'isDevMode').and.returnValue(false);
+      const spy = spyOn(console, 'warn');
+
+      const key = selectIdValue(AClockworkOrange, (book: any) => book.foo);
+
+      expect(spy).not.toHaveBeenCalled();
+    });
+
+    it('should not warn when key is undefined in dev mode', () => {
+      spyOn(ngCore, 'isDevMode').and.returnValue(false);
```
brandonroberts: `in prod mode`

PR #1171, opened 2018-07-04
feat(Store): add an overload to createFeatureSelector to provide better type checking

### LINKED ISSUE #1136: Enforce or drop store type parameter
The type parameter `T` in `Store<T>` is not actually enforced when piping a selector onto it, so you can basically apply a `MemoizedSelector<U,V>` even when T does not extend U. At least part of the problem seems to be the object/any types of `createFeatureSelector` here:
https://github.com/ngrx/platform/blob/55a0488c7e456d74affa2354a12cc20db12e5f38/modules/store/src/selector.ts#L272

I think it would be better to either actually enforce T extends U, or drop the type parameter from Store to acknowledge that it's not strictly typed.

As this is not enforced, we currently always use `Store<{}>`, and just rely on the programmer to use the appropriate selector that can deal with our application-wide store. This works, but seems a bit like a hack.

[ ] No
[x] Maybe :) In general, yes, but I have not yet tried to fix this and am not sure I can. I can give it a shot.

tdeschryver:
Closes #1136

### DIFF
--- a/docs/store/selectors.md
+++ b/docs/store/selectors.md
@@ -104,6 +104,24 @@ export const selectFeatureCount = createSelector(
 );
 ```
 
+There is also an overload on the `createFeatureSelector` method which allows a better typechecking.
+The example from above would become:
+
+```ts
+export const selectFeature = createFeatureSelector<AppState, FeatureState>(
+  'feature'
+);
+```
+
+Where `AppState` is the type from the top level feature state and `FeatureState` is the type from the feature slice to select.
+Meaning that the following would not compile because `foo` is not a feature slice of `AppState`.
+
+```ts
+export const selectFeature = createFeatureSelector<AppState, FeatureState>(
+  'foo'
+);
+```
+
 ## Reset Memoized Selector
 
 The selector function returned by calling `createSelector` or `createFeatureSelector` initially has a memoized value of `null`. After a selector is invoked the first time its memoized value is stored in memory. If the selector is subsequently invoked with the same arguments it will return the memoized value. If the selector is then invoked with different arguments it will recompute and update its memoized value. Consider the following:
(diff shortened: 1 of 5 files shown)


docs/store/selectors.md
```diff
@@ -104,6 +104,24 @@ export const selectFeatureCount = createSelector(
 );
 ```
 
+There is also an overload on the `createFeatureSelector` method, which allows a better typechecking.
```
brandonroberts: Instead of adding this as an additional section, let's just make this the default in the example above.

docs/store/selectors.md
```diff
@@ -104,6 +104,24 @@ export const selectFeatureCount = createSelector(
...
+
+```ts
+export const selectFeature = createFeatureSelector<AppState, FeatureState>(
+  'feature'
+);
+```
+
+Where `AppState` is the type from the top level feature state and `FeatureState` is the type from the feature slice to select.
+Meaning that the following would not compile because `foo` is not a feature slice of `AppState`.
+
```
brandonroberts: Keep the section starting with `Meaning` as a note.

docs/store/selectors.md
tdeschryver: Yea I wasn't sure about this, but you just confirmed my thinking.
Should I also update the example app?

brandonroberts: Yes

docs/store/selectors.md
```diff
@@ -97,13 +97,23 @@ export interface AppState {
...
+  'feature'
+);
 export const selectFeatureCount = createSelector(
   selectFeature,
   (state: FeatureState) => state.counter
 );
 ```
 
+Meaning that the following would not compile because `foo` is not a feature slice of `AppState`.
+
```
brandonroberts: Change to `The following selector below would not ...`

tdeschryver: No problem! Thanks for the reviews!

PR #1173, opened 2018-07-04
Prepare for TS 2.9

alex-okrushko:
At Google we already switched to 2.9. This PR makes ngrx/platform compatible with 2.9 as well (while still running fine at 2.7.2)


alex-okrushko:
btw, prettier config can/should be changed back once TS releases version that includes https://github.com/Microsoft/TypeScript/issues/24628 and we'll hope devs will skip 2.9.2 altogether :/

Issue #1157, opened 2018-06-26
change defaultMemoize's isEqual to compare results instead of arguments and rename to comparerFn


alex-okrushko:
I'll start with this: I'm NOT suggesting to remove argument comparison, **arguments should be compared anyway and comparing them based on === should be sufficient** to start the result evaluation.

However the provided [`isEqual`](https://github.com/ngrx/platform/blob/master/modules/store/src/selector.ts#L21) function should be comparing the results (lastResult + new result), when the arguments comparison doesn't pass.

### Reasoning:

1. **Comparing arguments is almost useless, especially if there are two or more arguments coming in.**
Consider the following:

```ts
enum OrderType {
  EXEC_LEVEL,
  INTERNAL,
  PUBLIC,
}

interface Order {
  type: OrderType, // <-- this is the enum above
  id: string,
  totalCost: string,
}

// Provides Order[] with all existing orders.
const getAllOrders = createSelector(someSelectorWithOrders, (state) => state.orders);

// Provides currently selected OrderType (the one from the enum).
const getSelectedType = createSelector(someSelectorWithTypes, (state) => state.currentType);

// Assume we have the following comparer, which would just LOGs what that it gets to compare.
function isMyCustomComparer(a: any, b: any) {
  console.log('a:', a, 'b:', b);
  return a === b; // this is what is currently isEqualCheck is doing anyway.
}

// This is how we use the default Memoizer, but are providing custom `isEquals`,
// which is a second argument in the defaultMemoizer function.
const customMemoizer = (aFn) => defaultMemoize(aFn, isMyCustomComparer);

// The actual select of interest. Returns Orders filtered by Current type.
const getOrdersBySelectedType = 
    createSelectorFactory(customMemoizer)(getAllOrders, getSelectedType, 
       (orders: Order[], currentOrderType: OrderType) => 
            orders.filter(order => order.type === currentOrderType));
```

Let's say we have the following global state 
```ts
{
   ...
   orders: [{type: OrderType.INTERNAL, id: '1', totalCost: '10.00'},
            {type: OrderType.PUBLIC, id: '2', totalCost: '100.00'}],
   ...
   currentType: OrderType.INTERNAL,
}
```

Now let's call it twice and see what `console.log('a:', a, 'b:', b);` inside `isMyCustomComparer` will produce.

```ts
getOrdersBySelectedType({...originalState});
getOrdersBySelectedType({...originalState});
```
The output we get is:
[...]

PR #1175, opened 2018-07-05
Change what defaultMemoize compares with isEqual

alex-okrushko:
Also renamed isEqual to isResultEqual to be clear what is compared
This is the fix for https://github.com/ngrx/platform/issues/1157

### DIFF
--- a/modules/store/spec/selector.spec.ts
+++ b/modules/store/spec/selector.spec.ts
@@ -4,6 +4,7 @@ import {
   createFeatureSelector,
   defaultMemoize,
   createSelectorFactory,
+  MemoizedProjection,
 } from '@ngrx/store';
 import { map, distinctUntilChanged } from 'rxjs/operators';
 
@@ -250,16 +251,14 @@ describe('Selectors', () => {
     it('should allow a custom memoization function', () => {
       const projectFn = jasmine.createSpy('projectionFn');
       const anyFn = jasmine.createSpy('t').and.callFake(() => true);
-      const equalFn = jasmine.createSpy('isEqual').and.callFake(() => true);
-      const customMemoizer = (aFn: any = anyFn, eFn: any = equalFn) =>
-        defaultMemoize(anyFn, equalFn);
+      const customMemoizer = (aFn: any = anyFn) => defaultMemoize(anyFn);
       const customSelector = createSelectorFactory(customMemoizer);
 
       const selector = customSelector(incrementOne, incrementTwo, projectFn);
       selector(1);
       selector(2);
 
-      expect(anyFn.calls.count()).toEqual(1);
+      expect(anyFn.calls.count()).toEqual(2);
     });
 
     it('should allow a custom state memoization function', () => {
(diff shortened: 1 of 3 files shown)


alex-okrushko:
@brandonroberts 
To be 100% compatible with currentVersion, instead of replacing `isEqual` with `isResultEqual` I can just add it as a 3-rd argument, but I think it's not needed, because `isResultEqual` will be catching results even before they are passed into other sub-selectors.
Let me know what you think.

tdeschryver:
I'm going to nitpick a bit and say that I find the name `defaultMemoize` a bit confusing, I would rather go for just `memoize`. This wouldn't mean that `defaultMemoize` would disappear tho.

Let me show you want I'm trying to say:

```
function memoize(projectionFn: AnyFn, isResultEqual: (a: any, b: any) => boolean) {
  // ...
}

function defaultMemoize(projectionFn: AnyFn) {
  return memoize(projectionFn, isEqualCheck);
}
```

The `defaultMemoize` will be used as the default in NgRx, as the name says. 
But if you rather want to use your own, you can use `memoize` instead of `defaultMemoize`.

If this makes sense... 😅

alex-okrushko:
> I would rather go for just `memoize`

`memoize` is already a function inside `defaultMemoize`. If anything I would've changed the name to `defaultMemoizer` with **r** at the end.

> The defaultMemoize will be used as the default in NgRx, as the name says.
> But if you rather want to use your own, you can use memoize instead of defaultMemoize.

I'd rather not duplicate the code with the second `memoize` there as well. 
First of all, there are very similar. The only difference is applied the results comparison.
Secondly, that would unnecessary increase the size of ngrx :P 

`createSelectorFactory` is used within Google in a number of places. I'm patching this PR and running all the tests globally on entire codebase. Let's see if it breaks anything. (I'll report back).

tdeschryver:
I'm not sure if I'm following, how would this increase the size? 
I'm not suggesting to duplicate the code, but to call the `memoize` function from inside `defaultMemoize`.

alex-okrushko:
So it failed a few tests that were testing memoization with custom equality
Ok, I'll adjust so it doesn't have any breaking changes

dummdidumm:
I think adding the isResultEqual as a new third argument would be better. No breaking changes + maybe someone really wants to compare arguments and not results.

alex-okrushko:
I've extracted result checking into its own argument, so now it's 100% backwards compatible.

PTAL :)

brandonroberts:
@alex-okrushko I will take a look. I still think this would be a change in behavior. Selectors that fire a second time before would not if the results were the same, correct? Have you seen any impact on large sets of data?

alex-okrushko:
@brandonroberts I don't think there would be a change in behavior. All previously written tests are exactly the same now. If 3-rd argument `isResultEqual` is not passed to `defaultMemoize`, it behaves exactly as before.

I'm reruning all the tests within G again.

> Have you seen any impact on large sets of data?
If `isResultEqual` is used? It helps not to re-render filtered lists.

alex-okrushko:
Thanks!

PR #1183, opened 2018-07-12
Add smarter type inference for ofType pipe.

### LINKED ISSUE #860: Update Signature of Actions and Actions#ofType to infer action interface in Effects
Using the new mapped types feature in TypeScript 2.8 we can now write a type safe version of `Actions#ofType` that can correctly infer the interface of the action based on the provided action type:

```ts
type Action<Type extends string> = { type: Type };

class Actions<T extends Action> extends Observable<T> { }

type ActionsOfType<A, T extends string> = A extends Action<T> ? A : never;

function ofType<V, T1 extends string>(t1: T1): OperatorFunction<V, ActionsOfType<V, T1>>;
function ofType(...types: string[]) {
  return function (source: Observable<any>) {
    return source.pipe(filter(action => types.indexOf(action.type) !== -1)) as any;
  }
}
```

This will be a major breaking change for Effects. To minimize the impact we should add an `@ngrx/effects/compat` submodule that exports a backwards compatible version of `Actions` and `ofType` so that developers can incrementally upgrade. We will remove `@ngrx/effects/compat` in 7.0.

rkirov:
It is based on TS2.8 introduced conditional types. Upgrade package.json
to that version.

Also remove the deprecated non-pipable version of `ofType` so that we
don't have to keep complicated types in two places.

Fixes some tests post removal of non-pipable version.

Original implementation by @mtaran-google.

Fixes #860


modules/effects/compat/rollup.config.js
```diff
@@ -0,0 +1,12 @@
+export default {
```
rkirov: I copied those from `testing` but I have not idea whether they are correct in the context of compat.

timdeschryver:
I think we should allow an "infite" number of `ofType` arguments?

mtaran-google:
That would be ideal, but the type system in TS 2.9 isn't powerful enough to check that many. 3.0 (in RC now) [improves on this](https://blogs.msdn.microsoft.com/typescript/2018/07/12/announcing-typescript-3-0-rc/#tuples-and-parameters) but I'm not sure if it'll be enough. It'll be interesting to check out when it lands.

timdeschryver:
@mtaran-google I see, thanks.

Just one more think, I thought the following would give me compile errors but they don't:
```ts
// this is OK
ofType<Search>(BookActionTypes.Search),

// I would expect an error here
ofType<Search>('sss'),

// I would expect an error here, because the payload of `SearchComplete` is an array
ofType<Search>(BookActionTypes.SearchComplete),
map(action => action.payload),
  switchMap(query => {
    if (query === '') {
        return empty();
     }
```

I fiddled with it in the past and came up with [this](https://github.com/timdeschryver/platform/commit/6cca61c08a1aa23b999bdaa0756965c9f15be794) which is giving compile errors.

rkirov:
@timdeschryver That's a good point. Let's recap the desired outcomes:

1) ofType('x') to just work based in inference
2) ofType<SomeType>('someActionType') to keep working as is.
3) ofType<SomeType>('typo') to error if SomeType.type != 'typo' (both of your examples boil down to this).

My current PR has 2) but not 3), and yours I think has 3) but not 2), because you have two free type variables "export function ofType<V extends Action, Type extends string>"  and no defaults. Maybe I am missing something, but I don't think you can have 2) and 3) at the same time.

rkirov:
However, based on your your example it looks like the T1,T2,..., etc expansions are not necessary. I think the following works the same.

```typescript
export function ofType<
  V extends Extract<U, { type: T }>,
  T extends string = string,
  U extends Action = Action
>(...t: T[]): OperatorFunction<U, V>;
```

Also one final thing, my work here is prompted by attempting to turn on --strictFunctionTypes and hitting a lot of ngrx issues. Please do your testing with that flag on. I think any solution that doesn't have at least two parameters U -> V to capture the broader observable input and the narrowed out observable output, will not work with --strictFunctionTypes.

alex-okrushko:
`Extract` showed up in 2.8, while ngrx is still at 2.7

rkirov:
@alex-okrushko That is the blocker for all this. TS2.8 is required for conditional types, Maksym's original change a custom alias for `Extract`.

timdeschryver:
@rkirov Yep you're completely right.
The solution I came up with doesn't support `ofType<SomeType>('someActionType')`.

And the snippet you posted has the same outcome.
```ts
export function ofType<
  V extends Extract<U, { type: T }>,
  T extends string = string,
  U extends Action = Action
>(...t: T[]): OperatorFunction<U, V>;
```

rkirov:
Ok, I just discovered one more issue that might be the deal breaker here. A lot of teams do not pass they type argument on their actions observable is it just of type - `Actions`. Now, if that property is `Actions<AllPossibleActions>` this change is great.

However, for unqualifed `Actions` this change (in both proposed forms) is actually a regression. That means if one has `private actions: Actions;` then `this.actions.pipe(ofType('something'))` creates `Observable<never>` instead of at least `Observable<Action>`.

This kinda expected because `Extract<{type: string}, {type: 'a'}>` is `never` and not `{type: 'a'}`. In a more general form `Extract<string, 'a'>` is `never` though one can expect it to be 'a'. Not sure if TS has plans to change that.

rkirov:
I updated this PR with what is my final proposal for the future types of 'ofType'. Added comments that should capture the upside and downsides of that approach. Hopefully, that would make the type soup a bit more maintainable.

Internally, I appears we have an agreement  that this is good and are going to change it with a patch before this PR lands here. Here it appears to be stuck behind angular, bazel, typescript versioning hell, that I can't resolve.

@MikeRyanDev and @brandonroberts please at least verify that there are no major concerns with this proposal, otherwise we would end up forking out internal google users from the mainstream branch here.

MikeRyanDev:
@rkirov This proposed change looks good to me.

modules/effects/src/actions.ts
```diff
@@ -1,8 +1,11 @@
 import { Inject, Injectable } from '@angular/core';
 import { Action, ScannedActionsSubject } from '@ngrx/store';
-import { Observable, Operator, OperatorFunction } from 'rxjs';
+import { Observable, OperatorFunction, Operator } from 'rxjs';
 import { filter } from 'rxjs/operators';
 
+// Removes types from T that are not assignable to U.
+export type Filter<T, U> = T extends U ? T : never;
```
timdeschryver: Because we're using `Extract` this can be removed now.

modules/effects/spec/actions.spec.ts
```diff
@@ -58,10 +66,10 @@ describe('Actions', function() {
...
-  it('should let you filter out actions', function() {
-    const actions = [ADD, ADD, SUBTRACT, ADD, SUBTRACT];
-    const expected = actions.filter(type => type === ADD);
+  const actions = [ADD, ADD, SUBTRACT, ADD, SUBTRACT];
+  const expected = actions.filter(type => type === ADD);
 
+  it('should let you filter out actions', function() {
     actions$
       .pipe(
         ofType(ADD),
```
timdeschryver: TS is complaining about the next line `map(update => update.type)`
This can be solved if we define our actions as `let actions$: Actions<AddAcction | SubtractAction>;`

modules/effects/src/actions.ts
```diff
@@ -19,17 +22,79 @@ export class Actions<V = Action> extends Observable<V> {
     observable.operator = operator;
     return observable;
   }
-
-  ofType<V2 extends V = V>(...allowedTypes: string[]): Actions<V2> {
```
timdeschryver: This may sound stupid, but here it goes anyways:
Since we have the fallback `ofType`, could we not remove this but mark it as deprecated?
With the next release we can remove it.
I say this because while the `compat` solution works, the user would have to change the import statement?

package.json
```diff
@@ -143,7 +143,7 @@
     "tslib": "1.6.0",
     "tslint": "^5.0.0",
     "tsutils": "2.20.0",
-    "typescript": "~2.7.2",
+    "typescript": "~2.8.4",
```
alex-okrushko: I think this means that angular version would need a bump as well:
```
@angular/compiler-cli@6.0.9 requires typescript@'>=2.7.0 <2.8.0' but 2.8.4 was found instead.
```

rkirov: Indeed, at this point this PR is doing too much. Another PR is needed to get to TS 2.8, with all the complications involved around angular, bazel, router(?). I can't do that upgrade, but will rebase this PR once that happens.

modules/effects/src/actions.ts
rkirov: I have no opinions on backwards compat (google operates on one-version policy), and will do whatever you guys decide.

modules/effects/src/actions.ts
rkirov: Indeed.

modules/effects/spec/actions.spec.ts
rkirov: Sorry, I haven't even been able to run the typechecker on the whole PR, because of the other deps. Fixed.

rkirov:
Opened https://github.com/ngrx/platform/issues/1220 to track the compiler upgrade. This PR is blocked (will respond to review comments though) until that gets resolved.

modules/effects/src/actions.ts
brandonroberts: I was planning on doing a 6.1 release, so maybe we can deprecate the instance operators before release and then move forward with the removal for the next major release.

rkirov:
rebased the PR, will take a look at the CI results tomorrow.

rkirov:
CI is green, please proceed with a regular review. I don't fully understand the compat strategy, mostly copy/pasted from other directories.

modules/effects/spec/actions.spec.ts
rkirov: CI is green. I assume this means it is not longer an issue.

timdeschryver: LGTM 👍

timdeschryver:
I think the `compat` version won't be needed. The `store.ofType` (and `store.select`) operators are now marked as deprecated and will be removed the next version.
Before you change anything, wait untill @brandonroberts 's confirmation tho.

brandonroberts:
@rkirov @timdeschryver I think its safe to remove the `compat` package also. The commit message does need to reflect the breaking change of removing the instance operator.

alex-okrushko:
@rkirov Hey Rado. Would you be able to take a look at this PR again?

dimavt:
Any update on this? I can finalize this PR, if you don't mind.

For someone who wants to try this feature in your project today, you can add this to your codebase:
```ts
declare module '@ngrx/effects' {
  export function ofType<
    V extends Extract<U, { type: T }>, 
    T extends string = string, 
    U extends Action = Action
  >(...actions: T[]): OperatorFunction<U, V>;
}
```

timdeschryver:
If you want I can finish this one up @rkirov but you'll have to give maintainers write access to your branch.

rkirov:
I will rebase it today. Sorry for the delay.

Btw, we have been running this with change internally in google for some time and I haven't heard any complaints.

modules/effects/spec/actions.spec.ts
```diff
@@ -11,12 +11,20 @@ import { hot, cold } from 'jasmine-marbles';
 import { of } from 'rxjs';
 
 describe('Actions', function() {
-  let actions$: Actions;
+  let actions$: Actions<AddAcction | SubtractAction>;
```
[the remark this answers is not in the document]
rkirov: done

rkirov:
Rebased, removed compat bits and updated commit message with 'BREAKING CHANGE'. Please take another look.

brandonroberts:
Thanks @rkirov!

PR #1184, opened 2018-07-13
docs(store): fix/clarify angularJS downgrade of Store

mikelseverson:
Small clarification/fix for @ngrx/store angularJS downgrade instructions.
Change is to specify "Store" instead of "provider" when calling downgradeInjectable.


brandonroberts:
Thanks @mikelseverson!

PR #1193, opened 2018-07-21
Adjust ngrx to be fully compatible with ts27 and ts29 at the same time

alex-okrushko:
Prior TS 2.9 keyof T has the type string. Since 2.9 An index type keyof T for some type T is a subtype of string | number | symbol.

2.8 introduced Extract<keyof T, string> that helps to narrow the type down back to string, but ngrx/platform is at 2.7 still, thus this change is needed.

### DIFF
--- a/modules/effects/src/effects_metadata.ts
+++ b/modules/effects/src/effects_metadata.ts
@@ -3,7 +3,8 @@ import { compose } from '@ngrx/store';
 const METADATA_KEY = '__@ngrx/effects__';
 
 export interface EffectMetadata<T> {
-  propertyName: keyof T;
+  // Once TS is >= 2.8 replace with <Key extends Extract<keyof T, string>>
+  propertyName: string;
   dispatch: boolean;
 }
 
@@ -29,7 +30,9 @@ function setEffectMetadataEntries<T>(
 }
 
 export function Effect<T>({ dispatch = true } = {}): PropertyDecorator {
-  return function(target: T, propertyName: keyof T) {
+  // Once TS is >= 2.8 replace with <Key extends Extract<keyof T, string>>
+  // for propertyName.
+  return function(target: T, propertyName: string) {
     const metadata: EffectMetadata<T> = { propertyName, dispatch };
     setEffectMetadataEntries<T>(target, [metadata]);
   } as (target: {}, propertyName: string | symbol) => void;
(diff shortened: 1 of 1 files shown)


modules/effects/src/effects_metadata.ts
```diff
@@ -43,7 +46,9 @@ export function getSourceMetadata<T>(instance: T): Array<EffectMetadata<T>> {
   return compose(getEffectMetadataEntries, getSourceForInstance)(instance);
 }
 
-export type EffectsMetadata<T> = { [key in keyof T]?: { dispatch: boolean } };
+// Once TS is >= 2.8 replace with
+// {[key in <Key extends Extract<keyof T, string>>]?:  { dispatch: boolean } };
+export type EffectsMetadata<T> = { [key: string]: { dispatch: boolean } };
```
brandonroberts: Why is this no longer optional?

alex-okrushko: Indexable doesn't need ? for the values to be optional

[Here is the playground link](https://www.typescriptlang.org/play/#src=interface%20MyInt%20%7B%20%0D%0A%20%20%20%20prop1%3A%20string%0D%0A%7D%0D%0A%0D%0Atype%20WithKeyOfOptional%3CT%3E%20%3D%20%7B%20%5Bkey%20in%20keyof%20T%5D%3F%3A%20%7B%20dispatch%3A%20boolean%20%7D%20%7D%0D%0Atype%20WithKeyOf%3CT%3E%20%3D%20%7B%20%5Bkey%20in%20keyof%20T%5D%3A%20%7B%20dispatch%3A%20boolean%20%7D%20%7D%20%2F%2F%20Doesn't%20have%20%3F%0D%0Atype%20SimpleIndexable%20%3D%20%7B%5Bkey%3A%20string%5D%3A%20%7Bdispatch%3A%20boolean%7D%7D%20%2F%2F%20Doesn't%20need%20%3F%0D%0A%0D%0Aconst%20withKeyOptional%3A%20WithKeyOfOptional%3CMyInt%3E%20%3D%20%7B%7D%3B%0D%0Aconst%20withKey%3A%20WithKeyOf%3CMyInt%3E%20%3D%20%7B%7D%3B%0D%0Aconst%20simpleIndecable%20%3D%20%7B%7D%3B)

brandonroberts: 👍

PR #1194, opened 2018-07-23
feat: allow properties in selectors

### LINKED ISSUE #1152: Allow Selector to accept arguments besides state
Thanks for this awesome library!

With [Reselect](https://github.com/reduxjs/reselect#accessing-react-props-in-selectors), we can call a selector with arguments like below (almost copied from Reselect's README):

```ts
import { createSelector } from 'reselect'

const getVisibilityFilter = (state, props) =>
  state.todoLists[props.listId].visibilityFilter

const getTodos = (state, props) =>
  state.todoLists[props.listId].todos

const getVisibleTodos = createSelector(
  [ getVisibilityFilter, getTodos ],
  (visibilityFilter, todos) => {
    switch (visibilityFilter) {
      case 'SHOW_COMPLETED':
        return todos.filter(todo => todo.completed)
      case 'SHOW_ACTIVE':
        return todos.filter(todo => !todo.completed)
      default:
        return todos
    }
  }
)

const todos = getVisibleTodos(state, props)
```

but ngrx's selector does not seem to have such a feature. It will be helpful for me if it's possible.

## Other information:

In the above example of Reselect, the props is not used in `createSelector`, but actually it also can be used like below:

```ts
import { createSelector } from 'reselect'

const shootingById = createSelector(
  [shootings,(state, shootingId) => shootingId],
  (shootings, shootingId) => head(shootings.filter(s => s._id === shootingId))
);

const shootings = shootingById(state, someId)
```

ref: https://github.com/reduxjs/reselect/issues/272#issuecomment-316754662 

[ ] No

timdeschryver:
This allows to provide provide extra properties inside selectors.

```ts
const selector = createSelector(
  state => state.count * 10,
  (state, props) => state.count + props.value,
  (one, two) => one + two
);

const result = selector(
  {
    count: 10,
  },
  { value: 47 }
);

// Result = 157.
```

If we agree on this I can add a commit to enhance the docs.

Closes #1152.

### DIFF
--- a/modules/store/spec/selector.spec.ts
+++ b/modules/store/spec/selector.spec.ts
@@ -224,6 +224,51 @@ describe('Selectors', () => {
     });
   });
 
+  describe('createSelector with only a props selector', () => {
+    it('should deliver the state and the props to the projection function', () => {
+      const projectFn = jasmine.createSpy('projectionFn');
+      const state = { counter: {} };
+      const props = { value: 47 };
+      const selector = createSelector(projectFn)(state, props);
+      expect(projectFn).toHaveBeenCalledWith(state, props);
+    });
+
+    it('should be possible to use a projector fn', () => {
+      const projectFn = jasmine.createSpy('projectionFn');
+      const selector = createSelector(projectFn);
+      selector.projector('foo', 'bar');
+      expect(projectFn).toHaveBeenCalledWith('foo', 'bar');
+    });
+
+    it('should call the projector function when the state changes', () => {
+      const projectFn = jasmine.createSpy('projectionFn');
+      const selector = createSelector(projectFn);
+
+      const firstState = { first: 'state' };
+      const secondState = { second: 'state' };
+      const props = { foo: 'props' };
+      selector(firstState, props);
+      selector(firstState, props);
+      selector(secondState, props);
+      expect(projectFn).toHaveBeenCalledTimes(2);
+    });
+
+    it('should allow you to release memoized arguments', () => {
+      const state = { first: 'state' };
+      const props = { first: 'props' };
+      const projectFn = jasmine.createSpy('projectionFn');
+      const selector = createSelector(projectFn);
+
+      selector(state, props);
+      selector(state, props);
+      selector.release();
+      selector(state, props);
+      selector(state, props);
+
+      expect(projectFn).toHaveBeenCalledTimes(2);
+    });
+  });
+
   describe('createSelector with arrays', () => {
     it('should deliver the value of selectors to the projection function', () => {
       const projectFn = jasmine.createSpy('projectionFn');
--- a/modules/store/spec/selector.spec.ts
+++ b/modules/store/spec/selector.spec.ts
@@ -125,6 +125,105 @@ describe('Selectors', () => {
     });
   });
 
+  describe('createSelector with props', () => {
+    it('should deliver the value of selectors to the projection function', () => {
+      const projectFn = jasmine.createSpy('projectionFn');
+
+      const selector = createSelector(
+        incrementOne,
+        incrementTwo,
+        (state: any, props: any) => props.value,
+        projectFn
+      );
+
+      selector({}, { value: 47 });
+      expect(projectFn).toHaveBeenCalledWith(countOne, countTwo, 47);
+    });
+
+    it('should be possible to test a projector fn independent from the selectors it is composed of', () => {
+      const projectFn = jasmine.createSpy('projectionFn');
+      const selector = createSelector(
+        incrementOne,
+        incrementTwo,
+        (state: any, props: any) => {
+          fail(`Shouldn't be called`);
+          return props.value;
+        },
+        projectFn
+      );
+      selector.projector('', '', 47);
+
+      expect(incrementOne).not.toHaveBeenCalled();
+      expect(incrementTwo).not.toHaveBeenCalled();
+      expect(projectFn).toHaveBeenCalledWith('', '', 47);
+    });
+
+    it('should call the projector function when the state changes', () => {
+      const projectFn = jasmine.createSpy('projectionFn');
+      const selector = createSelector(
+        incrementOne,
+        (state: any, props: any) => props.value,
+        projectFn
+      );
+
+      const firstSate = { first: 'state' };
+      const props = { foo: 'props' };
+      selector(firstSate, props);
+      selector(firstSate, props);
+      expect(projectFn).toHaveBeenCalledTimes(1);
+
+      const secondState = { second: 'state' };
+      selector(secondState, props);
+      expect(projectFn).toHaveBeenCalledTimes(2);
+    });
+
+    it('should memoize the function', () => {
+      let counter = 0;
+
+      const firstState = { first: 'state' };
+      const secondState = { second: 'state' };
+      const props = { foo: 'props' };
+
+      const projectFn = jasmine.createSpy('projectionFn');
+      const selector = createSelector(
+        incrementOne,
+        incrementTwo,
+        (state: any, props: any) => {
+          counter++;
+          return props;
+        },
+        projectFn
+      );
+
+      selector(firstState, props);
+      selector(firstState, props);
+      selector(firstState, props);
+      selector(secondState, props);
+
+      expect(counter).toBe(2);
+      expect(projectFn).toHaveBeenCalledTimes(2);
+    });
+
+    it('should allow you to release memoized arguments', () => {
+      const state = { first: 'state' };
+      const props = { foo: 'props' };
+      const projectFn = jasmine.createSpy('projectionFn');
+      const selector = createSelector(
+        incrementOne,
+        (state: any, props: any) => props,
+        projectFn
+      );
+
+      selector(state, props);
+      selector(state, props);
+      selector.release();
+      selector(state, props);
+      selector(state, props);
+
+      expect(projectFn).toHaveBeenCalledTimes(2);
+    });
+  });
+
   describe('createSelector with arrays', () => {
     it('should deliver the value of selectors to the projection function', () => {
       const projectFn = jasmine.createSpy('projectionFn');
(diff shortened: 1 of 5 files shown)


modules/store/src/models.ts
```diff
@@ -33,6 +33,9 @@ export interface StoreFeature<T, V extends Action = Action> {
   metaReducers?: MetaReducer<T, V>[];
 }
 
-export interface Selector<T, V> {
-  (state: T): V;
-}
+export type Selector<State, Result> = (state: State) => Result;
+
+export type SelectorWithProperty<State, Property, Result> = (
```
brandonroberts: Property should be plural because we're passing a set of them. `Props` or `Properties`?

modules/store/spec/selector.spec.ts
```diff
@@ -123,6 +123,106 @@ describe('Selectors', () => {
     });
   });
 
+  describe('createSelector with props', () => {
```
brandonroberts: We need some integration tests using `store.select` also.

modules/store/src/models.ts
timdeschryver: I would say `Props`?

modules/store/spec/selector.spec.ts
timdeschryver: Thanks for reminding me!
Creating this test made me realize we also have to change the `select` function.
Doing this I bumped into a small problem, see my comment in the following commit.

modules/store/src/store.ts
```diff
@@ -175,20 +177,25 @@ export function select<
...
   ...paths: string[]
 ) {
-  return function selectOperator(source$: Observable<T>): Observable<K> {
+  return function selectOperator(source$: Observable<State>): Observable<K> {
     let mapped$: Observable<any>;
 
     if (typeof pathOrMapFn === 'string') {
-      mapped$ = source$.pipe(pluck(pathOrMapFn, ...paths));
+      mapped$ =
+        typeof props === 'string'
```
timdeschryver: I think we can't support having props on a string selector e.g. `this.store.select('counter1')`, right?

brandonroberts: Right

modules/store/spec/selector.spec.ts
brandonroberts: I think only having props for selector functions is reasonable.

timdeschryver:
I just did a rebase and some of the tests are failing, let me fix these first.

timdeschryver:
I think I found a bug in the existing selector.

This fails, because it gets called 3 times

```
it('should memoize the function', () => {
      const firstState = { first: 'state' };
      const secondState = { second: 'state' };
      const projectFn = jasmine.createSpy('projectionFn');
      const selector = createSelector(
        incrementOne,
        incrementTwo,
        incrementThree,
        projectFn
      );

      selector(firstState);
      selector(firstState);
      selector(firstState);
      selector(secondState);
      selector(secondState); // if I add this line the test fails, I would expect it to succeed?

      expect(incrementOne).toHaveBeenCalledTimes(2);
      expect(incrementTwo).toHaveBeenCalledTimes(2);
      expect(incrementThree).toHaveBeenCalledTimes(2);
      expect(projectFn).toHaveBeenCalledTimes(2);
    });
```

modules/store/src/selector.ts
```diff
@@ -276,8 +504,19 @@ export function createSelectorFactory(
       return projector.apply(null, selectors);
     });
 
-    const memoizedState = defaultMemoize(function(state: any) {
-      return options.stateFn.apply(null, [state, selectors, memoizedProjector]);
+    const memoizedState = defaultMemoize(function(state: any, props: any) {
+      // createSelector works directly on state
```
brandonroberts: Is this related or a different feature? Feels like this should be a separate PR

modules/store/src/store.ts
```diff
@@ -175,20 +177,23 @@ export function select<
...
-export function select<T, K = any>(
+export function select<State, Props = any, Result = any>(
+  propsOrPath: Props,
   ...paths: string[]
-): (source$: Observable<T>) => Observable<K>;
-export function select<T, K>(
-  pathOrMapFn: ((state: T) => any) | string,
+): (source$: Observable<State>) => Observable<Result>;
+export function select<State, Props, K>(
+  pathOrMapFn: (state: State, props?: Props) => any | string,
```
brandonroberts: `(state ...) => any` should be wrapped in parenthesis

modules/store/src/store.ts
```diff
@@ -175,20 +177,23 @@ export function select<
...
   ...paths: string[]
 ) {
-  return function selectOperator(source$: Observable<T>): Observable<K> {
+  return function selectOperator(source$: Observable<State>): Observable<K> {
     let mapped$: Observable<any>;
 
     if (typeof pathOrMapFn === 'string') {
-      mapped$ = source$.pipe(pluck(pathOrMapFn, ...paths));
+      const pathSlices = [<string>props, ...paths].filter(Boolean);
+      mapped$ = source$.pipe(pluck(pathOrMapFn, ...pathSlices));
```
brandonroberts: Don't pass in props with a string selector

modules/store/src/store.ts
```diff
@@ -175,20 +177,23 @@ export function select<
...
     } else if (typeof pathOrMapFn === 'function') {
-      mapped$ = source$.pipe(map(pathOrMapFn));
+      mapped$ = source$.pipe(map(source => pathOrMapFn(source, <Props>props)));
```
brandonroberts: Why do `props` need to be typed here?

modules/store/src/store.ts
```diff
@@ -109,11 +109,13 @@ export class Store<T> extends Observable<T> implements Observer<Action> {
 
 export const STORE_PROVIDERS: Provider[] = [Store];
 
-export function select<T, K>(
-  mapFn: (state: T) => K
-): (source$: Observable<T>) => Observable<K>;
+export function select<State, Result, Props>(
```
brandonroberts: Leave the types as `T, K` in this PR. Its confusing to switch between `T, K` and `State, Result`

modules/store/src/selector.ts
```diff
@@ -276,8 +504,19 @@ export function createSelectorFactory(
...
+      // e.g. createSelector((state, props) => ...)
+      if (selectors.length === 0) {
```
brandonroberts: Is this needed in this PR?

modules/store/src/models.ts
```diff
@@ -33,6 +33,9 @@ export interface StoreFeature<T, V extends Action = Action> {
   metaReducers?: MetaReducer<T, V>[];
 }
 
-export interface Selector<T, V> {
```
brandonroberts: Leave as `T, V` in this PR.

brandonroberts:
We don't want to change the tests to match our intended result. If #1175 is related, lets figure out why it causes the tests to fail.

modules/store/src/selector.ts
timdeschryver: I thought you meant this with your comment "I think only having props for selector functions is reasonable."
But you probably meant:
```
createSelector(
   getX,
   (props) => props...
)
```


Is this possible to do? Or do shouldn't we pass the state to the 'props selectors'?

brandonroberts: What I meant was we don't need to try to support using string selectors with props like below.

```
select('something', 5);
```

modules/store/src/selector.ts
```diff
@@ -223,23 +400,65 @@ export function createSelector<State, S1, S2, S3, S4, S5, S6, S7, S8, Result>(
...
+  ...input: any[]
+): Selector<any, any> | SelectorWithProps<any, any, any> {
   return createSelectorFactory(defaultMemoize)(...input);
 }
 
 export function defaultStateFn(
   state: any,
-  selectors: Selector<any, any>[],
+  props: any,
+  selectors: any[],
```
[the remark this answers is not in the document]
timdeschryver: I think this isn't possible because we're adding the props when we invoke the select.
`fn(state, props)`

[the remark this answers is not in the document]
brandonroberts: For `selectors`?

modules/store/src/selector.ts
timdeschryver: Ah gotcha 😅 ! 
Do you think having the option to create a selector solely based on props would be worth it?
```
createSelector((state, props) => ...)
```

If so I can pull out [the commit](https://github.com/ngrx/platform/pull/1194/commits/4605ff965dbfd05230d25033bbc7ae96b140663c) out of this PR and create a new one, or I could clean up these commits into 2 commits.

modules/store/src/store.ts
timdeschryver: Because the type was defined as `Props | string`. 
Therefor I had to cast it to `Props`.

modules/store/src/selector.ts
brandonroberts: 2 commits is fine

modules/store/src/selector.ts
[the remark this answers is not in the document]
timdeschryver: Yea, we're invoking every selector with `state` and `props`.

The selector type only accepts one argument, which is the state.

modules/store/src/store.ts
brandonroberts: Gotcha. Just remove the filter then

modules/store/src/selector.ts
timdeschryver: It is, this will be part of the second commit to allow `createSelector((state, props) => ...)`.
This is needed because `createSelector`  only has one argument which is the projection function (there are no selectors). Meaning the `selectors` would be empty, thus our projector wouldn't be invoked.

modules/store/src/store.ts
timdeschryver: We can't if we only pass one path, the `props` value is undefined, we have to filter this one out.

`this.store.select('foo')`

brandonroberts: Right 👍

PR #1210, opened 2018-07-28
feat(Store): createSelector pass props to projector function

timdeschryver:
This is a follow up of #1194.

I was writing the docs and found it a bit awkward I had to create a selector to only pass the props

```ts
        getTodos,
        (state: TodoAppSchema, { id }: { id: number }) => id,
        (todos, id) => todos.find(todo => todo.id === id)
);
```
Therefor this PR, it also passes the props to the projector function.
Meaning the props will be passed through all the selectors + the projection function.

```ts
        getTodosState,
        (todos: Todo[], { id }: { id: number }) =>
          todos.find(todo => todo.id === id)
 );
```

### DIFF
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -90,7 +90,7 @@ export function createSelector<State, S1, Result>(
 ): MemoizedSelector<State, Result>;
 export function createSelector<State, Props, S1, Result>(
   s1: SelectorWithProps<State, Props, S1>,
-  projector: (s1: S1) => Result
+  projector: (s1: S1, props: Props) => Result
 ): MemoizedSelectorWithProps<State, Props, Result>;
 export function createSelector<State, S1, Result>(
   selectors: [Selector<State, S1>],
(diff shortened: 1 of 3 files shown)


modules/store/src/selector.ts
```diff
@@ -447,12 +483,19 @@ export function createSelector(
...
 export function defaultStateFn(
   state: any,
-  selectors: any[],
+  selectors: Selector<any, any>[] | SelectorWithProps<any, any, any>[],
   props: any,
   memoizedProjector: MemoizedProjection
 ): any {
-  const args = selectors.map(fn => fn(state, props));
-  return memoizedProjector.memoized.apply(null, args);
+  if (props === undefined) {
```
timdeschryver: This check is added because it would break some tests otherwise.
The tests that would fail are the ones that check what the projector is receiving.
I don't know if these are heavily tested, hence the check (I don't want to break any tests on existing code bases).

jpduckwo:
Any chance you could provide an example of how to pass props when using a selector? It's not in the docs...

timdeschryver:
@jpduckwo We will, most likely I'll open up a PR later this week.

timdeschryver:
@jpduckwo see #1233 feel free to give feedback.

PR #1212, opened 2018-07-28
docs(Effects): document defining effects as function

### LINKED ISSUE #1198: Effects: Document effects as functions

[ ] No

Effects can be defined in two ways but only one way is documented.

Property Assignment
```
@Effect()
this.effect$ = this.actions$.pipe(ofType('SOME_ACTION'));
```

Using a function that is called without arguments and returns the effect.
```
@Effect()
effect$() {
  return this.actions$.pipe(ofType('SOME_ACTION'));
}
```

Defining effects using a function allows you to use defaults that can be overridden during testing. This could be a useful pattern in providing test schedulers or custom values during testing.

```
@Effect()
effect$(waitTime = 3000, scheduler: Scheduler | undefined = undefined) {
  return this.actions$.pipe(ofType('SOME_ACTION')).pipe(delay(waitTime, scheduler));
}
```

timdeschryver:
Closes #1198

### DIFF
--- a/docs/effects/testing.md
+++ b/docs/effects/testing.md
@@ -104,3 +104,43 @@ describe('My Effects', () => {
   });
 });
 ```
+
+### Effects as functions
+
+Effects can be defined as functions as well as variables. Defining an effect as a function allows you to define default values while having the option to override these variables during tests. This without breaking the functionality in the application.
+
+The following example effect debounces the user input into from a search action.
+
+```ts
+@Effect()
+search$ = this.actions$.pipe(
+  ofType<Search>(BookActionTypes.Search),
+  debounceTime(300, asyncScheduler),
+  switchMap(...)
+```
+
+The same effect but now defined as a function, would look as follows:
+
+```ts
+@Effect()
+// refactor as input properties and provide default values
+search$ = ({
+  debounce = 300,
+  scheduler = asyncScheduler
+} = {}) => this.actions$.pipe(
+  ofType<Search>(BookActionTypes.Search),
+  debounceTime(debounce, scheduler),
+  switchMap(...)
+```
+
+Within our tests we can now override the default properties:
+
+```ts
+const actual = effects.search$({
+  debounce: 30,
+  scheduler: getTestScheduler(),
+});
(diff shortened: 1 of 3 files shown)


docs/effects/testing.md
```diff
@@ -104,3 +104,44 @@ describe('My Effects', () => {
   });
 });
 ```
+
+### Effects as functions
+
+Sometimes testing a more complex effect becomes tricky, for example the effect may depend on external functionality like RxJS schedulers, may have an high `interval`/`throttleTime`/`debounceTime`, or ... you probably get the idea.
```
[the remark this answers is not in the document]
brandonroberts: "... you probably get the idea" is too vague. We should at least describe what kind of operators these are.

docs/effects/testing.md
```diff
@@ -104,3 +104,43 @@ describe('My Effects', () => {
...
+
+```ts
+const actual = effects.search$({
+  debounce: 30,
+  scheduler: getTestScheduler(),
+});
+expect(actual).toBeObservable(expected);
+```
+
+Doing this has the extra benefit of hiding implementation details, making your tests less prone to break due to implementation details changes. If you would change the default `debounceTime` you wouldn't have to change your tests, these would still pass.
```
timdeschryver: Should mention this?

docs/effects/testing.md
```diff
@@ -104,3 +104,43 @@ describe('My Effects', () => {
   });
 });
 ```
+
+### Effects as functions
+
+Effects can be defined as functions as well as variables. Defining an effect as a function allows you to define default values while having the option to override these variables during tests. This without breaking the functionality in the application.
+
+For example, if we would have an effect that handles a typeahead we may want to debounce the user's input:
```
brandonroberts: Update to describe the example. 


"The following example effect debounces the user input into from a search action."

timdeschryver: Thanks for being so patient with me 😅

brandonroberts: No problem at all 👍

PR #1218, opened 2018-07-31
build: check schematics-core files in CI

### LINKED ISSUE #1214: Schematics: Check copy:schematics task by CI
## Other information:
Now, we must run `yarn copy:schematics` after modifying files in the schematics-core folder but we can merge pull request without it.
So I suggest add CI diff check job such as the following code.

```
diff-check:
  docker:
    - image: circleci/node:latest-browsers
  steps:
    - checkout
    - restore_cache:
        key: *cache_key
    - run: yarn
    - run: yarn copy:schematics
    - run: git diff --exit-code .
```

Please consider this change, thank you.

[ ] No

kouMatsumoto:
## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] CI related changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

## Why is this change necessarily?
Now, we must run `yarn copy:schematics` after modifying files in the schematics-core folder but we can merge pull request without it.
To prevent it, I add CI a new job.

## Changes I added
I newly added ci the `schematics-core-copy` job.
this job checks whether we executed `yarn copy:schematics` script and its changes are committed surely.
I configure this job to run before test job.

## How does this job work
CI runs `git diff --name-only --exit-code ./modules` command after `yarn copy:schematics` script.
It fails if our commits lacks changes of `yarn copy:schematics` script.

Closes #1214

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,7 @@ jobs:
             - ~/.cache/yarn
             - node_modules
 
-  schematics-core-copy:
+  check-schematics-core-copied:
     <<: *run_in_node_latest_browsers
     steps:
       - checkout
@@ -88,11 +88,6 @@ jobs:
       - run: yarn
       - run: yarn copy:schematics
       - run: git diff --name-only --exit-code ./modules
-      - save_cache:
-          key: *cache_key
-          paths:
-            - ~/.cache/yarn
-            - node_modules
 
   deploy:
     <<: *run_in_ngcontainer
@@ -113,11 +108,11 @@ workflows:
     jobs:
       - lint
       - bazel
-      - schematics-core-copy
+      - check-schematics-core-copied
       - test
       - deploy:
           requires:
-            - schematics-core-copy
+            - check-schematics-core-copied
             - test
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,9 +17,6 @@ var_2: &run_in_ngcontainer
 var_3: &set_bazel_options
   run:
     command: sudo cp .circleci/bazel.rc /etc/bazel.bazelrc
-var_4: &run_in_node_latest_browsers
-  docker:
-  - image: circleci/node:latest-browsers
 
 jobs:
 
@@ -64,12 +61,15 @@ jobs:
       - run: bazel query --output=label //... | xargs bazel test
 
   test:
-    <<: *run_in_node_latest_browsers
+    docker:
+      - image: circleci/node:latest-browsers
     steps:
       - checkout
       - restore_cache:
           key: *cache_key
       - run: yarn
+      - run: yarn copy:schematics
+      - run: git diff --name-only --exit-code ./modules
       - run: yarn run ci
       - run: yarn run example:build:prod
       - run: yarn run example:test --watch=false
@@ -78,17 +78,6 @@ jobs:
           paths:
             - ~/.cache/yarn
             - node_modules
-
-  check-schematics-core-copied:
-    <<: *run_in_node_latest_browsers
-    steps:
-      - checkout
-      - restore_cache:
-          key: *cache_key
-      - run: yarn
-      - run: yarn copy:schematics
-      - run: git diff --name-only --exit-code ./modules
-
   deploy:
     <<: *run_in_ngcontainer
     steps:
@@ -108,11 +97,9 @@ workflows:
     jobs:
       - lint
       - bazel
-      - check-schematics-core-copied
       - test
       - deploy:
           requires:
-            - check-schematics-core-copied
             - test
           filters:
             branches:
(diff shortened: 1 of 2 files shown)


timdeschryver:
This will make the build fail when the schematics aren't copied, right?

kouMatsumoto:
@timdeschryver 
Yes. it's exactly as you said.
Please see the description of this pull request since I updated it, thankyou.

.circleci/config.yml
```diff
@@ -76,6 +78,22 @@ jobs:
...
+  schematics-core-copy:
+    <<: *run_in_node_latest_browsers
+    steps:
+      - checkout
+      - restore_cache:
+          key: *cache_key
+      - run: yarn
+      - run: yarn copy:schematics
+      - run: git diff --name-only --exit-code ./modules
+      - save_cache:
```
timdeschryver: We don't have to save the cache I believe.

.circleci/config.yml
```diff
@@ -76,6 +78,22 @@ jobs:
           paths:
             - ~/.cache/yarn
             - node_modules
+
+  schematics-core-copy:
```
timdeschryver: I would change this because this seems like it copies the file. 
Something as `check-schematics-core-copied`?

kouMatsumoto: Exactly, it seems more easier to understand, thank you.

.circleci/config.yml
timdeschryver: Actually, after giving it more thought:

The test job is running in parallel with your new build step and these rely both on the yarn cache.
Personally it makes more sense to me to first check if the files are copied and if these are,l to run the tests. Otherwise the tests may result in a false positive - a test may turn green until you copied the schematic files. 

To fix this we can add the install step as you mentioned, then run the check, then run the tests.
Or we could add the git diff check in the current test step.

Perhaps we could add this in the BAZEL flow, but I don't know anything about it 😄

kouMatsumoto: @timdeschryver 
I consider two ways to resolve the matter you mentioned.
The first way is configuring `check-schematics-core-copied` as dependency of `test`.
We check if the files are copied and after that run tests.

```
workflows:
  version: 2
  build-test-deploy:
    jobs:
      - lint
      - bazel
      - check-schematics-core-copied
      - test
          requires:
            - check-schematics-core-copied
      - deploy:
          requires:
            - test
          filters:
            branches:
              only: master
```

In this case, we need move `save_cache` to `check-schematics-core-copied` from 'test' because `check-schematics-core-copied` is executed certainly before `test`.

The second way is create new job to install and save npm dependencies as I mentioned before.

```
yarn-install:
  <<: *run_in_node_latest_browsers
  steps:
    - checkout
    - restore_cache:
        key: *cache_key
    - run: yarn
    - save_cache:
        key: *cache_key
        paths:
        - ~/.cache/yarn
        - node_modules

workflows:
  version: 2
  build-test-deploy:
    jobs:
      - lint
      - bazel
      - check-schematics-core-copied
          requires:
            - yarn-install
      - test
          requires:
            - yarn-install
      - deploy:
          requires:
            - check-schematics-core-copied
            - test
          filters:
            branches:
              only: master
```

I think it's preffer to use this way because it can apply to other jobs created in the future.
Please tell me your opinion :grinning:

timdeschryver: Yep I agree, and I also like option 2 more.

You could do the following to prevent false positives:

```
- test
  requires:
    - check-schematics-core-copied
```

And to be clear, what I meant with the second option I mentioned in my comment above is to modify the test step to the following:
```
test:
    <<: *run_in_node_latest_browsers
    steps:
      - checkout
      - restore_cache:
          key: *cache_key
      - run: yarn
      - run: yarn run ci
      - run: yarn copy:schematics
      - run: git diff --name-only --exit-code ./modules
      - run: yarn run example:build:prod
      - run: yarn run example:test --watch=false
      - save_cache:
          key: *cache_key
          paths:
            - ~/.cache/yarn
            - node_modules
````

I think this is the preferred way because the current steps aren't split either.

kouMatsumoto: Indeed, it's more simpler and understandable for me :smile:.
Please see my update since I push additionally.

In the example code, `run: yarn run ci` executed before `yarn copy:schematics` but I think it should run after `yarn copy:schematics` because it executes test command internally.
Is this correct?

MatsumotoKou:
It seems to be failed test job as expected.

https://circleci.com/gh/ngrx/platform/2669?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

https://github.com/ngrx/platform/pull/1218/commits/6e667efc04a35dd0250a109c645f65c31b29e2f1

.circleci/config.yml
MatsumotoKou: @timdeschryver 
Could you please reply to me? :man-bowing:

timdeschryver: 😅 sorry @kouMatsumoto

PR #1219, opened 2018-08-01
feat(Effects): stringify action on invalid action

timdeschryver:
To have a more meaningful error when the effects retrieve an invalid action we could stringify the action. That way the dev would immediately know which action is wrong, this could be helpful for newcomers.

Message before: 
```
ERROR Error: Effect "AuthEffects.login$" dispatched an invalid action: [object Object]
```

Message after: 
```
ERROR Error: Effect "AuthEffects.login$" dispatched an invalid action: [{"payload":{"user":{"name":"User"}},"type":"[Auth] Login Success"}]
```

### DIFF
--- a/modules/effects/src/effect_notification.ts
+++ b/modules/effects/src/effect_notification.ts
@@ -37,7 +37,7 @@ function reportInvalidActions(
         new Error(
           `Effect ${getEffectName(
             output
-          )} dispatched an invalid action: ${action}`
+          )} dispatched an invalid action: ${stringify(action)}`
         )
       );
     }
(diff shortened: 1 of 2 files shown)


modules/effects/src/effect_notification.ts
```diff
@@ -57,3 +57,11 @@ function getEffectName({
 
   return `"${sourceName}.${propertyName}${isMethod ? '()' : ''}"`;
 }
+
+function stringify(action: any) {
```
brandonroberts: Type to `Action | null | undefined` and add a test

PR #1222, opened 2018-08-01
fix(schematics): Remove peer dependencies on Angular DevKit

### LINKED ISSUE #1206: Schematics: Remove peer dependencies on Angular devkit

[ ] No

There isn't a reasonable solution to this problem and it prevents us from using `ng update` without the `--force` option. We also need to come up with some integration tests that validate across different versions of the CLI schematics because. We can rely on the implicit dependencies under the assumption that if you're using our schematics, you have these libraries installed.

brandonroberts:
The peer dependencies prevent you from using `ng update` since they are pinned to specific versions. We'll verify the schematics work across versions and rely on the dependencies provided implicitly with the CLI

Closes #1206

### DIFF
--- a/modules/schematics/package.json
+++ b/modules/schematics/package.json
@@ -24,9 +24,5 @@
   "ng-update": {
     "packageGroup": "NG_UPDATE_PACKAGE_GROUP",
     "migrations": "NG_UPDATE_MIGRATIONS"
-  },
-  "peerDependencies": {
-    "@angular-devkit/core": "NG_DEVKIT_VERSION",
-    "@angular-devkit/schematics": "NG_DEVKIT_VERSION"
   }
 }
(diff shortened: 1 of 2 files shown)


modules/schematics/package.json
```diff
@@ -24,9 +24,5 @@
   "ng-update": {
     "packageGroup": "NG_UPDATE_PACKAGE_GROUP",
     "migrations": "NG_UPDATE_MIGRATIONS"
-  },
-  "peerDependencies": {
-    "@angular-devkit/core": "NG_DEVKIT_VERSION",
```
timdeschryver: I think this is the only place where we were using `NG_DEVKIT_VERSION`, thus we could also remove it from `defaults.bzl`.

PR #1223, opened 2018-08-01
fix: deprecate instance operators in Store and Effects

brandonroberts:
Adding deprecation notices since we'll be removing these at some point before the next major release.


timdeschryver:
Is there a way to have build warnings when these are used?
If not, I believe peoples would not notice that these operators are deprecated.

We could log a warning inside these methods when they are used? 
And/or if we add an ng-upgrade script we could also log that these are deprecated.

PR #1233, opened 2018-08-06
docs(store): add docs for selector with props

timdeschryver:


davidtlee:
@timdeschryver I have a question just to see if I'm understanding this. What would the benefit of the "props" syntax be over just passing parameters into a selector function? For example, what's the difference between:

Version 1:
```
this.store.pipe(select(getCount, { multiple: 10 }))
```
and
```
export getCount = createSelector(
  getCounterValue,
  (counter, props) => counter * props.multiply
)
```

Version 2
```
this.store.pipe(select(getCount(10)))
```
and
```
export getCount = (multiply) => createSelector(
  getCounterValue,
  (counter) => counter * multiply
)
```

timdeschryver:
In the example you wrote the behavior will be the same. 
The only difference is that `props` will also get passed to the `getCounterValue` selector.

davidtlee:
Ah, got it, thanks!

PR #1240, opened 2018-08-11
feat(Store): dispatch one update action when multiple features are added or removed

timdeschryver:
BREAKING CHANGE:

BEFORE:
```ts
{type: '@ngrx/store/update-reducers', feature: 'feat1'}
{type: '@ngrx/store/update-reducers', feature: 'feat2'}
```

AFTER:
```ts
{type: '@ngrx/store/update-reducers', features: ['feat1', 'feat2']}
```


modules/store/src/reducer_manager.ts
```diff
@@ -93,12 +93,17 @@ export class ReducerManager extends BehaviorSubject<ActionReducer<any, any>>
   private updateReducers(featureKeys: string[]) {
     this.next(this.reducerFactory(this.reducers, this.initialState));
 
-    featureKeys.forEach(feature => {
+    if (featureKeys.length === 1) {
       this.dispatcher.next(<Action>{
         type: UPDATE,
-        feature,
+        feature: featureKeys[0],
```
brandonroberts: There should be an array dispatched either way, with one more feature keys in the array.

timdeschryver: 👍 got it, so always a `features` array. I'll update it tomorrow.

govi2010:
Issue still there for Version 7.

Issue #1146, opened 2018-06-23
Add documentation on ng add for supported packages


brandonroberts:
## Other information:

Store and Effects can be added via `ng add`, but its not documented. Need to figure out a clear way to provide instructions on cli/npm/yarn installation

## I would be willing to submit a PR for the docs :heart:

[ ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

PR #1243, opened 2018-08-11
Update readme to show that you can use ng add

BeanHeaDza:


timdeschryver:
I wonder if we should add what this command does (setup the store, etc etc) - see #1146.

PR #1249, opened 2018-08-13
Effects: Add support for effects of different instances of same class

### LINKED ISSUE #1246: Effects: Instantiating Effects multiple times, only first instance is run

When instantiating an effect multiple times and adding it to EffectSources, only the first instance is run, all others are ignored.

https://stackblitz.com/edit/angular-cfqrvf

## Expected behavior:

Adding instances through EffectSources.addEffects should work when providing multiple instances of the same class, so this should work:

```
class ConfigurableEffect {
  @Effect({ dispatch: false })
  doSth$ = timer(1000).pipe(
    tap(() => console.log(this.message))
  );

  constructor(private message: string) { }
}

@NgModule({
  imports: [BrowserModule, FormsModule, StoreModule.forRoot(() => true), EffectsModule.forRoot([])],
  declarations: [AppComponent, HelloComponent],
  bootstrap: [AppComponent]
})
export class AppModule {
  constructor(effectSources: EffectSources) {
    effectSources.addEffects(new ConfigurableEffect("configA"));
    effectSources.addEffects(new ConfigurableEffect("configB"));
    // -> should run both "configA" and "configB" after 1 second, but only "configA" is run
  }
}
```

Latest

## Other information:

This is essentially blocking us from using effects in a configurable way.

[ ] No

dummdidumm:
Fixes #1246 

With this change, it is possible to add different instances of the same class to the effects and have them running:

```
class ConfigurableEffect {
  @Effect({ dispatch: false })
  doSth$ = timer(1000).pipe(
    tap(() => console.log(this.message))
  );

  constructor(private message: string) { }
}

@NgModule({
  imports: [BrowserModule, FormsModule, StoreModule.forRoot(() => true), EffectsModule.forRoot([])],
  declarations: [AppComponent, HelloComponent],
  bootstrap: [AppComponent]
})
export class AppModule {
  constructor(effectSources: EffectSources) {
    effectSources.addEffects(new ConfigurableEffect("configA"));
    effectSources.addEffects(new ConfigurableEffect("configB"));
    // -> this did not work before, second effects instance was never run
  }
}
```

`EffectSources` still filters out effects which are already running, so no danger of running the same effects instance multiple times.

### DIFF
--- a/modules/effects/src/effect_sources.ts
+++ b/modules/effects/src/effect_sources.ts
@@ -11,7 +11,6 @@ import {
 } from 'rxjs/operators';
 
 import { verifyOutput } from './effect_notification';
-import { getSourceForInstance } from './effects_metadata';
 import { resolveEffectSource } from './effects_resolver';
 
 @Injectable()
@@ -29,7 +28,7 @@ export class EffectSources extends Subject<any> {
    */
   toActions(): Observable<Action> {
     return this.pipe(
-      groupBy(getSourceForInstance),
+      groupBy(source => source),
       mergeMap(source$ =>
         source$.pipe(
           exhaustMap(resolveEffectSource),
--- a/modules/effects/spec/effect_sources.spec.ts
+++ b/modules/effects/spec/effect_sources.spec.ts
@@ -92,12 +92,23 @@ describe('EffectSources', () => {
     });
 
     it('should ignore duplicate sources', () => {
+      const sources$ = cold('--a--a--a--', {
+        a: new SourceA(),
+      });
+      const expected = cold('--a--------', { a });
+
+      const output = toActions(sources$);
+
+      expect(output).toBeObservable(expected);
+    });
+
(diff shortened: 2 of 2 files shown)


modules/effects/src/effect_sources.ts
```diff
@@ -29,22 +21,42 @@ export class EffectSources extends Subject<any> {
    */
   toActions(): Observable<Action> {
     return this.pipe(
-      groupBy(getSourceForInstance),
-      mergeMap(source$ =>
-        source$.pipe(
-          exhaustMap(resolveEffectSource),
-          map(output => {
-            verifyOutput(output, this.errorHandler);
+      scan(
```
timdeschryver: I might be overlooking something but won't the following accomplish the same thing?

```ts
groupBy(source => source),
```

dummdidumm: You are completely right, thanks for noticing! I will update the PR.

Issue #1263, opened 2018-08-20
Router-Store: Provide configuration for pre/post router navigation action


brandonroberts:
## If accepted, I would be willing to submit a PR for this feature

[ ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

Currently, router-store dispatches a `ROUTER_NAVIGATION` action as a pre-activation hook which is run before guards and resolvers. We should provide an option to let this be configured to run as a `post-activation` hook that is run after guards and resolvers.

Also add an action that is dispatched when the navigation cycle ends successfully and the current route is active.

dummdidumm:
#1010 related

dummdidumm:
Looking at the angular router docs, there is not much in between "after guards and resolvers" and before "navigation end". There is `ActivationEnd` but that can be fired multiple times, resulting in multiple dispatched actions, same for `ChildActivationEnd`. After those two events, there is only NavigationEnd. So `ROUTER_NAVIGATION` will be dispatched at the same time like a possible `ROUTER_NAVIGATED`

brandonroberts:
Fixed via https://github.com/ngrx/platform/pull/1267

Issue #1262, opened 2018-08-20
Router-Store: Add custom serializer to config object


brandonroberts:
## If accepted, I would be willing to submit a PR for this feature

[ ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

Currently you have to use the provider object syntax to override the default router state serializer. This can be easier by exposing a `serializer` option in the configuration with the custom serializer class. If none is provided, use the default.

dummdidumm:
#1010 related

timdeschryver:
Can I take this one? Or are you already working on it?

dummdidumm:
I have PR in the making implementing this and the other enhancement suggestions. I will request the merge soon.

cyberbeast:
Is this available in ngrx/router-store ^6.1.0?

timdeschryver:
This isn't available in 6.1. You could install ngrx from the nightly builds, these builds have the latest features.

cyberbeast:
@timdeschryver Thank you for that information. I'll check that out.

Issue #975, opened 2018-04-17
NgRx router integration dispatches the navigation event before the preactivation phase (Guards and Resolvers) executes.


russoturisto:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
</code></pre>

Following the advice of vsavkin from nrwl/nx to open an issue here (couldn't find an open issue and I know the problem is still there in 5.2.0):

https://github.com/nrwl/nx/issues/258

## What is the current behavior?

"DataPersistence.navigation relies on the NgRx router integration. The integration dispatches the navigation event before the preactivation phase (Guards and Resolvers) executes."

## Expected behavior:

"NgRx router integration dispatches the navigation event AFTER the preactivation phase (Guards and Resolvers) executes.

## Minimal reproduction of the problem with instructions:

please see:
https://github.com/nrwl/nx/issues/258

## Version of affected browser(s),operating system(s), npm, node and ngrx:

5.2.0

## Other information:

We are stuck on 5.2.X due to customer commitments, is it possible to fix this in 5.2.X?

Thanks, :)

bbaia:
Relates to #816

dummdidumm:
Maybe we should think about enhancing the features of the router module in general.

One of the things that comes to my mind is to rework the actions that are dispatched and at what time they are dispatched:

- Navigation Request -> during first navigation lifecycle event.
- Navigation Done -> after last navigation lifecycle event but not if its an error or cancel.
- Navigation Error -> on error
- Navigation Cancel -> on cancel

redjab:
Any response/update on this issue? I'm running into this recently, though with a slightly different scenario.
Imo, different actions should be dispatched depending on the router event, and only on NavigationEnd, NavigationError, and NavigationCancel should the params & queryParams in the router store gets updated, because only then you know for sure that these are going to be correct.

brandonroberts:
Fixed via https://github.com/ngrx/platform/pull/1267

PR #1267, opened 2018-08-21
feat(Router-Store): New actions, config enhancement

### LINKED ISSUE #1010: RFC: RouterStore changes & enhancements
Related to #975 I'd like to propose some changes / enhancements to the RouterStore.

# Breaking Changes:
- The new default state serializer is that of (or similar to) the example app. It's all you ever need in 90% of the cases. Old defaultSanitizer can be used via flag (see API enhancements).

# API enhancements:
new forRoot properties:
`useEnhancedRouterState`  (boolean):  if true, use old default serializer. Defaults to false.
`routerStateSanitizer` (function): shortcut for `{ provide: RouterStateSerializer, useClass: CustomRouterStateSerializer }`. Defaults to default sanitizer.

# New actions in line with router navigation events
`ROUTER_REQUEST` - fired on NavigationStart
`ROUTER_NAVIGATED` - fired on NavigationEnd
`ROUTER_CANCEL` -> fired on NavigationCancel (same behavior as today)
`ROUTER_ERROR` -> fired on NavigationError (same behavior as today)

The `ROUTER_NAVIGATION` action could stay as is or be an alias for ROUTER_NAVIGATED and/or marked as deprecated to remove it in the next major release.

# Other enhancements / changes:
1. Setting the router reducer in StoreModule.forRoot is no longer necessary (but can still be done)
2. Define a base interface for the router state which all reducer implementations should follow. Right now there is no definition when writing your own and the RouterStore relies on a certain structure. That's a little brittle.
3. It should be possible to redirect/abort navigation inside the reducer without throwing errors
4. To achieve 3, the state is enhanced with a "request" property which contains the desired next route. Will be set on `ROUTER_REQUEST` actions and cleared on other actions. The RouterStore uses this to compare the request object of the state with that from the navigation it just did. If it changed, it will issue a redirect/cancel.

These are all just suggestions and not all of these (or any) need to be implemented. What do you think?

dummdidumm:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

Closes #1010, #975, #1262, #1263 

- New actions ROUTER_REQUEST and ROUTER_NAVIGATED that are dispatched at the very start/end of the navigation cycle
- ROUTER_NAVIGATION action can now be configured to be dispatched after guards and resolvers
- Custom router state serializer can now be passed as part of config
- Internals: No longer using internal Angular router hook, instead official router events 

```
[x] Yes
[] No
```

StoreRouterConfigFunction is removed. It is no longer possible to pass a function returning a StoreRouterConfig to StoreRouterConnectingModule.forRoot

If you still need this, pass a provider like this:
```
{
      provide: ROUTER_CONFIG,
      useFactory: _createRouterConfig // your function
}
```

## Other information

I will update the documentation if the changes are approved.


modules/router-store/src/serializer.ts
```diff
@@ -1,10 +1,18 @@
 import { ActivatedRouteSnapshot, RouterStateSnapshot } from '@angular/router';
 
-export abstract class RouterStateSerializer<T> {
+/**
+ * Simple router state.
+ * All custom router states / state serializers should have at least this property.
+ */
+export interface RouterState {
```
brandonroberts: This interface needs a different name as the `@angular/router` already exposes a `RouterState`. It also needs to be added to `src/index.ts`

modules/router-store/src/router_store_module.ts
```diff
@@ -9,18 +9,38 @@ import {
...
-import { of } from 'rxjs';
 
 import {
   DefaultRouterStateSerializer,
   RouterStateSerializer,
   SerializedRouterStateSnapshot,
+  RouterState,
 } from './serializer';
 
+/**
```
brandonroberts: These new exports need to be added to `src/index.ts`

modules/router-store/src/router_store_module.ts
```diff
@@ -118,6 +175,14 @@ export function routerReducer<T = SerializedRouterStateSnapshot>(
...
 export interface StoreRouterConfig {
   stateKey?: string;
+  serializer?: new () => RouterStateSerializer;
+  /**
+   * By default, ROUTER_NAVIGATION is dispatched before guards and resolvers run.
+   * Therefore, the action could run too soon, for example
+   * there may be a navigation cancel due to a guard saying the navigation is not allowed.
+   * To run ROUTER_NAVIGATION after guards and resolvers, set this property to true.
+   */
+  dispatchNavActionOnEnd?: boolean;
```
brandonroberts: I think this should be more flexible. 

```
enum NavigationAction {
  PreActivation = 1,
  PostActivation = 2
};

navigationAction?: NavigationAction
```

The default would be PreActivation

modules/router-store/src/router_store_module.ts
```diff
@@ -141,12 +206,28 @@ export function _createDefaultRouterConfig(
...
+): RouterStateSerializer {
+  // This function gets handed a complete config-object from _createRouterConfig,
+  // so we know the serializer property exists
+  return new config.serializer!();
+}
+
 export type StoreRouterConfigFunction = () => StoreRouterConfig;
 
+enum RouterTrigger {
+  NONE,
```
brandonroberts: Provide explicit values

modules/router-store/src/router_store_module.ts
```diff
@@ -141,12 +206,28 @@ export function _createDefaultRouterConfig(
...
     ..._config,
   };
 }
 
+export function _createSerializer(
+  config: StoreRouterConfig
+): RouterStateSerializer {
+  // This function gets handed a complete config-object from _createRouterConfig,
+  // so we know the serializer property exists
+  return new config.serializer!();
```
brandonroberts: This doesn't support dependencies being injected into the serializer

dummdidumm: Is this even possible without jumping through fire hoops? The problem is that we don't know the dependencies, so we somehow have to get them dynamically. But to my knowledge, getting these and making sure they exist (and those who are transitively required, too) is not really supported by Angular through some public API. It is only supported to statically define the dependencies.

So we would end up either using Angular internals which is brittle, or write the DI-mechanisms again.

If this is not really possible, maybe we should ditch the serializer option..

@timdeschryver any ideas?

timdeschryver: I didn't have the time to take this all in, I'll try to get back at this later today.

modules/router-store/src/serializer.ts
```diff
@@ -1,10 +1,20 @@
 import { ActivatedRouteSnapshot, RouterStateSnapshot } from '@angular/router';
 
-export abstract class RouterStateSerializer<T> {
+/**
+ * Simple router state.
+ * All custom router states / state serializers should have at least this property.
+ */
+export interface SimpleRouterState {
```
brandonroberts: What about BaseRouterState?

modules/router-store/src/router_store_module.ts
brandonroberts: It's already being injected in the constructor, which will resolve its dependencies. Just use the config to tell it which one to provide. Remove the `StoreRouterConfigFunction` and move all the providers into the `forRoot` method. It will be breaking, but more consistent

```
  static forRoot(
    config: StoreRouterConfig | StoreRouterConfigFunction = {}
  ): ModuleWithProviders {
    return {
      ngModule: StoreRouterConnectingModule,
      providers: [
          { provide: _ROUTER_CONFIG, useValue: config },
          { provide: RouterStateSerializer, useClass: config.serializer ? config.serializer : DefaultRouterStateSerializer }
     ],
    };
  }
```

modules/router-store/src/serializer.ts
brandonroberts: Or BaseRouterStoreState

modules/router-store/src/router_store_module.ts
dummdidumm: Ok when there is no function to execute it will work (with AOT, too).

brandonroberts:
Thanks @dummdidumm!

brandonroberts:
@sbkpilot1 it won't go out until the next release, which isn't ready yet. If you want to use it now you'll need to use the nightly builds

jasanst:
Did the documentation get released before the functionality? 

I'm having trouble using the serializer in StoreRouterConnectingModule.forRoot as shown in the api - it doesn't seem to exist yet?

timdeschryver:
@jasanst yea it's an unreleased feature. 
If you want you can install the packages from the nightly builds.

jasanst:
Thanks, just getting started so would rather see documentation for the current prod version instead, is this still available?

brandonroberts:
@jasanst you can get the latest release on npm through the tag. https://github.com/ngrx/platform/tree/6.1.0

sadsa:
bump

DonatasD:
Is this fix available, in any stable release?

Issue #1336, opened 2018-09-17
Docs: Rewrite overview guide for Effects


brandonroberts:
## I would be willing to submit a PR for the docs :heart:

[ ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

Introduction (What and Why)
How to write effects
Example

PR #1338, opened 2018-09-22
[docs/effects/api.md] defer can be lead to misconception

SerkanSipahi:
The part of the description `... executed directly -> after <- the effect...` can lead to misconception in relation to the `defer` observable. It doesnt matter whether the `defer` observable is there or not! The code will always executed directly after the effect class is loaded.
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/docs/effects/api.md
+++ b/docs/effects/api.md
@@ -146,8 +146,8 @@ export class SomeEffectsClass {
   constructor(private actions$: Actions) {}
 
   @Effect({ dispatch: false })
-  init$: Observable<any> = of(null)).pipe(
-    tap(() => console.log('init$')
+  init$: Observable<any> = of(null).pipe(
+    tap(() => console.log('init$'))
   );
 }
 ```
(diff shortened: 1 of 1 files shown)


kouMatsumoto:
I think `defer` is correct as intently. "After the effect class is loaded" means "After the effect is subscribed by EffectsRunner".

Consider below

```
@Injectable()
export class SomeEffectsClass {
  constructor(private actions$: Actions, private) {}

  // console.log will be executed on SomeEffectsClass instance created and property initialized.
  @Effect({ dispatch: false })
  init$: Observable<any> = of(null).pipe(
    tap(() => console.log('on property initialize!'))
  );

  // console.log will be executed on SomeEffectsClass instance created and loaded by EffectsRunner (actually enabled).
  @Effect({ dispatch: false })
  init$: Observable<any> = defer(() => of(null)).pipe(
    tap(() => console.log('on effect loaded!'))
  );
}
```

timdeschryver:
I wonder if we can leave this part out because we now have our init actions and init feature actions.
@brandonroberts

brandonroberts:
Yes, this should be updated to use the effects init actions.

timdeschryver:
Closing this in favor of https://github.com/ngrx/platform/issues/1336.

PR #1355, opened 2018-10-07
docs: clarify that docs are for master branch

thefliik:
## PR Checklist

## PR Type

```
```

Currently, the docs linked to from the platform README are for the master branch / nightly builds. I often come back to the docs ahead of a major angular release (4->5, 5->6, 6->7) and I believe that every time I've looked at them they've reflected an unreleased api. This might be bad luck on my part, but I look at the docs infrequently enough that I always forget that they don't apply to the package version I'm using.

## What is the new behavior?

This PR adds a helpful message to the readme reminding folks that it's linking to the master branch version of the docs, and explaining how to find the docs for a specific release.

```
[x] Yes
[ ] No
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,12 @@ Become a sponsor and get your logo on our README on Github with a link to your s
 
 ## Packages
 
+***Note:** The links below will take you to the master branch version of the docs.
+To view the docs corresponding to a specific release (which is probably what you want),
+you need to look up the commit ID for a release and browse the snapshot of this repo at that point
+in time. One way to do this is to view the ["releases"](https://github.com/ngrx/platform/releases)
+tab in github, click the commit hash link, and then click "browse files".*
+
 - [@ngrx/store](./docs/store/README.md) - RxJS powered state management for Angular applications, inspired by Redux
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,12 @@ Become a sponsor and get your logo on our README on Github with a link to your s
 
 ## Packages
 
+***Note:** The links below will take you to the master branch version of the docs.
```
timdeschryver: I'm up for this change, but it's easier to browse via `tags`.

image

thefliik: @timdeschryver right you are! Didn't know about that option. I updated the suggestion text.

thefliik:
Feedback addressed

brandonroberts:
We will be moving the docs to an official site soon, so these caveats will be obsolete. The docs will be based off the current published release instead of master.

thefliik:
@brandonroberts Awesome! Ok, feel free to close. Though if "soon" might turn out to be longer then anticipated, I'd advocate for merging this change now and then simply removing it when the time comes. Might help someone without an apparent downside.

timdeschryver:
I was thinking about putting this message until the site is published, because we're getting some issues/questions because users are using unreleased features.

Issue #1230, opened 2018-08-06
Effects: Write migration from V6 to V7 to handle compatibility changes


brandonroberts:
## If accepted, I would be willing to submit a PR for this feature

[ ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

With https://github.com/ngrx/platform/issues/860 and the removal of instance operators, we'll need a way to migrate users usage of the existing `Actions` from `@ngrx/effects` to `Actions` from `@ngrx/effects/compat` and to the pipeable `ofType` operator where necessary.

This also covers migrating from `store.select()` to `store.pipe(select())`

timdeschryver:
I've modified `rxjs-tslint` to our needs at [ngrx-tslint](https://github.com/timdeschryver/ngrx-tslint).
I would appreciate it if some would try it out and give their feedback, **be sure to commit your work before running it** 😅.

Usage:
- Fork the repo: `git clone https://github.com/timdeschryver/ngrx-tslint.git`
- Install deps: `npm install`
- Create a release: `npm run release`
- Create an npm pack: `cd dist && npm pack && cd ..` - this should create a file `ngrx-tslint-versionNumber.tgz`
- Use the created file to install the linter in your project: `npm install path-to-file --save-dev`
- Add the rules to your `tslint.json`
- Run the linter with the `--fix` flag: `ng lint projectName --fix`

```json
{
  "rulesDirectory": ["node_modules/ngrx-tslint"],
  "rules": {
    "ngrx-store-operators": true,
    "ngrx-effects-operators": true,
    "ngrx-chained-pipes": true
  }
}
```

If you need help, feel free to ping me.

Issue #639, opened 2017-12-08
ngrx for communicating  ng1 and ng5 side-by-side applications 


dankrz:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ x] Support request
</code></pre>

## Expected behavior:

Hello all,

I would like to ask if it possible to share on state between two separate apps one in ng1 and another in ng5. They are not connected with downgrade mechanism.

Best regard,
Krzysztof

brandonroberts:
This isn't possible out of the box without the downgrade to link the apps. You'd have to create some synching mechanism between the two apps.

alex-okrushko:
Having two separate apps doesn't make a lot of sense to me. You'd want eventually to have the whole app as Angular, right?

If they are connected via downgrade it's possible to use Store in angularjs. In fact, that's what we do for Firebase console.

here is an example (code that we have differs a bit, but general idea is the same):

```typescript
// store_service.ts
export class StoreService {
  static NG_NAME = 'storeService';

  private storePromise: ng.IPromise<Store<GlobalState>>;
  private resolveStorePromise: (store: Store<GlobalState>) => void;

  /**
   * @ngInject
   */
  constructor($q: ng.IQService) {
    this.storePromise = $q((resolve) => this.resolveStorePromise = resolve);
  }

  setStore(store: Store<GlobalState>) {
    this.resolveStorePromise(store);
  }

  /**
   * redirect dispatch to the store.
   */
  dispatch(action: Action) {
      this.storePromise.then(store => store.dispatch(action));
    }
 }
```

Wiring of the store service in angularjs:
```typescript
// store_module.ts
export const storeModule =
    angular
        .module('firebase.console.storemodule')
        .service(StoreService.NG_NAME, StoreService);
```

Now providing and setting the store in the Angular. This is the main module that is being downgraded into angularjs app (via `downgradeModule(AppModuleNgFactory)`)
```typescript
// app_module.ts

// we have it in a separate file.
export function firealertsServiceFactory(
    injector: ng.auto.IInjectorService, store: Store<GlobalState>) {
  const service = injector.get<StoreService>(StoreService.NG_NAME);
  service.setStore(store); // passing the store here.
  return service;
}

@NgModule({
  imports: [
    ... // our imports
  StoreModule.forRoot({}),
  ],
  providers: [{
    provide: StoreService,
[...]

PR #1361, opened 2018-10-11
feat(Store): remove select function on store

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1230

```
[x] Yes
[ ] No
```

BREAKING CHANGES:

The store select functions are removed in favor of pipeable operators

BEFORE:
```ts
this.user = this.store.select(getUser);
```
AFTER:
```ts
this.user = this.store.pipe(select(getUser));
```

### DIFF
--- a/modules/effects/spec/effects_feature_module.spec.ts
+++ b/modules/effects/spec/effects_feature_module.spec.ts
@@ -172,7 +172,7 @@ class FeatureEffects {
 
   @Effect()
   effectWithStore = this.actions.ofType('INCREMENT').pipe(
-    withLatestFrom(this.store.select(getDataState)),
+    withLatestFrom(this.store.pipe(select(getDataState))),
     map(([action, state]) => ({ type: 'INCREASE' }))
   );
 }
--- a/modules/store/spec/store.spec.ts
+++ b/modules/store/spec/store.spec.ts
@@ -147,8 +147,8 @@ describe('ngRx Store', () => {
       counterSteps.subscribe(action => store.dispatch(action));
 
       const counterStateWithString = feature
-        ? (store as any).select(feature, 'counter1')
-        : store.select('counter1');
+        ? (store as any).pipe(select(feature, 'counter1'))
+        : store.pipe(select('counter1'));
 
       const counter1Values = { i: 1, w: 2, x: 0, y: 1, z: 2 };
 
--- a/modules/store/src/store.ts
+++ b/modules/store/src/store.ts
@@ -19,60 +19,6 @@ export class Store<T> extends Observable<T> implements Observer<Action> {
     this.source = state$;
   }
 
-  /**
-   * @deprecated from 6.1.0. Use the pipeable `select` operator instead.
-   */
-  select<K>(mapFn: (state: T) => K): Observable<K>;
-  select<a extends keyof T>(key: a): Observable<T[a]>;
-  select<a extends keyof T, b extends keyof T[a]>(
-    key1: a,
-    key2: b
-  ): Observable<T[a][b]>;
-  select<a extends keyof T, b extends keyof T[a], c extends keyof T[a][b]>(
-    key1: a,
-    key2: b,
-    key3: c
-  ): Observable<T[a][b][c]>;
-  select<
-    a extends keyof T,
-    b extends keyof T[a],
-    c extends keyof T[a][b],
-    d extends keyof T[a][b][c]
-  >(key1: a, key2: b, key3: c, key4: d): Observable<T[a][b][c][d]>;
-  select<
-    a extends keyof T,
-    b extends keyof T[a],
-    c extends keyof T[a][b],
-    d extends keyof T[a][b][c],
-    e extends keyof T[a][b][c][d]
-  >(key1: a, key2: b, key3: c, key4: d, key5: e): Observable<T[a][b][c][d][e]>;
-  select<
-    a extends keyof T,
-    b extends keyof T[a],
-    c extends keyof T[a][b],
-    d extends keyof T[a][b][c],
-    e extends keyof T[a][b][c][d],
-    f extends keyof T[a][b][c][d][e]
-  >(
-    key1: a,
-    key2: b,
-    key3: c,
-    key4: d,
-    key5: e,
-    key6: f
-  ): Observable<T[a][b][c][d][e][f]>;
-  /**
-   * This overload is used to support spread operator with
-   * fixed length tuples type in typescript 2.7
-   */
-  select<K = any>(...paths: string[]): Observable<K>;
-  select(
-    pathOrMapFn: ((state: T) => any) | string,
-    ...paths: string[]
-  ): Observable<any> {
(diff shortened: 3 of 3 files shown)


MikeRyanDev:
CC'ing the NgRx contributors here for a minute to discuss this some more. @alex-okrushko @rkirov @brandonroberts 

I was speaking to Brandon about this in some hallway (@alex-okrushko's favorite) and I don't think we've reasonably answered these questions:

1. Do users want to use `select` on non-Store observables?
2. Is this worth making `spyOn(Store, 'select')` harder?

I'm really worried about #2 in particular. If you want to mock out specific streams, spying on `select` is pretty straightforward and has helped us avoid introducing a Store-specific testing module.

I know RxJS is planning on ditching prototype operators altogether in the future, but I don't think we need to sacrifice the ergonomics of the NgRx-specific operators (including `ofType`).

My position is that we un-deprecate `Store#select` and `Actions#ofType`.

rkirov:
I don't have enough experience with testing ngrx, but I am surprised that mocking `select` is needed that often. Isn't using Angular's DI to inject a fake Store not a good alternative? Users are "paying" for angular's DI in complexity, so might as well rely on it more?

In general, I feel suspicious of classes extending `Observable<T>`, similarly to extending `Array<T>` or `Promise<T>`. The contract surface of these object is large, that it is risky to override any of it, without butchering something. Even if your implementation doesn't override any methods just adds new ones (like store), I find it much more reassuring if the API presents itself as just `Observable<T>` or `Array<T>`.  I would argue that has much greater "ergonomics" to developers and removing `Store#select` and `Actions#ofType` is a step in that direction - nrgx being just a way of constructing Observables & operators for them.

I feel a lot of it goes back to the old saying 

> "OOP makes code understandable by encapsulating moving parts. FP makes code understandable by minimizing moving parts."

Likely this discussions will devolve into FP vs OOP aesthetics.

timdeschryver:
I'm with @rkirov here and would like to add:

- I think doing it the "RxJS way" will be helpful for newcomers. They will see the Store as "just" an Observable instead of a "magic blackbox".
- I myself didn't really had the need to mock out the store so far (working on smaller projects). Not mocking the store also helped us out by spotting bugs early on, which wouldn't be the case (I believe) if we would have mocked out the store or its selectors.
- Because of your second point @MikeRyanDev, I would see that un-deprecating `Store#select` has its value. But I don't see a value of supporting `Actions#ofType`.
- FWIW, I took a look at [redux-observable](https://redux-observable.js.org/) and they don't support `actions.ofType`

alex-okrushko:
1) Does `Store` need to extend `Observable`? It contains `actionsObserver`, maybe that should be it? I'd rather store to have a simple API. Moreover, I'm preaching the generation of ViewModels through selectors and thus most of the `select` calls are `this.store.pipe(select(selectors.getFoo));` without any further piping.

2) Even if `select` is removed, `dispatch` is not going anywhere.

3) Also, it would make it a bit harder to link it AngularJS side (https://github.com/ngrx/platform/issues/639#issuecomment-350472083)

So I'm for un-depricating `select` as well.

MikeRyanDev:
@alex-okrushko Yeah, I think you've hit the nail on the head: I don't think Store needs to be an observable either. I'd be totally fine if it just implemented the Observer interface and had select/dispatch methods on it.

michaeljota:
I understand that `Store#select` is more friendly that `select` as a pipe operator, but I find `select` function more useful than the prototype version, as you can in your select nested properties in a type safe way. I think you should take advantage of using `rxjs`. Any Angular developer should now the basic usage of that library, as even Angular cover it in their tutorial. If using a library that the framework use is hard, I don't think is any less complex adding a library to manage state.

alex-okrushko:
@michaeljota I don't think anyone disagrees that `RxJS` is great or it is not taken advantage of, after all the entire NgRx is based on it. Both `select` method and operator provide the same level of type safety.
One of the concerns is, would you have any other operators in a pipe prior to `select`? Does `store.pipe(map(...), select(...))` even make sense?

michaeljota:
> Does store.pipe(map(...), select(...)) even make sense?

That's a fair point, however, as I said, being able to use store.pipe(select(...), select(...)) is an advantage in some scenarios.

alex-okrushko:
> being able to use store.pipe(select(...), select(...)) is an advantage in some scenarios.

I don't think I've ever seen anything like that. Selectors typically go from root/feature states down. Could you provide an example?

brandonroberts:
@timdeschryver can we close this and open a PR to un-deprecate `store.select`?

timdeschryver:
@brandonroberts we sure can!

alex-okrushko:
@benlesh I think the team wants to remove the extension of RxJS class in favor of containment. That's unrelated to the `select` being either the method of the Store or a stand-alone operator.

icepeng:
I think `Store` should not extend `Observable`. It should work like `FormGroup` in Angular.

We can subscribe a value from `formGroup.valueChanges.subscribe()`, and to subscribe children `FormControl` (or `AbstractControl`), `formGroup.controls.formControl.valueChanges.subscribe()`.

Store is a "State tree", and `select()` means "Selecting a partial tree". It is similar to selecting children form controls in `@angular/forms`.

The difference is `formGroup.valueChanges` only emits on value changes, but `store.select()` works like `BehaviorSubject`, replaying last emit on new subscription.

So my proposal is:
```
// Dispatch
store.dispatch()

// Subscribe
store.state.subscribe()
store.select(fromRoot.getState).state.subscribe()
```

```
class Store<T> {
  state: Observable<T>
  dispatch(action: Action) {...}
  select<T, Props, K>(mapFn: (state: T, props: Props) => K, props?: Props): Store<V> {
    ...
  }
}
```

defernan:
so is the direction moving forward both in favor of store.select and store.pipe(select(...))? 
What is the likelihood one will become deprecated?

justbaum30:
Also wondering what the future direction is. Which is preferred? `store.select` or `store.pipe(select())`. I would like to make sure our project is using the preferred way.

alex-okrushko:
I'm currently recommending `store.select`

GitFlip:
Is `store.select` then still the preferred way? We might find some push against suggesting its use on our projects if all the examples of selectors on the NgRx doc pages use `store.pipe` instead:

https://ngrx.io/guide/store/selectors
https://ngrx.io/guide/store

"Yes we know the docs show examples of using `store.pipe`, but there is this one PR on GitHub that has a comment that says `store.select` is recommended."

alex-okrushko:
Yes, `store.select` is still recommended.

GitFlip:
> Yes, `store.select` is still recommended.

Thanks for the quick reply! Any indication on whether or not the examples on the docs will be updated to reflect the recommendation?

alex-okrushko:
@GitFlip Thanks
Opened an issue to refactor it - help wanted 😀

PR #1370, opened 2018-10-16
refactor(entity): update types to interfaces

alex-okrushko:
## PR Checklist

- [x ] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[ ] Feature
[ x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Entities are using `export type` a lot. While they should work similar to `interface`, the latter is a preferred choice.
Moreover, `tsickle` tool has an issue type-annotating abstract classes that implement `type` at the moment: https://github.com/angular/tsickle/issues/912

## What is the new behavior?

The behavoir is unchanged, except it's _more correct_.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -1,20 +1,10 @@
-export interface ComparerStr<T> {
-  (a: T, b: T): string;
-}
-
-export interface ComparerNum<T> {
-  (a: T, b: T): number;
-}
+export type ComparerStr<T> = (a: T, b: T) => string;
+export type ComparerNum<T> = (a: T, b: T) => number;
 
 export type Comparer<T> = ComparerNum<T> | ComparerStr<T>;
 
-export interface IdSelectorStr<T> {
-  (model: T): string;
-}
-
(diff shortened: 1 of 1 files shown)


modules/entity/src/models.ts
```diff
@@ -1,40 +1,40 @@
-export type ComparerStr<T> = {
+export interface ComparerStr<T> {
   (a: T, b: T): string;
-};
+}
```
evmar: I think even simpler is to write this:

type ComparerStr<T> = (a: T, b: T) => string;

Here and below for the other function types.

(tsickle chokes on the old code, but I think the new code won't help there anyway, so you might as well write it in the simplest format, which tsickle also won't choke on.)

alex-okrushko: It takes new code, I verified it. But, yeah, for function types it looks even simpler with `type`

PR #1384, opened 2018-10-28
fix(RouterStore): serialize routeConfig inside the default serializer

### LINKED ISSUE #1363: DefaultRouterStateSerializer: path property missing in routeConfig

Example: https://stackblitz.com/edit/ngrx-seed-nl8b8q?file=src%2Fapp%2Fapp-routing.module.ts

Routing effect based on: https://github.com/vsavkin/state_management_ngrx4

## Expected behavior:

routerConfig should contain path / pathMatch (if it exists in original routerConfig) to allow easy identification of Router rule that was used to match used route. This way of matching is used in above mentioned example.

In example above, if routes in `app-routing.module.ts` are defined as shown in example, it's not possible to distinguish in `effects/routing.effects.ts`, function `handleNavigation` what rule in router definition was used (`data/:id` or `test/:id`) since path / pathMatch keys are missing. This has worked before since routeConfig from ActivatedRouteSnapshot was used directly and it contains these keys.

NgRx: 6.1.0
Angular: 6.1.1
Node: 8.9.4

## Other information:
Current behaviour is caused by: https://github.com/TypescriptID/platform/pull/102/commits/7917a276fa199b19d6d8f2e3d2fa1708e3481c2b

[x] No

timdeschryver:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `routeConfig` isn't being serialized in `DefaultRouterStateSerializer`.

Issue Number: #1363

## What is the new behavior?
The `routeConfig` is being serialized in `DefaultRouterStateSerializer`.
Note that not all properies are being serialized:

```ts
{
    // was already being serialized
    component?: Type<any>;

	// added properties to be serialized
    path?: string;
    pathMatch?: string;   
    redirectTo?: string;
    outlet?: string;

    // not being serialized
    matcher?: UrlMatcher;  
    canActivate?: any[];
    canActivateChild?: any[];
    canDeactivate?: any[];
    canLoad?: any[];
    data?: Data;
    resolve?: ResolveData;
    children?: Routes;
    loadChildren?: LoadChildren;
    runGuardsAndResolvers?: RunGuardsAndResolvers;
}
```

```
[x] Yes
[ ] No
```

The default router serializer now returns a `null` value for `routeConfig` when `routeConfig` doesn't exist on the `ActivatedRouteSnapshot` instead of an empty object.
 BEFORE:
 ```json
{
  "routeConfig": {}
}
```
 AFTER:
 ```json
{
  "routeConfig": null
}
```

## Other information
If this doesn't belong in the default serializer, feel free to close.
It's always possible for the impacted users to create there own serializer.

Closes #1363

### DIFF
--- a/modules/router-store/src/serializer.ts
+++ b/modules/router-store/src/serializer.ts
@@ -39,9 +39,15 @@ export class DefaultRouterStateSerializer
       data: route.data,
       url: route.url,
       outlet: route.outlet,
-      routeConfig: {
-        component: route.routeConfig ? route.routeConfig.component : undefined,
-      },
+      routeConfig: route.routeConfig
+        ? {
+            component: route.routeConfig.component,
+            path: route.routeConfig.path,
+            pathMatch: route.routeConfig.pathMatch,
+            redirectTo: route.routeConfig.redirectTo,
+            outlet: route.routeConfig.outlet,
+          }
+        : null,
       queryParams: route.queryParams,
       queryParamMap: route.queryParamMap,
       fragment: route.fragment,
--- a/modules/router-store/spec/serializer.spec.ts
+++ b/modules/router-store/spec/serializer.spec.ts
@@ -1,0 +1,101 @@
+import { RouterStateSnapshot } from '@angular/router';
+import { DefaultRouterStateSerializer } from '../src';
+
+describe('serializer', () => {
+  it('should serialize all properties', () => {
+    const serializer = new DefaultRouterStateSerializer();
+    const snapshot = createSnapshot();
+    const routerState = {
+      url: 'url',
+      root: snapshot,
+    } as RouterStateSnapshot;
+
+    const actual = serializer.serialize(routerState);
+    const expected = {
+      url: 'url',
+      root: createExpectedSnapshot(),
+    };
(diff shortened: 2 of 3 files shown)


modules/router-store/src/serializer.ts
```diff
@@ -39,9 +39,15 @@ export class DefaultRouterStateSerializer
       data: route.data,
       url: route.url,
       outlet: route.outlet,
-      routeConfig: {
-        component: route.routeConfig ? route.routeConfig.component : undefined,
-      },
+      routeConfig: route.routeConfig
```
brandonroberts: Shouldn't the `routeConfig` remain an empty object if its not defined to keep the existing behavior?

brandonroberts: Flag is as a breaking change. I'd rather the behavior be consistent and this would be the time to break it.

timdeschryver: Something like this? I have some troubles to properly word this.

brandonroberts: 👍

PR #1445, opened 2018-11-24
feat(store): support meta reducer factory for feature (#1414)

### LINKED ISSUE #1414: support meta reducer factory for feature

[x ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

itayod:
## PR Checklist

- [ x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[x ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1414 

```
[ ] Yes
[ x] No
```

### DIFF
--- a/modules/store/spec/store.spec.ts
+++ b/modules/store/spec/store.spec.ts
@@ -1,4 +1,4 @@
-import { ReflectiveInjector } from '@angular/core';
+import { ReflectiveInjector, InjectionToken } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { hot } from 'jasmine-marbles';
 import {
@@ -10,12 +10,16 @@ import {
   ReducerManagerDispatcher,
   UPDATE,
   REDUCER_FACTORY,
+  ActionReducer,
+  Action,
 } from '../';
+import { StoreConfig } from '../src/store_module';
 import {
   counterReducer,
   INCREMENT,
   DECREMENT,
   RESET,
+  counterReducer2,
 } from './fixtures/counter';
 import Spy = jasmine.Spy;
 import any = jasmine.any;
@@ -33,7 +37,10 @@ describe('ngRx Store', () => {
   let store: Store<TestAppSchema>;
   let dispatcher: ActionsSubject;
 
-  function setup(initialState: any = { counter1: 0, counter2: 1 }) {
+  function setup(
+    initialState: any = { counter1: 0, counter2: 1 },
+    metaReducers: any = []
+  ) {
     const reducers = {
       counter1: counterReducer,
       counter2: counterReducer,
@@ -41,7 +48,7 @@ describe('ngRx Store', () => {
     };
 
     TestBed.configureTestingModule({
-      imports: [StoreModule.forRoot(reducers, { initialState })],
+      imports: [StoreModule.forRoot(reducers, { initialState, metaReducers })],
     });
 
     store = TestBed.get(Store);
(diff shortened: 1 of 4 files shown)


itayod:
please let me know if you like the solution...
if so I will add some test and update the pull request.

A code example of how to use this feature:

```
import {NgModule, InjectionToken} from '@angular/core';
import {StoreModule} from '@ngrx/store';

export const FEATURE_META_REDUCER_TOKEN = new InjectionToken('Feature Meta Reducers');

export function getMetaReducers() {
  // array of meta reducers
  return [function(reducer) {
    return function (state, action) {
      return reducer(state, action);
    }
  }];
}

@NgModule({
  imports: [
    StoreModule.forFeature('feature', {}, {metaReducers: <any>FEATURE_META_REDUCER_TOKEN}),
  ],
  providers: [
    {
      provide: FEATURE_META_REDUCER_TOKEN,
      useFactory: getMetaReducers,
    },
  ],
})
export class FeatureModule { }

```

modules/store/src/store_module.ts
```diff
@@ -165,7 +166,7 @@ export class StoreModule {
       ngModule: StoreFeatureModule,
       providers: [
         {
-          provide: STORE_FEATURES,
+          provide: _FEATURE_STORE,
```
timdeschryver: Revert:

```suggestion
          provide: STORE_FEATURES,
```

modules/store/src/store_module.ts
```diff
@@ -176,6 +177,12 @@ export class StoreModule {
             initialState: config.initialState,
           },
         },
+        {
+          provide: STORE_FEATURES,
```
timdeschryver:
```suggestion
          provide: _STORE_FEATURES,
```

and change ` @Inject(STORE_FEATURES) private features: StoreFeature<any, any>[],` to ` @Inject(_STORE_FEATURES) private features: StoreFeature<any, any>[],` in the `StoreFeatureModule` constructor.

modules/store/src/store_module.ts
```diff
@@ -176,6 +177,12 @@ export class StoreModule {
...
+          multi: false,
+          deps: [Injector, _FEATURE_STORE],
```
timdeschryver:
```suggestion
          deps: [Injector, STORE_FEATURES],
```

modules/store/src/tokens.ts
```diff
@@ -24,6 +24,11 @@ export const _STORE_REDUCERS = new InjectionToken(
 export const _FEATURE_REDUCERS = new InjectionToken(
   '@ngrx/store Internal Feature Reducers'
 );
+
+export const _FEATURE_STORE = new InjectionToken(
```
timdeschryver:
```suggestion
export const _STORE_FEATURES = new InjectionToken(
```

modules/store/src/tokens.ts
```diff
@@ -24,6 +24,11 @@ export const _STORE_REDUCERS = new InjectionToken(
...
+  '@ngrx/store Internal Feature Store'
```
timdeschryver:
```suggestion
  '@ngrx/store Internal Store Features'
```

modules/store/src/tokens.ts
timdeschryver: Add the token to the exports in the `src/index.ts` file

modules/store/src/store_module.ts
itayod: Are you sure about that?
I think it will be more appropriate if the `StoreFeatureModule` will receive the `STORE_FEAUTRES` token  as it already receives the`FEATURE_REDUCERS` and not the `_FEATURE_REDUCERS.`.

Please let me know what are thinking.

timdeschryver: The reasoning behind is because users could use `STORE_FEATURES` to add features, while `_STORE_FEATURES` is meant for internal use only. That being said you make a valid point, all of the injected items are public.

itayod: I see your point, I will update the pull request.

brandonroberts:
@timdeschryver I think if we're going to allow an injection token, it should at the configuration level and not the meta-reducer level. It would prevent us from having an injection token for each thing you want to configure using DI.

itayod:
> @timdeschryver I think if we're going to allow an injection token, it should at the configuration level and not the meta-reducer level. It would prevent us from having an injection token for each thing you want to configure using DI.

Having that said,
My solution allow users to define their own injection token for meta reducers,
It is also be easy to extend it for the rest of the configuration fields (e.g `initialState`) by adding it to the 
`_createFeatureStore` function.

But of course it is for you to decide rather you like the idea or not.

brandonroberts:
Yes, something like that. You wouldn't spread the config if its an InjectionToken though since it contains the entire feature object. We also need some tests to verify this correctly merges configs from different NgModules, including lazy loaded ones.

itayod:
so the api will look something like that?

```javascript
    export const MY_FEATURE_CONFIG_TOKEN = new InjectionToken('Feature Config');
    ...

    StoreModule.forFeature('feature', {}, MY_FEATURE_CONFIG_TOKEN),
```

brandonroberts:
Yes

itayod:
I have made some changes in another commit (just in case...)
@brandonroberts is that solution is what you meant?

brandonroberts:
Its getting there. I'd prefer to keep `STORE_FEATURES` as a multi-provider though. Tim already mentioned this, but I agree that using `_STORE_FEATURES` internally to resolve the feature configs is a better option.

itayod:
I see, the reason I changed `STORE_FEATURES` to not be a multi provider is because this is the only solution I thought about any better solution in terms of support injection token for config.

I think that if we must support multi provider for `STORE_FEATURES` there are 2 possible solutions,
1 is to change `StoreFeatureModule` constructor (which I am not sure if we want to...).
2 is to create a token for each of the `StoreFeature` feilds and provide them by use existing, which by the way will also allow users to create an injection token for a specific field (e.g the initial state, meta reducer, store key...) .

@brandonroberts @timdeschryver please let me know what do think, or if you have some other solutions, I will be happy to implement them :)

timdeschryver:
I think the implementation is good, just a few details:

- based on https://github.com/ngrx/platform/pull/1445#issuecomment-442831153, I think we want to "force" the user to provide all properties instead of a partial config. So the implementation would look something as:

```ts
export function _createFeatureStore(injector: Injector, features: Array<any>) {
  return features.map(feat => {

    if(feat.config instanceof InjectionToken) {
      const config = injector.get(feat.config);
      return {
        key: feat.key,
        reducerFactory: config.reducerFactory,
        metaReducers: config.metaReducers,
        initialState: config.initialState
      }
    }

    return {
      key: feat.key,
      reducerFactory: feat.config.reducerFactory
        ? feat.config.reducerFactory
        : combineReducers,
      metaReducers: feat.config.metaReducers ? feat.config.metaReducers : [],
      initialState: feat.config.initialState,
    };
  });
}
```

- I think we should keep supporting the multi-provider for `STORE_FEATURES`. If it stings to inject an internal token in the constructor, we could also make this new token public and rename it to `STORE_FEATURES_COLLECTION` (or `STORE_FEATURES_CONFIG`) ?

brandonroberts:
There are a couple of options I see that we can do here.

1. Allow you to provide an InjectionToken for the StoreConfig, and fetch the config if needed in the constructor of the `StoreFeatureModule` where we are already mapping over feature.

2. Don't change anything and document using the `ReducerManager` to add features dynamically if you need dependency injection. We should do this anyway, but it may be a less risky solution.

itayod:
@brandonroberts I see, the first option could work, I will try implement it and make another commit.

brandonroberts:
@itayod is this ready? Will you rebase on master?

itayod:
almost, just wanted to make few code improvements that @timdeschryver advised me and I think we are good to go :)

itayod:
@brandonroberts @timdeschryver have you seen my update? think we can close that now?

modules/store/src/tokens.ts
```diff
@@ -24,6 +24,11 @@ export const _STORE_REDUCERS = new InjectionToken(
 export const _FEATURE_REDUCERS = new InjectionToken(
   '@ngrx/store Internal Feature Reducers'
 );
+
+export const FEATURE_NAME = new InjectionToken(
+  '@ngrx/store Internal Feature name'
```
timdeschryver: This shouldn't be internal?

modules/store/src/store_module.ts
```diff
@@ -58,19 +59,28 @@ export class StoreFeatureModule implements OnDestroy {
   constructor(
     @Inject(STORE_FEATURES) private features: StoreFeature<any, any>[],
     @Inject(FEATURE_REDUCERS) private featureReducers: ActionReducerMap<any>[],
+    @Inject(FEATURE_NAME) private featureName: string,
```
timdeschryver: `@Inject(FEATURE_NAME) private featureNames: string,`

modules/store/spec/store.spec.ts
```diff
@@ -471,4 +476,174 @@ describe('ngRx Store', () => {
...
+
+        return {
+          metaReducer1: metaReducer1,
+          metaReducer2: metaReducer2,
+        };
+      })();
+    });
+
+    it('should create a meta reducer for root and call it through', () => {
+      const metaReducerSpy: any = spyOn(
```
timdeschryver: Type this and the others to `: Spy`

modules/store/spec/store.spec.ts
```diff
@@ -471,4 +476,174 @@ describe('ngRx Store', () => {
...
+        metaReducerContainer,
+        'metaReducer1'
+      ).and.callThrough();
+      setup({ counter1: 0, counter2: 1 }, [metaReducerContainer.metaReducer1]);
```
timdeschryver: Since we don't do anything with the counter values, I think it will be clearer to use 

```ts
setup({ }, [metaReducerContainer.metaReducer1]);
```

modules/store/spec/store.spec.ts
```diff
@@ -471,4 +476,174 @@ describe('ngRx Store', () => {
...
+      const metaReducerSpy2: any = spyOn(
+        metaReducerContainer,
+        'metaReducer2'
+      ).and.callThrough();
+
+      TestBed.configureTestingModule({
+        imports: [
+          StoreModule.forRoot({}),
+          StoreModule.forFeature('counter1', counterReducer, {
+            metaReducers: [metaReducerContainer.metaReducer2],
```
timdeschryver: Wire up the "right" meta-reducer to prevent confusion:

```ts
metaReducers: [metaReducerContainer.metaReducer1],
```

modules/store/spec/store.spec.ts
```diff
@@ -471,4 +476,174 @@ describe('ngRx Store', () => {
...
+          }),
+          StoreModule.forFeature('counter2', counterReducer, {
+            metaReducers: [metaReducerContainer.metaReducer1],
```
timdeschryver: Wire up the "right" meta-reducer to prevent confusion:

```ts
metaReducers: [metaReducerContainer.metaReducer2],
```

modules/store/spec/store.spec.ts
```diff
@@ -471,4 +476,174 @@ describe('ngRx Store', () => {
...
+          }),
+        ],
+      });
+      const mockStore = TestBed.get(Store);
+      const action = { type: INCREMENT };
+      mockStore.dispatch(action);
+
+      expect(metaReducerSpy1).toHaveBeenCalledWith(counterReducer);
+      expect(metaReducerSpy2).toHaveBeenCalledWith(counterReducer);
```
timdeschryver: What about creating another `counterReducer` and check if the right meta-reducer is called with the correct reducer?

modules/store/spec/store.spec.ts
```diff
@@ -471,4 +476,174 @@ describe('ngRx Store', () => {
...
+    });
+  });
+
+  describe('Feature config token', () => {
+    it('should initial state with value', (done: DoneFn) => {
+      const FEATURE_CONFIG_TOKEN = new InjectionToken('Feature Config');
+      const initVal = { counter1: 1 };
```
timdeschryver: Rename to `initialState`

modules/store/spec/store.spec.ts
```diff
@@ -471,4 +476,174 @@ describe('ngRx Store', () => {
...
+        error: done,
+        complete: done,
+      });
+    });
+  });
+
+  it('should initial state with value for multi features', (done: DoneFn) => {
+    const FEATURE_CONFIG_TOKEN = new InjectionToken('Feature Config');
+    const FEATURE_CONFIG2_TOKEN = new InjectionToken('Feature Config2');
+    const initVal = 1;
```
timdeschryver: Rename to `initialState`, `initialState2`, `initialState3`

modules/store/spec/store.spec.ts
```diff
@@ -471,4 +476,174 @@ describe('ngRx Store', () => {
...
+        next(val: any) {
+          expect(val[featureKey]).toEqual(initVal);
+        },
+        error: done,
+        complete: done,
+      });
+    });
+  });
+
+  it('should initial state with value for multi features', (done: DoneFn) => {
```
timdeschryver: Move this to inside the `Feature config token` block

modules/store/spec/store.spec.ts
timdeschryver: Could we also move this creation of these spies to inside the `beforeEach` since we need them in every test

modules/store/src/store_module.ts
brandonroberts: This should be a `string[]` because its a multi-provider.

modules/store/src/store_module.ts
```diff
@@ -159,22 +169,29 @@ export class StoreModule {
...
-            reducerFactory: config.reducerFactory
-              ? config.reducerFactory
-              : combineReducers,
-            metaReducers: config.metaReducers ? config.metaReducers : [],
-            initialState: config.initialState,
-          },
+          useValue:
+            config instanceof InjectionToken
+              ? config
+              : <StoreFeature<any, any>>{
```
brandonroberts: The `StoreFeature` has `key` property also, but its not included here.

modules/store/src/tokens.ts
brandonroberts: Why does this need to be separate if the `key` is included in the `StoreFeature<T, V>`?

modules/store/src/store_module.ts
```diff
@@ -159,22 +169,29 @@ export class StoreModule {
...
       | ActionReducer<any, any>
       | InjectionToken<ActionReducer<any, any>>,
-    config: StoreConfig<any, any> = {}
+    config: StoreConfig<any, any> | InjectionToken<StoreConfig<any, any>> = {}
   ): ModuleWithProviders<StoreFeatureModule> {
     return {
       ngModule: StoreFeatureModule,
       providers: [
+        {
+          provide: FEATURE_NAME,
```
brandonroberts: Its easier to override `STORE_FEATURES` as a single token instead of having to override `STORE_FEATURES` and `FEATURE_NAME`

itayod:
I have fixed the tests,
about the implementation, I am thinking about going back to my previous commit...
thoughts?

itayod:
> I have fixed the tests,
> about the implementation, I am thinking about going back to my previous commit...
> thoughts?

I created another commit with the previous solution...

modules/store/src/store_module.ts
```diff
@@ -159,23 +161,39 @@ export class StoreModule {
...
+              !(config instanceof InjectionToken) && config.metaReducers
+                ? config.metaReducers
+                : [],
+            initialState:
+              !(config instanceof InjectionToken) && config.initialState,
           },
         },
+        {
+          provide: _STORE_FEATURES,
+          multi: false,
```
brandonroberts: `multi: false` is the default, so this isn't necessary.

modules/store/src/index.ts
```diff
@@ -43,12 +43,15 @@ export {
   _FEATURE_REDUCERS,
   FEATURE_REDUCERS,
   _FEATURE_REDUCERS_TOKEN,
+  _STORE_FEATURES,
```
brandonroberts: Don't add these to the public API.

modules/store/src/store_module.ts
```diff
@@ -159,23 +161,39 @@ export class StoreModule {
...
+            reducerFactory:
+              !(config instanceof InjectionToken) && config.reducerFactory
+                ? config.reducerFactory
+                : combineReducers,
+            metaReducers:
+              !(config instanceof InjectionToken) && config.metaReducers
+                ? config.metaReducers
+                : [],
+            initialState:
+              !(config instanceof InjectionToken) && config.initialState,
```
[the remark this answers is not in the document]
itayod: good point, what about:  ``` !(config instanceof InjectionToken) && config.initialState ? config.initialState : {} ```?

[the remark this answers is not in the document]
brandonroberts: Set it as undefined. If its an injection token, it will be overridden by the injected config value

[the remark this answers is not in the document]
itayod: Ok, I will make another commit right away

itayod:
> Couple more changes and we should be good to go.

done :)

brandonroberts:
@timdeschryver you have anymore feedback on this?

brandonroberts:
The `reducerFactory` should default to `combineReducers` if not provided when using an injection token. The `initialState` issue is a bug that's already been filed I believe.

timdeschryver:
👍 
The problem is the the `metaReducer` is called first without an initial state, resulting in the "default initial state". Thus the provided initial state is ignored.

I'll take a look at the issues, if it isn't filed - I will create a new issue.
But that's something for next year.

In advance, a happy new year 🎉 !

itayod:
@brandonroberts 
ok, but I thought about using the metaReducer suit to test this bug fix...
should I wait for this pr to be merged? or should I create a separate suit?

brandonroberts:
🎉

Issue #1447, opened 2018-11-27
UpdateEffects during prod build not behaving as intended 


timdeschryver:
## Minimal reproduction of the bug/regression with instructions:

The following snippet works while working in dev mode, but fails in prod mode.
This is because constructor names are mangled by the CLI during a build, thus the check `action.effects.includes('SomeEffectsClass')` would always fail.

```ts
@Effect()
init = this.actions.pipe(
  ofType(UPDATE_EFFECTS),
  filter(action => action.effects.includes('SomeEffectsClass')),
  map(action => ...)
);
```

## Expected behavior:

In order to get this working we should dispatch the `UPDATE_EFFECTS` action with an array of effect instances instead of the constructor names. This could be compared to the Angular router events. The above code snippet would become:

```ts
@Effect()
init = this.actions.pipe(
  ofType(UPDATE_EFFECTS),
  filter(action => action.effects.some(x => x instanceof SomeEffectsClass)),
  map(action => ...)
);
```

## Versions of NgRx, Angular, Node, affected browser(s) and operating system(s):
NgRx 7.0.0-beta.0

## I would be willing to submit a PR to fix this issue

[x] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

timdeschryver:
The downside of this would be that the action payload would be an effect resulting in a big payload. E.g. the `AuthEffects` from the example app would have a 125K lines payload.

timdeschryver:
Another approach would be to use something as mentioned in https://github.com/ngrx/platform/issues/683#issuecomment-370300048:

```
EffectsModule.forFeature([
  Effect1,
  Effect2
], {
  initAction: { type: 'EffectsRegistedAction' }
})
```

timdeschryver:
Or adding a `initAction` property in the effect self, which will be used to dispatch the init action:

```
@Injectable()
export class AuthEffects {
  initAction = 'MY_INIT_ACTION';
  
  ...
}
```

brandonroberts:
We're going into lifecycle madness with all these hooks 😂 . Are there other options are there to reasonably getting the Effects class name in prod? I'm hesitant to introduce another stream for "Effects Events" as init would be the only one.

PR #1448, opened 2018-11-27
feat(effects): add OnIdentifyEffects interface to register multiple effect instances

### LINKED ISSUE #1246: Effects: Instantiating Effects multiple times, only first instance is run

### LINKED ISSUE #1443: Effects: Issue with effects instantiating multiple times creates multiple same handlers

https://stackblitz.com/edit/ngrx-seed-ljumtv

## Expected behavior:

Not sure what is considered expected behavior in this case, but I think this is a breaking change. The issue is caused in 7.0.0-beta.0 after the addition of a new feature that allows instantiating the same effect class multiple times.

It should be change from this PR: https://github.com/ngrx/platform/pull/1249

The issue occurs when you have a shared module that will group some store logic, specifically effects.

In versions before 7.0.0-beta.0, if you import that shared store to 2 different lazy loaded modules, only one effect is instantiated thus when action is dispatched we only handle that action once.

In 7.0.0-beta.0 after the addition of that new feature that allows instantiating the same effect it will instantiate both of those effects and thus we will have 2 same handlers.

We are just now upgrading our enterprise application, its using a monorepo approach and has a lot of those shared stores as libraries for manipulating different collections that are shared across features.

On top of that, we utilize lazy loading of each feature, and those are also separated into libraries. Since most of those libraries are plug and play, they hide their implementation details, so when feature depends on shared store logic it just imports it, it does not expect that it is already provided. 

That way:
1) they can be bundled together if nothing else depends on the same logic.
2) App module does not need to know on what does that feature depend. 

Before the upgrade, this worked without a problem because effects were created once, similarly how reducer is being created if using the same principle.

After the upgrade if you start going through features it will instantiate the same effect multiple times, so for example, if you want to update/delete an entity, it will trigger the call for each handler that was created.

A possible solution is instead of importing those shared store modules into features, we could just add them to the app module, so they are created once, but that way it creates a leaky abstraction, since app module needs to know on which store part certain feature depends. Also it will be bundled together with the main bundle and those might not be necessarily needed at that point in time.

ngrx/effects: 7.0.0-beta.0
Not affected by angular, node or browser

[ ] No

timdeschryver:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1443

## What is the new behavior?

This change reverts #1249 in order to only instantiate effects once (e.g. when imported in different features). 
This PR adds a way to identify effects while using the `addEffects` method in order to resolve #1246.

```
[ ] Yes
[x] No
```


timdeschryver:
I think that should also work, something like this right?

```ts
export class FooEffects implements EffectIdentifier {
  identifier = 'Foo';
}
```

brandonroberts:
Yep

dummdidumm:
I think different effect classes should be able to get instantiated even if they have the same identifier.
So if class EffectsA has identifier "abc" and EffectsB also has identifier "abc", they should both get instantiated. With the current implementation this is not the case I think.

dummdidumm:
I also think that getting the identifier from the class instance is not a good idea. People now have to know that "identifier" is a reserved property that the effects library uses. If someone does not implement the interface but still uses a variable called identifier for something else, strange things could happen.

timdeschryver:
@dummdidumm I agree with:

> I think different effect classes should be able to get instantiated even if they have the same identifier.
So if class EffectsA has identifier "abc" and EffectsB also has identifier "abc", they should both get instantiated. With the current implementation this is not the case I think.

I think `identifier` is perhaps a bit too general, `ngrxIdentifier` or `effectIdentifier` might be better.

timdeschryver:
@dummdidumm Why should it be `a,a,b,b`?

dummdidumm:
SourceWithIdentifier is instantiated with identifier, "a" two times, SourceWithIdentifier 2 is instantiated with identifier "b" two times . It should be "a","b" for each effect class.

modules/effects/src/lifecycle_hooks.ts
```diff
@@ -0,0 +1,6 @@
+export interface EffectIdentifier {
```
brandonroberts: Need some more doc here for the API page 😉 

```
/**
 * @description 
 * Interface to set an identifier for effect instances.
 * 
 * By default, each Effects class will only be registered
 * once regardless of how many times the Effect class
 * is loaded. By implementing this interface, you define
 * a unique identifier to register an Effects class instance
 * multiple times.
 * 
 * @usageNotes
 * 
 * ### Set an identifier for an Effects class
 * 
 * ```
 * Example code here
 * ```
 */
export interface EffectIdentifier {
  /**
   * @description
   * String identifier to differentiate effect instances.
   */
  effectIdentifier: string;
}
```

dummdidumm:
I still think that setting the identifier on the class is not the best possible solution.
- people could use a property of the same name in a different context without knowing and get strange results
- angular's dependency injection needs to know where the constructor property comes from. So you cannot provide that effect class to the effectsmodule through forFeature/forRoot without workarounds
- conceptually it is not the responsibility of the class itself to know "who it is"

What are your arguments against going with the initial proposal ?

timdeschryver:
I don't have a favorite way of solving this (the first proposal vs the interface).

You're right with your second point @dummdidumm , but I'm not quite agreeing with your first point. Angular does more or less do the same with their lifecycle hooks, plus it opens up opportunities to introduce more of these in NgRx e.g. in #1447.

timdeschryver:
I forgot about `ngrxOnRunEffects` for a second 😅 .

To stay consistent I would want to rename the property `effectIdentifier` to `ngrxEffectIdentifier` and also move the `OnRunEffects` interface to the `lifecycle_hooks`, would that be OK @brandonroberts ?

brandonroberts:
@timdeschryver that's fine with me because it doesn't change the public API. I'd rather see this implemented as a method if we're being similar to `ngrxOnRunEffects` and have it be `ngrxOnIdentify` which is a little more descriptive.

timdeschryver:
👍 I'll implement it the way you suggested. 
Thanks for the feedback @dummdidumm and @brandonroberts !

timdeschryver:
What about `ngrxOnIdentifyEffects` (sorry, I know that I'm annoying 😅 )?

brandonroberts:
😂 

That's fine. You're identifying these set of effects 👍

brandonroberts:
@timdeschryver one more minor nit. Will you fix the first commit message with `OnIdentifyEffects`. It currently says `OnIdentityEffects`. I'm going to land the commits as-is.

PR #1452, opened 2018-12-02
docs: fixed usage of selector with .select()

luchsamapparat:
## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

### DIFF
--- a/projects/ngrx.io/content/guide/store/selectors.md
+++ b/projects/ngrx.io/content/guide/store/selectors.md
@@ -278,7 +278,7 @@ import { map, filter } from 'rxjs/operators';
 
 store
   .pipe(
-    select(selectValues(state)),
+    select(selectValues),
     filter(val => val !== undefined)
   )
   .subscribe(/* .. */);


luchsamapparat:
My change would be equivalent to `select(state => selectValues(state)),`, but what's currently in the docs (`select(selectValues(state)),`) is just wrong. Where would the `state` variable even come from?

Please have another look.

brandonroberts:
@luchsamapparat state is mapped to the selector for you. See

https://github.com/ngrx/platform/blob/master/modules/store/src/store.ts#L199

with usage

https://github.com/ngrx/platform/blob/master/projects/example-app/src/app/books/containers/collection-page.component.ts.

luchsamapparat:
The example you posted does not pass `state` to the selector:

`store.pipe(select(fromBooks.getBookCollection));`

If it were equivalent to what I'm trying to fix here, it would be

`store.pipe(select(fromBooks.getBookCollection(state)));`

Your example proves my point that it's wrong what's in the docs

brandonroberts:
Thanks!

PR #1305, opened 2018-08-28
feat(Effects): dispatch init feature effects action on init

timdeschryver:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #683

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

## Other information
If we agree on the API, I'll also add these changes to the docs.
I also would want to change the root effects action name (this will be a breaking change) to make it more clear that it's initiated by the root effects: from `@ngrx/effects/init` to `@ngrx/root-effects/init` .

Closes #683

PR #1467, opened 2018-12-13
feat(Effects): add OnInitEffects interface to dispatch an action on initialization

### LINKED ISSUE #1447: UpdateEffects during prod build not behaving as intended

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

As mentioned in the issue, the solution we have now to dispatch an `INIT` effect action doesn't work in prod builds because the constructor name is mangled.

Closes #1447

## What is the new behavior?

This PR adds a new lifecycle hook `OnInitEffects` that dispatches a user-defined action on effect registration. I couldn't find another way to access the effect's name.

Personally, I also find this solution to be cleaner and more flexible than the previous one.

Right now I've made it so that only one action can be dispatched, but this can also be an array of actions if it's needed. Imho one action is sufficient and also follows our best practices to treat an action as a unique event.

```
[ ] Yes
[ ] No
```

## Other information

The first commit reverts #1305 
The second commit adds the `OnInitEffects` hook. In this commit, I also changed the `OnIdentifyEffects` signature to be consistent with the other signatures.

### DIFF
--- a/docs/effects/api.md
+++ b/docs/effects/api.md
@@ -50,31 +50,6 @@ Usage:
 export class FeatureModule {}
 ```
 
-### UPDATE_EFFECTS
-
-After feature effects are registered, an `UPDATE_EFFECTS` action is dispatched.
-
-```ts
-type UpdateEffects = {
-  type: typeof UPDATE_EFFECTS;
-  effects: string[];
-};
-```
-
-For example, when you register your feature module as `EffectsModule.forFeature([SomeEffectsClass, AnotherEffectsClass])`,
-it has `SomeEffectsClass` and `AnotherEffectsClass` in an array as its payload.
-
-To dispatch an action when the `SomeEffectsClass` effect has been registered, listen to the `UPDATE_EFFECTS` action and use the `effects` payload to filter out non-important effects.
-
-```ts
-@Effect()
-init = this.actions.pipe(
-  ofType<UpdateEffects>(UPDATE_EFFECTS)
-  filter(action => action.effects.includes('SomeEffectsClass')),
-  map(action => ...)
-);
-```
-
 ## Actions
 
 Stream of all actions dispatched in your application including actions dispatched by effect streams.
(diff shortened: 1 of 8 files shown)


timdeschryver:
Fyi, it seems like the failing test command is "because" of yarn - https://github.com/yarnpkg/yarn/issues/6746 and will be fixed in the next release.

brandonroberts:
@timdeschryver instead of introducing a new hook, how about adding a configuration when registering effects to give the set a `name`? It could be the same as the class name, but we would use the array of names with the `UpdateEffects` action.

timdeschryver:
I think the downsides of using that approach are:

- With only a name, a user couldn't define a payload, if this is needed
- It won't be possible to trigger an `init` while manually adding effects, but is this even needed?

The better part would be, that it would be consistent with the config from `ngrx/store`.

brandonroberts:
Yea, I thought about manually adding effects but I think that's an advanced use case. I'd rather provide the name and use a consistent Update action then have to come up with a different one for each effect.

brandonroberts:
Scratch that last comment. I wasn't thinking about how the effects are registered with more than one class. I'm good with the init function.

PR #1498, opened 2019-01-04
fix(store): call metareducer with the user's config initial state

### LINKED ISSUE #1464: Ngrx/entity and metaReducers with localStorageSync
Hi Everyone,

I just would like to ask that it is possible to use ngrx/entity with localStorageSync?

My experience is that everything is working with normal states but with entity states no.

This is an example how I would like to register my entity state into local storage.
```
export function userPreferencesLocalStorageSyncReducer(reducer: ActionReducer<any>): ActionReducer<any> {
  return localStorageSync({
    keys: [USER_PREFERENCES_STORE_KEY],
    rehydrate: true,
  })(reducer);
}

@NgModule({
  imports: [StoreModule.forFeature(USER_PREFERENCES_STORE_KEY, reducer, { initialState, metaReducers })],
})
export class UserPreferencesStoreModule {}
```

Thanks for any advice

itayod:
## PR Checklist

## PR Type

```
[ x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`metaReducers` for feature are called without the user's config initial state.

Closes #1464

```
[ ] Yes
[ x] No
```

### DIFF
--- a/modules/store/spec/store.spec.ts
+++ b/modules/store/spec/store.spec.ts
@@ -14,6 +14,7 @@ import {
   Action,
 } from '../';
 import { StoreConfig } from '../src/store_module';
+import { combineReducers } from '../src/utils';
 import {
   counterReducer,
   INCREMENT,
@@ -373,11 +374,11 @@ describe('ngRx Store', () => {
     });
 
     it('should dispatch an update reducers action when a feature is added', () => {
-      reducerManager.addFeature(
-        createFeature({
-          key: 'feature1',
-        })
-      );
+      reducerManager.addFeature({
+        key: 'feature1',
+        reducers: {},
+        reducerFactory: <any>combineReducers,
+      });
 
       expect(reducerManagerDispatcherSpy).toHaveBeenCalledWith({
         type: UPDATE,
@@ -387,12 +388,16 @@ describe('ngRx Store', () => {
 
     it('should dispatch an update reducers action when multiple features are added', () => {
       reducerManager.addFeatures([
-        createFeature({
+        {
           key: 'feature1',
-        }),
-        createFeature({
+          reducers: {},
+          reducerFactory: <any>combineReducers,
+        },
+        {
           key: 'feature2',
-        }),
+          reducers: {},
+          reducerFactory: <any>combineReducers,
+        },
       ]);
 
       expect(reducerManagerDispatcherSpy).toHaveBeenCalledTimes(1);
@@ -546,12 +551,41 @@ describe('ngRx Store', () => {
           }),
         ],
       });
+
       const mockStore = TestBed.get(Store);
       const action = { type: INCREMENT };
+
       mockStore.dispatch(action);
 
       expect(metaReducerSpy1).toHaveBeenCalledWith(counterReducer);
       expect(metaReducerSpy2).toHaveBeenCalledWith(counterReducer2);
+    });
+
+    it('should initial state with value', (done: DoneFn) => {
+      const counterInitialState = 2;
+      TestBed.configureTestingModule({
+        imports: [
+          StoreModule.forRoot({}),
+          StoreModule.forFeature(
+            'counterState',
+            { counter: counterReducer },
+            {
+              initialState: { counter: counterInitialState },
+              metaReducers: [metaReducerContainer.metaReducer1],
+            }
+          ),
+        ],
+      });
+
+      const mockStore = TestBed.get(Store);
+
+      mockStore.pipe(take(1)).subscribe({
+        next(val: any) {
+          expect(val['counterState'].counter).toEqual(counterInitialState);
+        },
+        error: done,
+        complete: done,
+      });
     });
   });
 
--- a/modules/store/src/utils.ts
+++ b/modules/store/src/utils.ts
@@ -4,6 +4,7 @@ import {
   ActionReducerFactory,
   ActionReducerMap,
   MetaReducer,
+  InitialState,
 } from './models';
 
 export function combineReducers<T, V extends Action = Action>(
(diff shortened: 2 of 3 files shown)


modules/store/spec/store.spec.ts
```diff
@@ -534,24 +534,44 @@ describe('ngRx Store', () => {
       expect(metaReducerSpy2).toHaveBeenCalled();
     });
 
-    it('should create a meta reducer for feature and call it with the expected reducer', () => {
+    it('should create a meta reducer for feature and call it with the expected reducer', done => {
```
brandonroberts:
```suggestion
    it('should create a meta reducer for feature and call it with the expected reducer', (done: any) => {
```

modules/store/src/utils.ts
```diff
@@ -107,6 +115,7 @@ export function createFeatureReducerFactory<T, V extends Action = Action>(
       : (r: ActionReducer<T, V>) => r;
 
   return (reducer: ActionReducer<T, V>, initialState?: T) => {
+    // I think this is also bug, shouldn't it be: reducerFactory(reducer, initialState)?
```
itayod: This is quite rare... 
It should affect only users that use `reducerFactory` for feature and pass reducer as function instead of dict.

timdeschryver: No it is not.
This is a reducer factory, it accepts one reducer as argument.
We wrap this reducer with our reducer and simply pass the arguments received, with the extra check on the state, to the "real" reducer.

itayod: Ok I see, you'r right...
I thought this was a reducerFactory that possibly comes from the user,
I will remove the comment :)

modules/store/src/utils.ts
```diff
@@ -92,10 +93,17 @@ export function createReducerFactory<T, V extends Action = Action>(
...
+    reducerFactory = compose.apply(null, [...metaReducers, reducerFactory]);
   }
 
-  return reducerFactory;
+  return (rm: ActionReducerMap<T, V>, initialState?: InitialState<T>) => {
+    const reducer = reducerFactory(rm, initialState);
+
+    return (state: T | undefined, action: V) => {
+      state = state === undefined ? (initialState as T) : state;
+      return typeof reducer === 'function' ? reducer(state, action) : state;
```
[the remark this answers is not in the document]
itayod: @timdeschryver 
I think it's a valid point, but keep in mind that the `reducerFactory` can be provided by the user.
anyway let me know what you have decided :)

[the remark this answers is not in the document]
timdeschryver: Let's leave it like this.

itayod:
@brandonroberts @timdeschryver 
do you think we can close it?

modules/store/src/utils.ts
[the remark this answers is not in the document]
timdeschryver: Sorry, I was to fast... leaving it as is, is the safer option, but I personally would want to remove it. Redux is more or less just the `combineReducers` function, which should return a reducer function. Since this is a hot code path (called multiple times with each action dispatch), I believe we should remove the check. If there are users providing their own factory function and are experiencing issues we can put in the check.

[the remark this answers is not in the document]
itayod: Ok, so I will remove the check and update the failing test :)

modules/store/spec/store.spec.ts
```diff
@@ -534,24 +534,44 @@ describe('ngRx Store', () => {
...
+
       const mockStore = TestBed.get(Store);
       const action = { type: INCREMENT };
+
       mockStore.dispatch(action);
 
       expect(metaReducerSpy1).toHaveBeenCalledWith(counterReducer);
-      expect(metaReducerSpy2).toHaveBeenCalledWith(counterReducer2);
+      expect(metaReducerSpy2).toHaveBeenCalled();
+      expect(counterReducer2Spy).toHaveBeenCalledWith(2, action);
```
timdeschryver: We should let this test as it was, to check if the right meta reducer was called with the reducer.
We should create a new test to verify the initial state.

itayod: Ok, should the test should be as part of the metareducer suite?

modules/store/spec/store.spec.ts
```diff
@@ -373,11 +374,11 @@ describe('ngRx Store', () => {
...
     it('should dispatch an update reducers action when a feature is added', () => {
-      reducerManager.addFeature(
-        createFeature({
-          key: 'feature1',
-        })
-      );
+      reducerManager.addFeature({
+        key: 'feature1',
+        reducers: {},
+        reducerFactory: <any>combineReducers,
```
itayod: I had a type miss match here with combineReducer which Is kind of weird,
should I try to fix that?

timdeschryver: I think it should be part of this PR.
Is it causing errors? If so, you could use a jasmine spy instead.

itayod: I also think it shouldn't be part of this PR,
it will cause only compile error, meaning the test are still passing.
and I think jasmine spy is redundant here because this test isn't about the reducer factory here.

modules/store/src/utils.ts
```diff
@@ -92,10 +93,16 @@ export function createReducerFactory<T, V extends Action = Action>(
   metaReducers?: MetaReducer<T, V>[]
 ): ActionReducerFactory<T, V> {
   if (Array.isArray(metaReducers) && metaReducers.length > 0) {
-    return compose.apply(null, [...metaReducers, reducerFactory]);
+    reducerFactory = compose.apply(null, [...metaReducers, reducerFactory]);
   }
 
-  return reducerFactory;
+  return (reducers: ActionReducerMap<T, V>, initialState?: InitialState<T>) => {
+    const reducer = reducerFactory(reducers, initialState);
```
timdeschryver: Passing the `initialState` to the factory brings some confusion for me 😅 

We can remove it because we're already providing an initial state via the following code, making it obsolete.
```
state = state === undefined ? (initialState as T) : state;
```

Besides this, it LGTM!


PS: Sorry for doing the review partially.

itayod: Thanks!
I see that, you'r right.
It just breaks one of the tests... so I will remove that and update the failing test :)

PR #1506, opened 2019-01-07
docs: update example app

timdeschryver:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[x] Documentation content changes
[ ] Other... Please describe:
```

Would it be OK to update the example app deployed on GitHub?

Things I've changed to make it imho easier for users taking a look:
- enable actions in the devtools 
- also enable login with `ngrx`, which is also prefilled in the username textbox

We could also make these changes on the master branch?

I didn't update the StackBlitz example, because it can't handle the aliased paths defined in `tsconfig.app.json`.

Related: this is in order to update the NgRx documentation for the Redux Dev Tools extension, which were based on NgRx version 2 - https://github.com/zalmoxisus/redux-devtools-extension/pull/623

PR #1508, opened 2019-01-09
feat(Example): make the example app more user friendly

timdeschryver:
## PR Checklist

## PR Type

- allow username 'ngrx' to log in
- prefill username with ngrx
- enable devtools

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

## Other information

See https://github.com/ngrx/platform/pull/1506

### DIFF
--- a/projects/example-app/src/app/app.module.ts
+++ b/projects/example-app/src/app/app.module.ts
@@ -17,7 +17,6 @@ import { reducers, metaReducers } from '@example-app/reducers';
 import { schema } from '@example-app/db';
 
 import { AppComponent } from '@example-app/core/containers/app.component';
-import { environment } from '../environments/environment';
 import { AppRoutingModule } from '@example-app/app-routing.module';
 
 @NgModule({
@@ -55,7 +54,9 @@ import { AppRoutingModule } from '@example-app/app-routing.module';
      */
     StoreDevtoolsModule.instrument({
       name: 'NgRx Book Store App',
-      logOnly: environment.production,
+
+      // In a production build you would want to disable the Store Devtools
+      // logOnly: environment.production,
     }),
 
     /**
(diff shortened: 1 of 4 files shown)


projects/example-app/src/app/app.module.ts
```diff
@@ -55,7 +54,9 @@ import { AppRoutingModule } from '@example-app/app-routing.module';
      */
     StoreDevtoolsModule.instrument({
       name: 'NgRx Book Store App',
-      logOnly: environment.production,
```
brandonroberts: I would match this instead  https://github.com/ngrx/platform/blob/master/modules/schematics/src/store/index.ts#L102. That way you still have it during development

timdeschryver: But then the deployed version at https://ngrx.github.io/platform/example-app/#/login won't have the devtools enabled since it's a production build? The main reason for this change would be to have it enabled on the deployed example app, which is included in the redux devtools docs.

The maintainer of redux devtools mentioned he would write e2e tests on the extension used in our example app.

Would it be OK to mention the snippet as a comment? Or should we indeed not include the store devtools?

brandonroberts: I still want to include it as not something we have to flip manually when deploying the example. If we need to deploy it with the devtools enabled, we can do that using configuration and the environment variables.

timdeschryver: Oh yea definitely! I also don't want to introduce a (manual) toggle on builds.
I wanted to always build the app with the devtools fully enabled but still wanted to mention that as a user, you wouldn't want to include the dev tools in a production build.

PR #1530, opened 2019-01-25
feat(Schematics): add success/failure effects/actions to ng generate feature

### LINKED ISSUE #1524: Effects generated by Schematics cause 100% CPU / Frozen tab

1. Run ng generate @ngrx/schematics:feature
1. Start the server
1. Open page
1. Observe browser freeze

This is because in the generated module the generated action is dispatched. The effects listen for that dispatched action however does not filter the action so it gets dispatched as a result of passing through the effects.

## Expected behavior:

Dispatch loop does not occur.

```
"@angular/core": "7.2.2",
"@ngrx/effects": "7.0.0",
"@ngrx/router-store": "^7.0.0",
"@ngrx/store": "7.0.0",
"@ngrx/store-devtools": "7.0.0",
Node v10  
```

[ X ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

### LINKED ISSUE #1529: Schematics: add common success/failure actions to feature schematic
When a feature schematic is run, we scaffold out `[Something] Load Something`. In lots of cases you end up with, `[Something] Load Something`, `[Something] Load Something Success`, `[Something] Load Something Failure`.

We should provide some type of option that adds the success/error actions to the action/reducer/effect when generating a feature. The piece that's unknown is what to put in the *Map for the effect.

```ts
@Injectable()
export class SomethingEffect {
  @Effect()
  this.loadSomething$ = this.actions$.pipe(
    oftype('SomeAction'),
    (concat/exhaust/blah)Map(() => 
      (/** some observable to be replaced here */).pipe(
        map(() => new LoadSomethingSuccess()),
        catchError(() => of(new LoadSomethingFailure))
      )
    )
  )
}
```

wesleygrimes:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1529 
Should resolve #1524 

## What is the new behavior?

The PR enhances the default effect generated for feature.

- Adds 2 new actions: `LoadFeatureSuccess` and `LoadFeatureFailure`
- Enhances feature state interface and initialState with data, loading and error properties the
- Enhances feature reducer adding switch cases for each action type
  - `LoadFeature` action sets loading to true
  - `LoadFeatureSuccess` action sets loading to false, nulls error property and updates data with payload. 
  - `LoadFeatureFailure` action sets loading to false, populates error from payload
- Enhances default `loadFeatureEffect$` to demonstrate the idea of a request action, and then based on the return of an observable dispatch a success or failure action.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/schematics/src/action/files/__name@dasherize@if-flat__/__name@dasherize__.actions.ts
+++ b/modules/schematics/src/action/files/__name@dasherize@if-flat__/__name@dasherize__.actions.ts
@@ -12,7 +12,7 @@ export class Load<%= classify(name) %>s implements Action {
 <% if (api) { %>
 export class Load<%= classify(name) %>sSuccess implements Action {
   readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>sSuccess;
-  constructor(public payload: { data: any[] }) { }
+  constructor(public payload: { data: any }) { }
 }
 
 export class Load<%= classify(name) %>sFailure implements Action {
--- a/modules/schematics/src/action/files/__name@dasherize@if-flat__/__name@dasherize__.actions.ts
+++ b/modules/schematics/src/action/files/__name@dasherize@if-flat__/__name@dasherize__.actions.ts
@@ -2,14 +2,14 @@ import { Action } from '@ngrx/store';
 
 export enum <%= classify(name) %>ActionTypes {
   Load<%= classify(name) %>s = '[<%= classify(name) %>] Load <%= classify(name) %>s',
-  Load<%= classify(name) %>sSuccess = '[<%= classify(name) %>] Load <%= classify(name) %>s Success',
-  Load<%= classify(name) %>sFailure = '[<%= classify(name) %>] Load <%= classify(name) %>s Failure',
+  <% if (api) { %>Load<%= classify(name) %>sSuccess = '[<%= classify(name) %>] Load <%= classify(name) %>s Success',<% } %>
+  <% if (api) { %>Load<%= classify(name) %>sFailure = '[<%= classify(name) %>] Load <%= classify(name) %>s Failure',<% } %>
 }
 
 export class Load<%= classify(name) %>s implements Action {
   readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>s;
 }
-
+<% if (api) { %>
 export class Load<%= classify(name) %>sSuccess implements Action {
   readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>sSuccess;
   constructor(public payload: { data: any[] }) { }
@@ -19,5 +19,6 @@ export class Load<%= classify(name) %>sFailure implements Action {
   readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>sFailure;
   constructor(public payload: { error: any }) { }
 }
-
-export type <%= classify(name) %>Actions = Load<%= classify(name) %>s | Load<%= classify(name) %>sSuccess | Load<%= classify(name) %>sFailure;
+<% } %>
+<% if (api) { %>export type <%= classify(name) %>Actions = Load<%= classify(name) %>s | Load<%= classify(name) %>sSuccess | Load<%= classify(name) %>sFailure;<% } %>
+<% if (!api) { %>export type <%= classify(name) %>Actions = Load<%= classify(name) %>s;<% } %>
--- a/modules/schematics/src/action/files/__name@dasherize@if-flat__/__name@dasherize__.actions.ts
+++ b/modules/schematics/src/action/files/__name@dasherize@if-flat__/__name@dasherize__.actions.ts
@@ -1,11 +1,23 @@
 import { Action } from '@ngrx/store';
 
 export enum <%= classify(name) %>ActionTypes {
-  Load<%= classify(name) %>s = '[<%= classify(name) %>] Load <%= classify(name) %>s'
+  Load<%= classify(name) %>s = '[<%= classify(name) %>] Load <%= classify(name) %>s',
+  Load<%= classify(name) %>sSuccess = '[<%= classify(name) %>] Load <%= classify(name) %>s Success',
+  Load<%= classify(name) %>sFailure = '[<%= classify(name) %>] Load <%= classify(name) %>s Failure',
 }
 
 export class Load<%= classify(name) %>s implements Action {
   readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>s;
 }
 
-export type <%= classify(name) %>Actions = Load<%= classify(name) %>s;
+export class Load<%= classify(name) %>sSuccess implements Action {
+  readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>sSuccess;
+  constructor(public payload: { data: any[] }) { }
+}
+
+export class Load<%= classify(name) %>sFailure implements Action {
+  readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>sFailure;
+  constructor(public payload: { error: any }) { }
+}
+
+export type <%= classify(name) %>Actions = Load<%= classify(name) %>s | Load<%= classify(name) %>sSuccess | Load<%= classify(name) %>sFailure;
--- a/modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
+++ b/modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
@@ -1,11 +1,12 @@
 import { Injectable } from '@angular/core';
 import { Actions, Effect<% if (feature) { %>, ofType<% } %> } from '@ngrx/effects';
-<% if (feature) { %>import { catchError, map, switchMap } from 'rxjs/operators';<% } %>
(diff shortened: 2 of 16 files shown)


wesleygrimes:
I'm updating the unit tests on this schematic to match the new code

wesleygrimes:
Looks like the bazel build step failed but it was a timeout issue:
```bash
#!/bin/bash -eo pipefail
yarn
yarn install v1.10.1
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
Too long with no output (exceeded 10m0s)
```

wesleygrimes:
> We'll investigate if it keeps timing out. Some initial feedback:
> 
> * This should be behind an option for a feature (--api maybe??), so its opt-in at least for 7.x.
> * The current tests should remain as-is, with new tests for verifying it works as intended.
> * We don't want to make any assumptions on state structure, but provide actions and default returns for the reducers.

That all sounds good to me. Just one question regarding tests....

How should I handle a current test that is going to fail now because the `import { } from` line it's checking has changed to include the new Success and Failure actions?

wesleygrimes:
@brandonroberts regarding tests, nevermind, I get what you're saying, I will create new tests that test the schematics with the new flag enabled. The current tests will be reverted to match exactly as before as they are testing with the flag disabled by default.

Sorry for the confusion.

wesleygrimes:
@brandonroberts latest commit includes a new flagged added named `api` we can change this if needed. all existing tests have been left untouched. I have created new tests that tests schematics with the `api` flag enabled. in addition, the templates have been updated to utilize the new `api` flag and only define and implement the success and failure actions if the api flag is true.

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
```diff
@@ -1,10 +1,33 @@
...
+<% if (feature && !api) { %>import { <%= classify(name) %>ActionTypes } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
+<% if (feature && api) { %>import { catchError, map, switchMap } from 'rxjs/operators';<% } %>
+<% if (feature && api) { %>import { Load<%= classify(name) %>sFailure, Load<%= classify(name) %>sSuccess, <%= classify(name) %>ActionTypes } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
 
 @Injectable()
 export class <%= classify(name) %>Effects {
-<% if (feature) { %>
+<% if (feature && api) { %>
+
+  private _testObservable = of([
```
timdeschryver: What about using `EMPTY` instead?

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
```diff
@@ -1,10 +1,33 @@
...
+    { myDataProperty: 1100, myOtherDataProperty: 'more-text' }]
+  )
+    .pipe(
+      delay(1000), // simulate latency of 1 second
+      // (_ => { throw new Error('oops'); }) // uncomment to test error handling
+    );
+
+  @Effect()
+  load<%= classify(name) %>s$ = this.actions$.pipe(
+    ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>),
```
timdeschryver: With NgRx 7 we don't need to type the `ofType` operator, should we use the new syntax in the schematics?

It would be:

```
ofType(LoadFoos),

// and the actions defined as:
constructor(private actions$: Actions<FooActions>) {}
```

timdeschryver:
```suggestion
    ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name)s %>),
```

modules/schematics/src/reducer/files/__name@dasherize@if-flat__/__name@dasherize__.reducer.ts
```diff
@@ -14,8 +14,13 @@ export function reducer(state = initialState, action: <% if(feature) { %><%= cla
 <% if(feature) { %>
     case <%= classify(name) %>ActionTypes.Load<%= classify(name) %>s:
       return state;
+<% if(api) { %>
```
timdeschryver: Could you remove the import `import { Action } from '@ngrx/store';`, it isn't in use?

modules/schematics/src/action/files/__name@dasherize@if-flat__/__name@dasherize__.actions.ts
```diff
@@ -1,11 +1,24 @@
...
+  <% if (api) { %>Load<%= classify(name) %>sFailure = '[<%= classify(name) %>] Load <%= classify(name) %>s Failure',<% } %>
 }
 
 export class Load<%= classify(name) %>s implements Action {
   readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>s;
 }
+<% if (api) { %>
+export class Load<%= classify(name) %>sSuccess implements Action {
+  readonly type = <%= classify(name) %>ActionTypes.Load<%= classify(name) %>sSuccess;
+  constructor(public payload: { data: any[] }) { }
```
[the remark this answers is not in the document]
wesleygrimes: That makes sense, keeping it agnostic.

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
wesleygrimes: I am good with changing to the `EMPTY` constant from `rxjs`

brandonroberts: Few things:

- I suggest using concatMap instead of switchMap
- We don't need a private variable here, just inline the created observable inside the the *Map pipe. You're going to replace this with a service call. Maybe an observable of an empty array since we don't know what the state structure is.

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
wesleygrimes: @timdeschryver 
I will update the constructor, but the Effect should already output the following:
```typescript
 @Effect()
  loadFoos$ = this.actions$.pipe(
    ofType(FooActionTypes.LoadFoos),
    switchMap(() =>
      EMPTY.pipe(
        map(data => new LoadFoosSuccess({ data })),
        catchError(error => of(new LoadFoosFailure({ error })))
      )
    )
  );
```
Is that not correct?

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
```diff
@@ -1,10 +1,33 @@
...
 @Injectable()
 export class <%= classify(name) %>Effects {
-<% if (feature) { %>
+<% if (feature && api) { %>
+
+  private _testObservable = of([
+    { myDataProperty: 1000, myOtherDataProperty: 'some-text' },
+    { myDataProperty: 1100, myOtherDataProperty: 'more-text' }]
+  )
+    .pipe(
```
brandonroberts: No need for stub code here

modules/schematics/src/reducer/files/__name@dasherize@if-flat__/__name@dasherize__.reducer.ts
wesleygrimes: I'll add conditional logic to only import this if `!feature`

wesleygrimes: This should be resolved in the latest commit.

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
wesleygrimes: This should be resolved now.

wesleygrimes: this should be resolved now with the latest commit.

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
```diff
@@ -1,12 +1,30 @@
...
 
 @Injectable()
 export class <%= classify(name) %>Effects {
-<% if (feature) { %>
+<% if (feature && api) { %>
+  @Effect()
+  load<%= classify(name) %>s$ = this.actions$.pipe(
+    ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s),
+    concatMap(() =>
+      EMPTY.pipe(
```
brandonroberts: I feel like we need some comment here to tell the user to replace the empty observable?

```suggestion
      /** An EMPTY observable only emits completion. Replace with your own observable API request */
      EMPTY.pipe(
```

wesleygrimes: That makes sense to me.

brandonroberts:
LGTM. Docs for https://ngrx.io/guide/schematics/feature can be done separately.

PR #1539, opened 2019-02-03
 feat(Store): define meta-reducers with a token

### LINKED ISSUE #1441: RFC: Adding meta reducers with injection tokens
This RFC is based on the plugin system of NGXS, where you can import plugins (meta reducers). An example of this is the following:

```ts
@NgModule({
  imports: [
    NgxsModule.forRoot([]),
    NgxsLoggerPluginModule.forRoot()
  ]
})
export class AppModule {}
```

This was from what I have seen, well received by the people using NGXS. 
So I was thinking of adding this into NgRx, eg:

```ts
@NgModule({
      imports: [
        StoreModule.forRoot(
          {},
          {
            metaReducers: [logReducer],
          }
        ),
      ],
      providers: [
        {
          provide: META_REDUCERS,
          useValue: storeFreeze,
          multi: true,
        },
        {
          provide: META_REDUCERS,
          useValue: localStorageSync,
          multi: true,
        },
      ],
    });
```

Currently, it's possible to provide meta reducers by using the config option within `StoreModule.forRoot()` or to provide a meta reducer by overriding the `META_REDUCERS` token.

## Other information:
This change should be backward compatible, where you could still provide the meta reducers via the config, but as well via providers.

[ ] No

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1441

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/store.spec.ts
+++ b/modules/store/spec/store.spec.ts
@@ -1,4 +1,9 @@
-import { ReflectiveInjector, InjectionToken } from '@angular/core';
+import {
+  ReflectiveInjector,
+  InjectionToken,
+  NgModule,
+  ModuleWithProviders,
+} from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { hot } from 'jasmine-marbles';
 import {
@@ -591,7 +596,7 @@ describe('ngRx Store', () => {
     });
 
     describe('Via tokens', () => {
-      const createReducer = ({
+      const createMetaReducer = ({
         logText,
         log,
       }: {
@@ -612,12 +617,12 @@ describe('ngRx Store', () => {
           providers: [
             {
               provide: META_REDUCERS,
-              useValue: createReducer({ logText: 'first', log }),
+              useValue: createMetaReducer({ logText: 'token1', log }),
               multi: true,
             },
             {
               provide: META_REDUCERS,
-              useValue: createReducer({ logText: 'second', log }),
+              useValue: createMetaReducer({ logText: 'token2', log }),
               multi: true,
             },
           ],
@@ -625,9 +630,9 @@ describe('ngRx Store', () => {
 
         store = TestBed.get(Store) as Store<any>;
 
-        expect(log).toEqual(['first', 'second']);
+        expect(log).toEqual(['token1', 'token2']);
         store.dispatch({ type: 'FOO' });
-        expect(log).toEqual(['first', 'second', 'first', 'second']);
+        expect(log).toEqual(['token1', 'token2', 'token1', 'token2']);
       });
 
       it('should respect the order of reducers', () => {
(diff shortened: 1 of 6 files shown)


modules/store/spec/store.spec.ts
```diff
@@ -587,6 +589,204 @@ describe('ngRx Store', () => {
         complete: done,
       });
     });
+
+    describe('Via tokens', () => {
+      const createReducer = ({
```
brandonroberts: For clarity this should be called `createMetaReducer`

modules/store/spec/store.spec.ts
```diff
@@ -587,6 +589,204 @@ describe('ngRx Store', () => {
...
+        expect(log).toEqual(['first', 'second']);
+        store.dispatch({ type: 'FOO' });
+        expect(log).toEqual(['first', 'second', 'first', 'second']);
+      });
+
+      it('should respect the order of reducers', () => {
+        const log: string[] = [];
+        TestBed.configureTestingModule({
+          imports: [
+            StoreModule.forRoot(
```
brandonroberts: This should include a test with an additional NgModule that contains a provided meta-reducer. NgModules have priority over ModuleWithProviders so it will change the order

modules/store/spec/store.spec.ts
```diff
@@ -587,6 +589,204 @@ describe('ngRx Store', () => {
...
+        store = TestBed.get(Store) as Store<any>;
+
+        expect(log).toEqual(['first', 'second']);
+        store.dispatch({ type: 'FOO' });
+        expect(log).toEqual(['first', 'second', 'first', 'second']);
+      });
+
+      it('should respect the order of feature reducers', () => {
+        const log: string[] = [];
+        TestBed.configureTestingModule({
```
brandonroberts: Same comment about NgModules changing the order

timdeschryver:
You're right. I haven't thought of this...

I think with the current setup we can't distinguish the meta reducers defined with the config and the meta reducers defined with a token. Thus we can't provide a proper solution? 

In order to solve this, we could add a new `USER_DEFINED_META_REDUCERS` token. Based on these meta reducers and the meta reducers provided via the config we have more control over it. We could build up the meta reducers composition as follows

```ts
[userDefinedMetaReducer1, userDefinedMetaReducer2, configReducer1, configReducer2]

// or with an example
[persistState, logger, store-freeze]
```

Another question that pops up, is this feature is really needed or would it be confusing for our users that we provide 2 different ways to define meta reducers.

brandonroberts:
It depends. If order was not a concern, it would be fine. If the goal is just to provide a way for developers to provide meta-reducers as `SomeModule.forRoot()`, I don't think its worth it. Its going to cause more headaches with them being causing weird order combinations.

timdeschryver:
I follow you completely 😄 
My goal was to be able to let users provide meta reducers as 'NGXS plugins'. But now that I think of it I think it would lead to confusion, instead of a handy feature.

timdeschryver:
FWIW, https://github.com/ngrx/platform/commit/c2b857845dfa6c4f9c94763a33885102086b0aeb shows how the order of meta reducers.

brandonroberts:
I think its a reasonable change for those who want to override the defaults. I also think its a breaking change for people currently overriding the root meta-reducers with a token. We'll put it on the list for V8.

timdeschryver:
👍 , I've re-pushed c2b857845dfa6c4f9c94763a33885102086b0aeb

timdeschryver:
This feature is implemented via #1613

PR #1576, opened 2019-02-21
fix(schematics): type actions and avoid endless loop in effect

### LINKED ISSUE #1573: Schematics: Default effect leads to an endless loop
The command `ng generate feature` does create an effects class with the following effect (example):

```ts
@Effect()
loadBooks$ = this.actions$.pipe(ofType(BookActionTypes.LoadBooks));
```
(see https://github.com/ngrx/platform/blob/master/modules/schematics/src/effect/files/__name%40dasherize%40if-flat__/__name%40dasherize__.effects.ts#L24)

If left untouched, this will lead to a [loop of death](https://angular.schule/blog/2018-06-5-useful-effects-without-actions#5-loop-of-death-) as soon as the first `LoadBooks` action is being dispatched.

## Possible solutions
* Comment out the `@Effect()` decorator by default
* don't provide a default effect after creation
* set `{ dispatch: false }`

I don't like any of them – but I ran into that the infinite loop problem multiple times now 😆
Will prepare a PR if there is consensus about how we can solve this.

[ ] No

fmalcher:
Two minor fixes for effects generated by `ng g feature`:

- Make the generated effect for 'ng g feature'  harmless by not dispatching any action
- type the 'Actions' injectable by default in feature effect

Closes #1573

## PR Checklist

- [ ] ~~Documentation has been added / updated (for bug fixes / features)~~

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
+++ b/modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
@@ -1,9 +1,13 @@
 import { Injectable } from '@angular/core';
 import { Actions, Effect<% if (feature) { %>, ofType<% } %> } from '@ngrx/effects';
-<% if (feature && api) { %>import { catchError, map, concatMap } from 'rxjs/operators';<% } %>
-<% if (feature && api) { %>import { EMPTY, of } from 'rxjs';<% } %>
-<% if (feature && api) { %>import { Load<%= classify(name) %>sFailure, Load<%= classify(name) %>sSuccess, <%= classify(name) %>ActionTypes, <%= classify(name) %>Actions } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
-<% if (feature && !api) { %>import { <%= classify(name) %>ActionTypes } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
+<% if (feature && api) { %>import { catchError, map, concatMap } from 'rxjs/operators';
+import { EMPTY, of } from 'rxjs';
+import { Load<%= classify(name) %>sFailure, Load<%= classify(name) %>sSuccess, <%= classify(name) %>ActionTypes, <%= classify(name) %>Actions } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';
+<% } %>
+<% if (feature && !api) { %>import { concatMap } from 'rxjs/operators';
+import { EMPTY } from 'rxjs';
+import { <%= classify(name) %>ActionTypes, <%= classify(name) %>Actions } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';
+<% } %>
 
 @Injectable()
 export class <%= classify(name) %>Effects {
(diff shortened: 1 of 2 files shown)


modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
```diff
@@ -3,7 +3,9 @@ import { Actions, Effect<% if (feature) { %>, ofType<% } %> } from '@ngrx/effect
 <% if (feature && api) { %>import { catchError, map, concatMap } from 'rxjs/operators';<% } %>
 <% if (feature && api) { %>import { EMPTY, of } from 'rxjs';<% } %>
 <% if (feature && api) { %>import { Load<%= classify(name) %>sFailure, Load<%= classify(name) %>sSuccess, <%= classify(name) %>ActionTypes, <%= classify(name) %>Actions } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
-<% if (feature && !api) { %>import { <%= classify(name) %>ActionTypes } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';<% } %>
+<% if (feature && !api) { %>import { concatMap } from 'rxjs/operators';<% } %>
```
timdeschryver: Nit, I think it would be more readable if we would group these `if` statements into a single one?
The same counts for `feature && api`.

fmalcher: I thought about this, but saw it similarly in one of the other schematics. I'll change this

modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts
```diff
@@ -21,9 +23,12 @@ export class <%= classify(name) %>Effects {
 <% } %>
 <% if (feature && !api) { %>
   @Effect()
-  load<%= classify(name) %>s$ = this.actions$.pipe(ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s));
+  load<%= classify(name) %>s$ = this.actions$.pipe(
+    ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s),
```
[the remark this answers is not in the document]
fmalcher: done

modules/schematics/src/effect/index.spec.ts
```diff
@@ -250,12 +250,18 @@ describe('Effect Schematic', () => {
...
+      /import { FooActionTypes, FooActions } from '\.\/foo.actions';/
     );
     expect(content).toMatch(/export class FooEffects/);
+    expect(content).toMatch(/loadFoos\$ = this\.actions\$.pipe\(/);
+    expect(content).toMatch(/ofType\(FooActionTypes\.LoadFoos\)/);
+    expect(content).toMatch(/concatMap\(\(\) => EMPTY\)/);
+
     expect(content).toMatch(
-      /loadFoos\$ = this\.actions\$.pipe\(ofType\(FooActionTypes\.LoadFoos\)\);/
+      /constructor\(private actions\$: Actions<FooActions>\) {}/
```
brandonroberts: Instead of removing the existing line, add an additional check under it.

modules/schematics/src/effect/index.spec.ts
```diff
@@ -274,7 +280,7 @@ describe('Effect Schematic', () => {
     );
     expect(content).toMatch(/export class FooEffects/);
     expect(content).not.toMatch(
-      /loadFoos\$ = this\.actions\$.pipe\(ofType\(FooActionTypes\.LoadFoos\)\);/
+      /loadFoos\$ = this\.actions\$.pipe\(ofType\(FooActionTypes\.LoadFoos/
```
brandonroberts: Why were these removed?

modules/schematics/src/effect/index.spec.ts
fmalcher: This is no removal, the replacement for the old line 258 is now in 259–261

brandonroberts: Ok, I see 👍

modules/schematics/src/effect/index.spec.ts
fmalcher: I think we need to rethink this line completely. The purpose is to make sure the root effects class doesn't contain ANY effect at all – no matter whether this is an API effect, the `concatMap(() => EMPTY)` one or the old endless loopy one.
Therefore, should we change this to

```
expect(content).not.toMatch(
  /loadFoos\$ = this\.actions\$.pipe\(/
)
```

to make sure none of the effects is contained?

Issue #1501, opened 2019-01-04
CreateSelectorFactory returns entire state when provided an empty array of selectors


defernan:
## Explanation
Providing createSelectorFactory n selectors yields inconsistent results between v6 and v7. 

In version 6 if createSelectorFactory was provided an array of empty selectors, the projectorFunction would be supplied an empty array.

In version 7 the projectorFunction is supplied an array of the entire state.

## Minimal reproduction of the bug/regression with instructions:
Refer to console in each stackblitz

v6
https://stackblitz.com/edit/typescript-createselectorfactory-v6

v7
https://stackblitz.com/edit/typescript-createselectorfactory-v7

## Expected behavior:

Retain functionality and return an empty array if no selectors were provided. 
Is there a better way to get the result of a dynamic number of selectors?

## Versions of NgRx, Angular, Node, affected browser(s) and operating system(s):
ngrx v6 -> v7

## Use case:
I have x members at any time which i need to query information on and display statuses, which is why i need dynamic selectors.

timdeschryver:
I think that the current implementation is correct, but also that the previous implementation was correct. The reasoning behind it is the following:

- When you don't have any selectors/getters defined, it should invoke the projector function. This causes the behavior in your v7 example to return the state.
- A `createSelector` without a selector/getter wasn't possible in v6. The only reason why it returns an empty array in your v6 example is because you're using the spread operator in your selector. Change the implementation to `(statuses) => statuses` and it will log out `undefined`, because there are no selectors/getter to return anything. 

The change that causes this:
The selector with props feature, introduced a `createSelector` function without the need to define a selector/getter. With this change the following lines 
https://github.com/ngrx/platform/blob/master/modules/store/src/selector.ts#L554 were added to directly call the projector in this case. In your case, this also means that we directly call the projector function and that the state will be passed to the projector function (instead of `undefined` because there are no selectors).

These changes maked the following possible:

```ts
const getTodosById = createSelector(
  (state: TodoAppSchema, id: number) => state.todos.find(p => p.id === id)
);
```

Instead of the need to create a selector/getter:
```ts
const getTodosById = createSelector(
  (state: TodoAppSchema) => state,
  (todos: Todo[], id: number) => todos.find(p => p.id === id)
);
```

[...]

PR #1579, opened 2019-02-25
fix(Store): selector with only a projector

### LINKED ISSUE #1558: Single createSelector function doesn't get state passed in.
Hello,
This took me some time to find but I've ran into this issue.
When I have only simple selector
```
const somethingSelector = createSelector(
  (state: StoreState) => {
    return {
      state.something
    };
  }
);
```
It will not get the state passed in.

I've found that this PR: #1515 has introduced this issue.
Not very sure why the function is not called with state as an argument but it broke my application immediately.

## Expected behavior:
Nothing should be broken by a not-breaking release.
Use of createSelector without any previous selectors is broken.

[ X] No. I think reverting the previous PR should be done. And a proper fix for the issue it was intented to fix

timdeschryver:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1558

```
[x] Yes
[ ] No
```

BREAKING CHANGE:

Selectors with only a projector function aren't valid anymore.
This change will make the usage more consistent.

BEFORE:

```
  (state: TodoAppSchema, id: number) => state.todos.find(p => p.id === id)
);
```

AFTER:

```
  (state: TodoAppSchema) => state.todos,
  (todos: Todo[], id: number) => todos.find(p => p.id === id)
);
```


modules/store/spec/selector.spec.ts
```diff
@@ -143,17 +143,6 @@ describe('Selectors', () => {
       });
     });
 
-    it('should not short circuit to the projector fn if there are no selectors and props', () => {
```
timdeschryver: This test was to verify the fix for #1501.
Since we don't short circuit anymore, this test became obsolete.

brandonroberts:
Will you put in another PR that deprecates the old method overload?

PR #1593, opened 2019-03-02
test(schematics): add @ngrx/schematics to Bazel tests and tweak CI config

brandonroberts:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[x] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Closes #

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

PR #1596, opened 2019-03-03
refactor(schematics):entity schematics should update the state to plural

### LINKED ISSUE #1412: Entity schematics should update the state with a plural instead of a singular

1. Create a new Angular project (V7)
2. Generate the intial state management:
```
ng generate @ngrx/schematics:store State --statePath store --root --module app.module.ts
```
3. Generate a new entity:
```
ng generate @ngrx/schematics:entity user/User -m --reducers ../store/index.ts app.module.ts
```

Check [this repository](https://github.com/jdjuan/ngrx-plural-bug/blob/master/src/app/store/index.ts#L11-L17) for the generated files. But here is the interesting part:

```typescript
...
export interface State {
  user: fromUser.State;
}

export const reducers: ActionReducerMap<State> = {
  user: fromUser.reducer,
};
...
```

## Expected behavior:

According to the documentaiton [here](https://github.com/ngrx/platform/blob/master/docs/entity/adapter.md#entity-selectors), it seems like entities should be defined as plural in the state:

```typescript
export interface State {
  users: fromUser.State;
}

export const reducers: ActionReducerMap<State> = {
  users: fromUser.reducer,
};
```

- Angular: 7.0.3
- Node: 8.11.2
- OS: Windows

[ ] No

santoshyadav198613:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

currently the state variable is singular updating the same to plural
Closes #1412

```
[ ] Yes
[x] No
```


santoshyadav198613:
@brandonroberts  ,

Need help here don't know why CI is getting failed for e2e testing.

timdeschryver:
@santoshyadav198613 , the failing e2e test isn't related to your change.
I will take a look at it.

timdeschryver:
The problem is that we are caching the cypress installation in the wrong build step.
This should be solved when #1593 gets merged.

santoshyadav198613:
> d it needs some tests for common scenarios.

But this code is part of common function inside schematics-core , can you guide me how should we handle it?

brandonroberts:
You create a test for a generated schematic and check that the output is what you expect. For example, your test would use the entity schematic to generate a `user` collection. The schematic should generate a property named `users`, with the associated files.

santoshyadav198613:
> Should we use the `pluralize` method at this point?
> Because I'm a bit afraid this would lead to inconsistencies, for example, while scaffolding the effects we append the name with an **s**.

Ok let me check it.

santoshyadav198613:
> You create a test for a generated schematic and check that the output is what you expect. For example, your test would use the entity schematic to generate a `user` collection. The schematic should generate a property named `users`, with the associated files.

ok will update and let you know.

santoshyadav198613:
> You create a test for a generated schematic and check that the output is what you expect. For example, your test would use the entity schematic to generate a `user` collection. The schematic should generate a property named `users`, with the associated files.

Hi @brandonroberts ,

Added a test case for the scenario this is what i have done.
Added one store and entity and then verified if state variable is plural or not same for reducers.

santoshyadav198613:
@brandonroberts and @timdeschryver let me know about the usage of plural function should i remove it.

santoshyadav198613:
Hi @brandonroberts ,
Can you verify again, wrote the test case as per our discussion.

brandonroberts:
@santoshyadav198613 we don't want to modify existing behavior for other areas. Instead, add an `options` object to the ngrx-utils methods as the second argument that's set to { pluralize: false }`. For entity schematics, pass `{ pluralize: true }` and use it whether to call `pluralize` or just `camelize` inside those functions.

timdeschryver:
I'm a bit hesitant to add `pluralize` (for now) because this would lead to inconsistencies, e.g. city would be pluralized to cities, and in the other schematics to citys. 

What about keeping this PR scoped to making the entities pluralized (adding an `s`), and to create a separate issue to pluralize all the schematics via this function?

brandonroberts:
@timdeschryver yes, scoping this to entities is what I had in mind also.

santoshyadav198613:
> @santoshyadav198613 we don't want to modify existing behavior for other areas. Instead, add an `options` object to the ngrx-utils methods as the second argument that's set to { pluralize: false }`. For entity schematics, pass `{ pluralize: true }`and use it whether to call`pluralize`or just`camelize` inside those functions.

Thanks @brandonroberts ,
Will make the changes.

santoshyadav198613:
Hi @brandonroberts and @timdeschryver ,
Done with the changes as per discussion, test cases are also added/modified.
The change is restricted to entity only, also appending only 's', keeping pluralize let me know if i should remove it.

modules/schematics/src/entity/index.spec.ts
```diff
@@ -159,4 +161,34 @@ describe('Entity Schematic', () => {
       files.indexOf(`${projectPath}/src/app/reducers/foo.reducer.spec.ts`)
     ).toBeGreaterThanOrEqual(0);
   });
+
+  it('should create all files of an entity within grouped and nested folders', () => {
```
timdeschryver: This spec name has to be updated

modules/schematics-core/utility/ngrx-utils.ts
```diff
@@ -152,11 +154,13 @@ export function addReducerToActionReducerMap(
 
   let node = actionReducerMap.initializer;
 
+  const state =
+    options.plural === undefined
```
timdeschryver: This check isn't sufficient, if we set `plural` to `false` we would still pluralize the name and this might be confusing 😄

modules/schematics/src/entity/schema.json
```diff
@@ -49,6 +49,11 @@
       "description":
         "Group actions, reducers and effects within relative subfolders",
       "aliases": ["g"]
+    },
+    "plural": {
```
timdeschryver: This isn't needed, we always pluralize - https://github.com/ngrx/platform/blob/757d04ebeb35a8811c214a3b3c9e02318e650034/modules/schematics/src/entity/index.ts#L34

modules/schematics/src/entity/schema.ts
```diff
@@ -36,4 +36,9 @@ export interface Schema {
    */
 
   group?: boolean;
+  /**
+   * Specifies if this is state and reducer variable will be plural
+   */
+
+  plural?: boolean;
```
timdeschryver: Same here

modules/schematics-core/utility/ngrx-utils.ts
santoshyadav198613: Yes i was thinking about same, after i went to sleep. thanks for catching it.

modules/schematics/src/entity/index.spec.ts
santoshyadav198613: done

modules/schematics/src/entity/schema.json
santoshyadav198613: removed

modules/schematics/src/entity/schema.ts
santoshyadav198613: done

modules/schematics-core/utility/ngrx-utils.ts
```diff
@@ -90,11 +90,13 @@ export function addReducerToStateInterface(
     return new NoopChange();
   }
 
+  let state = stringUtils.camelize(options.name);
```
brandonroberts:
```suggestion
  let state = options.plural ? stringUtils.pluralize(options.name) : stringUtils.camelize(options.name);
```

Let's use the pluralize method instead of tacking on an `s`.

modules/schematics-core/utility/strings.ts
```diff
@@ -110,6 +110,25 @@ export function capitalize(str: string): string {
...
 }
 
+/**
+ Returns the Pluralize form of a string
+
+ ```javascript
+ 'innerHTML'.pluralize()         // 'InnerHTMLs'
+ 'action_name'.pluralize()       // 'actionNames'
+ 'css-class-name'.pluralize()    // 'cssClassNames'
+ 'regex'.capitalize()            // 'regexes'
```
brandonroberts:
```suggestion
 'regex'.pluralize()            // 'regexes'
```

modules/schematics-core/utility/strings.ts
```diff
@@ -110,6 +110,25 @@ export function capitalize(str: string): string {
...
+ 'user'.capitalize()             // 'users'
```
brandonroberts:
```suggestion
 'user'.pluralize()             // 'users'
```

modules/schematics/schematics-core/utility/strings.ts
```diff
@@ -110,6 +110,25 @@ export function capitalize(str: string): string {
   return str.charAt(0).toUpperCase() + str.substr(1);
 }
 
+/**
+ Returns the Pluralize form of a string
```
brandonroberts:
```suggestion
 Returns the plural form of a string
```

santoshyadav198613: done

modules/schematics-core/utility/strings.ts
santoshyadav198613: done

modules/schematics-core/utility/strings.ts
santoshyadav198613: done

modules/schematics-core/utility/ngrx-utils.ts
santoshyadav198613: done

santoshyadav198613:
Hi @brandonroberts @timdeschryver ,
All changes are done please review.

PR #1613, opened 2019-03-09
feat(Store): runtime checks

### LINKED ISSUE #857: Add Mutation and Serialization Checks into Store
If we choose to implement #856 we should consider adding mutation checks directly into store. Think ngrx-state-freeze but built-in and it also checks if state can be serialized.

Considerations:
* State should only be checked when Angular is in development mode
* Checking should be the default but it should be able to be turned off when configuring the Store
* @ngrx/router-store should ship with a custom serializer turned on by default that allows the state to be frozen and serialized

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #857

## What is the new behavior?

- [x] State serialization checks
- [x] Action serialization checks
- [x] Mutation checks

- [x] Remove ngrx-store-freeze ?
- [x] Remove non-serializable properties on the router ?

```
[x] Yes
[ ] No
```

BREAKING CHANGE:

The META_REDUCERS token has been renamed to USER_PROVIDED_META_REDUCERS
The META_REDUCERS token has become a multi token and can be used by
library authors

## Other information

Based of [feat/serialization-and-mutation-checks](https://github.com/ngrx/platform/tree/feat/serialization-and-mutation-checks)

This PR also affects #1539

### DIFF
--- a/modules/store/src/index.ts
+++ b/modules/store/src/index.ts
@@ -46,6 +46,7 @@ export {
   _FEATURE_REDUCERS,
   FEATURE_REDUCERS,
   _FEATURE_REDUCERS_TOKEN,
+  USER_PROVIDED_META_REDUCERS,
 } from './tokens';
 export {
   StoreModule,
--- a/modules/store/src/models.ts
+++ b/modules/store/src/models.ts
@@ -26,7 +26,7 @@ export interface ActionReducerFactory<T, V extends Action = Action> {
   ): ActionReducer<T, V>;
 }
 
-export type MetaReducer<T, V extends Action = Action> = (
+export type MetaReducer<T = any, V extends Action = Action> = (
   reducer: ActionReducer<T, V>
 ) => ActionReducer<T, V>;
 
@@ -57,3 +57,8 @@ export type FunctionWithParametersType<P extends unknown[], R = void> = (
 export type ParametersType<T> = T extends (...args: infer U) => unknown
   ? U
   : never;
+export interface RuntimeChecks {
+  strictStateSerializability: boolean;
+  strictActionSerializability: boolean;
+  strictImmutability: boolean;
+}
--- a/modules/store/src/runtime_checks.ts
+++ b/modules/store/src/runtime_checks.ts
@@ -1,0 +1,95 @@
+import { isDevMode, Provider } from '@angular/core';
+import {
+  stateSerializationCheckMetaReducer,
+  actionSerializationCheckMetaReducer,
+  immutabilityCheckMetaReducer,
+} from './meta-reducers';
+import { RuntimeChecks, MetaReducer } from './models';
+import {
+  _USER_RUNTIME_CHECKS,
+  _ACTIVE_RUNTIME_CHECKS,
+  META_REDUCERS,
+} from './tokens';
+
+export function createActiveRuntimeChecks(
+  runtimeChecks?: Partial<RuntimeChecks>
+): RuntimeChecks {
+  if (isDevMode()) {
+    if (runtimeChecks === undefined) {
+      console.warn(
+        '@ngrx/store: runtime checks are currently opt-in but will be the default in the next major version, see https://ngrx.io/guide/migration/v8 for more information.'
+      );
+    }
+    return {
+      strictStateSerializability: false,
+      strictActionSerializability: false,
+      strictImmutability: false,
+      ...runtimeChecks,
+    };
+  }
+
+  return {
+    strictStateSerializability: false,
+    strictActionSerializability: false,
+    strictImmutability: false,
+  };
+}
+
+export function createStateSerializationCheckMetaReducer({
+  strictStateSerializability,
+}: RuntimeChecks): MetaReducer {
+  return reducer =>
+    strictStateSerializability
+      ? stateSerializationCheckMetaReducer(reducer)
+      : reducer;
+}
+
+export function createActionSerializationCheckMetaReducer({
+  strictActionSerializability,
+}: RuntimeChecks): MetaReducer {
+  return reducer =>
+    strictActionSerializability
+      ? actionSerializationCheckMetaReducer(reducer)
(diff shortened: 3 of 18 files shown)


modules/store/src/state.ts
```diff
@@ -47,9 +47,12 @@ export class State<T> extends BehaviorSubject<any> implements OnDestroy {
...
 
-    this.stateSubscription = stateAndAction$.subscribe(({ state, action }) => {
-      this.next(state);
-      scannedActions.next(action);
+    this.stateSubscription = stateAndAction$.subscribe({
+      next: ({ state, action }) => {
+        this.next(state);
+        scannedActions.next(action);
+      },
+      error: err => this.error(err),
```
timdeschryver: This was needed in order for me to test the case with an invalid next state.

timdeschryver:
Is this the API we had in mind?

brandonroberts:
Yes, I like the API. A few thoughts:

- I wonder how much router-store would be impacted from removing paramMap and queryParamMap?
- I think this has the potential to uncover many accidental mutations in peoples apps (break them).
- What's the migration strategy for those overriding the META_REDUCERS token today?

timdeschryver:
> I wonder how much router-store would be impacted from removing paramMap and queryParamMap?

We would stay have `params` and `queryParams` which is the same object (?) but without some getters or util functions.

> I think this has the potential to uncover many accidental mutations in peoples apps (break them).

What about making these checks opt-in v8 and log a warning on start up in dev that these checks will be the default in v9 or v8.x?

> What's the migration strategy for those overriding the META_REDUCERS token today?

Is it correct that if we would flatten out the meta reducers inside the `concatMetaReducers` function this shouldn't be a problem? The only issue would be that these reducers would be registered as an internal or library meta-reducer
To test this out I will create a test 😄

timdeschryver:
See fcd21bb, I lied. The user also has to add `multi: true`.

But now that I'm thinking about it, the user has to change the token from `META_REDUCERS` to `USER_PROVIDED_META_REDUCERS`.

timdeschryver:
For an easy transition we could provide a linter and possibly add it to the migration script?  
Something like https://github.com/timdeschryver/tslint-playground/blob/master/src/rules/noMetaReducersRule.ts.

modules/store/src/index.ts
```diff
@@ -6,6 +6,7 @@ export {
   MetaReducer,
   Selector,
   SelectorWithProps,
+  RuntimeChecks,
```
brandonroberts: These don't need to be exported publicly except for `USER_PROVIDED_META_REDUCERS`

projects/example-app/src/app/app.module.ts
```diff
@@ -35,7 +35,13 @@ import { AppRoutingModule } from '@example-app/app-routing.module';
      * meta-reducer. This returns all providers for an @ngrx/store
      * based application.
      */
-    StoreModule.forRoot(reducers, { metaReducers }),
+    StoreModule.forRoot(reducers, {
+      metaReducers,
+      runtimeChecks: {
+        strictStateSerializabilityChecks: false,
+        strictActionSerializabilityChecks: false,
```
brandonroberts: Where are we mutating actions? We should clean this up.

brandonroberts:
I think we should do a separate PR to remove the non-serializable parts (paramMap, queryParamMap) from router-store so we aren't breaking too many things here.

modules/store/src/models.ts
```diff
@@ -39,3 +39,9 @@ export type SelectorWithProps<State, Props, Result> = (
   state: State,
   props: Props
 ) => Result;
+
+export interface RuntimeChecks {
+  strictStateSerializabilityChecks: boolean;
```
brandonroberts: We are already calling these `RuntimeChecks`, so we can remove `Checks` from the properties themselves.

modules/router-store/spec/utils.ts
```diff
@@ -29,7 +29,14 @@ export function createTestModule(
   TestBed.configureTestingModule({
     declarations: [AppCmp, SimpleCmp],
     imports: [
-      StoreModule.forRoot(opts.reducers),
+      StoreModule.forRoot(opts.reducers, {
```
brandonroberts: Remove this as its opt-in

modules/store/src/runtime_checks.ts
```diff
@@ -0,0 +1,104 @@
...
+  META_REDUCERS,
+} from './tokens';
+
+export function createActiveRuntimeChecks(
+  runtimeChecks?: Partial<RuntimeChecks>
+): RuntimeChecks {
+  if (isDevMode()) {
+    if (runtimeChecks === undefined) {
+      console.warn(
+        '@ngrx/store: we added immutability and serializability runtime checks in @ngrx/store.\n' +
```
brandonroberts: I think we should keep this message short and add information about the checks to a page we link to. Another idea is to make this a comment for the `RootStoreConfig` so it shows up in their IDE and makes the console less noisy.

timdeschryver: Yea, you're right 😅 the link to a page is a good idea 👍 
My intention was to log it so it will certainly be visible, but this can quickly become annoying. It will also log out the messages in people's tests... On the other hand I'm just afraid this change will be unnoticed by teams if we put in a comment on `RootStoreConfig`. This log can be disabled by providing a config, this can just be an empty object.

brandonroberts: 👍

brandonroberts:
@timdeschryver will you rebase so we can land this?

brandonroberts:
Does this need to land as a rebase? Are there any breaking change commits?

timdeschryver:
I've rebased and added a BREAKING CHANGE comment

PR #1643, opened 2019-03-20
docs: adding description to action page for NGRX

softchris:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

## What is the new behavior?
Added explanation to what an action is

```
[ ] Yes
[x] No
```


projects/ngrx.io/content/guide/store/actions.md
```diff
@@ -6,6 +6,23 @@ Actions are one of the main building blocks in NgRx. Actions express _unique eve
 
 Actions are used in many areas of NgRx. Actions are the inputs and outputs of many systems in NgRx. Actions help you to understand how events are handled in your application. This guide provides general rules and examples for writing actions in your application.
 
+## An action is an object
```
brandonroberts: How is this different from the `Action interface` section below? They basically say the same thing.

softchris: hi Brandon :) So my idea was to mention this as early as possible. My general idea with the docs is to present the vanilla choice as early as possible and then show in the next section what benefits it holds to use interfaces and classes.. I feel a lot of the confusion about NGRX comes from people not understanding what happens in Vanilla, their lack of typescript and then they get confused

alex-okrushko:
I was about to say that this was copied from Chris Noring's blog articles and this realized that it was you 🙂

softchris:
> I was about to say that this was copied from Chris Noring's blog articles and this realized that it was you 🙂

haha.. Yea me and Brandon was discussing me helping the NGRX docs.. We said what I wrote made sense as architecture. So I wrote a first sample on architecture to see if you liked the style, wdyt?

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+This is where we create an action representing our intent and because we want to change to something specific we need to have a payload with the value `french`. An action is nothing more than an object looking like this"
+
+<code-example header="">
+const action { type: '[Language] change', payload: 'french' }
+</code-example>
+
+At this point we want to *dispatch* this action so our intended change ends up the store. As we said before Reducers are the guards of the store. Therefore we need to pass our action to a Reducer. 
+
+Reducers are the guards of the store as has been mentioned a couple of times. So what other properties to they have? We know they are functions but a thing that makes Redux really great is that it's predictable. What do we mean by that? Predictability is something we achieve by having our Reducers be pure functions. Ok, another concept, what does pure functions mean? Pure functions are functions that act like mathematical functions is that given the same input data they produce the same output data. Let's see an example:
```
timdeschryver:
```suggestion
Reducers are the guards of the store as has been mentioned a couple of times. So what other benefits do they have? We know they are functions but a thing that makes Redux really great is that it's predictable. What do we mean by that? Predictability is something we achieve by having our Reducers be pure functions. Ok, another concept, what does pure functions mean? Pure functions are functions that act like mathematical functions, that given the same input data they produce the same output data. Let's see an example:
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+<code-example header="A pure function">
+function sum(lhs, rhs) { 
+  return lhs + rhs; 
+}
+
+sum(1,1); // 2
```
timdeschryver: Maybe use the same arguments as below?

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+  }
+}
+</code-example>
+
+Ok, so we have a reducer, so how do we use it? Well, that's just a matter of invoking the reducer, like so:
+
+<code-example header="invoking our reducer">
+let state = reducer('', { type: '[Language] change', payload: 'english' }); //english
+
+state = reducer(state, { type: '[Language] change', payload: 'french' })
```
timdeschryver:
```suggestion
state = reducer(state, { type: '[Language] change', payload: 'french' }) //french
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+  compute(state, action) {
+    return {
+      language: languageReducer(state.language, action),
+      products: productsReducer(state.products, action) 
+    }
+  }
+}
+</code-example>
+We have on purpose left out the part on how we subscribe to the store and how we broadcast a new store state after it has been changed by a call to `dispatch()`. The reason for that is that it's implementation specific to what Redux implementation we are talking about. In some implementations EventEmitters are being used, in NGRX it's RxJS.
```
timdeschryver:
```suggestion

We have on purpose left out the part on how we subscribe to the store and how we broadcast a new store state after it has been changed by a call to `dispatch()`. The reason for that is that it's implementation specific to what Redux implementation we are talking about. In some implementations EventEmitters are being used, in NGRX it's RxJS. 
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+    return {
+      language: languageReducer(state.language, action),
+      products: productsReducer(state.products, action) 
+    }
+  }
+}
+</code-example>
+We have on purpose left out the part on how we subscribe to the store and how we broadcast a new store state after it has been changed by a call to `dispatch()`. The reason for that is that it's implementation specific to what Redux implementation we are talking about. In some implementations EventEmitters are being used, in NGRX it's RxJS. 
+
+At this point we have mentioned all major concepts in Redux like `action`, `reducer` and `store`. We've even shown a very naive implementation so you understand how all the pieces go together. Next is understanding where NGRX comes with its specific solution and especially why the heavy usage of Typescript makes sense and helps creating a better experience.
```
timdeschryver:
```suggestion
At this point we have mentioned all major concepts in Redux like `action`, `reducer` and `store`. We've even shown a very naive implementation so you understand how all the pieces go together. Next is understanding where NgRx comes with its specific solution and especially why the heavy usage of TypeScript makes sense and helps creating a better experience.
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+## Introducing NGRX and the need for type safety
+NGRX is a Redux implementation meant for Angular and it makes heave use of Typescript and RxJS. The biggest reason is probably typesafety.
```
timdeschryver:
```suggestion
NgRx is a Redux implementation meant for Angular and it makes heave use of TypSscript and RxJS. The biggest reason is type safety. 
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+      products: productsReducer(state.products, action) 
+    }
+  }
+}
+</code-example>
+We have on purpose left out the part on how we subscribe to the store and how we broadcast a new store state after it has been changed by a call to `dispatch()`. The reason for that is that it's implementation specific to what Redux implementation we are talking about. In some implementations EventEmitters are being used, in NGRX it's RxJS. 
+
+At this point we have mentioned all major concepts in Redux like `action`, `reducer` and `store`. We've even shown a very naive implementation so you understand how all the pieces go together. Next is understanding where NGRX comes with its specific solution and especially why the heavy usage of Typescript makes sense and helps creating a better experience.
+
+## Introducing NGRX and the need for type safety
```
timdeschryver:
```suggestion
## Introducing NgRx and the need for type safety
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+NGRX is a Redux implementation meant for Angular and it makes heave use of Typescript and RxJS. The biggest reason is probably typesafety. 
+
+So why would we want to make them typesafe in the first place? Well, remember when we constructed an action in our previous section and assigned a string value to the `type` field? That is actually quite error prone, cause you might mispel that string value. It's maybe hard to understand why a mispel would be such a big deal but let's very briefly mention how actions are used. 
+
+### Dispatching an action
+Actions are being *dispatched* to reducer functions. What does that mean? Normally when you set up NGRX you do so be setting up your store like this:
```
timdeschryver:
```suggestion
Actions are being *dispatched* to reducer functions. What does that mean? Normally when you set up NgRx you do so by setting up your store like this:
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+  }
+}
+
+// app.module.ts
+StoreModule.forRoot({
+  products: productsReducer
+})
+</code-example>
+
+When you do that you say that `products` is a thing that will exist in a store and it's state is something we can either *read* or *change*. The way we *read* the value is by subscribing to it using the `Store` service inside of component usually like so:
```
timdeschryver:
```suggestion
When you do that you say that `products` is a piece of the state that will exist in the store and it's state is something we can either *read* or *change*. The way we *read* the value is by subscribing to it using the `Store` service inside of component usually like so:
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+<code-example header="app.module.ts">
+constructor(private store: Store<AppState>) {
+  this.store.pipe(
+    select('products')
+  ).subscribe(products => console.log('products', products));
+}
+</code-example>
+
+The way we *change* we change our data in the store is by calling the same `Store` service and its `dispatch()` method. The `dispatch()` method will *dispatch*, i.e send an action to the `store` and try to match that to a reducer and specifically a matching case in a reducer. So if we in our component where to do this:
```
timdeschryver:
```suggestion
The way we *change* our data in the store is by calling the same `Store` service and its `dispatch()` method. The `dispatch()` method will *dispatch*, i.e send an action to the `store` and try to match that to a reducer and specifically a matching case in a reducer. So if we in our component where to do this:
```

This reads like NgRx invokes the "correct" reducer, but it actually sends the action to all reducers. Just saying it here, because sometimes people have a confusion about this.

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+At this point we have mentioned all major concepts in Redux like `action`, `reducer` and `store`. We've even shown a very naive implementation so you understand how all the pieces go together. Next is understanding where NGRX comes with its specific solution and especially why the heavy usage of Typescript makes sense and helps creating a better experience.
+
+## Introducing NGRX and the need for type safety
+NGRX is a Redux implementation meant for Angular and it makes heave use of Typescript and RxJS. The biggest reason is probably typesafety. 
+
+So why would we want to make them typesafe in the first place? Well, remember when we constructed an action in our previous section and assigned a string value to the `type` field? That is actually quite error prone, cause you might mispel that string value. It's maybe hard to understand why a mispel would be such a big deal but let's very briefly mention how actions are used. 
+
+### Dispatching an action
```
timdeschryver: ~~Should it read better if this would be before `Introducing NGRX and the need for type safety`?~~

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+  return [...state, action.payload]
+</code-example>
+
+The end result would be that our `products` store property would now contain one more product.
+
+### When an action fails
+This all seems good right. Set up a property, associate to a reducer, dispatch an action and our state changes. All is good or? All is good UNTIL you mispel your `action.type`. Consider the following code below:
+
+<code-example header="Reducer example">
+function reducer(state = 0, action) {
```
timdeschryver: I think it would  be better to continue using the products reducer here.

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+  }
+}
+// this WON'T work as intended 
+reducer({ type: 'INREMENT' })
+
+// this WILL work, cause it's correctly spelled
+reducer({ type: 'INCREMENT' })
+</code-example>
+
+Above we are showcasing how actions are used, namely that they are objects being sent to a reducer functions that lets them jump through a `switch` construct. We are also demonstrating how the spelling of your action matters. If we spell it `INREMENT` it wont hit the intended case but if we correctly spell it `INCREMENT` it will hit the correct case.
```
timdeschryver: ~~It's also the payload that could be misspelled, or using the "wrong" payload. Don't know if it's worth adding or not...~~

This is added below 😄

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+The interface is made up of the following:
+
+<code-example header="Action Interface">
+interface Action {
+  type: string;
+}
+</code-example>
+
+The interface has a single property, the `type`, represented as a string. At it's core that's what an action is, an object containing a `type`. The reason the `payload` isn't a property on the `Action` interface is that it's considered *additional context* and is not mandatory. We can definitely add a `payload` property to the class implementing the above interface though.
```
timdeschryver: >  We can definitely add a `payload` property to the class implementing the above interface though.

Is worth adding? Considering the new `createAction` won't be a class.

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+<code-example header="login-page.actions.ts">
+import { Action } from '@ngrx/store';
+
+export class Login implements Action {
+  readonly type = '[Login Page] Login';
+
+  constructor(public payload: { username: string; password: string }) {}
+}
+</code-example>
+As you can see above we implemented the `Action` interface by creating the class `Login`. In doing so we also used the `readonly` keyword on the `type` field. Thereby we ensured that you can't change the value of `type` after we've instantiated an object from the `Login` class.
```
timdeschryver:
```suggestion

As you can see above we implemented the `Action` interface by creating the class `Login`. In doing so we also used the `readonly` keyword on the `type` field. Thereby we ensured that you can't change the value of `type` after we've instantiated an object from the `Login` class. 
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+<code-example header="Action in plain JavaScript">
+class Login {
+  constructor(payload) {
+    this.type = '[Login Page] Login';
+    this.payload = payload;
+  }
+}
+</code-example>
+As you can see above we lose quite a bit of features when using plain JavaScript like:
```
timdeschryver:
```suggestion

As you can see above we lose quite a bit of features when using plain JavaScript like: 
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+export class Login implements Action {
+  readonly type = '[Login Page] Login';
+
+  constructor(public payload: { username: string; password: string }) {}
+}
+</code-example>
+As you can see above we implemented the `Action` interface by creating the class `Login`. In doing so we also used the `readonly` keyword on the `type` field. Thereby we ensured that you can't change the value of `type` after we've instantiated an object from the `Login` class. 
+
+We did another thing as well, namely adding the `payload` to the constructor of the `Login` class and we used the keyword `public` to do it. Using an accessor like `public`, `private` or `protected` in a constructor is a nice short hand for creating the field as well as pass it in as a parameter into the constructor. The below code in plain Javascript looks like the following:
```
timdeschryver:
```suggestion
We did another thing as well, namely adding the `payload` to the constructor of the `Login` class and we used the keyword `public` to do it. Using an accessor like `public`, `private` or `protected` in a constructor is a nice short hand for creating the field as well as pass it in as a parameter into the constructor. The below code in plain JavaScript looks like the following:
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+    this.payload = payload;
+  }
+}
+</code-example>
+As you can see above we lose quite a bit of features when using plain JavaScript like: 
+- having fields like `payload` created for us via the constructor, 
+- ensuring we can't change the `type` value after instantiation. 
+- we would also need to perform checks on the `payload` being passed in to ensure it contains a `username` and `password`. 
+
+At this point I hope you want to return to the nice and typesafe world of Typescript. Lets talk about Action creation next.
```
timdeschryver:
```suggestion
At this point I hope you want to return to the nice and type safe world of TypeScript. Lets talk about Action creation next. 
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+### Creating an Action
+Ok, time has come to instantiate an object from our `Login` class and thereby create an action. One of the common places to create an action is in the context of being inside of a component and responding to a user event. So imagine we are inside of `login-page.component.ts` and the user has just entered their `username` and `password` and clicked on a button to login. At this point the `clicked()` method is invoked on our component. The code in the component looks as follows:
+
+<code-example header="login-page.component.ts">
+click(username: string, password: string) {
+  store.dispatch(new Login({ username: username, password: password }));
+}
+</code-example>
+We do two things here:
```
timdeschryver:
```suggestion

We do two things here:
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+- instantiate an action and pass in an object literal containing our username and password
+- dispatch the action by calling `store.dispatch()`
+
+When we instantiate our action `Login` we see that we pass an object literal. Wasn't that what we were trying to get away from by using Typescript? Actually the compiler helps us here because we declare that the constructor will accept a parameter `payload` that requires a `username` and `password` as properties. Let's show that part again from the definition of our `Login` action:
```
timdeschryver:
```suggestion
When we instantiate our action `Login` we see that we pass an object literal. Wasn't that what we were trying to get away from by using TypeScript? Actually the compiler helps us here because we declare that the constructor will accept a parameter `payload` that requires a `username` and `password` as properties. Let's show that part again from the definition of our `Login` action:
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+<code-example header="excerpt from Login action">
+  constructor(public payload: { username: string; password: string }) {}
+</code-example>
+As you can see Typescript saves us many times in our `click()` method in `login-component.ts`, it saves us by us instantiating `Login` over create an object literal like this `{ type: '[Login Page] Login', payload: { username, password } }`, we know how easy it is to mispel a string. We are also saved by defining a shape of the `payload` field to ensure our `password` and `username` properties are set.
```
timdeschryver:
```suggestion

As you can see, TypeScript saves us many times in our `click()` method in `login-component.ts`, it saves us by us instantiating `Login` over creating an object literal like this `{ type: '[Login Page] Login', payload: { username, password } }`. We already know how easy it is to mispel a string but we are also saved by defining a shape of the `payload` field to ensure our `password` and `username` properties are set. 
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+import { Action } from '@ngrx/store';
+
+export class Login implements Action {
+  readonly type = '[Login Page] Login';
+
+  constructor(public payload: { username: string; password: string }) {}
+}
+</code-example>
+
+We've discussed previously how the above code, using Typescript, provides a lot of benefits over creating Actions using plain JavaScript and object literals or even using classes. We can take this one step further though, by using enums. Right now our `type` is a string. Let's first create an enum like so:
```
timdeschryver:
```suggestion
We've discussed previously how the above code, using TypeScript, provides a lot of benefits over creating Actions using plain JavaScript and object literals or even using classes. We can take this one step further though, by using enums. Right now our `type` is a string. Let's first create an enum like so:
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+<code-example header="An ActionTypes enum">
+export enum ActionTypes {
+  Login = '[Login Page] Login',
+}
+</code-example>
+As you can see above we've moved our `type` value into the `ActionTypes` enum and now we can perform the next step which is to replace our `type` value with our enum value:
```
timdeschryver:
```suggestion

As you can see above we've moved our `type` value into the `ActionTypes` enum and now we can perform the next step which is to replace our `type` value with our enum value:
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+<code-example header="Login, Logout reducer">
+function reducer(state, action) {
+  switch(action.type) {
+    // define code here
+  }
+}
+</code-example>
+
+As you can see above we have no typesafety. We could be feeding this `reducer()` function any kind of parameters. So lets fix that shall we ?
```
timdeschryver:
```suggestion
As you can see above we have no type safety. We could be feeding this `reducer()` function any kind of parameters. So lets fix that shall we?
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+The first thing we are going to do is add something called a union type. Union types works in the following way:
+
+<code-example header="Example union">
+type Cars = 'Ferrari' | 'Wolkswagen' | 'Volvo';
```
timdeschryver:
```suggestion
type Cars = 'Ferrari' | 'Volkswagen' | 'Volvo';
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+function car(car: Cars) {
+  console.log(car);
+}
+
+car('Ferrari'); // OK
+car('Porsche'); // Not OK
+car('Volvo') // OK
+</code-example>
+
+As you can see from above code we are defining the type `Cars` and what we are doing is adding a number of values that it can assume. Each value it can assume are separated by a pipe, `|`. The example invocation of the method `car()` tells us that `Ferrari` and `Volvo` are ok to add but not `Porsche` as it's not part of the definition of our union type.
```
timdeschryver:
```suggestion
As you can see from the above code we are defining the type `Cars` and what we are doing is adding a number of values that it can assume. Each value it can assume are separated by a pipe, `|`. The example invocation of the method `car()` tells us that `Ferrari` and `Volvo` are ok to add but not `Porsche` as it's not part of the definition of our union type.
```

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+export class Login implements Action {
+  readonly type = ActionTypes.Login;
+
+  constructor(public payload: { username: string; password: string }) {}
+}
+</code-example>
+
+Above we've added the `Logout` type to our `ActionTypes` enum and we've also created our `Logout` action class. As you can see everything that goes together can be using the same enum.
+
+### Union types
```
timdeschryver: Should we point to the TS docs here?

projects/ngrx.io/content/guide/store/architecture.md
softchris: good point, I'll make that more clear..

projects/ngrx.io/content/guide/store/architecture.md
softchris: not sure... I mean ultimately it becomes an object { type: 'message', payload: 'something' }, whether you instantiate a class to accomplish it or createAction()... just so I understand is the Action interface disappearing?

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+
+## Redux
+As was stated earlier Redux is variant of a Pub Sub pattern and it uses certain constructs to make sure we can listen to data from one place and also change the data in a predictable way. 
+
+### Basic concepts
+For Redux to work properly there are some players in there that needs to exist, those are:
+
+- The store, this is the single source of truth, the one place where your data lives, your state. You read from it and if you want to make changes, this is the data that ultimately gets changed. 
+- Reducers, these are functions, guards if you will that ensures that changes happen to the store in a predictable way. Changes to the store can only happen through a reducer. 
+- Actions, these are messages that carry an intent and sometimes a payload. Typical intents are 'Adding a product', 'Login', 'Remove customer'. These are things you want to do. The payload is what we need to change the data to. In the case of 'Adding a product', the payload would be the product we would want to add. Actions gets *dispatched* to a Reducer that processes the action and the end result of that processing is a change to the state of the store
```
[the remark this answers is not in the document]
softchris: is it better that I write something like.. the payload contains extra information that we need to carry out our action. Sometimes it contains an `id` so we know what product to remove, sometimes it's the product itself that we need to add. The point is, the content of the payload differs depending on how it is used. Additionally, think about it this way, if the message type alone doesn't help you to carry out your action correctly, then you need to add extra meta information to the payload to see to it that the action can succeed with its intended change?

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
 # Architecture
+
+NGRX is based on the architectural pattern Redux. Redux in turn is based upon the general pattern Pub-Sub or Publish, Subscribe. This means we send a message to a central data store and a number of listeners on that data store are being told there is a new message.
```
wesleygrimes: As you read my reviews on grammar, you will find that I am big on brevity. These are just suggestions, but the end result should read a little better.

- Consider changing `are being told there is` to `are notified that there is`

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
 # Architecture
+
+NGRX is based on the architectural pattern Redux. Redux in turn is based upon the general pattern Pub-Sub or Publish, Subscribe. This means we send a message to a central data store and a number of listeners on that data store are being told there is a new message. 
+
+A very typical usage for this is a Chat application. In a Chat application the people in the chat room can communicate with specific participants or they can broadcast what they say to the entire chat room. It's messages being sent, some messages meant for specific people or for a group.
```
wesleygrimes: - Consider changing `A very typical usage` to `The Redux pattern is a great fit for a chat application.`.
- Missing comma after `In a Chat Application`
- Consider removing `what they say` from `they can broadcast what they say to the entire chat room`.
- `It's messages being sent, some messages meant for specific people or for a group` - Looks like there are some missing words here, should this be `Messages can be sent to either specific people or the entire group`?

projects/ngrx.io/content/guide/store/architecture.md
```diff
@@ -1 +1,418 @@
...
+{
+  "language": "english",
+  "products": [{ "id": 1, "name": "movie" }]
+}
+```
+If this is our state above that means the application is showing all its text in `english` and somewhere we are rendering a list of products. Lets say now the user changes the language in the applciation from `english` from `french`. How would we change this? 
+
+#### Changing our state in the store
+
+This is where we create an action representing our intent and because we want to change to something specific we need to have a payload with the value `french`. An action is nothing more than an object looking like this"
```
wesleygrimes: - Consider changing  `and because we want to change to something specific` to `and because our intent is to switch to a specific language`

Issue #1634, opened 2019-03-18
Store: improve action creators


alex-okrushko:
## Problem
1) _Actions_ are consumed by **STRING** or **ENUM** types, however created by **class/function** which makes them **hard to track through IDEs** or other search tools. At some point our team resorted to creating the visualizer to try to tackle that problem. "Good action hygiene" helps, but only in devtools.

```ts
// component.ts
this.store.dispatch(new MyAction('blah')); // uses Class symbol

// effect.ts
ofType(MY_ACTION_TYPE) // uses just string or enum
map(() => new MyOtherAction()) // uses Class symbol

// reducer.ts
case MY_ACTION_TYPE: // uses string or enum
```

There has been a number of takes on this already (https://github.com/ngrx/platform/issues/584, current function-base action creator https://github.com/ngrx/platform/issues/1480).

2) Class-based creators is the most used way to create Actions within Google's codebase (to the point where I couldn't find even a single place where function-based are used) and having const string `MY_ACTION_TYPE` (instead of type enum) to accompany the Class itself is also something that is frequently done.
However, recently new linter rule was added that requires any **exported** symbols to have the comment. This ruled turned any **Action** into the following:

```ts
/** comment is required for exported const */
export const MY_ACTION_TYPE = 'type1';
/** comment is required as well, which would duplicate the type comment, 
  * which is really annoying */
export class MyAction implements Action {
  readonly type = MY_ACTION_TYPE;
  
  constructor(readonly x: number) {}  
}
```

## Describe any alternatives/workarounds you're currently using

None, however there were proposals to duplicate `readonly type` with `static type` on the same class.

## Proposal

To address both of the issues I suggest to create function returning Class that would "enhance" the action creator.

```ts

/** comment is required for exported classes */
export class MyAction extends typedAction('type1') { // <--- do this instead of readonly type
  constructor(readonly x: number) {
    super(); // this is the only annoying part. Needed when Action constructor has parameters.
  }
}
```

`typedAction` here would create a class and add both `readonly type` and `static type` to the class.
Suggestions about the naming of that function are very welcomed. I heard `withAction`, however not sure if that provides enough context about what it does.

New usage would be
```ts
// component.ts
this.store.dispatch(new MyAction('blah')); // uses Class symbol

// effect.ts
ofType(MyAction.type) // uses Class symbol with static type prop
map(() => new MyOtherAction()) // uses Class symbol

// reducer.ts
case MyAction.type: // uses Class symbol with static type prop
```

[...]

PR #1654, opened 2019-03-28
feat(store, example): add action creator

alex-okrushko:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes # 

https://github.com/ngrx/platform/issues/1634

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/action_creator.spec.ts
+++ b/modules/store/spec/action_creator.spec.ts
@@ -1,0 +1,136 @@
+import { createAction, props, union } from '@ngrx/store';
+import { expecter } from 'ts-snippet';
+
+describe('Action Creators', () => {
+  let originalTimeout: number;
+  beforeEach(() => {
+    originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL;
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 2000;
+  });
+
+  afterEach(() => {
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = originalTimeout;
+  });
+
+  const expectSnippet = expecter(
+    code => `
+  // path goes from root
+  import {createAction, props, union} from './modules/store/src/action_creator';
+    ${code}`,
+    {
+      moduleResolution: 'node',
+      target: 'es2015',
+    }
+  );
+
+  describe('createAction', () => {
+    it('should create an action', () => {
+      const foo = createAction('FOO', (foo: number) => ({ foo }));
+      const fooAction = foo(42);
+      expect(fooAction).toEqual({ type: 'FOO', foo: 42 });
+    });
+
+    it('should narrow the action', () => {
+      const foo = createAction('FOO', (foo: number) => ({ foo }));
+      const bar = createAction('BAR', (bar: number) => ({ bar }));
+      const both = union({ foo, bar });
+      const narrow = (action: typeof both) => {
+        if (action.type === foo.type) {
+          expect(action.foo).toEqual(42);
+        } else {
+          throw new Error('Should not get here.');
+        }
+      };
+      narrow(foo(42));
+    });
+
+    it('should be serializable', () => {
+      const foo = createAction('FOO', (foo: number) => ({ foo }));
+      const fooAction = foo(42);
+      const text = JSON.stringify(fooAction);
+      expect(JSON.parse(text)).toEqual({ type: 'FOO', foo: 42 });
+    });
+
+    it('should enforce ctor parameters', () => {
+      expectSnippet(`
+        const foo = createAction('FOO', (foo: number) => ({ foo }));
+        const fooAction = foo('42');
+    `).toFail(/not assignable to parameter of type 'number'/);
+    });
+
+    it('should enforce action property types', () => {
+      expectSnippet(`
+          const foo = createAction('FOO', (foo: number) => ({ foo }));
+          const fooAction = foo(42);
+          const value: string = fooAction.foo;
+      `).toFail(/'number' is not assignable to type 'string'/);
+    });
+
+    it('should enforce action property names', () => {
+      expectSnippet(`
+          const foo = createAction('FOO', (foo: number) => ({ foo }));
+          const fooAction = foo(42);
+          const value = fooAction.bar;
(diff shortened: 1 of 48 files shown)


modules/store/spec/action_creator.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    ${code}`,
+    {
+      moduleResolution: 'node',
+      target: 'es2015',
+    }
+  );
+
+  describe('createAction', () => {
+    it('should create an action', () => {
+      const foo = createAction('FOO', (foo: number) => ({ foo }));
```
brandonroberts: The tests need a line break to separate setup, doing, and assertion.

projects/example-app/src/app/auth/actions/auth-api.actions.ts
```diff
@@ -1,26 +1,17 @@
...
-
-  constructor(public payload: { error: any }) {}
-}
-
-export class LoginRedirect implements Action {
-  readonly type = AuthApiActionTypes.LoginRedirect;
-}
-
-export type AuthApiActionsUnion = LoginSuccess | LoginFailure | LoginRedirect;
+const all = union({ loginSuccess, loginFailure, loginRedirect });
```
brandonroberts: Why not combine these two lines into one?

projects/example-app/src/app/books/reducers/books.reducer.spec.ts
```diff
@@ -1,4 +1,5 @@
 import { reducer } from '@example-app/books/reducers/books.reducer';
+import { union } from '@ngrx/store';
```
brandonroberts: unused import

projects/example-app/src/app/auth/reducers/auth.reducer.ts
```diff
@@ -14,14 +14,14 @@ export function reducer(
   action: AuthApiActions.AuthApiActionsUnion | AuthActions.AuthActionsUnion
 ): State {
   switch (action.type) {
-    case AuthApiActions.AuthApiActionTypes.LoginSuccess: {
+    case AuthApiActions.loginSuccess.type: {
```
brandonroberts: Already discussed, so just leaving as a note that `AuthApiActions.loginSuccess` for the case would be preferred but we need the action type.

projects/example-app/src/app/auth/actions/auth-api.actions.ts
alex-okrushko: Unfortunately, because of https://github.com/Microsoft/TypeScript/issues/6606 they cannot be combined (at least I didn't find an easy way to do it).

Alternatively, instead of 
```ts
 const all = union({ loginSuccess, loginFailure, loginRedirect });
export type AuthApiActionsUnion = typeof all;
```
we can do the following
```
// drop the union(...) altogether
export type AuthApiActionsUnion = ReturnType<
  typeof loginSuccess | typeof loginFailure | typeof loginRedirect
>;
```
This is basically what `union` fn does anyway.
 
WDYT?

projects/example-app/src/app/books/reducers/books.reducer.spec.ts
alex-okrushko: Done

projects/example-app/src/app/auth/actions/auth-api.actions.ts
alex-okrushko: The second alternative would be to export the `union` itself and use `typeof AuthApiActionsUnion` instead of `AuthApiActionsUnion`.
That's the best alternative from my point of view, but it might introduce some friction

modules/store/spec/action_creator.spec.ts
alex-okrushko: Not sure what you mean here. I added the few line breaks where I think they should be, but the rest mostly look right to me :)

projects/example-app/src/app/auth/actions/auth-api.actions.ts
brandonroberts: I'd prefer less friction with the union approach you already had

modules/store/spec/action_creator.spec.ts
brandonroberts: I'd prefer

```
      const foo = createAction('FOO', (foo: number) => ({ foo }));

      const fooAction = foo(42);

      expect(fooAction).toEqual({ type: 'FOO', foo: 42 });
```

Not something we specifically need to do now though

projects/example-app/src/app/auth/actions/auth-api.actions.ts
timdeschryver: Hmm, what about the following?

```ts
export const CollectionApiActions = {
  addBookSuccess: createAction(
    '[Collection/API] Add Book Success',
    props<{ book: Book }>()
  ),
  addBookFailure: createAction('[Collection/API] Add Book Failure'),
  loadBooksFailure,
};

const all = union(CollectionApiActions);
export type CollectionApiActionsUnion = typeof all;
```

At the barrel export in `index.ts` it would be:
```diff
-- import * as CollectionApiActions from './collection-api.actions';
++ export * from './collection-api.actions';
```

modules/store/spec/action_creator.spec.ts
alex-okrushko: Done

projects/example-app/src/app/auth/actions/auth-api.actions.ts
alex-okrushko: Personally, I don't see the need to wrap action creators with an object - that does feel like it saves us much :)
I typically do module import anyway and we don't use the barrel (aka entry-point) modules - they are OK for libraries, but for regular project they just add overhead without clear benefit, IMHO.

alex-okrushko: What I ended up doing is use the `union` function when there are 2+ action creators (except a single case, where I added the comments) within a file and just do the `ReturnType<typeof foo>` when there's only one.

PR #1666, opened 2019-03-29
fix(schematics): avoid lastIndexOf error while creating store schematics (#1659)

### LINKED ISSUE #1659: Schematics-Store: lastIndexOf
I try to check result and forgot write a store name:

```
$ ng g store -m modules/root-store -d
```

**Result:**
```
Cannot read property 'lastIndexOf' of undefined
```
image

there is no check if [options.path exist](https://github.com/ngrx/platform/blob/master/modules/schematics/src/store/index.ts#L136)

EnricoVogt:
fix(schematics): avoid lastIndexOf error while creating store schematics (#1659)

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1659

## What is the new behavior?

Now its possible to create a store schematic without a name if its called with the "--root" flag, otherwise a name is necessary.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/schematics/src/store/index.spec.ts
+++ b/modules/schematics/src/store/index.spec.ts
@@ -215,4 +215,27 @@ describe('Store Schematic', () => {
     );
     expect(content).toMatch(/export interface FeatureState {/);
   });
+
+  it('should fail if a feature state name is not specified', () => {
+    const options = {
+      ...defaultOptions,
+      name: undefined,
+      root: false,
+    };
+
+    expect(() => {
+      schematicRunner.runSchematic('store', options, appTree);
+    }).toThrowError('Please provide a name for the feature state');
+  });
+
+  it('should pass if a root state name is not specified', () => {
+    const options = {
+      ...defaultOptions,
+      name: undefined,
+    };
+
+    expect(() => {
+      schematicRunner.runSchematic('store', options, appTree);
+    }).not.toThrow();
+  });
 });
--- a/modules/schematics/src/store/index.ts
+++ b/modules/schematics/src/store/index.ts
@@ -131,9 +131,13 @@ function addImportToNgModule(options: StoreOptions): Rule {
 
 export default function(options: StoreOptions): Rule {
   return (host: Tree, context: SchematicContext) => {
+    if (!options.name && !options.root) {
+      throw new Error(`Please provide a name for the feature state`);
+    }
+
     options.path = getProjectPath(host, options);
 
-    const parsedPath = parseName(options.path, options.name);
+    const parsedPath = parseName(options.path, options.name || '');
     options.name = parsedPath.name;
     options.path = parsedPath.path;
 


King-i:
My 2 cents is if `name` is a required parameter then it may be better for it to be specified in `schema.json` under `required` which will prevent a user being able to run the schematic as stated in the schematics specifications, checking if the value was provided but having it declared as a required type in `schema.ts` defeats the purpose of the type system and the configuration via schema, if unit tests are required to confirm that `schema.json` matches `schema.ts` then that can be done via a AST read and compare?

EnricoVogt:
I think the name parameter should not be required. If you create a rootstate (--root) the name is definitely not used. For my part i wont write something that is not neccesarry.

I have an open mind for other solutions.
I will wait for a advice how to solve it.

@brandonroberts 
@timdeschryver

King-i:
If that's the case then no it shouldn't be set as required, however the definition says its required, so therefore the schema interface needs to be updated to have `name` set as optional, which may have caught this bug in the original unit tests in the first place.

modules/schematics/schematics-core/utility/parse-name.ts
```diff
@@ -5,7 +5,7 @@ export interface Location {
   path: Path;
 }
 
-export function parseName(path: string, name: string): Location {
+export function parseName(path: string, name: string = ''): Location {
```
timdeschryver: I think we would want to scope this change specific to the store shematics.
To do this, we should default to an empty string at https://github.com/ngrx/platform/blob/3d0642867d6f4443a3fcfce194bc1112546ededd/modules/schematics/src/store/index.ts#L134

EnricoVogt: I will change this.

EnricoVogt: @timdeschryver i changed it now

modules/schematics/src/store/index.spec.ts
```diff
@@ -215,4 +215,31 @@ describe('Store Schematic', () => {
...
+    expect(thrownError).toBeDefined();
+  });
+
+  it('should pass if a root state name is not specified', () => {
+    const options = {
+      ...defaultOptions,
+      name: undefined,
+    };
+
+    expect(function() {
```
timdeschryver:
```suggestion
    expect(() => {
```

modules/schematics/src/store/index.spec.ts
```diff
@@ -215,4 +215,31 @@ describe('Store Schematic', () => {
...
+      root: false,
+    };
+
+    let thrownError: Error | null = null;
+    try {
+      schematicRunner.runSchematic('store', options, appTree);
+    } catch (err) {
+      thrownError = err;
+    }
+    expect(thrownError).toBeDefined();
```
timdeschryver:
```suggestion
    expect(() => {
      schematicRunner.runSchematic('store', options, appTree);
    }).toThrowError('Please provide a name for the feature state');
```

modules/schematics/src/store/index.spec.ts
EnricoVogt: Done.

modules/schematics/src/store/index.spec.ts
EnricoVogt: Did this for my Test now. There is another Test "should fail if specified module does not exist" which use also the `try catch` - should i change this also?

timdeschryver: Not in the PR, you can open up a new PR if you would like to

PR #1676, opened 2019-03-31
feat(effects): allow ofType to handle ActionCreator

alex-okrushko:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

## What is the new behavior?

Allows to filter Actions with `ofType(actionCreator)` as well as maintains previous behavior `ofType(actionCreator.type)`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/src/actions.ts
+++ b/modules/effects/src/actions.ts
@@ -1,5 +1,10 @@
 import { Inject, Injectable } from '@angular/core';
-import { Action, ScannedActionsSubject } from '@ngrx/store';
+import {
+  Action,
+  ActionCreator,
+  Creator,
+  ScannedActionsSubject,
+} from '@ngrx/store';
 import { Observable, OperatorFunction, Operator } from 'rxjs';
 import { filter } from 'rxjs/operators';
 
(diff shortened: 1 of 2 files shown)


modules/effects/src/actions.ts
```diff
@@ -87,12 +108,27 @@ export function ofType<
...
 ): OperatorFunction<Action, V>;
 export function ofType(
-  ...allowedTypes: string[]
+  ...allowedTypes: Array<string | ActionCreator<string, Creator>>
 ): OperatorFunction<Action, Action> {
-  return filter((action: Action) =>
-    allowedTypes.some(type => type === action.type)
-  );
+  return filter((action: Action) => {
+    for (let typeOrActionCreator of allowedTypes) {
```
timdeschryver: What about checking on action first and next on the action creator type?

```ts
return allowedTypes.some(type => {
  // Comparing the string to type
  if (typeof type === 'string') {
    return type === action.type;
  }

  // We are filtering by ActionCreator
  return type && type.type === action.type;
});
```

alex-okrushko: You are right, reads much better. Also I think I reimplemented `some` in my previous solution 🤦‍♂️

PR #1704, opened 2019-04-05
docs: update usage of actions to use action creators

### LINKED ISSUE #1671: Docs: add docs for action creators
Follow up to #1480 

1. Updates the existing documentation to use new action creators feature
2. Adds a section for current usage of action classes

brandonroberts:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1671 

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngrx.io/content/examples/store/src/app/counter.actions.ts
+++ b/projects/ngrx.io/content/examples/store/src/app/counter.actions.ts
@@ -1,20 +1,6 @@
 // #docregion
-import { Action } from '@ngrx/store';
+import { createAction } from '@ngrx/store';
 
-export enum ActionTypes {
-  Increment = '[Counter Component] Increment',
-  Decrement = '[Counter Component] Decrement',
-  Reset = '[Counter Component] Reset',
-}
-
-export class Increment implements Action {
-  readonly type = ActionTypes.Increment;
-}
-
-export class Decrement implements Action {
-  readonly type = ActionTypes.Decrement;
-}
-
-export class Reset implements Action {
-  readonly type = ActionTypes.Reset;
-}
+export const increment = createAction('[Counter Component] Increment');
+export const decrement = createAction('[Counter Component] Decrement');
+export const reset = createAction('[Counter Component] Reset');
--- a/projects/ngrx.io/content/guide/effects/testing.md
+++ b/projects/ngrx.io/content/guide/effects/testing.md
@@ -38,8 +38,8 @@ describe('My Effects', () => {
   });
 
   it('should work', () => {
-    const action = new MyActions.ExampleAction();
-    const completion = new MyActions.ExampleActionSuccess();
+    const action = MyActions.exampleAction();
+    const completion = MyActions.exampleActionSuccess();
 
     // Refer to 'Writing Marble Tests' for details on '--a-' syntax
     actions = hot('--a-', { a: action });
@@ -81,10 +81,10 @@ describe('My Effects', () => {
 
   it('should work', () => {
     actions = new ReplaySubject(1);
-    actions.next(new MyActions.ExampleAction());
+    actions.next(MyActions.exampleAction());
 
     effects.someSource$.subscribe(result => {
-      expect(result).toEqual(new MyActions.ExampleActionSuccess());
+      expect(result).toEqual(MyActions.exampleActionSuccess());
     });
   });
 });
--- a/projects/ngrx.io/content/guide/entity/adapter.md
+++ b/projects/ngrx.io/content/guide/entity/adapter.md
@@ -106,115 +106,45 @@ export interface User {
 </code-example>
 
 <code-example header="user.actions.ts">
-import { Action } from '@ngrx/store';
+import { createAction, props, union } from '@ngrx/store';
 import { Update } from '@ngrx/entity';
 
 import { User } from '../models/user.model';
 
-export enum UserActionTypes {
-  LOAD_USERS = '[User] Load Users',
-  ADD_USER = '[User] Add User',
-  UPSERT_USER = '[User] Upsert User',
-  ADD_USERS = '[User] Add Users',
-  UPSERT_USERS = '[User] Upsert Users',
-  UPDATE_USER = '[User] Update User',
-  UPDATE_USERS = '[User] Update Users',
-  MAP_USERS = '[User] Map Users',
-  DELETE_USER = '[User] Delete User',
-  DELETE_USERS = '[User] Delete Users',
-  DELETE_USERS_BY_PREDICATE = '[User] Delete Users By Predicate',
-  CLEAR_USERS = '[User] Clear Users',
-}
+export const loadUsers = createAction('[User/API] Load Users', props&lt;{ users: User[] }&gt;());
+export const addUser = createAction('[User/API] Add User', props&lt;{ user: User }&gt;());
+export const upsertUser = createAction('[User/API] Upsert User', props&lt;{ user: User }&gt;());
+export const addUsers = createAction('[User/API] Add Users', props&lt;{ user: User }&gt;());
+export const upsertUsers = createAction('[User/API] Upsert Users', props&lt;{ users: User[] }&gt;());
+export const updateUser = createAction('[User/API] Update User', props&lt;{ user: Update&lt;User&gt; }&gt;());
+export const updateUsers = createAction('[User/API] Update Users', props&lt;{ users: Update&lt;User&gt;[] }&gt;());
+export const mapUsers = createAction('[User/API] Map Users', props&lt;{ entityMap: EntityMap&lt;User&gt; }&gt;());
+export const deleteUser = createAction('[User/API] Delete User', props&lt;{ id: string }&gt;());
+export const deleteUsers = createAction('[User/API] Delete Users', props&lt;{ id: string[] }&gt;());
+export const deleteUsersByPredicate = createAction('[User/API] Delete Users By Predicate', props&lt;{ predicate: Predicate&lt;User&gt; }&gt;());
+export const clearUsers = createAction('[User/API] Clear Users');
 
-export class LoadUsers implements Action {
-  readonly type = UserActionTypes.LOAD_USERS;
-
-  constructor(public payload: { users: User[] }) {}
-}
-
-export class AddUser implements Action {
-  readonly type = UserActionTypes.ADD_USER;
-
-  constructor(public payload: { user: User }) {}
-}
-
-export class UpsertUser implements Action {
-  readonly type = UserActionTypes.UPSERT_USER;
-
-  constructor(public payload: { user: User }) {}
-}
-
-export class AddUsers implements Action {
-  readonly type = UserActionTypes.ADD_USERS;
-
-  constructor(public payload: { users: User[] }) {}
-}
-
-export class UpsertUsers implements Action {
-  readonly type = UserActionTypes.UPSERT_USERS;
-
-  constructor(public payload: { users: User[] }) {}
-}
-
-export class UpdateUser implements Action {
-  readonly type = UserActionTypes.UPDATE_USER;
-
-  constructor(public payload: { user: Update&lt;User&gt; }) {}
-}
-
-export class UpdateUsers implements Action {
-  readonly type = UserActionTypes.UPDATE_USERS;
-
-  constructor(public payload: { users: Update&lt;User&gt;[] }) {}
-}
-
-export class MapUsers implements Action {
-  readonly type = UserActionTypes.MAP_USERS;
-
-  constructor(public payload: { entityMap: EntityMap&lt;User&gt; }) {}
-}
-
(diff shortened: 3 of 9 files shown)


projects/ngrx.io/content/guide/store/actions.md
```diff
@@ -55,52 +53,58 @@ Following these guidelines helps you follow how these actions flow throughout yo
...
-  constructor(public payload: { username: string; password: string }) {}
-}
+export const login = createAction(
+  '[Login Page] Login',
+  props<{ username: string; password: string }>()
+);
 </code-example>
 
-Actions are written as classes to provide a type-safe way to construct an action when it's being dispatched. The `Login` action implements the `Action` interface to adhere to its structure. The `payload` in this example is an object of a username and password, that is additional metadata needed for the handling of the action.
+The `createAction` function returns a function, that when called returns an object in the shape of the `Action` interface. The `props` method is used to define any additional metadata needed for the handling of the action. Actions creators provide a consistent, type-safe way to construct an action that is being dispatched.
```
timdeschryver: Action creators provide a consistent, type-safe way to construct an action that is being dispatched.

projects/ngrx.io/content/guide/store/reducers.md
```diff
@@ -116,7 +101,7 @@ export function reducer(
...
+In the example above, the reducer is handling 3 actions: `[Scoreboard Page] Home Score`, `[Scoreboard Page] Away Score`, and `[Scoreboard Page] Reset`. Each action is strongly-typed based on the provided `ActionsUnion`. Each action handles the state transition immutably. This means that the state transitions are not modifying the original state, but are returning a new state object using the spread operator. The spread syntax copies the properties from the current state into the object, creating a new reference. This ensures that a new state is produced with each change, preserving the purity of the change. This also promotes referential integrity, guaranteeing that the old reference was discarded when a state change occurred.
```
timdeschryver: What about adding a link to MDN for the spread syntax? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

projects/ngrx.io/content/guide/store/reducers.md
```diff
@@ -20,29 +20,14 @@ and the associated reducer function.
...
-export class IncrementAway implements Action {
-  readonly type = ActionTypes.IncrementAway;
-}
-
-export class Reset implements Action {
-  readonly type = ActionTypes.Reset;
-
-  constructor(public payload: { home: number; away: number }) {}
-}
+export const homeScore = createAction('[Scoreboard Page] Home Score');
```
timdeschryver: No `map` used here, is this intentional?

projects/ngrx.io/content/examples/store/src/app/counter.actions.ts
```diff
@@ -1,20 +1,6 @@
...
-}
-
-export class Decrement implements Action {
-  readonly type = ActionTypes.Decrement;
-}
-
-export class Reset implements Action {
-  readonly type = ActionTypes.Reset;
-}
+export const increment = createAction('[Counter Component] Increment');
```
timdeschryver: Just to be sure, I think we're keeping this as simple as possible.
But here we're not using a `map` to define actions, nor the `union`.

projects/ngrx.io/content/guide/store/testing.md
```diff
@@ -141,13 +140,13 @@ describe('My Component', () => {
...
   it('should dispatch an action to load data when created', () => {
-    const action = new DataActions.LoadData();
+    const action = DataActions.loadData();
 
     expect(store.dispatch).toHaveBeenCalledWith(action);
   });
 
   it('should dispatch an action to refreshing data', () => {
-    const action = new DataActions.RefreshData();
+    const action = new DataActions.refreshData();
```
timdeschryver: const action = DataActions.refreshData();

projects/ngrx.io/content/guide/store/testing.md
```diff
@@ -156,7 +155,7 @@ describe('My Component', () => {
 
   it('should display a list of items after the data is loaded', () => {
     const items = [1, 2, 3];
-    const action = new DataActions.LoadDataSuccess({ items });
+    const action = new DataActions.loadDataSuccess({ items });
```
timdeschryver: const action = DataActions.loadDataSuccess({ items });

projects/ngrx.io/content/guide/effects/testing.md
```diff
@@ -38,8 +38,8 @@ describe('My Effects', () => {
   });
 
   it('should work', () => {
-    const action = new MyActions.ExampleAction();
-    const completion = new MyActions.ExampleActionSuccess();
+    const action = MyActions.exampleAction();
+    const completion = new MyActions.exampleActionSuccess();
```
timdeschryver: const completion = MyActions.exampleActionSuccess();

projects/ngrx.io/content/guide/entity/adapter.md
```diff
@@ -106,115 +106,34 @@ export interface User {
...
-
-export class ClearUsers implements Action {
-  readonly type = UserActionTypes.CLEAR_USERS;
-}
+export const UserActions = {
+  loadUsers: createAction('[User/API] Load Users', props<{ users: User[] }>()),
+  addUser: createAction('[User/API] Add User', props<{ user: User }>()),
+  upsertUser: createAction('[User/API] Upsert User', props<{ user: User }>()),
+  addUsers: createAction('[User/API] Add Users', props<{ user: User }>()),
+  upserUsers: createAction('[User/API] Upsert Users', props<{ users: User[] }>()),
```
timdeschryver: upsertUsers

projects/ngrx.io/content/guide/entity/adapter.md
```diff
@@ -228,53 +147,53 @@ export const initialState: State = adapter.getInitialState({
...
       return adapter.removeOne(action.payload.id, state);
     }
 
-    case UserActionTypes.DELETE_USERS: {
+    case UserActions.deleteUsers.type: {
       return adapter.removeMany(action.payload.ids, state);
     }
 
-    case UserActionTypes.DELETE_USERS_BY_PREDICATE: {
+    case UserActionss.deleteUsersByPredicate.type: {
```
timdeschryver: UserActions

projects/ngrx.io/content/guide/entity/adapter.md
```diff
@@ -228,53 +147,53 @@ export const initialState: State = adapter.getInitialState({
   selectedUserId: null,
 });
 
-export function reducer(state = initialState, action: UserActionsUnion): State {
+export function reducer(state = initialState, action: UserActions): State {
```
timdeschryver: The action's don't have a `payload` anymore

projects/ngrx.io/content/guide/store/reducers.md
brandonroberts: Yes, no map or unions for simplicity

projects/ngrx.io/content/examples/store/src/app/counter.actions.ts
brandonroberts: Same as above

projects/example-app/src/app/books/actions/book.actions.ts
```diff
@@ -1,9 +1,12 @@
...
-);
+export const BookActions = {
+  loadBook: createAction(
+    '[Book Exists Guard] Load Book',
+    props<{ book: Book }>()
+  ),
+};
 
-export type BookActionsUnion = ReturnType<typeof loadBook>;
+const all = union(BookActions);
```
alex-okrushko: For a single Action we could avoid the `union`.

projects/example-app/src/app/auth/actions/auth-api.actions.ts
```diff
@@ -1,20 +1,17 @@
-import { props, createAction } from '@ngrx/store';
+import { createAction, props, union } from '@ngrx/store';
 import { User } from '@example-app/auth/models/user';
 
-export const loginSuccess = createAction(
-  '[Auth/API] Login Success',
-  props<{ user: User }>()
-);
+export const AuthApiActions = {
```
alex-okrushko: I was think to add the docs myself as well 😛
Anyhow, that's not quite how I see new action creators to be used for a few reasons:
1) I think "module import" works well without the need to wrap with extra object.
2) Google's TS style guide says "When providing a structural-based implementation, explicitly include the type at the declaration of the symbol". So we won't be able to follow the pattern.
3) I'm thinking to change the union fn to take an array instead. That's what @cartant did with `ts-action` and it feels like passing a collection makes more sense from semantics perspective.
4) Didn't verify this personally, but I think unused action creators would be easier to "tree-shake" when they are by themselves (instead of part of the object - that too might be already working, but could depend on the code optimizer).

brandonroberts: Good points. That's why I left this as a separate commit.

1. I'm thinking about it more in terms of ergonomics. I think having 1 export to use across reducers and effects rather than TheseActions and TheseActionsUnion is nice. Idea about declaration merging from here https://medium.com/@martin_hotell/improved-redux-type-safety-with-typescript-2-8-2c11a8062575
2. Unless you're talking about schematics, I don't see how this prevents Google from following that pattern, and is common that you build apps in a different style either way 😉. 
3. Ok
4. Ok

alex-okrushko: 1. Even from ergonomics.
    -  Files like projects/example-app/src/app/auth/actions/index.ts become unnecessary (we actually never used them anyway)
    - name the union as `All` and leave action creators as is.
    - Import in the reducer file as
    ```ts
    import * as authApiActions from './path/to/auth-api.actions'

    function reducer(state, action: authApiActions.All) {
      switch(action.type) {
        case authApiActions.login.type: {...}
      }
    }
    ```
    It is still somewhat "1 export". I saw Martin's post. I cannot say that I agree with the final solution.
2. That's true that the style is different, but the goal is to keep it as close as possible, given that the style guide is driven by reasonable suggestions, right?
3. OK
4. OK (markdown won't let me skip these numbers)
5. It moves further from how `createSelector`s are organized (there's not containment object)
6. The **Action Union will hopefully be gone** soon-ish 🥳

timdeschryver: I don't know which I like the most.
I should be able to "refactor" the current app that I'm working on to use `createAction` during this week, and I'll have an opinion then 😄 

I was thinking about a 3rd option, which is don't use `unionize`.
- It isn't needed anymore in `Effects`
- We could use unionize directly in the reducers with only the use actions?
  If we take this a step further and if we would introduce a `createReducer` based on a specific reducer union, could we then check if all actions are handled within the reducer? For example if you would create a TODO reducer and add `CREATE TODO` and `REMOVE TODO`, if you wouldn't handle `REMOVE TODO` it would give errors at compile time.

timdeschryver: I'm kinda liking the thoughts I had here above.
As example, the collection reducer from the example app:

```ts
// define actions handled by this reducer
const actions = union({
  loadCollection,
  loadBooksSuccess,
  addBookSuccess,
  addBookFailure,
  removeBookSuccess,
  removeBookFailure,
});

export function reducer(state = initialState, act: Action): State {
  // type actions as mentioned in #951
  const action = act as typeof actions;

  switch (action.type) {
    case loadCollection.type: {
      return {
        ...state,
        loading: true,
      };
    }

    case loadBooksSuccess.type: {
      return {
        loaded: true,
        loading: false,
        ids: action.books.map(book => book.id),
      };
    }

    case addBookSuccess.type:
    case removeBookFailure.type: {
      if (state.ids.indexOf(action.book.id) > -1) {
        return state;
      }

      return {
        ...state,
        ids: [...state.ids, action.book.id],
      };
    }

    case removeBookSuccess.type:
    case addBookFailure.type: {
      return {
        ...state,
        ids: state.ids.filter(id => id !== action.book.id),
      };
    }

    default: {
      return state;
    }
  }
}
```

brandonroberts: What about going back to the regular const for the action creators and just using `ReturnType` whether there's one or many action creators? Keep the module imports and use the `Union` for reducers.

```ts
export type Union = ReturnType<
  | typeof loadCollection
  | typeof loadBooksSuccess
>;
```

alex-okrushko: I'm fine with that as well. Let's just keep the `union` fn in the codebase as well (readjusted to take a tuple - I'll take care of that).
Again, hopefully the unions won't be needed for too much longer and are their for backwards-compatibility only :)

PR #1719, opened 2019-04-09
feat(entity): add undefined to Dictionary's index signature

alex-okrushko:
## PR Checklist

## PR Type

Index signatures always return value and never return `undefined` 🤯

```ts
export class Dictionary {
  [id: string]: number;
}

const dict = new Dictionary();
dict[100] // <--- type is number
```
https://github.com/Microsoft/TypeScript/issues/13778

Typescript is telling me that I don't need to check because it's of type `number`. In reality it's actually `number | undefined`, and I should check for the `undefined`.

Suggestion in the bug from TS team is to add " `| undefined` at their definition sites".

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

This became particular painful when `MemoizedSelector<State, Result>` started to validate the return type of the selector to make sure it's actually `Result`.

Closes #

## What is the new behavior?

If one relied on `Dictionary<Entity>` to not return `undefined` they would have to adjust the code.

Yes, it is. See 👆

```
[x] Yes
[ ] No
```

### DIFF
--- a/modules/entity/spec/state_selectors.spec.ts
+++ b/modules/entity/spec/state_selectors.spec.ts
@@ -92,7 +92,7 @@ describe('Entity State Selectors', () => {
 
     it('should type single entity from Dictonary as entity type or undefined', () => {
       // MemoizedSelector acts like a type checker
-      const singleEnitity: MemoizedSelector<
+      const singleEntity: MemoizedSelector<
         EntityState<BookModel>,
         BookModel | undefined
       > = createSelector(selectors.selectEntities, enitites => enitites[0]);
--- a/modules/entity/spec/state_selectors.spec.ts
+++ b/modules/entity/spec/state_selectors.spec.ts
@@ -6,6 +6,7 @@ import {
   AnimalFarm,
   TheGreatGatsby,
 } from './fixtures/book';
+import { MemoizedSelector, createSelector } from '@ngrx/store';
 
 describe('Entity State Selectors', () => {
   describe('Composed Selectors', () => {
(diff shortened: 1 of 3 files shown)


modules/entity/spec/state_selectors.spec.ts
```diff
@@ -89,6 +90,14 @@ describe('Entity State Selectors', () => {
       expect(entities).toEqual(state.entities);
     });
 
+    it('should type single entity from Dictonary as entity type or undefined', () => {
```
Edouardbozon:
```suggestion
    it('should type single entity from Dictionary as entity type or undefined', () => {
```

brandonroberts:
What message should be displayed in the breaking change?

alex-okrushko:
> What message should be displayed in the breaking change?

message in the announcement or message in the error?

Dictionary could be producing `undefined` but previous typings were not explicit about it.

larrifax:
I know I'm a bit late to the party here, but I just wanted to chime in and say that this change makes it harder to create selectors that does something more than just picking a specific index out of the dictionary.

As pointed out in the Typescript-issue mentioned in the PR description (microsoft/TypeScript#13778), Typescript isn't smart enough to remove `undefined` from the type when mapping over the Dictionary, which is why the Typescript team hasn't made this change in their base libs. 

This means that `undefined` is included both in the type sent in to the mapping function as well as the return type of ramda's or lodash's `pickBy`, `values`, and I guess also `Object.values` / `Object.entries`.

Seems like we have to choose between two evils at the time being..

alex-okrushko:
> This means that undefined is included both in the type sent in to the mapping function as well as the return type of ramda's or lodash's pickBy, values, and I guess also Object.values / Object.entries.

@larrifax That's correct. In cases like that I either add `filter((value): value is NonNullable<typeof value> => !!value)` or sometime assertions (`value!`) if I'm looping over myself.

PR #1725, opened 2019-04-10
feat(schematics): add support for create effect to schematics

### LINKED ISSUE #1682: Schematics: Add schematics for effect creators
Follow up to #1667  

1. Adds an `effectCreators` option to the `effect`, and `feature` schematics that uses the new effect creator function instead of decorators
2. The `effectCreators` flag will use a different set of files for scaffolding. This keeps our current structure with `files` intact.
3. Adds tests to verify the behavior of the previously mentioned schematics

itayod:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1682 

## What is the new behavior?

add support for `createEffect` function in effect schematics

```
[ ] Yes
[ ] No
```

### DIFF
--- a/modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts.template
+++ b/modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts.template
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { Actions, Effect<% if (feature) { %>, ofType<% } %> } from '@ngrx/effects';
+import { Actions, <%= effectMethod %><% if (feature) { %>, ofType<% } %> } from '@ngrx/effects';
 <% if (feature && api) { %>import { catchError, map, concatMap } from 'rxjs/operators';
 import { EMPTY, of } from 'rxjs';
 import { Load<%= classify(name) %>sFailure, Load<%= classify(name) %>sSuccess, <%= classify(name) %>ActionTypes, <%= classify(name) %>Actions } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';
@@ -12,8 +12,7 @@ import { <%= classify(name) %>ActionTypes, <%= classify(name) %>Actions } from '
 @Injectable()
 export class <%= classify(name) %>Effects {
 <% if (feature && api) { %>
-  @Effect()
-  load<%= classify(name) %>s$ = this.actions$.pipe(
+  <%= effectStart %>
     ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s),
     concatMap(() =>
       /** An EMPTY observable only emits completion. Replace with your own observable API request */
@@ -21,15 +20,17 @@ export class <%= classify(name) %>Effects {
         map(data => new Load<%= classify(name) %>sSuccess({ data })),
         catchError(error => of(new Load<%= classify(name) %>sFailure({ error }))))
     )
-  );<% } %>
+  <%= effectEnd %>
+<% } %>
+
 <% if (feature && !api) { %>
-  @Effect()
-  load<%= classify(name) %>s$ = this.actions$.pipe(
+  <%= effectStart %>
     ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s),
     /** An EMPTY observable only emits completion. Replace with your own observable API request */
     concatMap(() => EMPTY)
-  );
+  <%= effectEnd %>
 <% } %>
+
 <% if (feature) { %>
   constructor(private actions$: Actions<<%= classify(name) %>Actions>) {}
 <% } else { %>
--- a/modules/schematics/src/effect/index.spec.ts
+++ b/modules/schematics/src/effect/index.spec.ts
@@ -27,6 +27,7 @@ describe('Effect Schematic', () => {
     feature: false,
     root: false,
     group: false,
+    effectCreators: false,
   };
 
   const projectPath = getTestProjectPath();
@@ -314,6 +315,44 @@ describe('Effect Schematic', () => {
 
     expect(content).toMatch(
       /constructor\(private actions\$: Actions<FooActions>\) {}/
+    );
+  });
+
+  it('should create an effect using creator function', () => {
+    const options = { ...defaultOptions, effectCreators: true, feature: true };
+
+    const tree = schematicRunner.runSchematic('effect', options, appTree);
+    const content = tree.readContent(
+      `${projectPath}/src/app/foo/foo.effects.ts`
+    );
+    expect(content).toMatch(
+      /import { Actions, createEffect, ofType } from '@ngrx\/effects';/
+    );
+    expect(content).not.toMatch(/@Effect\(\)/);
+    expect(content).toMatch(
+      /loadFoos\$ = createEffect\(\(\) => this.actions\$.pipe\(/
+    );
+  });
+
+  it('should create an api effect using creator function', () => {
+    const options = {
+      ...defaultOptions,
+      effectCreators: true,
+      api: true,
+      feature: true,
+    };
+
+    const tree = schematicRunner.runSchematic('effect', options, appTree);
+    const content = tree.readContent(
+      `${projectPath}/src/app/foo/foo.effects.ts`
+    );
+
+    expect(content).toMatch(
+      /import { Actions, createEffect, ofType } from '@ngrx\/effects';/
+    );
+    expect(content).not.toMatch(/@Effect\(\)/);
+    expect(content).toMatch(
+      /loadFoos\$ = createEffect\(\(\) => this.actions\$.pipe\(/
     );
   });
 });
--- a/modules/schematics/src/effect/index.ts
+++ b/modules/schematics/src/effect/index.ts
@@ -93,6 +93,21 @@ function addImportToNgModule(options: EffectOptions): Rule {
   };
 }
 
+function getEffectMethod(effectCreators?: boolean) {
+  return effectCreators ? 'createEffect' : 'Effect';
+}
+
+function getEffectStart(name: string, effectCreators?: boolean): string {
+  const effectName = stringUtils.classify(name);
+  return effectCreators
+    ? `load${effectName}s$ = createEffect(() => this.actions$.pipe(`
+    : '@Effect()\n' + `  load${effectName}s$ = this.actions$.pipe(`;
+}
+
+function getEffectEnd(effectCreators?: boolean) {
+  return effectCreators ? '));' : ');';
+}
+
 export default function(options: EffectOptions): Rule {
   return (host: Tree, context: SchematicContext) => {
     options.path = getProjectPath(host, options);
@@ -116,6 +131,9 @@ export default function(options: EffectOptions): Rule {
             options.flat ? '' : s,
             options.group ? 'effects' : ''
           ),
+        effectMethod: getEffectMethod(options.effectCreators),
+        effectStart: getEffectStart(options.name, options.effectCreators),
+        effectEnd: getEffectEnd(options.effectCreators),
         ...(options as object),
       } as any),
       move(parsedPath.path),
--- a/modules/schematics/src/effect/schema.json
+++ b/modules/schematics/src/effect/schema.json
@@ -63,6 +63,12 @@
       "description":
         "Specifies if effect has api success and failure actions wired up",
       "aliases": ["a"]
+    },
+    "effectCreators": {
+      "type": "boolean",
+      "default": false,
+      "description": "Specifies whether to use the effect creators function",
+      "aliases": ["ec"]
     }
   },
   "required": []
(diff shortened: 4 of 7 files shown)


modules/schematics/src/effect/files/__name@dasherize@if-flat__/__name@dasherize__.effects.ts.template
```diff
@@ -35,4 +35,22 @@ export class <%= classify(name) %>Effects {
 <% } else { %>
   constructor(private actions$: Actions) {}
 <% } %>
+
```
itayod: If there was an option to extend a template that could be a nice solution here @brandonroberts @timdeschryver do you know if it's possible?

brandonroberts: Not currently

brandonroberts:
We'll also need to account for using action creators with the new effect creators. Instead of using `*ActionTypes`, the action creator is used, and it calls the action creator instead of creating an instance of an action class.

itayod:
> We'll also need to account for using action creators with the new effect creators. Instead of using `*ActionTypes`, the action creator is used, and it calls the action creator instead of creating an instance of an action class.

Hmm I see, I will try to find a proper solution that will consider the action creator option possibility,
I wounder if there is some way of splicing the template... 🤔

itayod:
@brandonroberts I have updated the solution what do you think?

modules/schematics/src/effect/index.spec.ts
```diff
@@ -12,7 +12,7 @@ import {
   defaultAppOptions,
 } from '../../../schematics-core/testing';
 
-describe('Effect Schematic', () => {
+fdescribe('Effect Schematic', () => {
```
timdeschryver:
```suggestion
describe('Effect Schematic', () => {
```

modules/schematics/src/effect/index.ts
```diff
@@ -93,6 +93,16 @@ function addImportToNgModule(options: EffectOptions): Rule {
   };
 }
 
+export function getEffectMethod(
+  name: string,
+  effectCreator: boolean | undefined
```
timdeschryver:
```suggestion
  effectCreator?: boolean
```

modules/schematics/src/effect/index.ts
```diff
@@ -93,6 +93,16 @@ function addImportToNgModule(options: EffectOptions): Rule {
   };
 }
 
+export function getEffectMethod(
```
timdeschryver:
```suggestion
function getEffectMethod(
```

modules/schematics/src/effect/schema.json
```diff
@@ -63,6 +63,12 @@
       "description":
         "Specifies if effect has api success and failure actions wired up",
       "aliases": ["a"]
+    },
+    "effectCreator": {
+      "type": "boolean",
+      "default": false,
+      "description": "Specifies if the effect creation uses 'createEffect'",
+      "aliases": ["ec"]
```
timdeschryver: What abous using `c` as alias and reuse it for `createAction`?

modules/schematics/src/feature/schema.json
```diff
@@ -56,6 +56,12 @@
       "description":
         "Specifies if api success and failure actions, reducer, and effects should be generated as part of this feature.",
       "aliases": ["a"]
+    },
+    "effectCreator": {
+      "type": "boolean",
+      "default": false,
+      "description": "Specifies if the effect creation uses 'createEffect'",
+      "aliases": ["ec"]
```
timdeschryver: Same here

modules/schematics/src/effect/schema.json
timdeschryver: cc @brandonroberts , because I just saw you were using `ac` for the `createAction` schematics

brandonroberts:
@itayod it would be easier to generate parts of the template in the factory code and pass a variable to the template for output. That's probably as close as we'll get to partial templates.

modules/schematics/src/effect/schema.json
itayod: > @itayod it would be easier to generate parts of the template in the factory code and pass a variable to the template for output. That's probably as close as we'll get to partial templates.

I agree that's what I just did 😄

itayod:
> @itayod it would be easier to generate parts of the template in the factory code and pass a variable to the template for output. That's probably as close as we'll get to partial templates.

I agree that's what I just did 😄

modules/schematics/src/effect/schema.json
itayod: wouldn't it make conflicts when generating feature?

timdeschryver: This is indeed true, it would mean using the one of the two would result in both of the creators being used.

brandonroberts:
We can save some line breaks in the template by moving the if cases up. It was too much to type separately.

```
import { Actions, <% if (effectCreator) { %>createEffect<% } else { %>Effect<% } %><% if (feature) { %>, ofType<% } %> } from '@ngrx/effects';
<% if (feature && api) { %>import { catchError, map, concatMap } from 'rxjs/operators';
import { EMPTY, of } from 'rxjs';
import { Load<%= classify(name) %>sFailure, Load<%= classify(name) %>sSuccess, <%= classify(name) %>ActionTypes, <%= classify(name) %>Actions } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';
<% } %>
<% if (feature && !api) { %>import { concatMap } from 'rxjs/operators';
import { EMPTY } from 'rxjs';
import { <%= classify(name) %>ActionTypes, <%= classify(name) %>Actions } from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';
<% } %>

@Injectable()
export class <%= classify(name) %>Effects {<% if (feature && api) { %>
<% if (!effectCreator) { %>@Effect()<% }%>
  <%= effectMethod %>
    ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s),
    concatMap(() =>
      /** An EMPTY observable only emits completion. Replace with your own observable API request */
      EMPTY.pipe(
        map(data => new Load<%= classify(name) %>sSuccess({ data })),
        catchError(error => of(new Load<%= classify(name) %>sFailure({ error }))))
    )
  )<% if (effectCreator) { %>)<% }%>;
<% } %><% if (feature && !api) { %>
  <% if (!effectCreator) { %>@Effect()<% }%>
  <%= effectMethod %>
    ofType(<%= classify(name) %>ActionTypes.Load<%= classify(name) %>s),
    /** An EMPTY observable only emits completion. Replace with your own observable API request */
    concatMap(() => EMPTY)
  )<% if (effectCreator) { %>)<% }%>;
<% } %>
<% if (feature) { %>  constructor(private actions$: Actions<<%= classify(name) %>Actions>) {}<% } else { %>  constructor(private actions$: Actions) {}<% } %>
}

```

modules/schematics/src/effect/schema.json
brandonroberts: I think we should simplify here and just use `--creators` to use the new syntax for functions for actions and effects. Its unlikely that you'll use one without the other, so let's shift people in that direction.

modules/schematics/src/feature/schema.json
brandonroberts: I think we should simplify here and just use `--creators`  in the `feature` schema to use the new syntax for functions for actions and effects. Its unlikely that you'll use one without the other, so let's shift people in that direction.

itayod:
I have updated the code.
just one nit, what do you think about using **ts compiler** instead **regex** for the tests?

modules/schematics/src/effect/index.ts
```diff
@@ -93,6 +93,21 @@ function addImportToNgModule(options: EffectOptions): Rule {
...
 
+function getEffectMethod(effectCreator?: boolean) {
+  return effectCreator ? 'createEffect' : 'Effect';
+}
+
+function getEffectStart(name: string, effectCreator?: boolean): string {
+  const effectName = stringUtils.classify(name);
+  return effectCreator
+    ? `load${effectName}s$ = createEffect(() => this.actions$.pipe(`
+    : '@Effect()\n' + `\tload${effectName}s$ = this.actions$.pipe(`;
```
timdeschryver: The tab here isn't needed, currently it will output:

```ts
  @Effect()
        loadFoos$ = this.actions$.pipe(
    ofType(FooActionTypes.LoadFoos),
    /** An EMPTY observable only emits completion. Replace with your own observable API request */
    concatMap(() => EMPTY)
  );
```

timdeschryver:
> just one nit, what do you think about using ts compiler instead regex for the tests?

Not sure what you mean, how would a ts compiler test look like?

modules/schematics/src/effect/index.ts
itayod: hmm weird when I test it I am getting different result 🤔

modules/schematics/src/effect/schema.json
```diff
@@ -63,6 +63,12 @@
       "description":
         "Specifies if effect has api success and failure actions wired up",
       "aliases": ["a"]
+    },
+    "effectCreator": {
+      "type": "boolean",
+      "default": false,
+      "description": "Specifies if the effect creation uses 'createEffect'",
+      "aliases": ["c"]
```
brandonroberts: This one should alias to `ec`

modules/schematics/src/effect/schema.ts
```diff
@@ -48,4 +48,9 @@ export interface Schema {
    * Specifies if effect has api success and failure actions wired up
    */
   api?: boolean;
+
+  /**
+   * Specifies if the effect creation uses 'createEffect'
```
brandonroberts:
```suggestion
   * Specifies whether to use the effect creators function
```

modules/schematics/src/feature/schema.ts
```diff
@@ -44,4 +44,9 @@ export interface Schema {
    * should be generated as part of this feature.
    */
   api?: boolean;
+
+  /**
+   * Specifies if the effect creation uses 'createEffect'
+   */
+  effectCreator?: boolean;
```
brandonroberts:
```suggestion
  effectCreators?: boolean;
```

modules/schematics/src/effect/schema.ts
```diff
@@ -48,4 +48,9 @@ export interface Schema {
...
+   */
+  effectCreator?: boolean;
```
brandonroberts:
```suggestion
  effectCreators?: boolean;
```

itayod:
> > just one nit, what do you think about using ts compiler instead regex for the tests?
> 
> Not sure what you mean, how would a ts compiler test look like?

Well, I mean use ts compiler to test if the template has the expected result.
I have created a quick function to demonstrate how:

```javascript
function getImportSpecifiersFor(file: ts.Node, importName: string) {
  let imports;
  file.forEachChild((node) => {
    if (ts.isImportDeclaration(node)
        && node.moduleSpecifier.text === importName
        && node.importClause) {
      imports = (node.importClause.namedBindings as NamedImports).elements
          .map((importElem) => importElem.name.text)
    }
  })
}
```

and then the test looks:
```javascript
    const importsFromNgrx = getImportSpecifiersFor(file, '@ngrx/effects');
    expect(importsFromNgrx).toEqual([ 'Actions', 'createEffect', 'ofType' ]);
```

modules/schematics/src/effect/index.ts
itayod: apparently because I use tab as 2 spaces in my pc it looked good...
thanks for your notice!! :)

modules/schematics/src/effect/index.spec.ts
```diff
@@ -316,4 +317,42 @@ describe('Effect Schematic', () => {
       /constructor\(private actions\$: Actions<FooActions>\) {}/
     );
   });
+
+  it('should create an effect using creator function', () => {
+    const options = { ...defaultOptions, effectCreator: true, feature: true };
```
brandonroberts:
```suggestion
    const options = { ...defaultOptions, effectCreators: true, feature: true };
```

modules/schematics/src/effect/index.spec.ts
```diff
@@ -316,4 +317,42 @@ describe('Effect Schematic', () => {
...
+    expect(content).not.toMatch(/@Effect\(\)/);
+    expect(content).toMatch(
+      /loadFoos\$ = createEffect\(\(\) => this.actions\$.pipe\(/
+    );
+  });
+
+  it('should create an api effect using creator function', () => {
+    const options = {
+      ...defaultOptions,
+      effectCreator: true,
```
brandonroberts:
```suggestion
      effectCreators: true,
```

modules/schematics/src/feature/schema.ts
brandonroberts:
```suggestion
  effectCreators?: boolean;
```

Issue #1724, opened 2019-04-10
RFC: Add reducer factory to Store


brandonroberts:
We've discussed this in other places, and was mentioned here with the introduction of action creators.  https://blog.angularindepth.com/ngrx-action-creators-redesigned-d396960e46da

This is to open up the discussion about rationale. We've introduced two functions recently: `createAction` and `createEffect`, both with the purpose of preserving the type-safety of the things they create, while making them easier to use and consume. This lines up with what we already have in `createSelector` and `createEntityAdapter` from Store and Entity respectively. Naturally, next up would be `createReducer` that might make handling creating a reducer to handle state changes somewhat less verbose. 

What we promote today:

```ts
export function reducer(state, action) {
  switch(action.type) {
    case 'someAction1':
      return newState;
    case 'someAction2':
      return newState;
    default:
      return state;
  }
}
```

All current proposals include generating a reducer function using a map, array, etc. This does not include or make use of decorators to create them. Examples

https://github.com/reduxjs/redux-starter-kit/blob/master/src/createReducer.ts
https://github.com/cartant/ts-action/blob/master/source/reducer.ts
https://redux.js.org/recipes/reducing-boilerplate#generating-reducers

I still like using the switch statement. Its less magical, somewhat easier to glance at. That being said, at this point I think we're at least looking at having a `createReducer` function. Upfront I'll say that using it with AOT is gonna be a pain because you can't use generated functions within NgModule metadata without workarounds. I've tested this using `ts-action` and the `createReducer` method. You end up having to wrap the generated reducer in an exported function to be compliant. 

As mentioned, there are AOT workarounds to that approach, but it would add complexity to registering state with generated reducers. Ideally, we want the best of both worlds. Easy to use APIs that provide the same level of type safety and flexibility.

Feel free to comment with other proposals/suggestions.

timdeschryver:
To play the devils advocate here, what benefits would `createReducer` bring us, because the benefits of `createAction` and `createEffect` were known before we started the implementation.

This is my list:

- No default return case needed, resulting in no more "frustrations" when one would forget to handle it
- We could add a check that every (intended) action is being handled within the reducer
- Have the same naming with `createAction`, `createSelector` and `createEffect` (as described in the issue)

alex-okrushko:
my list would include the ability to drop Action unions entirely 🎉

> We could add a check that every (intended) action is being handled within the reducer

[...]

PR #1746, opened 2019-04-13
feat(store): add createReducer function

alex-okrushko:
## PR Checklist

## PR Type

Introduces `createReducer` function, that has the same functionality as `reducer` from `ts-actions` (created by @cartant).

It helps remove Action unions and dramatically reduce the code needed for reducers. Compare:

Screen Shot 2019-04-13 at 1 13 55 PM

However, there are a few caveats that are caused by https://github.com/Microsoft/TypeScript/issues/3755 and https://github.com/Microsoft/TypeScript/issues/7547#issuecomment-218017839 with the latter acknowledging current TS limitation. 

### Sometimes `reducer` functions in `on(...)` have to be explicitly typed with their return type. Let's look at the example:
``` ts
export interface State {
  showSidenav: boolean;
}

const initialState: State = {
  showSidenav: false,
};

export const reducer = createReducer<State>(
  [
    // Explicit 'State' return type for the 'on' functions is needed because
    // otherwise the return types are narrowed to showSidenav: false instead of
    // showSidenav: boolean (that State is asking for).
    on(LayoutActions.closeSidenav, (): State => ({ showSidenav: false })),
    on(LayoutActions.openSidenav, (): State => ({ showSidenav: true })),
  ],
  initialState
);
```
If the `State` in `on(LayoutActions.closeSidenav, (): State` is **not provided** it would narrow the type to `{showSidenav: false}` and `{showSidenav: true}` correspondingly, which would not match the interface of `State` (which is `{showSidenav: boolean}`).
TS highlights this as error, so providing the `State` is required

### Sometimes typos could be missed

Let's look at another example (reduced for brevity):
```ts
export interface State {
  error: string | null;
  pending: boolean;
}

export const initialState: State = {
  error: null,
  pending: false,
};

export const reducer = createReducer<State>(
  [
    on(AuthApiActions.loginFailure, (state, { error }) => ({
      ...state,
      error,
      ending: false, // <--- 'pending' with typo here - missed letter 'p'.
    })),
  ],
  initialState
);
```

Note the `ending: false`, that missed `p`. The original intention was to spread the state and **override** with `pending: false`, however because of the typo `pending` won't be overridden and instead extra `ending` property will be added.
This will not match the `State` interface, but `State` could be extended from this interface with type (because entire `state` of type `State` is spread).

switch-based reducer function catches these types of errors, because providing the return type of the `reducer` function is a norm: `function reducer(state: State = initialState, action: Action): State`.

**Solution:** This unfortunate error can be also caught by explicitly typing the `reducer` within `on`:
`on(AuthApiActions.loginFailure, (state, { error }): State => ({..})),`

For me, that's an acceptable solution. Maybe we should recommend to always type it. I wish the `createSelector<State>` would've caught it, but it's not currently possible due to TS limitations mentioned above.

Note: if the `State` typed could not be extracted from the returned object literal, the error would be displayed property even without explicit return type, e.g.:
```ts
 on(AuthApiActions.loginFailure, (state, { error }) => ({
      // ...state, comment out the spread
      error,
      ending: false, // <--- 'pending' with typo here - missed letter 'p'.
     // Error would be provided for this case, because State interface is not fulfilled.
    })),
``` 

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes # https://github.com/ngrx/platform/issues/1724

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/reducer_creator.ts
+++ b/modules/store/src/reducer_creator.ts
@@ -1,18 +1,15 @@
-import { ActionCreator, ActionType, Action } from './models';
+import { ActionCreator, ActionReducer, ActionType, Action } from './models';
 
 // Return type of the `on` fn.
 export interface On<S> {
-  reducer: Reducer<S>;
+  reducer: ActionReducer<S>;
   types: string[];
 }
 
-export type Reducer<S, A extends Action = Action> = (state: S, action: A) => S;
-
 // Specialized Reducer that is aware of the Action type it needs to handle
-export type OnReducer<S, C extends ActionCreator[]> = Reducer<
-  S,
-  ActionType<C[number]>
->;
+export interface OnReducer<S, C extends ActionCreator[]> {
+  (state: S, action: ActionType<C[number]>): S;
+}
 
 export function on<C1 extends ActionCreator, S>(
   creator1: C1,
@@ -52,8 +49,8 @@ export function on(
 export function createReducer<S>(
   ons: On<S>[],
   initialState: S
-): Reducer<S | undefined> {
-  const map = new Map<string, Reducer<S>>();
+): ActionReducer<S> {
+  const map = new Map<string, ActionReducer<S>>();
   for (let on of ons) {
     for (let type of on.types) {
       map.set(type, on.reducer);
--- a/modules/store/src/reducer_creator.ts
+++ b/modules/store/src/reducer_creator.ts
@@ -1,21 +1,28 @@
 import { ActionCreator, ActionType, Action } from './models';
 
-export type On<S> = (state: S, action: Action) => S;
-export type OnReducer<S, C extends ActionCreator[]> = (
-  state: S,
-  action: ActionType<C[number]>
-) => S;
-export type Reducer<S> = (state: S | undefined, action: Action) => S;
+// Return type of the `on` fn.
+export interface On<S> {
+  reducer: Reducer<S>;
+  types: string[];
+}
+
+export type Reducer<S, A extends Action = Action> = (state: S, action: A) => S;
+
+// Specialized Reducer that is aware of the Action type it needs to handle
+export type OnReducer<S, C extends ActionCreator[]> = Reducer<
+  S,
+  ActionType<C[number]>
+>;
 
 export function on<C1 extends ActionCreator, S>(
   creator1: C1,
   reducer: OnReducer<S, [C1]>
-): { reducer: On<S>; types: string[] };
+): On<S>;
 export function on<C1 extends ActionCreator, C2 extends ActionCreator, S>(
   creator1: C1,
   creator2: C2,
   reducer: OnReducer<S, [C1, C2]>
-): { reducer: On<S>; types: string[] };
+): On<S>;
 export function on<
   C1 extends ActionCreator,
   C2 extends ActionCreator,
@@ -26,11 +33,11 @@ export function on<
   creator2: C2,
   creator3: C3,
   reducer: OnReducer<S, [C1, C2, C3]>
-): { reducer: On<S>; types: string[] };
+): On<S>;
 export function on<S>(
   creator: ActionCreator,
   ...rest: (ActionCreator | OnReducer<S, [ActionCreator]>)[]
-): { reducer: On<S>; types: string[] };
+): On<S>;
 export function on(
   ...args: (ActionCreator | Function)[]
 ): { reducer: Function; types: string[] } {
@@ -43,10 +50,10 @@ export function on(
 }
 
 export function createReducer<S>(
-  ons: { reducer: On<S>; types: string[] }[],
+  ons: On<S>[],
   initialState: S
-): Reducer<S> {
-  const map = new Map<string, On<S>>();
+): Reducer<S | undefined> {
+  const map = new Map<string, Reducer<S>>();
   for (let on of ons) {
     for (let type of on.types) {
       map.set(type, on.reducer);
--- a/modules/store/src/reducer_creator.ts
+++ b/modules/store/src/reducer_creator.ts
@@ -1,0 +1,59 @@
+import { ActionCreator, ActionType, Action } from './models';
+
+export type On<S> = (state: S, action: Action) => S;
+export type OnReducer<S, C extends ActionCreator[]> = (
+  state: S,
+  action: ActionType<C[number]>
+) => S;
+export type Reducer<S> = (state: S | undefined, action: Action) => S;
+
+export function on<C1 extends ActionCreator, S>(
+  creator1: C1,
+  reducer: OnReducer<S, [C1]>
+): { reducer: On<S>; types: string[] };
+export function on<C1 extends ActionCreator, C2 extends ActionCreator, S>(
+  creator1: C1,
+  creator2: C2,
+  reducer: OnReducer<S, [C1, C2]>
+): { reducer: On<S>; types: string[] };
+export function on<
+  C1 extends ActionCreator,
+  C2 extends ActionCreator,
+  C3 extends ActionCreator,
+  S
+>(
+  creator1: C1,
+  creator2: C2,
+  creator3: C3,
+  reducer: OnReducer<S, [C1, C2, C3]>
+): { reducer: On<S>; types: string[] };
+export function on<S>(
+  creator: ActionCreator,
+  ...rest: (ActionCreator | OnReducer<S, [ActionCreator]>)[]
+): { reducer: On<S>; types: string[] };
+export function on(
+  ...args: (ActionCreator | Function)[]
+): { reducer: Function; types: string[] } {
+  const reducer = args.pop() as Function;
+  const types = args.reduce(
+    (result, creator) => [...result, (creator as ActionCreator).type],
+    [] as string[]
+  );
+  return { reducer, types };
+}
+
+export function createReducer<S>(
+  ons: { reducer: On<S>; types: string[] }[],
+  initialState: S
+): Reducer<S> {
+  const map = new Map<string, On<S>>();
+  for (let on of ons) {
+    for (let type of on.types) {
+      map.set(type, on.reducer);
+    }
+  }
+  return function(state: S = initialState, action: Action): S {
+    const reducer = map.get(action.type);
+    return reducer ? reducer(state, action) : state;
+  };
+}
--- a/projects/example-app/src/app/books/reducers/collection.reducer.ts
+++ b/projects/example-app/src/app/books/reducers/collection.reducer.ts
@@ -3,6 +3,7 @@ import {
   CollectionPageActions,
   CollectionApiActions,
 } from '@example-app/books/actions';
+import { createReducer, on } from '@ngrx/store';
 
 export interface State {
   loaded: boolean;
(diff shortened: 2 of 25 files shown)


projects/example-app/src/app/reducers/index.ts
```diff
@@ -16,25 +17,28 @@ import * as fromRouter from '@ngrx/router-store';
...
 /**
  * Our state is composed of a map of action reducer functions.
  * These reducer functions are called with each dispatched action
  * and the current or initial state and return a new immutable state.
  */
-export const reducers: ActionReducerMap<State> = {
-  layout: fromLayout.reducer,
-  router: fromRouter.routerReducer,
-};
+export function reducers(state: State | undefined, action: Action) {
```
brandonroberts: This isn't valid for the `StoreModule.forRoot()` signature that expects `ActionReducerMap<State, Action> | InjectionToken<ActionReducerMap<State, Action>>`. We're currently using the map of reducers and `combineReducers` to create the root reducer internally. 

We'll need some additional tests to verify this works with `StoreModule.forRoot()` also.

modules/store/src/reducer_creator.ts
```diff
@@ -0,0 +1,59 @@
...
+  creator1: C1,
+  creator2: C2,
+  creator3: C3,
+  reducer: OnReducer<S, [C1, C2, C3]>
+): { reducer: On<S>; types: string[] };
+export function on<S>(
+  creator: ActionCreator,
+  ...rest: (ActionCreator | OnReducer<S, [ActionCreator]>)[]
+): { reducer: On<S>; types: string[] };
+export function on(
```
brandonroberts: I'd like to see this function be `when` instead of `on`

brandonroberts:
The creation of reducers is very nice, and as you mentioned, having to add more explicit type checking on each function is unfortunate. Both that and having to wrap the reducers for each feature feels like a step back in ease of use upfront even if you gain a less verbose way of creating a reducer function.

projects/example-app/src/app/reducers/index.ts
alex-okrushko: You are right, it failed to start
```
ERROR in projects/example-app/src/app/app.module.ts(36,25): error TS2345: Argument of type '(state: State | undefined, action: Action) => { layout: State; router: RouterReducerState<BaseRouterStoreState>; }' is not assignable to parameter of type 'ActionReducerMap<State, Action> | InjectionToken<ActionReducerMap<State, Action>>'.
  Type '(state: State | undefined, action: Action) => { layout: State; router: RouterReducerState<BaseRouterStoreState>; }' is missing the following properties from type 'InjectionToken<ActionReducerMap<State, Action>>': _desc, ngInjectableDef
```

I'll see if i can make it take a function as well.

brandonroberts:
@alex-okrushko is this initial set of work ready to merge?

alex-okrushko:
> is this initial set of work ready to merge?

This one uses `InjectionToken` instead of passing `reducers` fn directly
Also, this one has `on` instead of `when` that you recommended. I'm unsure about that naming still. I also considered `case`, that should smoother the transition... by I like the brevity of `on`. Thoughts? @timdeschryver @MikeRyanDev

brandonroberts:
> > is this initial set of work ready to merge?
> 
> This one uses `InjectionToken` instead of passing `reducers` fn directly
> Also, this one has `on` instead of `when` that you recommended. I'm unsure about that naming still. I also considered `case`, that should smoother the transition... by I like the brevity of `on`. Thoughts? @timdeschryver @MikeRyanDev

We can land with the token now, and follow-up with the fn if you prefer.
I'd rather stay away from `case` as a reserved word. I like `when` because it provides more context. We don't refer to actions as "on this occurrence", but more of "when this event happens".

alex-okrushko:
FYI, 
`(): State => ({ showSidenav: false })`
is solved with
`state => ({ showSidenav: false })`

`state` argument is unused (which is unfortunate) but helps with inference.

projects/example-app/src/app/books/reducers/collection.reducer.ts
```diff
@@ -16,54 +17,43 @@ const initialState: State = {
...
-    case CollectionApiActions.addBookSuccess.type:
-    case CollectionApiActions.removeBookFailure.type: {
-      if (state.ids.indexOf(action.book.id) > -1) {
-        return state;
+export const reducer = createReducer<State>(
+  [
+    on(CollectionPageActions.loadCollection, state => ({
+      ...state,
+      loading: true,
+      blha: true,
```
timdeschryver: I think this can be removed? 😄

projects/example-app/src/app/books/reducers/search.reducer.ts
```diff
@@ -17,55 +18,37 @@ const initialState: State = {
...
-    default: {
-      return state;
-    }
-  }
-}
+export const reducer = createReducer<State>(
+  [
+    on(FindBookPageActions.searchBooks, (state, { query }) => {
+      return query === ''
+        ? {
```
timdeschryver: The ternary should be swapped here

projects/example-app/src/app/books/reducers/collection.reducer.ts
```diff
@@ -16,54 +17,43 @@ const initialState: State = {
...
+      loaded: true,
+      loading: false,
+      ids: books.map(book => book.id),
+    })),
+    // Supports handing multiple types of actions
+    on(
+      CollectionApiActions.addBookSuccess,
+      CollectionApiActions.removeBookFailure,
+      (state, { book }) => {
+        if (state.ids.indexOf(book.id) > -1) {
```
timdeschryver: Nit, use `includes` here instead of `indexOf`?

modules/store/src/reducer_creator.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export function createReducer<S>(
+  ons: { reducer: On<S>; types: string[] }[],
+  initialState: S
+): Reducer<S> {
+  const map = new Map<string, On<S>>();
+  for (let on of ons) {
+    for (let type of on.types) {
+      map.set(type, on.reducer);
```
timdeschryver: Not for now, but should we add some tooling here to notify if you would add the same action type twice? Because one could copy paste an action and forget to rename the type.

timdeschryver:
My vote goes to using `on`, mostly because some state machines I've worked with also use `on` as trigger.

timdeschryver:
How would we feel about having the initial state as first argument?

It would allow to skip the array, but perhaps it's useful to have the distinction between the two?
```ts
export const reducer = createReducer<State>(
  initialState,
  on(AuthApiActions.loginSuccess, (state, { user }) => ({ ...state, user })),
  on(AuthActions.logout, () => initialState)
);
```

Implementation:
```ts
export function createReducer<S>(
  initialState: S,
  ...ons: { reducer: On<S>; types: string[] }[]
): Reducer<S> {
  const map = new Map<string, On<S>>();
  for (let on of ons) {
    for (let type of on.types) {
      map.set(type, on.reducer);
    }
  }
  return function(state: S = initialState, action: Action): S {
    const reducer = map.get(action.type);
    return reducer ? reducer(state, action) : state;
  };
}
```

modules/store/src/reducer_creator.ts
```diff
@@ -0,0 +1,59 @@
...
+export type OnReducer<S, C extends ActionCreator[]> = (
+  state: S,
+  action: ActionType<C[number]>
+) => S;
+export type Reducer<S> = (state: S | undefined, action: Action) => S;
+
+export function on<C1 extends ActionCreator, S>(
+  creator1: C1,
+  reducer: OnReducer<S, [C1]>
+): { reducer: On<S>; types: string[] };
```
timdeschryver: Should we create a type for `: { reducer: On<S>; types: string[] };` ? We're using it several times here

alex-okrushko:
> How would we feel about having the initial state as first argument?
I thought about that as well, but the `Array.prototype.reduce` has it as a last argument, so I left it there. Definitely something that we can discuss.

brandonroberts:
As an additional note, you won't be able to  use `createReducer` with a single reducer in a feature such as `StoreModule.forFeature('feat', reducer)` without wrapping it there also.

```ts
export function reducer(state, action) {
  return createReducer(...);
}

StoreModule.forFeature('feat', reducer)
```

More caveats ..

modules/store/src/reducer_creator.ts
alex-okrushko: Very good point!

alex-okrushko:
>  I like when because it provides more context. We don't refer to actions as "on this occurrence", but more of "when this event happens".

We want to treat Actions like Events... In DOM when events listeners use 'on' prefix, so that's another ➕for `on`.

> As an additional note, you won't be able to use createReducer with a single reducer

Will look into that as well.

brandonroberts:
@alex-okrushko sounds like `on` is the better choice there. You don't have to do anything about the `forFeature`. With AOT that's just the way it is right now.

brandonroberts:
LGTM other than Tim's comments

projects/example-app/src/app/books/reducers/collection.reducer.ts
alex-okrushko: lol.
Done.

projects/example-app/src/app/books/reducers/search.reducer.ts
alex-okrushko: How did you notice that?

projects/example-app/src/app/books/reducers/collection.reducer.ts
alex-okrushko: Unrelated to this PR - let's keep them focused.

projects/example-app/src/app/reducers/index.ts
alex-okrushko: Working on it in another PR

modules/store/src/reducer_creator.ts
alex-okrushko: I moved things a bit there... should be better now.

alex-okrushko: I moved things around a bit

modules/store/src/reducer_creator.ts
alex-okrushko: opened https://github.com/ngrx/platform/issues/1758

alex-okrushko:
Should be ready to go 😀

alex-okrushko:
@timdeschryver @brandonroberts 
I'd like to revise this

> How would we feel about having the initial state as first argument? 

I was trying to remove the need to provide the generic for the `createReducer` function - and making the ` initialState` to be the first parameter makes the rest of the `on` functions correctly infer the type `State`.

I'd like to change it to the following now:

```ts
export const reducer = createReducer(
  initialState,
  on(AuthApiActions.loginSuccess, (state, { user }) => ({ ...state, user })),
  on(AuthActions.logout, () => initialState)
);
```

zetsnotdead:
I believe it would be benefitial if 

> As an additional note, you won't be able to use `createReducer` with a single reducer in a feature such as `StoreModule.forFeature('feat', reducer)` without wrapping it there also.
> 
> ```ts
> export function reducer(state, action) {
>   return createReducer(...);
> }
> 
> StoreModule.forFeature('feat', reducer)
> ```

I believe it would be beneficial if this was also included in the docs as an AoT comment.

brandonroberts:
Will you leave a comment about this in #1762?

zetsnotdead:
Will do :)

morenoisidro:
Hi @alex-okrushko ! Would mind giving a little background on why this change was made? When would you suggest using the injection token approach over the standard one?

image


Thanks!

PR #1847, opened 2019-05-09
feat(router-store): add routerState config option

### LINKED ISSUE #1834: `strictActionSerializability` does not work with RouterModule (and custom serializer)

Steps to reproduce:
- create a new angular project and add ngrx store & router modules
- use the router module with a custom serializer (the one from the docs is perfect)
- trigger any router navigation

Due to #1630, `strictActionSerializability` cannot be used with the DefaultRouterStateSerializer, but it should work with a custom one, for example the [one from the docs](https://next.ngrx.io/guide/router-store/configuration).  

Right now one get's an error: _'ERROR Error: "Detected unserializable action at "payload.event""''_

This looks to be triggered in the [isPlainObject method](https://github.com/ngrx/platform/blob/e836719ce6d7659c304878b5fd933169dfc85948/modules/store/src/meta-reducers/utils.ts#L96) that is used in the [getUnserializable method](https://github.com/ngrx/platform/blob/1fbd59cc265fce3e29642a683cbaf7810fbcdfff/modules/store/src/meta-reducers/utils.ts#L1). All the router events extend the base [RouterEvent](https://angular.io/api/router/RouterEvent) class so the prototype check will fail `targetPrototype === Object.prototype `.

[GitHub repo here](https://github.com/adrianfaciu/NgRxRouterModuleRuntimeChecks)

## Expected behavior:

Should be able to use `strictActionSerializability` with RouterModule.

Angular version: 8.0.0-rc.2
NgRx version: 8.0.0-beta.1

[ ] No

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

By default, the router event exists on the router-store action e.g. `NavigationStart` and `RoutesRecognized`. The problem is that these events are not serializable.

Closes #1834

## What is the new behavior?

This changes introduces a new config parameter, `routerState` which has two options : `RouterState.Full` and `RouterState.Minimal`. If set to `Full`, the current behavior remains. When set to `Minimal`, a new router serializer `MinimalRouterStateSerializer` will be used to serialize the router event, it will also only use the navigation id and the url from the router event to append to the `@ngrx/router-store` action.

If a serializer is already provided, it will continue to use the custom serializer.

```
[ ] Yes
[x] No
```


projects/example-app/src/app/app.module.ts
```diff
@@ -43,7 +43,9 @@ import { AppRoutingModule } from '@example-app/app-routing.module';
     /**
      * @ngrx/router-store keeps router state up-to-date in the store.
      */
-    StoreRouterConnectingModule.forRoot(),
+    StoreRouterConnectingModule.forRoot({
+      onlyEventNavigationId: true,
```
timdeschryver: This won't solve the serializability checks, as the `routerState` is not serializable. 
I think we should introduce a `SerializableRouterStateSerializer` router in the example app to follow our own best practices on keeping actions and state serializable. We could even create this router in `@ngrx/router-store`?

modules/router-store/src/router_store_module.ts
```diff
@@ -302,11 +308,15 @@ export class StoreRouterConnectingModule {
   ): void {
     this.trigger = RouterTrigger.ROUTER;
     try {
+      console.log(this.config.onlyEventNavigationId);
```
adrianfaciu: Probably want to remove this :)

projects/example-app/src/app/app.module.ts
brandonroberts: How about a `routerState` configuration flag that let's the user pick?

Full = DefaultRouterStateSerializer (what we have today),
Minimal = MinimalRouterStateSerializer (Basically this https://ngrx.io/guide/router-store/configuration#custom-router-state-serializer and the minimal route event info)

We could flip them at some point so the default is the fully serializable one.

timdeschryver: I like this idea! The only "catch" to it would be that it would be possible to say set the config to `Minimal` and provide a custom serializer (this would overrule the minimal config), which is a bit contradictory.

brandonroberts: True. I'd say we go with the minimal route event info either way, and override the custom one if they choose a `routerState`

timdeschryver:
Updated the whole PR with feedback given.

projects/ngrx.io/content/guide/router-store/configuration.md
```diff
@@ -88,3 +90,52 @@ StoreRouterConnectingModule.forRoot({
...
+
+<code-example header="app.module.ts">
+StoreRouterConnectingModule.forRoot({
+  routerState: RouterState.Full,
+});
+</code-example>
+
+### RouterState.Minimal
+
+`RouterState.Minimal` will use the `MinimalRouterStateSerializer` serializer to serialize Angular router event.
```
brandonroberts:
```suggestion
`RouterState.Minimal` will use the `MinimalRouterStateSerializer` serializer to serialize the Angular Router's `RouterState` and `RouterEvent`.
```

projects/ngrx.io/content/guide/router-store/configuration.md
```diff
@@ -88,3 +90,52 @@ StoreRouterConnectingModule.forRoot({
...
+
+The metadata on the action consists of the navigation id and the url.
+
+```ts
```
brandonroberts:
```suggestion
```json
```

projects/ngrx.io/content/guide/router-store/configuration.md
```diff
@@ -88,3 +90,52 @@ StoreRouterConnectingModule.forRoot({
...
+
+## routerState
+
+This property decides which router serializer should be used. If there is a custom serializer provided, it will use the provided serializer. `routerState` also sets the metadata on dispatched `@ngrx/router-store` action.
+
+### RouterState.Full
+
+When this property is set to `RouterState.Full`, `@ngrx/router-store` will use the `DefaultRouterStateSerializer` serializer to serialize the Angular router event.
+
+The metadata on the action will contain the Angular router event, e.g. NavigationStart` and `RoutesRecognized`.
```
brandonroberts:
```suggestion
The metadata on the action will contain the Angular router event, e.g. `NavigationStart` and `RoutesRecognized`.
```

projects/ngrx.io/content/guide/router-store/configuration.md
```diff
@@ -5,12 +5,14 @@ interface StoreRouterConfig {
   stateKey?: string | Selector&lt;any, RouterReducerState&lt;T&gt;&gt;;
   serializer?: new (...args: any[]) => RouterStateSerializer;
   navigationActionTiming?: NavigationActionTiming;
+  state?: RouterState;
```
brandonroberts:
```suggestion
  routerState?: RouterState;
```

modules/router-store/src/router_store_module.ts
```diff
@@ -50,6 +60,12 @@ export interface StoreRouterConfig<
    * set this property to NavigationActionTiming.PostActivation.
    */
   navigationActionTiming?: NavigationActionTiming;
+  /**
+   * Decides which router serializer should be used, if there is none provided, and the metadata on the dispatched @ngrx/router-store action payload.
+   * Set to `Full` to use the `DefaultRouterStateSerializer` and to set the angular router events as payload.
+   * Set to `Minimal` to use the `MinimalRouterStateSerializer` and to set a minimal router event with the nagivation id and url as payload.
```
brandonroberts:
```suggestion
   * Set to `Minimal` to use the `MinimalRouterStateSerializer` and to set a minimal router event with the navigation id and url as payload.
```

modules/router-store/spec/router_store_module.spec.ts
```diff
@@ -134,4 +139,78 @@ describe('Router Store Module', () => {
...
+        const { serializer } = setup(
+          RouterState.Full,
+          MinimalRouterStateSerializer
+        );
+        expect(serializer).toEqual(new MinimalRouterStateSerializer());
+      });
+    });
+
+    describe('Minimal', () => {
+      it('should dispatch the navidation id with url', async () => {
```
brandonroberts:
```suggestion
      it('should dispatch the navigation id with url', async () => {
```

timdeschryver:
Thanks Brandon!

modules/router-store/src/serializers/minimal_serializer.ts
```diff
@@ -0,0 +1,33 @@
+import { RouterStateSnapshot, ActivatedRouteSnapshot } from '@angular/router';
+import { BaseRouterStoreState, RouterStateSerializer } from './shared';
+
+export interface MinimalRouteSnapshot {
+  params: RouterStateSnapshot['root']['params'];
```
brandonroberts: Should all these be the same?

projects/ngrx.io/content/guide/router-store/configuration.md
```diff
@@ -88,3 +90,52 @@ StoreRouterConnectingModule.forRoot({
   navigationActionTiming: NavigationActionTiming.PostActivation,
 });
 </code-example>
+
+## routerState
```
brandonroberts:
```suggestion
## Router State Snapshot
```

modules/router-store/src/serializers/minimal_serializer.ts
```diff
@@ -0,0 +1,33 @@
...
+  url: RouterStateSnapshot['root']['params'];
+  queryParams: RouterStateSnapshot['root']['params'];
+  data: RouterStateSnapshot['root']['params'];
```
brandonroberts:
```suggestion
  data: RouterStateSnapshot['root']['data'];
```

modules/router-store/src/serializers/minimal_serializer.ts
timdeschryver: No they should not 😐

modules/router-store/src/serializers/minimal_serializer.ts
```diff
@@ -0,0 +1,33 @@
...
+  queryParams: RouterStateSnapshot['root']['queryParams'];
+  data: RouterStateSnapshot['root']['data'];
+}
+
+export interface MinimalRouterStateSnapshot extends BaseRouterStoreState {
+  root: MinimalRouteSnapshot;
+  url: string;
+}
+
+export class MinimalRouterStateSerializer
```
brandonroberts: After taking a second look at this, I think this should match what the custom serializer does in the docs. If we just give them the root snapshot, the params will be empty, the data will be empty, only leaving the url and global queryParams.

timdeschryver: That would make more sense

timdeschryver: I've pushed these changes. One caveat is that users won't be able to access params from higher up. Another implementation I've encountered a few times is to loop over all of the children and pluck the params from each child and build up an object with all of the params.

modules/router-store/spec/serializers.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+      },
+      firstChild: undefined,
+    };
+
+    // properties that aren't serializable
+    delete snapshot.paramMap;
+    delete snapshot.queryParamMap;
+    delete snapshot.component;
+
+    // properties that do not exist on the minimal serializer
```
timdeschryver: Is this OK to leave out? 
I did this because these are always `undefined`

modules/router-store/src/serializers/minimal_serializer.ts
```diff
@@ -0,0 +1,53 @@
+import { RouterStateSnapshot, ActivatedRouteSnapshot } from '@angular/router';
+import { BaseRouterStoreState, RouterStateSerializer } from './shared';
+
+export interface MinimalActivatedRouteSnapshot {
+  routeConfig: ActivatedRouteSnapshot['routeConfig'];
```
timdeschryver: Is there a better way to type this?

modules/router-store/spec/serializers.spec.ts
brandonroberts: Yep

modules/router-store/src/serializers/minimal_serializer.ts
brandonroberts: Other than to use its actual type `Route`?

timdeschryver: Yea, I tried to just use `ActivatedRouteSnapshot` and override some properties but I'm not sure if this is possible.

brandonroberts: I'm not sure either

PR #1891, opened 2019-05-25
 feat(store): separate store/action immutability checks

### LINKED ISSUE #1890: splitting up strictImmutability into separate store/action checks

Not mutating store anyway.

## Other information:

In cases where an action has a mutable payload, setting strictImmutability throws an error when the payload mutates.  This is by design, and an overall good thing.

Here is an example where the Increment class has a payload that mutates after a timeout:

https://stackblitz.com/edit/ngrx-seed-3ytt7p 

For cases where developers cannot control the creation/dispatch of an action I would propose (like with serialization), adding strictStoreImmutability and strictActionImmutability (both of which strictImmutability would use in maintaining its current behavior).

[ ] No

ludicast:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The strictImmutability runtime check enforces the immutability of both actions and the store without letting the developer choose only one.

In certain cases we need to enforce immutability of either store or
action, but not both.  

Issue #1890

## What is the new behavior?

This commit adds strictActionImmutability and
strictStateImmutability runtime checks.  It keeps the original
strictImmutability check, which now composes its state and action variants.

Closes #1890

```
[ ] Yes
[x] No
```


timdeschryver:
Thanks for opening the issue and this PR.
I'm going to close this PR in favor of https://github.com/ngrx/platform/pull/1894 because we wanted to refactor the serializability meta-reducer with this change.
A small difference in that solution and this solution is that we completely dropped `strictImmutability`

```diff
-strictImmutability: true,
+strictStateImmutability: true,
+strictActionImmutability: true,
```

PR #1914, opened 2019-06-04
fix(action): Stop passing a function to store.dipatch and stop effect…

### LINKED ISSUE #1906: Strange behavior when passing action function into store.dispatch
I called someAction (see below simple example code) without function call "()" when it passed to store.dipatch(....)! The problem, it has not become an error. The action even arrived the reducer. Unfortunately it was not logged in the ReduxDevTools (it freezed, all other actions e.g. router will be ignored in the ReduxDevTools). Only with a MetaReducer logger I see what happens. After I called closeMainSidebarAction() with "()" the ReduxDevTools worked again. 

We should prevent this behavior by throwing an error.

That was really frustrating! I think it took me more than 2 hours to realize that I forgot to call the function call. I think that the potential is high that another one will run into this strange behavior.

I will create a PR for it!

Code Example:
```
const someAction = createAction('Some Action');
store.dispatch(closeMainSidebarAction);
```

## Expected behavior:

It should throw an error that to pass a function into dispatch is not allwed?

ngrx v8 rc0

[ ] No

SerkanSipahi:
… that is returns a function

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1906

```
[ ] Yes
[x] No
```


modules/store/src/actions_subject.ts
```diff
@@ -13,12 +25,7 @@ export class ActionsSubject extends BehaviorSubject<Action>
   }
 
   next(action: Action): void {
-    if (typeof action === 'undefined') {
```
SerkanSipahi: There were no tests for this block. Should I write some for it?

modules/effects/src/effects_runner.ts
```diff
@@ -17,6 +19,7 @@ export class EffectsRunner implements OnDestroy {
     if (!this.effectsSubscription) {
       this.effectsSubscription = this.effectSources
         .toActions()
+        .pipe(tap(action => throwErrorOnInvalidAction(action)))
```
timdeschryver: We don't need to to verify the action here. The action returned will be dispatched to the store, and will be verified there.

modules/store/src/actions_subject.ts
```diff
@@ -5,6 +5,18 @@ import { Action } from './models';
 
 export const INIT = '@ngrx/store/init' as '@ngrx/store/init';
 
+export const throwErrorOnInvalidAction = (action: Action) => {
+  if (typeof action === 'undefined') {
+    throw new TypeError(`Actions must be objects`);
+  } else if (typeof action === 'function') {
+    throw new TypeError(`
+      Actions as function are not now allowed.
```
timdeschryver:
```suggestion
      Actions as function are not allowed.
```

modules/store/src/actions_subject.ts
```diff
@@ -5,6 +5,18 @@ import { Action } from './models';
...
+      Make sure your action is called e.g someAction()`);
```
timdeschryver: Should we make this more explicit?

```
Dispatch expected an object, instead it received xxx
If you're using the createAction function, make sure to invoke the function before dispatching the action. E.g. someAction should be someAction().
```

modules/effects/src/effects_runner.ts
SerkanSipahi: You are absolutely right. The circle closes at the place! 
`EffectsRunner.start() => Store.next() => ActionsSubject.next()` :)

SerkanSipahi: Its removed!

modules/store/src/actions_subject.ts
SerkanSipahi: Like this ?
```js
next(action: Action): void {
  if(typeOf(action) !== 'object') {
    throw new TypeError(`
      Dispatch expected an object, instead it received ${action}.
      If you're using the createAction function, make sure to invoke the function
      before dispatching the action. E.g. someAction should be someAction().`);    
  } else if (typeof action.type === 'undefined') {
    throw new TypeError(`Actions must have a type property`);
  }
  super.next(action);
}
```

brandonroberts:
There should be an added check here also https://github.com/ngrx/platform/blob/master/modules/effects/src/effect_notification.ts#L33

SerkanSipahi:
Is there any place in ngrx where we can put helper methods? e.g. this:
```js
const throwErrorOnInvalidAction = action => {
  if(typeOf(action) !== 'object') {
    throw new TypeError(`
      Dispatch expected an object, instead it received ${action}.
      If you're using the createAction function, make sure to invoke the function
      before dispatching the action. E.g. someAction should be someAction().`);    
  } else if (typeof action.type === 'undefined') {
    throw new TypeError(`Actions must have a type property`);
  }
}
```

This can be used can also in https://github.com/ngrx/platform/blob/master/modules/effects/src/effect_notification.ts#L33

brandonroberts:
No. In this case they will just have to be separate. I don't think the function needs to exist in the public API for either library.

SerkanSipahi:
@brandonroberts I hope I understood you correctly! Like this?
```
// this check because `[]`(array) is also an object when checking with `typeof action`!
const typeOf = action =>  ({}).toString.call(action).slice(8, -1).toLowerCase();

function isAction(action: any): action is Action {
  return typeOf(action) === 'object' && action && action.type && typeof action.type === 'string';
}
```

modules/store/src/actions_subject.ts
SerkanSipahi: Its more explicit now (commited)

SerkanSipahi:
@brandonroberts its separated now (commited)

modules/store/src/actions_subject.ts
```diff
@@ -13,12 +19,14 @@ export class ActionsSubject extends BehaviorSubject<Action>
   }
 
   next(action: Action): void {
-    if (typeof action === 'undefined') {
-      throw new TypeError(`Actions must be objects`);
+    if (typeOf(action) !== 'object') {
```
alex-okrushko: would `typeof action !== 'object'` not work?

modules/effects/src/effect_notification.ts
```diff
@@ -30,8 +30,19 @@ export function reportInvalidActions(
   }
 }
 
+const typeOf = (value: any) =>
```
alex-okrushko: Generally these should be structured as function declarations, e.g.
```ts
function typeOf(value: any) {...}
```

modules/store/src/actions_subject.ts
SerkanSipahi: There is a edge case when you pass accidently an array instead of object. 
```js
typeof [] // object
```
```js
({}).toString.call([]).slice(8, -1).toLowerCase() // array
```
I just wanted a strict treatment regardless of whether we typescript or not.

alex-okrushko: It's true that array is type of `object` as well, however:
- we are trying to catch the case where `function` is passed instead of `object`, for example action creator is passed without calling it
- even if array doesn't run into this check, it will be caught with the next if-statement `typeof action.type === 'undefined'`

Also, I find the code snippet `({}).toString.call([]).slice(8, -1).toLowerCase()` to be somewhat "crafty" 😁and not very straightforward/readably. I'd rather avoid it.

timdeschryver: I agree with @alex-okrushko.
To keep it readable I'm in favor of using `typeof`.

Plus we already have a compile error if we try to dispatch an array. The `createAction` is only the problem here.

image

SerkanSipahi: I had the same thoughts (compile error) after reading what @alex-okrushko wrote. I was to strict! I have to find a balance between typescript and javascript :)

SerkanSipahi: `({}).toString.call([]).slice(8, -1).toLowerCase()` its removed. There is now a simple the for function.

modules/effects/src/effect_notification.ts
SerkanSipahi: The `typeOf ` is removed. We use `typeof xxx` as you suggested.

SerkanSipahi:
@timdeschryver @alex-okrushko all code parts are adjusted your code review suggestions.

alex-okrushko:
Looks good. Thanks Serkan!

modules/store/src/actions_subject.ts
```diff
@@ -13,12 +13,16 @@ export class ActionsSubject extends BehaviorSubject<Action>
   }
 
   next(action: Action): void {
-    if (typeof action === 'undefined') {
+    if (typeof action === 'function') {
+      throw new TypeError(`
+        Dispatch expected an object, instead it received function.
+        If you're using the createAction function, make sure to invoke the function
+        before dispatching the action. E.g. someAction should be someAction().`);
```
brandonroberts: Nit: Everyone doesn't know what `e.g.` means.

```suggestion
        before dispatching the action. For example, someAction should be someAction().`);
```

modules/store/src/actions_subject.ts
```diff
@@ -13,12 +13,16 @@ export class ActionsSubject extends BehaviorSubject<Action>
   }
 
   next(action: Action): void {
-    if (typeof action === 'undefined') {
+    if (typeof action === 'function') {
+      throw new TypeError(`
+        Dispatch expected an object, instead it received function.
```
timdeschryver:
```suggestion
        Dispatch expected an object, instead it received a function.
```

SerkanSipahi: Done

modules/store/src/actions_subject.ts
SerkanSipahi: Done

modules/effects/src/effect_notification.ts
```diff
@@ -31,7 +31,12 @@ export function reportInvalidActions(
 }
 
 function isAction(action: any): action is Action {
-  return action && action.type && typeof action.type === 'string';
+  return (
+    typeof action !== 'function' &&
+    action &&
+    action.type &&
+    typeof action.type === 'string'
+  );
```
alex-okrushko: I think you don't need to wrap the return in brackets. This could be reverted to what it was before.

SerkanSipahi: Oh, sorry, I didn't realize that! I had already wondered why prettier always jumped on when I committed something on NgRx. That doesn't happen with my private repos.  I thought it was something like a hooks coming from ngrx repo.

PR #1925, opened 2019-06-06
perf: fine tune schematics

timdeschryver:
Exclude node_modules
Only commit file updates when there are changes

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: Performance
```

Closes #

```
[ ] Yes
[x] No
```

This PR modifies the v8 migrations.

While running the `ng update` schematics, I noticed these were taking a long time.
Initially I thought this was because we did not exclude `node_modules`, but the main reason was that we committed changes on every ts file.

That's why I extracted the common logic into schematics-core, so we don't have this problem in the future.

### DIFF
--- a/modules/effects/schematics-core/utility/change.ts
+++ b/modules/effects/schematics-core/utility/change.ts
@@ -136,12 +136,16 @@ export class ReplaceChange implements Change {
 
 export function createReplaceChange(
   sourceFile: ts.SourceFile,
-  path: Path,
   node: ts.Node,
   oldText: string,
   newText: string
 ): ReplaceChange {
-  return new ReplaceChange(path, node.getStart(sourceFile), oldText, newText);
+  return new ReplaceChange(
+    sourceFile.fileName,
+    node.getStart(sourceFile),
+    oldText,
+    newText
+  );
 }
 
 export function createChangeRecorder(
@@ -161,4 +165,14 @@ export function createChangeRecorder(
     }
   }
   return recorder;
+}
+
+export function commitChanges(tree: Tree, path: string, changes: Change[]) {
+  if (changes.length === 0) {
+    return false;
+  }
+
+  const recorder = createChangeRecorder(tree, path, changes);
+  tree.commitUpdate(recorder);
+  return true;
 }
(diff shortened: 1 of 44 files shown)


modules/effects/schematics-core/utility/change.ts
```diff
@@ -162,3 +166,13 @@ export function createChangeRecorder(
   }
   return recorder;
 }
+
+export function commitChanges(tree: Tree, path: string, changes: Change[]) {
+  if (changes.length === 0) {
```
timdeschryver: Only commit the file when there are changes, I forgot this in some of the v8 migrations making them slow (plus there was a information message on every file changes, even node_modules, which might confuse people).

modules/effects/schematics-core/utility/visit-utils.ts
```diff
@@ -10,24 +10,35 @@ export function visitTSSourceFiles<Result = void>(
...
-      path,
-      tree.read(path)!.toString(),
-      ts.ScriptTarget.Latest
-    );
+  return result;
+}
 
-    if (sourceFile.isDeclarationFile) {
-      return;
+function* visit(directory: DirEntry): IterableIterator<ts.SourceFile> {
```
timdeschryver: Don't use `tree.visit` for more flexibility (like ignore `node_modules`).
Inspiration from https://github.com/angular/angular-cli/blob/576119f2cab6a53f0bc67026b723fffbe8407258/packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts. I like this solution the most as we break out the `node_modules` folder as fast as possible.

Another option is to use the `ignore` package, like the nx workspace does - https://github.com/nrwl/nx/blob/3af31f1cd996db2f225773eb1888bd109efb9f76/packages/schematics/migrations/update-8-0-0/update-8-0-0.ts#L176

Or to simple check if `node_modules` is in the path.

tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "strictPropertyInitialization": false,
+    "downlevelIteration": true,
```
timdeschryver: Needed for `yield* visit(directory.dir(path));`

PR #1944, opened 2019-06-12
feat(data): pass tag to EntityDataService

JiaLiPassion:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Now, `EntityDataService` can only accept `Entity` as parameter, the `tag` from `EntityCollectionService` is missing, we may need this `tag` in some use cases.

For example, Sometimes you may want to dispatch to different backend endpoints in the same operation of the EntityDataService, for example, let's say in frontend, we need to get `most downloaded items` and `most starred items`, if the backend provides a unified endpoint, we can just overwrite the `getWithQuery` operation, but if the backend provides two endpoints, we still want to use `getWithQuery` mechanism, in this case, we can use `tag`.
## What is the new behavior?

In the case I describe above, we can implement the `CustomDataService` like this,

```typescript
export class HeroDataService extends DefaultDataService&lt;Hero&gt; {
  getWithQuery(params: string | QueryParams, options?: EntityDataServiceOptions): Observable&lt;Hero[]&gt; {
    let data$;
    if (options && options.tag) {
      if (options.tag === 'MOST_DOWNLOADED') {
        data$ = getData('${serverUrl}/mostdownloaded');
      } else if (options.tag === 'MOST_STARRED') {
        data$ = getData('${serverUrl}/moststarred');
      }
    }
    if (!data$) {
      data$ = super.getWithQuery(params);
    }
    return data$.pipe(map(heroes => heroes.map(hero => this.mapHero(hero))));
  }
}
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/data/spec/effects/entity-effects.spec.ts
+++ b/modules/data/spec/effects/entity-effects.spec.ts
@@ -31,16 +31,34 @@ describe('EntityEffects (normal testing)', () => {
   let logger: Logger;
   let dataService: TestDataService;
 
-  function expectCompletion(completion: EntityAction, done: DoneFn) {
+  function expectCompletionWithTag(
+    completion: EntityAction,
+    methodName: string | undefined,
+    done: DoneFn
+  ) {
     effects.persist$.subscribe(
       result => {
         expect(result).toEqual(completion);
+        if (methodName) {
+          expect(
+            expectDataServiceCalledWithTag(methodName, completion.payload.tag!)
+          );
+        }
         done();
       },
       error => {
         fail(error);
       }
     );
+  }
+
+  function expectCompletion(completion: EntityAction, done: DoneFn) {
+    expectCompletionWithTag(completion, undefined, done);
+  }
+
+  function expectDataServiceCalledWithTag(methodName: string, tag: string) {
+    const allArgs = (dataService as any)[methodName].calls.allArgs()[0];
+    expect(allArgs[allArgs.length - 1].tag).toEqual(tag);
   }
 
   beforeEach(() => {
@@ -116,7 +134,7 @@ describe('EntityEffects (normal testing)', () => {
     });
 
     actions$.next(action);
-    expectCompletion(completion, done);
+    expectCompletionWithTag(completion, 'getAll', done);
   });
 
   it('should perform QUERY_ALL when dispatch custom action w/ that entityOp', (done: DoneFn) => {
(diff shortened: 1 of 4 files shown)


wardbell:
I see the value of passing more information into the dataservice method so that a custom implementation of that method implementation can make a request based on that information.

I'm not convinced that the string `tag` is the best carrier of that extra information. What if I want to pass a number? an object? a function? What if I want the tag to do its main job of labeling the action - making it more specific rather than generic - that conveys who issued the action? How would that translate to a request URL? Wouldn't that entangle the action label with the server API? In short, the `tag` is over-determined and, thus, not safe for the purpose you've identified.

I'm also not convinced that we need a special interface such as `EntityDataServiceOptions`.  I realize that I could extend it with other properties than `tag` and thus resolve my difficulty with the overdetermined `tag`. But do I really need a separate interface for this purpose.

The `EntityAction` is already an extensible package of action information to guide the implementation. We could pass _that_ into these methods instead.

Of course the `tag` is already in the `EntityAction` so you _could_ exploit it as you've done here. But you can make a custom extension (subclass) of `EntityAction` that has additional properties suitable for your custom HTTP request. That will work fine as long as your extended `EntityAction` has a recognizable `EntityOp`.

So we won't need a new interface type and we won't be limited to a `tag` property. We can revise the like this:
```
export interface EntityCollectionDataService<T> {
  readonly name: string;
  add(entity: T, action?: EntityAction): Observable<T>;
  delete(id: number | string, action?: EntityAction): Observable<number | string>;
  getAll(action?: EntityAction): Observable<T[]>;
  getById(id: any, action?: EntityAction): Observable<T>;
  getWithQuery(params: QueryParams | string, action?: EntityAction): Observable<T[]>;	
  update(update: Update<T>, action?: EntityAction): Observable<T>;
  upsert(entity: T, action?: EntityAction): Observable<T>;
}
```

I figure we have to make the action parameter optional so as to not break existing custom implementations.

What do you think?

smorandi:
yes, the approach with passing the entityaction as the carrier of additional information sounds more reasonable and offers far greater flexibility for custom implementation (short of passing a json in the tag :) to guide a custom implementation.

wardbell:
@smorandi Sorry for delayed reply. The answer is "yes"

brandonroberts:
@JiaLiPassion do you have enough information to update this PR according to the discussion above?

acotelin-dbg:
Heya! Thank you all for putting this PR together! We've started using @ngrx/data and we're facing the same situation. Any idea when would get merged in?

JiaLiPassion:
@acotelin-dbg , thanks for the feedback, I am still working on this one, will post status later.

brandonroberts:
@JiaLiPassion just reopen this PR when you're ready. Thanks!

wardbell:
@JiaLiPassion Remember that when we do re-open, our solution will be different. We won't pass the tag. We'll pass the entire `EntityAction` so that the custom data service has access to the widest amount of available information in order to process the request in whatever way it needs.

JiaLiPassion:
@brandonroberts , @wardbell , Thank you for the review and comment, sorry I am working on the other project right now, I will come back to reopen this PR soon.

smorandi:
i think the reopen is still pending on this one. or hhas this been done already?

cam-m:
Ability for extended data services receive the full action and adjust requests accordingly would be awesome. Will this pull request be revived?  I'm happy to help out with a new PR if needed?

cam-m:
I've proposed a slightly different solution in #3663. This proposal offers a direct line to adding query parameters or headers to an api requests, but it also supports what's being proposed here. A custom data service could decide how to handle query parameters/headers, and completely change how the api is called as they want.

Interested in your thoughts, and hope its useful

PR #1962, opened 2019-06-19
docs: Clean Up Store Integration Testing Section

### LINKED ISSUE #1952: docs: Clean Up Integration Testing Examples
Now that we have updated the documentation with using MockStore + mock selectors, we should revisit the existing testing examples (prior to v6) that have been moved to "Integration Testing" to more accurately represent an integration test.  It looks like the [v6 testing docs](https://github.com/ngrx/platform/blob/6.1.2/docs/store/testing.md) were just moved into this section without updating them.

I propose updating the documentation accordingly:
- In the store/testing integration testing section, do not have an example that dispatches a `load_____Success` action.  These actions are dispatched by an Effect. In this example, a true integration test would import the EffectsModule and mock the Effect's downstream dependencies.  But since this section is only concerned about Store Integration, the example should not be of an app that uses effects, and the component test should only dispatch actions that the component can.
- Add an effects/testing integration testing section, which shows importing the StoreModule and EffectsModule and mocks the Effect's downstream dependencies.

[ ] No

jtcrowson:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The current section was outdated.  It replicated the [v6 general testing documentation](https://github.com/ngrx/platform/blob/6.1.2/docs/store/testing.md), which discussed conditioning states.  That should now be done by the MockStore.

Also simplified the example to not dispatch actions that are usually done by an effect.

Closes #1952 (partially)
I'd like to do a follow up PR to add an integration testing section to effects/testing.

## What is the new behavior?

Updated documentation to be specific to the integration testing use case.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngrx.io/content/examples/store/src/app/my-counter/my-counter.component.html
+++ b/projects/ngrx.io/content/examples/store/src/app/my-counter/my-counter.component.html
@@ -1,7 +1,7 @@
-<button (click)="increment()">Increment</button>
- 
+<button id="increment" (click)="increment()">Increment</button>
+
 <div>Current Count: {{ count$ | async }}</div>
 
-<button (click)="decrement()">Decrement</button>
+<button id="decrement" (click)="decrement()">Decrement</button>
 
-<button (click)="reset()">Reset Counter</button>
+<button id="reset" (click)="reset()">Reset Counter</button>
--- a/projects/ngrx.io/content/examples/store/src/app/tests/integration.spec.ts
+++ b/projects/ngrx.io/content/examples/store/src/app/tests/integration.spec.ts
@@ -1,0 +1,64 @@
+import { TestBed, async, ComponentFixture } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { StoreModule } from '@ngrx/store';
+import { MyCounterComponent } from '../my-counter/my-counter.component';
+import { counterReducer } from '../counter.reducer';
+
+describe('MyCounterComponent', () => {
+  let component: MyCounterComponent;
+  let fixture: ComponentFixture<MyCounterComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ MyCounterComponent ],
+      imports: [
+        StoreModule.forRoot({ count: counterReducer })
+      ]
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(MyCounterComponent);
+    component = fixture.debugElement.componentInstance;
+    fixture.detectChanges();
+  }));
+
+  it('should create the component', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should increment the counter value when increment is clicked', () => {
+    clickByCSS('#increment');
+
+    expect(getCounterText()).toBe(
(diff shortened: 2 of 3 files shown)


jtcrowson:
@timdeschryver I planned on including that in a separate PR to be included in the effects/testing section.  After that, we can then include a link here to point users to an integration testing with store + effects + effect's dependent services.  But I thought it was best not to include stuff about effects in the store testing documentation.

timdeschryver:
Sorry, one more thing. Often times when we have similar tests, we also have to provide some initial state. Would it add value here if we would provide it here?

jtcrowson:
No worries, want to make sure we’re on the same page. Why would we need an initial state? By importing the relevant reducer with the StoreModule, the initial state gets set. In our integration test, we should just be verifying that the Store is properly communicating with our component.

timdeschryver:
Correct. But sometimes a component needs data that it isn't responsible for, data that is already in the store. An example of this would be  that you have a countries selector to populate a list. The component is not responsible to fetch this data, but it's using a selector to select this part of the state tree.

jtcrowson:
Should the list component inject the NgRx store? Best practice is usually to have a container component inject the store and pass down relevant data. Container components often dispatch actions to fetch the data they need. 

If we were to need some sort of initial state, would that be done in the test by getting a store instance and dispatching an action to load the data into the store?

brandonroberts:
I think using a list of data provides a better example of the different things you would test against. Dispatching an action in the `ngOnInit` to load the data, dispatching when selecting an item from the list, or performing some action against the item, such as dispatching an action to delete one. If we're doing an integration test, I'd also recommend trigger an event from the template instead of just calling the component function.

jtcrowson:
One issue with a list-style component as you’ve described is that Effects would likely be used. A load action would trigger an effect that would dispatch a success or error action. I tried to have an example use case that uses store without effects. Then, add an Integration Testing section to the effects/testing docs that imports the EffectsModule for a list component like you described. 

I’ll add triggering an event from the template.

brandonroberts:
We should reference the example from getting started if we're using that as a baseline then. It doesn't provide any context of the reducer or selectors that are associated with the counter. 

I'd rather the example be pulled from a live example also instead of inclined

jtcrowson:
Perfect! That’s actually what I was thinking when I made the example, just didn’t remember it was already in the NgRx docs 🙌 I’ll refactor these docs based on that example and add the integration test to the live example.

jtcrowson:
@timdeschryver @brandonroberts ready for re-review.

timdeschryver:
Not related to this PR, but it's about the same page.
Should we move the sections a bit, as:

- testing reducers - this would be a new section
- testing selectors
- mock store
- mock selectors
- integration test

jtcrowson:
I was thinking:
- Store-Dependent Component Testing
  - Unit Testing
    - Mock Store
    - Mock Selectors
  - Integration Testing
- Testing Reducers
- Testing Selectors

I think component testing has priority, as I don't think as many devs are testing reducers/selectors in comparison.

PR #1984, opened 2019-07-01
docs: Update book search failure action with only error message

### LINKED ISSUE #1953: Example: Update book search failure action with only error message

strictActionSerializability throws error in case of an @angular/common/http/HttpErrorResponse

The goal is to cause an http error.
Modifying `searchBooks()` in `google-books.service.ts` in the ngrx's `example-app` is enough:

```ts
searchBooks(queryTitle: string): Observable<Book[]> {
    return this.http
      .get<{ items: Book[] }>('xxx') // 'xxx' endpoint doesn't exist, which will cause an error
      .pipe(map(books => books.items || []));
  }
```

## Expected behavior:

Since the `xxx` endpoint doesn't exist, the `search$` effect in `book.effects.ts` will catch the error, but since it's not serializable, the `strictActionSerializability` flag will kick in and throw an error.

Screen Shot 2019-06-17 at 15 00 52

the newest ngrx, since version 8.

## Other information:

The reason why this is failing is because `'normalizedNames`, `lazyInit`, `lazyUpdate` properties from `HttpHeaders` are of type `Map` or `function`.

I guess this is actually a desired behaviour, and `strictActionSerializability` works as expected (since the Angular's error response object is not serializable), but my concern is how to handle these errors?

[ ] No

evgenyfedorenko:
Closes #1953

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/example-app/src/app/books/effects/book.effects.spec.ts
+++ b/projects/example-app/src/app/books/effects/book.effects.spec.ts
@@ -61,7 +61,7 @@ describe('BookEffects', () => {
       const completion = BooksApiActions.searchFailure({
         errorMsg: 'Unexpected Error. Try again later.',
       });
-      const error = 'Unexpected Error. Try again later.';
+      const error = { message: 'Unexpected Error. Try again later.' };
 
       actions$ = hot('-a---', { a: action });
       const response = cold('-#|', {}, error);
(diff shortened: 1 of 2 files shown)


evgenyfedorenko:
fixing test...

PR #1980, opened 2019-06-27
docs: API level documentation for Creator Functions

jtcrowson:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[x] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Closes #1977, #1956

## What is the new behavior?

- Creates basic API documentation for Creator Functions.
- Does not attempt to provide the level of context that the main documentation provides.
- Provides simple usage examples.
- Provides implementation details.

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

PR #1987, opened 2019-07-02
docs:  add destructuring actions with props example for createReducer…

### LINKED ISSUE #1985: Docs: Add example to createReducer of destructuring actions with props
We need to add more documentation around using actions with `props` with destructuring in the `createReducer` examples here: [createReducer docs](https://ngrx.io/guide/store/reducers).

Maybe consider adding an additional action that has `props` and then update the `createReducer` to use those new props in the `on` method.

An example would be:

```typescript
// action
export const loadSuccess = createAction('[Items API] Load Success', props<{items: Item[]}>());

// reducer `on` parameter
const featureReducer = createReducer(
  initialState,
  on(featureActions.loadSuccess, (state, { items }) => ({ ...state, items }))
);
```

jordanpowell88:
… (#1985)

Closes #1985

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[X] Documentation content changes
[ ] Other... Please describe:
```

Currently their is no documentation around how to access actions with props in the `on` method of `createReducer`

Closes #1985

## What is the new behavior?
 1.) ad an action with a prop
 2.) show how to destructure the action to access the prop in the `on` method of `createReducer`

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/ngrx.io/content/guide/store/reducers.md
+++ b/projects/ngrx.io/content/guide/store/reducers.md
@@ -25,6 +25,7 @@ import { createAction } from '@ngrx/store';
 export const homeScore = createAction('[Scoreboard Page] Home Score');
 export const awayScore = createAction('[Scoreboard Page] Away Score');
 export const resetScore = createAction('[Scoreboard Page] Score Reset');
+export const setScores = createAction('[Scoreboard Page] Set Scores');
 
 </code-example>
 
@@ -72,7 +73,8 @@ const scoreboardReducer = createReducer(
   initialState,
   on(ScoreboardPageActions.homeScore, state => ({ ...state, home: state.home + 1 })),
   on(ScoreboardPageActions.awayScore, state => ({ ...state, away: state.away + 1 })),
-  on(ScoreboardPageActions.resetScore, state => ({ home: 0, away: 0 }))
+  on(ScoreboardPageActions.resetScore, state => ({ home: 0, away: 0 })),
+  on(ScoreboardPageActions.setScores, (state, { scores }) => ({ home: scores.home, away: scores.away }))
 );
 
 export function reducer(state: State | undefined, action: Action) {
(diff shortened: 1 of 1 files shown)


wesleygrimes:
Thanks for your submission @jordanpowell88! We will take a look and get back to you.

jordanpowell88:
I agree, I debated whether to do this in my initial commit as it seamed unclear to me. I will work on adding those changes

wesleygrimes:
Thank you @jordanpowell88! I think getting the new reducer function documented will be a super valuable PR.

timdeschryver:
fwiw, @jtcrowson documented the new API `create` functions at #1980. If you click on the `on` function in the code snippets, you'll be redirected to these API docs. As addressed in that same PR, we have to be careful to not document everything twice, as it will be harder to maintain.

I'm not saying we don't have to document it, I just want to raise some awareness that #1980  exists 🙂

wesleygrimes:
I agree we don't want to duplicate efforts, but we need a place to gently guide folks into the functionality. 

So, how about we use the API documentation from @jtcrowson as a baseline and pull out the highlights into this page? Then we create hard links to the API docs from this area so that we don't duplicate docs.

jtcrowson:
Agree that there is some duplication between here and #1980.  

Using Angular.io as inspiration, API-level docs explain the function and it's parameters, while general docs explain how to use the function in combination with others to accomplish application functionality.

For this case, I like that the general docs already show defining a set of actions, a state, an initial state, then using them in a reducer.  But I'm not sure we need a `createReducer` section to explain the function in isolation.

I'd advocate adding the additional example to the existing section and dropping the `createReducer` section in favor of #1980:

```typescript
const scoreboardReducer = createReducer(
  initialState,
  on(ScoreboardPageActions.homeScore, state => ({ ...state, home: state.home + 1 })),
  on(ScoreboardPageActions.awayScore, state => ({ ...state, away: state.away + 1 })),
  on(ScoreboardPageActions.resetScore, state => ({ home: 0, away: 0 })),
  on(ScoreboardPageActions.setScores, (state, { game }) => ({ home: game.home, away: game.away }))
);
```

wesleygrimes:
I am also in agreement that we don’t want duplication. We do, however, want this part of the guide/website to gently guide folks to the official API docs section on createReducer. As I stated earlier, we need to introduce the concept of the new style reducer here, add high level details on using the on method with props and show a good example of destructuring. Then in that high level example put a link to point folks directly to the more detailed API docs on createReducer.

jordanpowell88:
Do you feel my most recent commit accomplish this @wesleygrimes ?

wesleygrimes:
Headed on vacation for a week but check when I return. I think it needs some tweaks as @jtcrowson suggests. 😃

-Wes

brandonroberts:
#1980 has landed so this one can be revisited to see if these changes are still needed.

wesleygrimes:
OK, nice work @jtcrowson on #1980. I agree we don't need an entire section here. @jordanpowell88 let's simplify this PR and just add the example with destructuring. There is already a link to the API section for `createReducer` so that covers getting folks to the detailed API docs.

PR #2017, opened 2019-07-15
Schematics: Add and import feature key for reducer state

### LINKED ISSUE #1996: Schematics: Add and import feature key for reducer state
Currently, when we register a feature, we tie to it the reducer using a string, such as:

`StoreModule.forFeature('users', fromUsers.reducer)`

This would add an exported "feature key" to the reducer file and import it to prevent typos.

In the reducer file:

```ts
export const usersFeatureKey = 'users';
```

Where the feature is registered

```ts
import * as fromUsers from './users.reducer';

@NgModule({
  imports: [
    StoreModule.forFeature(fromUsers.usersFeatureKey, fromUsers.reducer)
  ]
})
```

Where the top-level selector is used

```ts
export const selectUsersState = createFeatureSelector<AppState, UsersState>(fromUsers.usersFeatureKey)
```

I'd propose updating the docs, and example app to use this convention also.

jtcrowson:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1996 

## What is the new behavior?

Update schematics to introduce a featureKey, which is used when registering feature state in a module or in a parent state.

Example App and Docs are also updated to reflect this best practice.

Here are the changes for each schematic:

### Store:
- When using the root option, does not add a featureKey.
- When using without root option, creates a featureKey constant in the reducer file based on the name, which is used to register the feature if a module option is passed.

### Reducer
- featureKey constant is always created in the reducer file based on the name.
- When using the module option, featureKey is used to register the feature.
- When using the reducers option, featureKey is used as the key:
```ts
export interface State {
  [fromUser.userFeatureKey]: fromUser.State;
}
export const reducers: ActionReducerMap<State> = {
  [fromUser.userFeatureKey]: fromUser.reducer,
};
```

### Entity
- featureKey constant is always created in the reducer file based on the pluralized version of the name.
- When using the module option, featureKey is used to register the feature.
- When using the reducers option, featureKey is used as the key.

### Feature
- feature constant is always created in the reducer file based on the name.
- When using the module option, featureKey is used to register the feature.
- When using the reducers option, featureKey is used as the key.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/data/schematics-core/index.ts
+++ b/modules/data/schematics-core/index.ts
@@ -7,6 +7,7 @@ import {
   group,
   capitalize,
   featurePath,
+  pluralize,
 } from './utility/strings';
 
 export {
@@ -64,6 +65,7 @@ export const stringUtils = {
   group,
   capitalize,
   featurePath,
+  pluralize,
 };
 
 export { updatePackage } from './utility/update';
--- a/modules/data/schematics-core/utility/ngrx-utils.ts
+++ b/modules/data/schematics-core/utility/ngrx-utils.ts
@@ -147,6 +147,7 @@ export function addReducerToActionReducerMap(
 
       return { initializer: variable.initializer, type };
     })
+    .filter(initWithType => initWithType.type !== undefined)
     .find(({ type }) => type.typeName.text === 'ActionReducerMap');
 
   if (!actionReducerMap || !actionReducerMap.initializer) {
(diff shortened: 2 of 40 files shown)


brandonroberts:
@jtcrowson will you amend and do another force push to kick off CI?

jtcrowson:
@brandonroberts amended and force pushed 👍

PR #2026, opened 2019-07-20
fix(effects): resubscribe every time and error occurs (#2023)

### LINKED ISSUE #2023: Resubscribe on Error should work multiple times

```
$myEffect = createEffect(() =>
  this.actions$.pipe(
    ofType('MY_TYPE'),
    tap(() => console.log('I am here')),
    switchMap(() => throwError('error'))
  )
)
```

The console log shows only twice.

On the first error it switches to a new observable, but without any safeguard for the next error.
https://github.com/ngrx/platform/blob/master/modules/effects/src/effects_resolver.ts#L32

## Expected behavior:

The resubscribe should work multiple times in case error happens in the stream
The error handling should be in a recursive function, something like this:

```
const handleErrors = obs$ => obs$.pipe(catchError(error => {
  if (errorHandler) errorHandler.handleError(error);
  // Return observable that produces this particular effect
  return handleErrors(obs$);
}));

const resubscribable$ = resubscribeOnError ? handleErrors(observable$) : observable$;
```

8.1.0

[ ] No

peterreisz:
## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2023

## What is the new behavior?

The effects will resubscribe every time an error occurs in the stream.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/src/effects_resolver.ts
+++ b/modules/effects/src/effects_resolver.ts
@@ -24,14 +24,19 @@ export function mergeEffects(
           ? sourceInstance[propertyName]()
           : sourceInstance[propertyName];
 
+      const resubscribeInCaseOfError = (
+        observable$: Observable<any>
+      ): Observable<any> =>
+        observable$.pipe(
+          catchError(error => {
+            if (errorHandler) errorHandler.handleError(error);
+            // Return observable that produces this particular effect
+            return resubscribeInCaseOfError(observable$);
+          })
+        );
+
       const resubscribable$ = resubscribeOnError
-        ? observable$.pipe(
-            catchError(error => {
-              if (errorHandler) errorHandler.handleError(error);
-              // Return observable that produces this particular effect
-              return observable$;
-            })
-          )
+        ? resubscribeInCaseOfError(observable$)
         : observable$;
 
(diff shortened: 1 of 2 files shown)


modules/effects/src/effects_resolver.ts
```diff
@@ -24,14 +24,19 @@ export function mergeEffects(
           ? sourceInstance[propertyName]()
           : sourceInstance[propertyName];
 
+      const resubscribeInCaseError = (
```
timdeschryver:
```suggestion
      const resubscribeInOfCaseError = (
```

brandonroberts: `resubscribeInCaseOfError`?

peterreisz: I prefer the last one, fixed.

peterreisz:
@brandonroberts @timdeschryver This PR was reverted due to some CI issue. Are you planning on fixing the issue? Can I help somehow?

timdeschryver:
@peterreisz see #2165

PR #2046, opened 2019-08-07
Ngrx component setup

BioPhoton:
# @ngrx/component

The planned release includes the following things:
## Alpa Release
- [x] [Design Doc - Push Pipe](https://hackmd.io/Uus0vFu3RmWRVGgmtzuUWQ?edit) In: HackMd, Michael Hladky.
- [x] [Design Doc - Let Directive](https://hackmd.io/8_3rp0A7RweSYJiulsifbQ?edit) In: HackMd, Michael Hladky.
## Beta Releases
**Improvements**
- [ ] cdAware creation factory
  - [ ] Ensure that it is in line with the current project
- [ ] Push Pipe
  - [ ] Ensure typing in the templeat
  - [ ] Tests for Ivy vs ViewEngine vs zone-full vs zone-less
  - [ ] Muntiple nested input binding tests
  - [ ] Nested ng-container tests
  - [ ] performance tests 
     - [ ] with a sync array that exhausts synchronously and count number of renders for all scenarios
- [ ] Let Directive
  - [ ] Ensure typing in the templeat
  - [ ] Tests for Ivy vs ViewEngine vs zone-full vs zone-less
  - [ ] Muntiple ng-container on same page with the same input

**Features**
- [ ] [Design Doc - Coalescing of Change Detection](https://hackmd.io/42oAMrzYReizA65AwQ7ZlQ) 
In: HackMd, Michael Hladky.
- [ ] operator implementation
  - [ ] scoping strategy tests
  - [ ] implement tests
  - [ ] implement docs
- [ ] Finalize feature set

## To Discuss

- [ ] [Design Doc - Store Service]() @TODO
- [ ] [Design Doc - State Service]() In: HackMd, Michael Hladky.


projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+- [Observable Input Bindings](#observable-input-bindings)
+- [Observable Output Bindings](#observable-output-bindings)
+- [Local State Management](#local-state-management)
+  * [selectSlices RxJS Operator](#selectslices-rxjs-operator)
+
+<!-- tocstop -->
+
+### Push Pipe
+
+An angular pipe similar to the `async` pipe but triggers `detectChanges` instead of `markForCheck`.
```
timdeschryver:
```suggestion
An Angular pipe similar to the `async` pipe but triggers `detectChanges` instead of `markForCheck`.
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+This is required to run zone-less. We render on every pushed message.
+(currently, there is an [isssue](https://github.com/angular/angular/issues/31438) with the `ChangeDetectorRef` in ivy so we have to wait for the fix.
+
+The pipe should work as template binding `{{thing$ | push}}` 
+as well as input binding `[color]="thing$ | push"` and trigger the changes of the host component.
```
timdeschryver:
```suggestion
as well as an input binding `[color]="thing$ | push"` and triggers the changes of the host component.
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+  color: {{thing.color}}
+  shape: {{thing.shape}}
+<div>
+
+<app-color [color]="(thing$ | push).color">
+</app-color>
+```
+
+**Included Features:**
+- subscription handling overview  life cycle
```
timdeschryver:
```suggestion
- subscription handling overview life cycle
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+- a unified way of handling null and undefined with streams
```
timdeschryver:
```suggestion
- a unified way of handling `null` and `undefined` values with streams
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+- optional flag to turn off scheduling over `AnimationFrameScheduler` (on by default)
+- change detection is done manually which allows it to work zone-less too
+
+### Let Structural Directive
+
+The `*rgnxLet` directive serves a convenient way of binding multiple observables in the same view context.
+It also helps with some extra features under the hood.
+
+The current way of handling subscriptions in the view looks like that:
```
timdeschryver:
```suggestion
The current way of handling subscriptions in the view looks like this:
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+<ng-container *ngrxLet="observables; color as c; shape as s; name as n">
+  <app-color [color]="c" [shape]="s" [name]="n">
+  </app-color>
+</ng-container>
+```
+
+**Included Features:**
+- binding is always present. (`*ngIf="{}"` normally effects it)
+- it takes away the multiple usages of the `async` pipe 
+- proper handling of null and undefined values
```
timdeschryver:
```suggestion
- proper handling of `null` and `undefined` values
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+    .pipe(
+      switchMapTo(interval(1000)),
+      map(_ => Date.now()),
+      takeUntil(this.onDestroy$)
+    )
+    .subscribe();
+```
+
+**Included Features**
+- it handles late subscribers.
```
timdeschryver:
```suggestion
- it handles late subscribers
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+- exposes only observables
+- respects single shot vs ongoing life cycles
+- subscription handling over the component lifetime
+- return the latest value when resubscribing
+
+#### selectChanges RxJS Operator
+
+An operators `selectChanges` to select one or many specific slices from `SimpleChange`.
```
timdeschryver:
```suggestion
An operator `selectChanges` to select one or many specific slices from `SimpleChange`. 
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+  onChanges$: Observable<SimpleChanges>;
+
+  @Input() state;
+  state$ = this.onChanges$.pipe(getChange('state'));
+}
+```
+
+Following things are done under the hood:
+- pull out `currentValue` from `SimpleChanges` object
+- optional it could have a parma for a custom comparison function
```
timdeschryver:
```suggestion
- optional it could have a param for a custom comparison function
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Observable Input Bindings
+
+A property decorator which turns component or directive input binding into an observable and assigned it to the related property.
```
timdeschryver:
```suggestion
A property decorator which turns component or directive input binding into an observable and assigns it to the related property.
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+- It is multi-casted to avoid multiple subscriptions
+- It works with WebComponents and AngularComponents
+
+### Observable Output Bindings
+
+TBD from Dominic Elm
+
+### Local State Management
+
+This extension is maybe the most interesting one. While we can
```
timdeschryver: I think it's missing something here?
While we can ...

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+
+This extension is maybe the most interesting one. While we can 
+
+A tiny logic that combines:
+- values over input bindings
+- The component class internal state
+- state rendered to view 
+- state from services.
+
+A way to connect events from the view and component as observable.
```
timdeschryver:
```suggestion
A way to connect events from the view and component as observables.
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+    .connectSlice('isNew', this.isNew$);
+  this.lS
+    .connectSlice('buttons', this.buttons$);
+}
+```
+
+#### selectSlices RxJS Operator
+
+A flexible way to query one or many state slices.
+It considers also a late subscriber.
```
timdeschryver:
```suggestion
It also considers late subscribers. 
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+    .connectSlice('buttons', this.buttons$);
+}
+```
+
+#### selectSlices RxJS Operator
+
+A flexible way to query one or many state slices.
+It considers also a late subscriber. 
+
+An operators `selectSlice` to select a specific slice from the managed state.
```
timdeschryver:
```suggestion
An operator `selectSlices` to select a specific slice from the managed state. 
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+A flexible way to query one or many state slices.
+It considers also a late subscriber. 
+
+An operators `selectSlice` to select a specific slice from the managed state. 
+This operator can be used to get slices from `this.lS$`.
+
+```typescript
+buttons$ = this.lS.state$
+  .pipe(
+    selectChange(['state', 'substate'])
```
timdeschryver: `selectSlices` ?

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,227 @@
...
+  );
+```
+
+Following things are done under the hood:
+- it handles late subscribers with `shareReplay(1)` 
+- it forwards only distinct values
```
timdeschryver:
```suggestion
- it only forwards distinct values
```

BioPhoton:
Thanks for the review @timdeschryver !

jtcrowson:
@BioPhoton Are you looking for feedback on the documentation here?  I'm happy to review, if so.

jtcrowson:
Also, if you add
```
{
      "title": "Component",
      "children": [
        {
          "title": "Overview",
          "url": "guide/component"
        }
      ]
    },
```
to `projects/ngrx.io/content/navigation.json`, you'll be able to navigate to the docs from the sidebar.

BioPhoton:
@jtcrowson Yes i look for feedback on the idea. Also the pocs could be looked at. But the main thing here is the idea and features included

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,238 @@
+# @ngrx/component
+
+Reactive Component/Decorator library for Angular applications.
+
+Included are decorators and directives that support a fully reactive in Angular Components.
```
Adrael: Do you mean a reactive `state`?

```suggestion
Included are decorators and directives that support a fully reactive state in Angular Components.  
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,238 @@
+# @ngrx/component
+
+Reactive Component/Decorator library for Angular applications.
+
+Included are decorators and directives that support a fully reactive in Angular Components.  
+
+## Introduction
+
+Parts of Angular like the `ReactiveFromsModule`, `RouterModule`, `HttpClientModule`, etc. are already reactive.
```
Adrael:
```suggestion
Parts of Angular like the `ReactiveFormsModule`, `RouterModule`, `HttpClientModule`, etc. are already reactive.
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,238 @@
...
+For those who prefer imperative code, it's a little effort to restrict it to a single subscription.
+
+On the other hand for those who prefer reactive code, it's not that easy. 
+A lot of conveniences are missing, and beside the `async` pipe there is pretty much nothing to take away the manual mapping to observables.  
+Furthermore, an increasing number of packages start to be fully observable based. A very popular and widely used example is [NgRx](https://ngrx.io/). It enables us to maintain global push-based state management based on observables. 
+Also, other well-known libraries, [angular material](https://material.angular.io/) provide a reactive way of usage.
```
Adrael:
```suggestion
Also, other well-known libraries, [Angular material](https://material.angular.io/) provide a reactive way of usage.
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,238 @@
...
+              shape: observable2$ | async,
+              name:  observable3$ | async
+            } as c">
+  {{c.color}}
+  <app-color [color]="c.color" [shape]="c.shape" [name]="c.name">
+  </app-color>  
+</ng-container>
+```
+
+The `*ngrxLet` directive take over several things and makes it more convenient and save to work with streams in the template
```
Adrael:
```suggestion
The `*ngrxLet` directive takes over several things and makes it more convenient and safe to work with streams in the template
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,238 @@
...
+`*ngrxLet="o$ as o;"`
```
Adrael:
```suggestion
`*ngrxLet="o$ as o"` 
```

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -0,0 +1,238 @@
...
+  onChanges$: Observable<SimpleChanges>;
+
+  @Input() state;
+  state$ = this.onChanges$.pipe(getChange('state'));
+}
+```
+
+Following things are done under the hood:
+- pull out `currentValue` from `SimpleChanges` object
+- optional it could have a param for a custom comparison function
```
Adrael:
```suggestion
- optional: it could have a param for a custom comparison function
```

LayZeeDK:
Go Michael! 👏

SerkanSipahi:
Nice!

BioPhoton:
Hi all,

ATM I'm working on the performance optimizations with zone-less applications and the new component package.
The feature that is needed here is the coalescing of CD (change detection) calls.

I'm trying to get a good overview before I start hacking so the following information may contain miss assumptions. Please correct me wherever needed!

I already tested out some techniques with @JiaLiPassion and created a configurable RxJS operator for it. This operator acts as a rate-limiting operator on the event-loop level.
 
I create 2 variants, one that filters and one more like a side effect operator like a tap that fires the side effect based on some condition.

```typescript
function observeOncePerAnimationFrame<T>(
  // the e.g. our CD call
  work: () => void,
  // configuration to define the scope and provide a reference to the animationFrame reference (zone patched it so we cant use `requestAnimationFrame` directly
  cfg: {
    // by default the coalescing works app-wide.
    // Key to this problem will be to find the right scoping here
    context: window,
   // in Angular zone patches the animation frame so we have to provide the unpatched one in our case
    requestAnimationFrameRef: requestAnimationFrame,
  }
)
}

// == Usage ===
changes$.pipe(
  observeOncePerAnimationFrame(this.cdRef.detectChanges)
).subscribe();
```

The big challenge is to find the right approach for scoping the CD coalescing.

What does it mean?

**Problem:**
We can not use e.g. the window as the scope in which we ignore all but one call, we have to do it per component or even more per template scope. 
Multiple pipes or let directives on the same page trigger CD individually.

If we run zone-full `markDirty` takes over this problem and coalesces multiple calls IN a animationFrame or uses A animationFrame to do the chek. (currently, I don't have everything at the top of my head, but @LayZeeDK was so kind and reverse engineered a bit. So the correct information is here.)
However, if we run zone-less, and many motivation of | push especially, and *let, is to enable us easy to write zone-less applications.

When running zone-less Angulars can't detect the changes and so can't coalesce them, so we need to come up with our own strategy.

So if multiple changes happen at the same moment they will fire multiple CD calls.
Here we have to solve the above-mentioned problem of coalescing the CD calls within a component on our own.

**Research:**
Inside the pipe  (or the directive), and this is independent of Ivy's `ɵdetectChanges` or ViewEngines `ChangeDetectorRef.detectChanges` we use EmbeddedViewRef. 

If we start to dig in we realize there are different situations, the EmbeddedViewRef is eighter a reference to the parent view or the child component or template scope, depending on the usage.

- Template Interpolation:
```html
{{observable$ | push}}
``` 
This situation is valid for the `push` pipe. Here the pipe is used to render values to the HTML.
Inside the pipe, if we log we should see something like 'parentSomthingSomething' if we log the EmbeddedViewRef.

- Template context:
```html
<ng-container *let="observable$ as o"></ng-container>
<simple-component [value]="observable$ | push as o"></simple-component>
``` 
This situation is valid for the `push` pipe and `*let` directive. Here the pipe or directive is used to bind  values to a specific HTML scope (the ng-content example) or to a components input binding (the simple-component example).  If we log we should see something related to the child scope (please excuse me if I get vague here, but ATM I just try to put thoughts down and don't have the code open) or similar if we log the EmbeddedViewRef. 

With this information, let's think about the coalescing problem again.
In the case of template interpolation, all events fired inside the scope of the component we should fire CD only once even if we fire it 3 times at the same moment. 
In the case of template context, we fire CD for every scope once. Meaning multiple bindings of the same component updated at the same time should fire only once per tick.

**First Thoughts:**
Depending on the what our EmbeddedViewRef ref is we have to use a different scoping for the coalescing. 

In the case of the push pipe placed as template interpolation, we could coalesce all CD calls within the component as only one call is needed, no matter if we bind the same observable multiple times or multiple different observables. 

In the case of *let and the push pipe use as template binding, we can use it's EmbeddedViewRef pointing to the child as a reference and coalesce multiple changes within a template scope. Other *let directives next to it will trigger separately and could fire in the same tick to update it's scope of the template. This is good if we have other bindings and bad if we bind the same observable at another place in the template.

LayZeeDK:
Unlike `ChangeDetectorRef#markForCheck` and `ChangeDetectorRef#detectChanges`, `markDirty` can work in zoneless (`NoopNgZone`) projects. I think `markDirty` solves all the problems you mention in Ivy. With or without an `NgZone`.

# What does `markDirty` do?
Similar to `ChangeDetectorRef#markForCheck`, `markDirty` marks the component's view and all its ancestors all the way up to the root component's view as dirty. Unlike `ChangeDetectorRef#markForCheck`, `markDirty` also schedules a tick to perform a change detection cycle which dirty checks bindings based on the change detection mode (`CheckOnce`/`CheckAlways`) and dirty flag of individual component views. This tick is scheduled using a scheduler. The default scheduler for this is `requestAnimationFrame`, but can be configured to use a different scheduler when using `renderComponent` to bootstrap a new component tree.

# In which cases does `markDirty` not solve the problems?
1. In View Engine applications ([it's recommended to keep supporting View Engine in Angular versions 9 and 10](https://indepth.dev/the-angular-ivy-guide-for-library-authors/)).
1. There's a difference between `markDirty` and `detectChanges`. `detectChanges` synchronously runs a change detection cycle starting at the passed component's view and continues down through its descendants in this sub-component tree. I can't determine whether there might be some edge cases here. Angular's own `PushPipe` from `CommonModule` uses `ChangeDetectorRef#markForCheck` (still the case in Angular version 9) though.

# How does `markDirty` coalesce calls into a single change detection cycle per VM turn?
`markDirty`--or rather a function it indirectly depends on called [`scheduleTick`](https://github.com/angular/angular/blob/304584c291227d1878ccbdc5a4db3b4a93e633aa/packages/core/src/render3/instructions/shared.ts#L1770-L1806) (which is unfortunately not publicly exposed)--sets a single bit the first time in a VM turn that it is called. In the same breath, it schedules a tick using the relevant scheduler (default is ´requestAnimationFrame`). In consequent calls, it checks whether the bit is set. If so, it's basically a no-op.

`markDirty` first executes the statement `markViewDirty(getComponentViewByInstance(component));` before calling `scheduleTick`.

[`markViewDirty](https://github.com/angular/angular/blob/304584c291227d1878ccbdc5a4db3b4a93e633aa/packages/core/src/render3/instructions/shared.ts#L1744-L1767) is the one doing the work of traversing the component tree upwards and marking views as dirty (need dirty checking, regardless of `CheckOnce`/`CheckAlways`).

modules/component/package-lock.json
```diff
@@ -0,0 +1,5 @@
+{
```
timdeschryver: This file can be deleted

modules/component/src/component.module.ts
```diff
@@ -1,15 +1,15 @@
...
-const EXPORTS = [DECLARATIONS];
+export const EXPORTS = [DECLARATIONS];
 
 @NgModule({
   declarations: [DECLARATIONS],
   imports: [],
+  // @TODO ivy compatible
   exports: [EXPORTS],
 })
 export class NgRxComponentModule {}
```
timdeschryver:
```suggestion
export class ComponentModule {}
```

We don't prefix with `NgRx`

modules/component/src/index.ts
```diff
@@ -1,11 +1,8 @@
...
-export { selectChange } from './hook$/operators/selectChange';
-export { HostListener$ } from './host-listener$/host-listener$.decorator';
-export { Input$ } from './input$/input$.decorator';
-export { LocalStateService } from './local-state/local-state';
-export { selectSlice } from './local-state/operators/selectSlice';
-export { Push$Pipe } from './push$/push$.pipe';
-export { Async$Pipe } from './push$/async$.pipe';
-export { detectChanges } from './push$/operators/detectChanges';
-export { LetDirective } from './let/let.directive';
+export * from './core/operators/observeOncePerAnimationFrame';
```
timdeschryver: Can you revert this, and make it specific exports please?

For example.

```ts
export { PushPipe } from './push';
```

We only want to export the public stuff

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -1,3 +1,28 @@
...
+ 
+This library provides a set of primitives making it more convenient to work in a reactive architecture.
+  
+## Installation 
+
+Detailed installation instructions can be found on the [Installation](guide/component/install) page.
+
+## Included
+
+- [Push Pipe](#guide/component/push)
```
timdeschryver:
```suggestion
- [Push Pipe](guide/component/push)
```

In order to build the docs

projects/ngrx.io/content/guide/component/index.md
```diff
@@ -1,3 +1,28 @@
...
+- [Let Structural Directive](#guide/component/let)
```
timdeschryver:
```suggestion
- [Let Structural Directive](guide/component/let)
```

Same

modules/component/src/index.ts
```diff
@@ -1,11 +1,8 @@
-export { Hook$ } from './hook$/hook$.decorator';
```
timdeschryver: Inside the `hooks.decorator` file, we have to change the import from `ɵNG_COMPONENT_DEF` to `ɵNG_COMP_DEF`, as this has been renamed.

modules/component/src/state/state.ts
```diff
@@ -0,0 +1,249 @@
...
+   */
+  ngOnDestroy(): void {
+    this.teardown();
+  }
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class GlobalState<T> extends State<T> {}
```
timdeschryver: Move this into another file, otherwise bazel won't be able to build it. 
(don't forget to add it to the `index.ts` file afterwards 🙂)

modules/component/src/index.ts
LayZeeDK: This was renamed in v9. It's marked as private so it can be renamed in even a minor or patch release of Angular. It's risky to depend on unstable APIs.

BioPhoton: There is no hooks decorator!

BioPhoton: Something went wrong with this merge!!

BioPhoton:
@timdeschryver Is it possible to revert your latest changes? There must be something completely wrong as you comment on files I deleted long time ago.

BioPhoton:
@LayZeeDK 

The described situation works for zone-full applications for sure.

For zone-less I'm not sure. I will invest in this in the next days. 

Also markDirty schedules an app.tick. This would trigger rerendering (or at least the check for it) everywhere? right?

LayZeeDK:
>Also markDirty schedules an app.tick. This would trigger rerendering (or at least the check for it) everywhere? right?

Yes, the internal `scheduleTick` function used by `ɵmarkDirty` will schedule a change detection cycle for the full component tree, starting at the root component. `ChangeDetectorRef#detectChanges` and `ɵdetectChanges` will run a change detection cycle at the sub component tree that has the specified component at its root.

LayZeeDK:
@JoostK @alexzuza What do you think about `detectChanges` vs. `markDirty` with/without `NgZone`? Performance? Synchronicity of `detectChanges`? Coalescing to a single change detection cycle per VM turn?

BioPhoton:
Important to say here is that we can coalesce the detectChanges call within a compoent or embedded view: https://github.com/ngrx/platform/pull/2046#issuecomment-579056504

LayZeeDK:
>within a compoent or embedded view
Why is that, Michael?

LayZeeDK:
>If we run zone-full markDirty takes over this problem and coalesces multiple calls IN a animationFrame or uses A animationFrame to do the chek. (currently, I don't have everything at the top of my head, but @LayZeeDK was so kind and reverse engineered a bit. So the correct information is here.)
However, if we run zone-less, and many motivation of | push especially, and *let, is to enable us easy to write zone-less applications.
>
>When running zone-less Angulars can't detect the changes and so can't coalesce them, so we need to come up with our own strategy.

Not true. `ɵmarkDirty` would work equally well with `NoopNgZone`. `ChangeDetectorRef#markForCheck` would not though.

BioPhoton:
>>within a component or embedded view

>Why is that, Michael?

Please correct me if needed. 
```html
{{observable$ | push}}
```
revieves a different cdRev (EmbeddedView) than
```html
<ng-container *ngIf="observable$ | push as o">
{{o}}
</ng-container>
```
does.

LayZeeDK:
> Please correct me if needed.
> 
> ```
> {{observable$ | push}}
> ```
> 
> revieves a different cdRef (EmbeddedView) than
> 
> ```
> <ng-container *ngIf="observable$ | push as o">
> {{o}}
> </ng-container>
> ```
> 
> does.
> 
> The first is meant by component the second by embedded view.

I just ran another experiment to rediscover what I learned when we were trying to manage change detection without Zone.js.

# In Ivy

| Part | `ChangeDectorRef._cdRefInjectingView` |
| ---- | ---- |
| Component | `LComponentView_*` |
| Pipe in template expression | `LComponentView_*` |
| Pipe in `NgIf` expression | `LEmbeddedView_*` |

They each have a separate instance of a `ChangeDetectorRef` implementation (this is also true in View Engine)

BioPhoton:
@LayZeeDK 

Ha!! I had the right instinct!

> They each have a separate instance of a ChangeDetectorRef implementation (this is also true in View Engine)

What about directives (*ngrxLet)?

Also, I don't see the difference between component and pipe in template expression.

It's hard for me to find but I assume that `ChangeDectorRef` will use `ɵdetectChanges` under the hood. So I will focus to solve everything with cdRef.

LayZeeDK:
*ngrxLet will be similar to *ngIf. No difference between regular template expression or pipe in template expression, it seems. Except that we can't inject a ViewRef in a pipe, isn't that what we discovered?

LayZeeDK:
Everything uses detectChanges in the end. The difference is that the when and where. markDirty results in detectChanges starting on the root component, but scheduled with requestAnimationFrame. detectChanges and ChangeDetectorRef#detectChanges executes immediately, starting at the component in context. ChangeDetectorRef#markForCheck (used by AsyncPipe) doesn't run detectChanges, but relies on detectChanges being scheduled and run at the root component by a subscription to NgZone that is subscribed in ApplicationRef and runs when the NgZone microtask queue is empty.

modules/component/src/index.ts
timdeschryver: Seems like the hooks decorator was already added to the ngrx branch by Brandon.
I will push the change to get the build resolved so it isn't blocking, we'll ask Brandon what the plans are with the hooks decorator.

alexzuza:
Hey @LayZeeDK.

> Everything uses detectChanges in the end. The difference is that the when and where. markDirty results in detectChanges starting on the root component, but scheduled with requestAnimationFrame. detectChanges and ChangeDetectorRef#detectChanges execute immediately, starting at the view in question. ChangeDetectorRef#markForCheck (used by AsyncPipe) doesn't run detectChanges, but relies on detectChanges being scheduled and run from the root view by a subscription to NgZone that is subscribed in ApplicationRef and runs when the NgZone microtask queue is empty.

Agree with that

> I just ran another experiment to rediscover what I learned when we were trying to manage change detection without Zone.js.
> # In Ivy
> Part 	`ChangeDetectorRef._cdRefInjectingView`
> Component 	`LComponentView_*`
> Pipe in template expression 	`LComponentView_*` (the same object as for the component)
> Pipe in `NgIf` expression 	`LEmbeddedView_*`
> 
> They are all separate instances of a `ChangeDetectorRef` implementation (this is also true in View Engine)

Can you please provide a simple example of it?

AFAIK, they are all should be using reference to the same component view. 

Here's is a [demo](https://ng-run.com/edit/vbcTXFp9cVQtsigapDs2?open=app%2Fapp.component.html&aot=true)

In the console we can see:



But I agree they are all different instances. meaning ChangeDetectorRef injected from component !== ChangeDetectorRef injected in Pipe. (Taken from https://github.com/angular/angular/blob/9bd959076730c4e22ceadda73694198b4f01b9e0/packages/core/src/render3/VIEW_DATA.md#expando-and-injecting-special-objects **NOTE: An interesting thing about these objects is that they are not memoized** ).

Some time ago I posted related issue https://github.com/angular/angular/issues/33677 which broke my Pipe

See also https://github.com/angular/angular/pull/31438

BioPhoton:
Thanks for this feedback @alexzuza!
I'll work on the document for coalesceing tomorrow. This input is essential for it!

alexzuza:
@LayZeeDK @BioPhoton 
Seems I found the case when insertion view differs from declaration view

**app.component.html**

```
<insertion [template]="ref"></insertion>

<ng-template #ref>
  <span>{{obs$ | push}}</span>
</ng-template>
 
```

**insertion.component.html**
```
<ng-container [ngTemplateOutlet]="template"> </ng-container>
```




Demo https://ng-run.com/edit/jZXinQ0E4kqtfOOchd0L?open=app%2Fpush.pipe.ts

See also related pull request
https://github.com/angular/angular/pull/33074

LayZeeDK:
I just nested a pipe expression inside the template of a simple *ngIf. Will set up an example later.

alexzuza:
> 
> 
> I just nested a pipe expression inside the template of a simple *ngIf. Will set up an example later.

@LayZeeDK

Got it:

```
<ng-container *ngIf="1">
  {{ obs$ | push }}
</ng-container> 
```

BioPhoton:
Thank's  @alexzuza and @LayZeeDK!!

I started 2 documents summarizing all the collected information.

- [Angular-Change-Detection-in-ViewEngine-and-Ivy.md](https://gist.github.com/BioPhoton/42b5a830fd93b2ef7683f3cb6b9a400d)
- [Angular-ComponentView-and EmbeddedView-in-ViewEngine-and-Ivy.md](https://gist.github.com/BioPhoton/9b63a868d6809d2d0727f61fb6e84d0c)

If you have any examples snippets structure advises information .. any help is welcome. I need some time to understand all the latest findings and prepare for the coalescing Design Doc.

Thanks again for your time and interest! It motivates me a lot to get some help here.

LayZeeDK:
Exactly, @alexzuza. And you say that `_cdRefInjectingView` is used by which API? `ChangeDetectorRef#markForCheck`? How about `ChangeDetectorRef#detect changes`?

BioPhoton:
Hi @LayZeeDK and @alexzuza 

based on the above explorations and some reading in the code I made the following assumption:

- cdRef will stay in Ivy
- cdRef .detectChanges will use `ɵdetectChanges` under the hood

This assumptions lead to following CD handling:

- **ViewEngine** 
  - in zone-full mode `ChangeDetectorRef.markForCheck` 
  - in zone-less mode `ChangeDetectorRef.detectChanges`
- **Ivy**
  - in zone-full mode `ɵmarkDirty` 
  - in zone-less mode `ChangeDetectorRef.detectChanges`

A thing I still need to clarify is if  `ChangeDetectorRef` will use introduce a new method `markDirty` which uses `ɵmarkDirty` under the hood. 

If this is true I could implement it as suggested above and in the future, if the API appears I can refactor to just:

 - **ViewEngine/Ivy** 
  - in zone-full mode `ChangeDetectorRef.markDirty` 
  - in zone-less mode `ChangeDetectorRef.detectChanges`


Is this realistic?

LayZeeDK:
markDirty could probably be used in all cases, but is still private, experimental and not as performant as detectChanges. Buy detectChanges will not work great in all cases.

BioPhoton:
> markDirty could probably be used in all cases
AFAICT not in zone-less mode. And this is the main goal of those to (push. let)

> But *detectChanges will not work if apps rely on NgZone or shared mutable state.
For `NgZone` I implemented a switch. 

For mutable state, they can just switch to CDS.Default instead of OnPush I assume.
IMHO reactive programming in Angular should motivate for immutable programming but should also wort when passing the same instance.

LayZeeDK:
Do you have a demo app so I can prove you wrong? I would think markDirty works for all cases, since it schedules a full change detection cycle from the current component and all the way up to the root component. ChangeDetectorRef#markForCheck would have worked without NgZone if it scheduled a tick like markDirty.

LayZeeDK:
ChangeDetectionStrategy.Default will not work without NgZone.

BioPhoton:
> Do you have a demo app so I can prove you wrong? I would think markDirty works for all cases, since it schedules a full change detection cycle from the current component and all the way up to the root component. ChangeDetectorRef#markForCheck would have worked without NgZone if it scheduled a tick like markDirty.

Got ya!

But still, isn't the detectChanges way noticeable more performant?
I assume even without the coalescing feature?

LayZeeDK:
I would think so.

BioPhoton:
> I would think so. 
I discussed it with @alexzuza and also tested in his tool.

The theory holds. For the coalescing scope, I still have to do a lot of considerations.
But this is an additional feature besides using detectChanges IMHO

LayZeeDK:
Agreed, that optimization option could be added later.

BioPhoton:
@LayZeeDK 

I try to get rid of this line of code.

```typescript 
// Angular 8 View Engine
  if (w.ng && w.ng.probe) {
    return false;
  }
```

If we rely on it I have to implement an extra service to respect SSR and testing.

Do you think there is another way? Maybe also over the AsyncPipe class? Because it is used already

```typescript
// Angular 8 Ivy
  if ((pipe as any).ngPipeDef) {
    return true;
  }

  // Angular 9 Ivy
  if ((pipe as any).ɵpipe) {
    return true;
  } 
```

LayZeeDK:
```
function isIvy(): boolean {
  const ng: any = ((self || global || window) as any).ng;

  return ng === undefined
    || ng.getComponent !== undefined
    || ng.applyChanges !== undefined;
}
```

BioPhoton:
Thanks for your instant answer! 
So you see the window check as essential?

LayZeeDK:
> Thanks for your instant answer!
> So you see the window check as essential?

It's looking for the global `ng` object in different environments.
```typescript
(self || global || window)
```
is a backwards compatible way of resolving the global object (which in modern environments can be accessed using `globalThis`).

I haven't tried it out with server-side rendering or web workers, but web workers usually shouldn't be aware of Ivy or View Engine now that the web worker renderer has been definitively removed.

> Also `ng === undefined || ng.prop !== undefinde `
> should be `ng !== undefined && ng.prop !== undefinde` right?

No, it is actually checking for:
1. "Is the global `ng` object is unavailable" = Ivy production mode since View Engine has the `ng` object both in development mode and production mode.
2. "Is the global `ng` object available with the new debugging API?" = Ivy development mode.

as per the View Engine/Ivy/v8/v9/development mode/production mode overview I made for you:
https://1drv.ms/x/s!Ak8uxvn342s9nxrTjQ41A0iqOXll?e=Nwq79B

> Also, should I consider dev mode vs prod mode?
The check above attempts to detect Ivy both in development mode and production mode.

BioPhoton:
**Update:**
@brandonroberts 
The current state of this PR is ready to review.

modules/component/src/component.module.ts
BioPhoton: done

modules/component/spec/core/operators/processCdAwareObservables.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+    });
+
+    it('should take a promise', () => {
+      const observable: Observable<Promise<any>> = of(Promise.resolve(42));
+      observable
+        .pipe(processCdAwareObservables(o$ => o$, o$ => o$, o$ => o$))
+        .subscribe(v => expect(v).toBe(42));
+    });
+
+    it('should take a undefined', () => {
```
chrisguttandin: I think it should be "should take an undefined". It appears in some other places as well.

modules/component/spec/core/utils/is-zone-less.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { isZoneLess } from '../../../src/core/utils';
+
+class NoopNgZone {}
+
+describe('isZoneLess', () => {
+  it('should return false is something else than noop zone is passed', () => {
+    expect(isZoneLess({})).toBe(false);
+  });
+
+  it('should return false is something else than noop zone is passed', () => {
```
chrisguttandin: This should probably say something like "should return true if a noop zone is passed".

modules/component/src/core/cd-aware.abstract.ts
```diff
@@ -0,0 +1,95 @@
...
+  );
+
+  constructor(
+    protected readonly cdRef: ChangeDetectorRef,
+    protected readonly ngZone: NgZone
+  ) {}
+
+  work(): void {
+    this.handleChangeDetection(
+      // cast is needed to make is work for typescript.
```
chrisguttandin: "make is work" should probably be "make it work"

BioPhoton:
Thanks, @chrisguttandin a lot for your time and interest. 
I updated the sources based on your suggestions.

modules/component/src/component.module.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const DECLARATIONS = [LetDirective, PushPipe];
+export const EXPORTS = [DECLARATIONS];
+
+@NgModule({
+  declarations: [DECLARATIONS],
+  imports: [],
+  exports: [EXPORTS],
+})
+export class ComponentModule {}
```
brandonroberts: As this name is probably being used pretty commonly, we should change this to something like `ReactiveComponentModule`

modules/component/src/let/let.directive.ts
```diff
@@ -0,0 +1,133 @@
...
+      error: error => (this.ViewContext.$error = error),
+      complete: () => (this.ViewContext.$complete = true),
+    };
+  }
+
+  getConfigurableBehaviour<T>(): remainHigherOrder<T> {
+    return pipe(
+      withLatestFrom(this.config$),
+      map(([value$, config]: [Observable<any>, NgRxLetConfig]) => {
+        return !config.optimized
```
brandonroberts: Just leaving a note to say we discussed this, and are going to leave it in the codebase for now.

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.configSubject.next(config);
+    this.observablesSubject.next(obs);
+    return this.renderedValue;
+  }
+
+  getConfigurableBehaviour<T>(): remainHigherOrder<T> {
+    return pipe(
+      withLatestFrom(this.config$),
+      map(([value$, config]) => {
+        return !config.optimized
```
brandonroberts: Just leaving a note to say we discussed this, and are going to leave it in the codebase for now.

modules/component/src/component.module.ts
```diff
@@ -0,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { LetDirective } from './let';
+import { PushPipe } from './push';
+
+export const DECLARATIONS = [LetDirective, PushPipe];
+export const EXPORTS = [DECLARATIONS];
+
+@NgModule({
+  declarations: [DECLARATIONS],
+  imports: [],
```
brandonroberts: nit: remove empty `imports` array

modules/component/spec/let/let.directive.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+let letDirectiveTestComponent: {
+  cfg: CoalescingConfig;
+  value$: Observable<any> | undefined | null;
+};
+let componentNativeElement: any;
+
+describe('LetDirective', () => {
+  describe('when nexting values', () => {
+    beforeEach(async(() => {
+      TestBed.configureTestingModule({
```
brandonroberts: The setup could be wrapped in a helper function, but not necessary to change these

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+import { NextObserver, Observable, PartialObserver, pipe, Subject } from 'rxjs';
+import { distinctUntilChanged, map, tap, withLatestFrom } from 'rxjs/operators';
+import {
+  CdAware,
+  CoalescingConfig as NgRxPushPipeConfig,
+  remainHigherOrder,
+  STATE_DEFAULT,
+} from '../core';
+
+@Pipe({ name: 'ngrxPush', pure: false })
```
brandonroberts: I may have brought this up before, but I think this should just be `push$`
```suggestion
@Pipe({ name: 'push$', pure: false })
```

modules/component/src/index.ts
```diff
@@ -1 +1,3 @@
-export * from './placeholder';
+export { PushPipe } from './push';
+export { LetDirective } from './let';
+export { ComponentModule } from './component.module';
```
brandonroberts:
```suggestion
export { ReactiveComponentModule } from './component.module';
```

modules/component/src/push/push.pipe.ts
alex-okrushko: I think we should stick with "vendor" prefix of "ngrx" to:

- avoid any potential naming collisions
- be clear where it's coming from
- similar to `ng`

brandonroberts: That's fine with me

modules/component/src/core/utils/is-ivy.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  // Is the global ng object is unavailable?
+  // ng === undefined in Ivy production mode
+  // View Engine has the ng object both in development mode and production mode.
+  return (
+    ng === undefined ||
+    // in case we are in dev mode in ivy
+    // `probe` property is available on ng object we use View Engine.
+    ng.probe === undefined
+  );
```
alex-okrushko: these comments are hard to understand for me.
I think this would've helped me
```ts

// Table for ng global presence in ViewEngine and Ivy for prod/dev modes:
//
// | render     |  ViewEngine    |  ViewEngine    |      Ivy          |      Ivy          |     
// | mode       |     prod       |      dev       |      prod         |      dev          |
// | ng         |     present    |     present    |     undefined     |     present       |
// | ng.probe   |     present    |     present    |     undefined     |     undefined     |
//
// So for Ivy we need to make sure that ng is undefined or, 
// in case of dev environment, ng.probe is undefined

return ng?.probe === undefined 
```

modules/component/src/component.module.ts
BioPhoton: regarding prefix:
https://github.com/ngrx/platform/pull/2046#discussion_r373784703

BioPhoton: @brandonroberts 
> As this name is probably being used pretty commonly, we should change this to something like ReactiveComponentModule

It will also affect directives, pipes, and services IMHO. Maybe more general: `ReactiveFeaturesModule`?

modules/component/src/let/let.directive.ts
BioPhoton: Added Comments

modules/component/src/push/push.pipe.ts
BioPhoton: Added Comments

modules/component/src/push/push.pipe.ts
BioPhoton: So I keep it.

modules/component/src/core/cd-aware.abstract.ts
BioPhoton: done

modules/component/spec/core/utils/is-zone-less.spec.ts
BioPhoton: done

modules/component/spec/core/operators/processCdAwareObservables.spec.ts
BioPhoton: done

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
+import { ChangeDetectorRef, NgZone, Pipe, PipeTransform } from '@angular/core';
+import { NextObserver, Observable, PartialObserver, pipe, Subject } from 'rxjs';
+import { distinctUntilChanged, map, tap, withLatestFrom } from 'rxjs/operators';
+import {
+  CdAware,
+  CoalescingConfig as NgRxPushPipeConfig,
```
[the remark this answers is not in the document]
BioPhoton: Just for now. It will get replaced by a real interface
https://github.com/ngrx/platform/pull/2046#discussion_r390936540

modules/component/src/index.ts
BioPhoton: Let's talk there: https://github.com/ngrx/platform/pull/2046#discussion_r391677873

modules/component/src/push/push.pipe.ts
chrisguttandin: These are just my two cents, but shouldn't it be `ngRxPush` since pipe names should be lowerCamelCase? https://angular.io/guide/styleguide#pipe-names

modules/component/spec/let/let.directive.spec.ts
BioPhoton: Done https://github.com/ngrx/platform/pull/2046/commits/e101edd38b51213d5b334608517940b941a2ccda

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+  remainHigherOrder,
+  STATE_DEFAULT,
+} from '../core';
+
+@Pipe({ name: 'ngrxPush', pure: false })
+export class PushPipe extends CdAware implements PipeTransform {
+  private renderedValue: any = STATE_DEFAULT;
+
+  configSubject = new Subject<NgRxPushPipeConfig>();
```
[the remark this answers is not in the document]
BioPhoton: Done

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+  config$ = this.configSubject.asObservable().pipe(distinctUntilChanged());
```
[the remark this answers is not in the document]
BioPhoton: done

modules/component/src/core/utils/is-zone-less.ts
```diff
@@ -0,0 +1,3 @@
+export function isZoneLess(z: any): boolean {
```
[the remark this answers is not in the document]
BioPhoton: Does the function test anything for ngZone?  :D
How can I improve?

modules/component/src/core/utils/get-change-detection-handling.ts
```diff
@@ -0,0 +1,22 @@
...
+} from '@angular/core';
+
+import { isIvy } from './is-ivy';
+import { isZoneLess } from './is-zone-less';
+
+export function getChangeDetectionHandler(
+  ngZone: NgZone,
+  cdRef: ChangeDetectorRef
+): <T>(component?: T) => void {
+  if (isIvy()) {
```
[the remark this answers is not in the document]
BioPhoton: Yes. To implement the zone-less features and configuration. This was also present back then.

modules/component/src/core/utils/get-request-animation-frame-in-angular.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (
+    'requestAnimationFrame' in globalThis &&
+    globalThis.requestAnimationFrame !== undefined
+  ) {
+    return globalThis.requestAnimationFrame;
+  }
+  // if on server return empty function
+  return (cb: Function) => {
+    cb && cb();
+    return Math.random();
```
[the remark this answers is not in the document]
BioPhoton: Added comments:
https://github.com/ngrx/platform/pull/2046/commits/c151a453c6433ee629aea3e42e96b21a84595313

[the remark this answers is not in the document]
BioPhoton: Added comments:
https://github.com/ngrx/platform/pull/2046/commits/c151a453c6433ee629aea3e42e96b21a84595313

modules/component/src/core/utils/is-ivy.ts
BioPhoton: Added the comments.
https://github.com/ngrx/platform/pull/2046/commits/9181a10854ec7848bd4049996fbbfc7a47183b85
Thanks for the table!

modules/component/src/component.module.ts
```diff
@@ -0,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { LetDirective } from './let';
+import { PushPipe } from './push';
+
+export const DECLARATIONS = [LetDirective, PushPipe];
```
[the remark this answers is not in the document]
BioPhoton: done. I think it was related to old problems in builds..

https://github.com/ngrx/platform/pull/2046/commits/241bfddfb908b0e848fb5541fd49cf0b8de7f036

modules/component/src/component.module.ts
BioPhoton: Done:
https://github.com/ngrx/platform/pull/2046/commits/241bfddfb908b0e848fb5541fd49cf0b8de7f036

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+import { distinctUntilChanged, map, tap, withLatestFrom } from 'rxjs/operators';
+import {
+  CdAware,
+  CoalescingConfig as NgRxPushPipeConfig,
+  remainHigherOrder,
+  STATE_DEFAULT,
+} from '../core';
+
+@Pipe({ name: 'ngrxPush', pure: false })
+export class PushPipe extends CdAware implements PipeTransform {
```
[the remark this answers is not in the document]
BioPhoton: I agree totally. In the first session with @brandonroberts, I discussed it. We agreed on keeping it atm and after a release change it.

I'll provide a solution to discuss after resolving the other comments

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.observablesSubject.next(obs);
+    return this.renderedValue;
+  }
+
+  getConfigurableBehaviour<T>(): remainHigherOrder<T> {
+    return pipe(
+      withLatestFrom(this.config$),
+      map(([value$, config]) => {
+        return !config.optimized
+          ? value$.pipe(tap(_ => this.work()))
```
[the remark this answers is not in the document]
BioPhoton: done: 
https://github.com/ngrx/platform/pull/2046/commits/7b8f577c74553c224cb55daf3319bc630e734c7b

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor(cdRef: ChangeDetectorRef, ngZone: NgZone) {
+    super(cdRef, ngZone);
+    this.subscription = this.observables$.subscribe();
+  }
+
+  transform<T>(obj: null | undefined, config?: NgRxPushPipeConfig): null;
+  transform<T>(obj: Observable<T>, config?: NgRxPushPipeConfig): T;
+  transform<T>(
+    obs: Observable<T> | null | undefined,
```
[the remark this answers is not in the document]
BioPhoton: done:
https://github.com/ngrx/platform/pull/2046/commits/0bbc7303b00fe468190a65b6c4484c9e812bb741

modules/component/src/core/utils/typing.ts
```diff
@@ -0,0 +1,46 @@
+import { isObservable, Observable, OperatorFunction } from 'rxjs';
+
+export type potentialObservableValue<T> =
+  | Promise<T>
+  | Observable<T>
+  | undefined
+  | null;
+
+export type remainHigherOrder<T> = (
```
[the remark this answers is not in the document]
BioPhoton: done https://github.com/ngrx/platform/pull/2046/commits/b02b1308942cbf2cdd077e416fbfe6fab1acbef5

modules/component/src/core/cd-aware.abstract.ts
```diff
@@ -0,0 +1,95 @@
...
+// This abstract class holds all the shared logic for the push pipe and the let directive
+// responsible for change detection
+// If you extend this class you need to implement how the update of the rendered value happens.
+// Also custom behaviour is something you need to implement in the extending class
+export abstract class CdAware implements OnDestroy {
+  handleChangeDetection: <T>(component?: T) => void = getChangeDetectionHandler(
+    this.ngZone,
+    this.cdRef
+  );
+  protected requestAnimationFrameRef: (
```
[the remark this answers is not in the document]
BioPhoton: https://github.com/ngrx/platform/pull/2046/commits/4896695b26372caa44f2b07238c463b6f51fc2ba

modules/component/src/core/cd-aware.abstract.ts
```diff
@@ -0,0 +1,95 @@
...
+export interface CoalescingConfig {
+  optimized: boolean;
+}
+
+// This abstract class holds all the shared logic for the push pipe and the let directive
+// responsible for change detection
+// If you extend this class you need to implement how the update of the rendered value happens.
+// Also custom behaviour is something you need to implement in the extending class
+export abstract class CdAware implements OnDestroy {
+  handleChangeDetection: <T>(component?: T) => void = getChangeDetectionHandler(
```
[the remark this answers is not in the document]
BioPhoton: done https://github.com/ngrx/platform/pull/2046/commits/4896695b26372caa44f2b07238c463b6f51fc2ba

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+            );
+      })
+    );
+  }
+
+  getUpdateViewContextObserver(): PartialObserver<any> {
+    return {
+      // assign value that will get returned from the transform function on the next change detection
+      next: (value: any) => (this.renderedValue = value),
+      error: (error: Error) => (this.renderedValue = error.message),
```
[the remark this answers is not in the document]
BioPhoton: done https://github.com/ngrx/platform/pull/2046/commits/d440b73e1b4bd2e407a320c8551fb8c8e475b78d

modules/component/src/core/operators/processCdAwareObservables.ts
```diff
@@ -0,0 +1,32 @@
...
+      resetContextBehaviour,
+      // Add behaviour to apply changes to context for new values
+      updateContextBehaviour,
+      // Add behaviour to apply configurable behaviour
+      configurableBehaviour,
+      // unsubscribe from previous observables
+      // then flatten the latest internal observables into the output
+      // @NOTICE applied behaviour (on the values, not the observable) will fire here
+      switchAll(),
+      // reduce number of emissions to distinct values compared to teh previous one
```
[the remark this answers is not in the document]
BioPhoton: done 
https://github.com/ngrx/platform/pull/2046/commits/a62c141851822814674dfd7df238b2f29c213321

modules/component/src/core/utils/pipe-from-array.ts
```diff
@@ -0,0 +1,20 @@
+import { noop, UnaryFunction } from 'rxjs';
+
+export function pipeFromArray<T, R>(
```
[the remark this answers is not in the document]
BioPhoton: removed it:
https://github.com/ngrx/platform/pull/2046/commits/bdae9fb014cfe843f2ad71762ee04418a7e9621f

modules/component/src/core/operators/processCdAwareObservables.ts
```diff
@@ -0,0 +1,32 @@
...
+  resetContextBehaviour: remainHigherOrder<T>,
+  updateContextBehaviour: remainHigherOrder<T>,
+  configurableBehaviour: remainHigherOrder<T>
+): MonoTypeOperatorFunction<T> {
+  return (o$: Observable<any>): Observable<T> => {
+    return o$.pipe(
+      // Ignore observables of the same instances
+      distinctUntilChanged(),
+      // try to convert it to values, throw if not possible
+      // @TODO fix any type
```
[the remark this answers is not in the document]
BioPhoton: Could you elaborate more?

modules/component/src/core/utils/pipe-from-array.ts
```diff
@@ -0,0 +1,20 @@
...
+  fns: Array<UnaryFunction<T, R>>
+): UnaryFunction<T, R> {
+  if (!fns) {
```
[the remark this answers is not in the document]
BioPhoton: removed:
https://github.com/ngrx/platform/pull/2046/commits/bdae9fb014cfe843f2ad71762ee04418a7e9621f

modules/component/src/core/utils/pipe-from-array.ts
```diff
@@ -0,0 +1,20 @@
...
+    return noop as UnaryFunction<any, any>;
```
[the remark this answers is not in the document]
BioPhoton: removed https://github.com/ngrx/platform/pull/2046/commits/bdae9fb014cfe843f2ad71762ee04418a7e9621f

modules/component/src/core/utils/state-default.ts
```diff
@@ -0,0 +1 @@
+export const STATE_DEFAULT: any = undefined;
```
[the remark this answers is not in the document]
BioPhoton: removed:
https://github.com/ngrx/platform/pull/2046/commits/6f7e0f97a03eaaffedffa2820c382be42af68a84

modules/component/src/core/utils/typing.ts
```diff
@@ -0,0 +1,46 @@
+import { isObservable, Observable, OperatorFunction } from 'rxjs';
+
+export type potentialObservableValue<T> =
```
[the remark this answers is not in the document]
BioPhoton: done:
https://github.com/ngrx/platform/pull/2046/commits/d8d9f1311041289e516c8d3a948517b9ff99a999

modules/component/spec/let/let.directive.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+        declarations: [LetDirectiveTestComponent, LetDirective],
+        providers: [
+          { provide: NgZone, useClass: NgZone },
```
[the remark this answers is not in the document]
BioPhoton: done:
https://github.com/ngrx/platform/pull/2046/commits/d8d9f1311041289e516c8d3a948517b9ff99a999

modules/component/src/core/operators/toObservableValue.ts
```diff
@@ -0,0 +1,27 @@
...
+  isPromiseGuard,
+} from '../utils';
+
+export function toObservableValue<T>(
+  potentialObservableValue$: Observable<T> | Promise<T> | undefined | null
+): Observable<T | undefined | null> {
+  if (
+    potentialObservableValue$ === undefined ||
+    // @NOTICE This check is here to mirror the async pipes behaviour
+    potentialObservableValue$ === null
```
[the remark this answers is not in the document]
BioPhoton: done: https://github.com/ngrx/platform/pull/2046/commits/d419ea71132a125f69d7cba22da9437bf8a0963a

modules/component/src/core/utils/get-request-animation-frame-in-angular.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+  // If function is executed on server or in zone-less mode return normal animationFrameRef
+  if (
+    'requestAnimationFrame' in globalThis &&
+    globalThis.requestAnimationFrame !== undefined
+  ) {
+    return globalThis.requestAnimationFrame;
+  }
+  // if on server return empty function
+  return (cb: Function) => {
```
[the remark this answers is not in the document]
BioPhoton: https://github.com/ngrx/platform/pull/2046/commits/d45a3a90adc37ae9480f51e7b265e91354d64dc1

modules/component/src/core/cd-aware.abstract.ts
```diff
@@ -0,0 +1,95 @@
...
+  Subscription,
+} from 'rxjs';
+import { map, tap } from 'rxjs/operators';
+import { processCdAwareObservables } from './operators';
+
+export interface CoalescingConfig {
+  optimized: boolean;
+}
+
+// This abstract class holds all the shared logic for the push pipe and the let directive
```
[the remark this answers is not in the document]
BioPhoton: To anybody, this is a good place to support IMHO.

@alex-okrushko , could you be so kind and create a change suggestion?

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  configSubject = new Subject<NgRxPushPipeConfig>();
+  config$ = this.configSubject.asObservable().pipe(distinctUntilChanged());
+
+  constructor(cdRef: ChangeDetectorRef, ngZone: NgZone) {
+    super(cdRef, ngZone);
+    this.subscription = this.observables$.subscribe();
+  }
+
+  transform<T>(obj: null | undefined, config?: NgRxPushPipeConfig): null;
```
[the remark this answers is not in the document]
BioPhoton: changed

modules/component/src/core/utils/argument-not-observable-error.ts
```diff
@@ -0,0 +1,24 @@
+// tslint:disable-next-line:no-empty-interface
+export interface ArgumentNotObservableError extends Error {}
+
```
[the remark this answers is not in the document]
BioPhoton: changed error message but keep error because this is a common thing in the future reactive code of this package.

modules/component/src/core/utils/get-global-this.ts
```diff
@@ -0,0 +1,3 @@
+export function getGlobalThis() {
```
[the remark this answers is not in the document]
BioPhoton: I added comments and typing:
https://github.com/ngrx/platform/pull/2046/commits/569f77a13c791a39df1cb3456feff37e5cd1fdd9
If you like you could help with some docs to speed up work?

modules/component/src/core/utils/coalesce-work.ts
```diff
@@ -0,0 +1,54 @@
...
+  cfg: CoalesceConfig = {
+    leading: false,
+    trailing: true,
+    context: (window as unknown) as CoalescingContext,
+    executionContextRef: requestAnimationFrame.bind(window),
+  }
+): CoalesceConfig {
+  return {
+    leading: true,
+    trailing: false,
```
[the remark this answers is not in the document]
BioPhoton: removed:
https://github.com/ngrx/platform/pull/2046/commits/9b016cfe51612d4e81b14004624b0db7a84a1c5b

modules/component/src/core/utils/coalesce-work.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function isScheduling(cfg: CoalesceConfig): boolean {
+  return (
+    cfg.context.executionContextId !== -1 &&
+    cfg.context.executionContextId !== undefined
+  );
+}
+
+export function coalesceWork(work: () => void, cfg: CoalesceConfig) {
```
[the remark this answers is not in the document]
BioPhoton: removed:
https://github.com/ngrx/platform/pull/2046/commits/9b016cfe51612d4e81b14004624b0db7a84a1c5b

modules/component/src/core/utils/coalesce-work.ts
```diff
@@ -0,0 +1,54 @@
...
+  trailing?: boolean;
+  context: CoalescingContext;
+  executionContextRef: (cb: () => void) => number;
+}
+
+export function getCoalesceWorkConfig(
+  cfg: CoalesceConfig = {
+    leading: false,
+    trailing: true,
+    context: (window as unknown) as CoalescingContext,
```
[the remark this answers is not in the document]
BioPhoton: removed:
https://github.com/ngrx/platform/pull/2046/commits/9b016cfe51612d4e81b14004624b0db7a84a1c5b

modules/component/src/core/utils/get-change-detection-handling.ts
```diff
@@ -0,0 +1,22 @@
...
+    return isZoneLess(ngZone) ? detectChanges : markDirty;
```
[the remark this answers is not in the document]
BioPhoton: flipped logic:
https://github.com/ngrx/platform/pull/2046/commits/41dc4639ae8ea6c4993832081760b0871d3a6940

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+import {
+  CdAware,
+  CoalescingConfig as NgRxPushPipeConfig,
+  remainHigherOrder,
+  STATE_DEFAULT,
+} from '../core';
+
+@Pipe({ name: 'ngrxPush', pure: false })
+export class PushPipe extends CdAware implements PipeTransform {
+  private renderedValue: any = STATE_DEFAULT;
```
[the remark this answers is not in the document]
BioPhoton: Nice. If you could sketch it somewhere this would be cool

modules/component/src/push/push.pipe.ts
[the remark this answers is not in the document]
BioPhoton: Needs discussion in slack

modules/component/src/core/cd-aware.abstract.ts
```diff
@@ -0,0 +1,95 @@
...
+// Also custom behaviour is something you need to implement in the extending class
+export abstract class CdAware implements OnDestroy {
+  handleChangeDetection: <T>(component?: T) => void = getChangeDetectionHandler(
+    this.ngZone,
+    this.cdRef
+  );
+  protected requestAnimationFrameRef: (
+    cb: () => void
+  ) => number = getRequestAnimationFrameInAngular().bind(getGlobalThis());
+  protected subscription = new Subscription();
```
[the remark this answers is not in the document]
BioPhoton: removed:
https://github.com/ngrx/platform/pull/2046/commits/9b016cfe51612d4e81b14004624b0db7a84a1c5b

modules/component/src/component.module.ts
BioPhoton: Changed to ReactiveComponentModule:
https://github.com/ngrx/platform/pull/2046/commits/d52467f8a17e784108ce2a0bcc36feacd692a277

modules/component/src/core/operators/processCdAwareObservables.ts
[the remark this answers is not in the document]
BioPhoton: Ok, I need you to help me with typing here. I tried to refactor it but it is not that easy for me ATM.

modules/component/src/core/utils/get-change-detection-handling.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { isIvy } from './is-ivy';
+import { hasZone } from './has-zone';
+
+export function getChangeDetectionHandler(
+  ngZone: NgZone,
+  cdRef: ChangeDetectorRef
+): <T>(component?: T) => void {
+  if (isIvy()) {
+    return !hasZone(ngZone) ? detectChanges : markDirty;
```
tonivj5: Just my two cents: all the places where hasZone is used are negated, it could be improved :+1:

```suggestion
    return hasZone(ngZone) ? markDirty : detectChanges;
```

BioPhoton: done

tonivj5: this is only one, the rest of hasZone's call are negated too

modules/component/src/core/utils/typing.ts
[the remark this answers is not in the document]
alex-okrushko: Same for `null`.

modules/component/src/core/operators/processCdAwareObservables.ts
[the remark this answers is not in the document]
alex-okrushko: Sounds good, I'll try to take a look into that.

modules/component/src/core/utils/argument-not-observable-error.ts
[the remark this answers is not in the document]
alex-okrushko: Could you please explain why the suggested example above does not meet the needs?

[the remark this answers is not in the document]
BioPhoton: removed the file:
https://github.com/ngrx/platform/pull/2046/commits/f47b5ea6e049e032479b72a19656d882e39f196e

modules/component/src/core/utils/typing.ts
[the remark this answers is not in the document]
BioPhoton: done: 
https://github.com/ngrx/platform/pull/2046/commits/3ede9f36743c14f359e2e37b448497c25d509fc5

modules/component/src/push/push.pipe.ts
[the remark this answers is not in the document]
BioPhoton: Docs added:
https://github.com/ngrx/platform/pull/2046/commits/ba79b766f902a510ec7efeeb246b79bee6476157

modules/component/src/core/operators/processCdAwareObservables.ts
```diff
@@ -0,0 +1,32 @@
+import { distinctUntilChanged, map, switchAll } from 'rxjs/operators';
+import { MonoTypeOperatorFunction, Observable } from 'rxjs';
+import { remainHigherOrder } from '../utils';
+import { toObservableValue } from './toObservableValue';
+
+export function processCdAwareObservables<T>(
```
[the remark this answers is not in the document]
BioPhoton: done:
https://github.com/ngrx/platform/pull/2046/commits/ab445b8637c6f962e3de1ee0fa90f62ea32c49fb

modules/component/src/core/cd-aware.abstract.ts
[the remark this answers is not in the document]
BioPhoton: done 
https://github.com/ngrx/platform/pull/2046/commits/ab445b8637c6f962e3de1ee0fa90f62ea32c49fb

modules/component/src/core/operators/processCdAwareObservables.ts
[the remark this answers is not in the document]
BioPhoton: I included your suggestion and removed the separate operator

BioPhoton:
Update:
@brandonroberts Done with resolving the review.
Please let me know how to proceed.

Demo here:
https://github.com/BioPhoton/angular-zone-less-examples

BioPhoton:
Hi @alexzuza,

The issue: https://github.com/angular/angular/issues/33677
is still a blocker. I followed all other links but the smart angular bot closed all issues so I can't follow up on the discussion. :D

There are 2 PRs I found provided a solution for it. None of them merged.

What is your status? Do you know more?

Thanks, Michael!

brandonroberts:
Thanks for the work @BioPhoton! As discussed, we are proceeding with merging this in, and following up with additional PRs for cleanup.

modules/component/src/push/push.pipe.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * Heavy dynamic and interactive UIs suffer from zones change detection a lot and can
+ * lean to bad performance or even unusable applications, but the `async` pipe does not work in zone-less mode.
+ *
+ * `ngrxPush` pipe solves that problem.
+ *
+ * Included Features:
+ *  - Take observables or promises, retrieve their values and render the value to the template
+ *  - Handling null and undefined values in a clean unified/structured way
+ *  - Triggers change-detection differently if `zone.js` is present or not (`detectChanges` or `markForCheck`)
```
eugene-stativka: @BioPhoton PushPipe in zone-less mode uses detectChanges which triggers change detection synchronously. Isn’t it better to markForCheck and schedule CD (appref.tick) on at least the next requestAnimationFrame?

BioPhoton: Hi @eugene-stativka !

Definitely!

I summed everything related to this problem in a document here: 
https://hackmd.io/42oAMrzYReizA65AwQ7ZlQ

Please let me know the feedback here: https://github.com/ngrx/platform/issues/2441

sameh-g:
Is it OK to start use this lib in my Project ?? Since its still under development or ??

PR #2074, opened 2019-08-19
fix(store): ensure type-safety of select()

### LINKED ISSUE #2070: Improve the type safety of select()
It would be nice if 
```
this.store.pipe(select('someProperty'))
```
would throw a compile error if `someProperty` does not exist in the state defined.
Currently, it always falls back to this signature
https://github.com/ngrx/platform/blob/214e7dc3080ec81afff29140882a695db22cc129/modules/store/src/store.ts#L178-L185

[ ] No

macjohnny:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

With e.g. 
```typescript
interface State {
  sub: {
    property: number;
  }
}
```
the invokation of `select()` is allowed with any list of strings (e.g. `select('sub', 'property')`), even if the properties do not exist in the `State` defined, e.g. 
```typescript
select('sub', 'somethingElse')
```

Closes #2070

## What is the new behavior?
The string path parts passed to `select()` must be a key of `State`.
```typescript
select('sub', 'property')
```

```
[x] Yes
[ ] No
```

any non-type-safe invokation of `select()` will throw a compilation error, e.g. `store.pipe(select('sub', 'nonExistentProperty')` will fail if `nonExistentProperty` is not a key of `sub.nonExistentProperty` is not a valid property of the `State`
This also applies to fixed-length tuple types e.g. `[string, string]`, which can be circumvented with a better typing e.g.
```typescript
type MyPath<T, a extends keyof T = keyof T, b extends keyof T[a] = keyof T[a]> = [a, b];
const path: MyPath<State> = ['sub', 'property'];
```

### DIFF
--- a/modules/store/src/store.ts
+++ b/modules/store/src/store.ts
@@ -62,16 +62,29 @@ export class Store<T> extends Observable<T> implements Observer<Action> {
     key5: e,
     key6: f
   ): Observable<T[a][b][c][d][e][f]>;
-  /**
-   * This overload is used to support spread operator with
-   * fixed length tuples type in typescript 2.7
-   */
-  select<K = any>(...paths: string[]): Observable<K>;
-  select<Props = any>(
-    pathOrMapFn: ((state: T, props?: Props) => any) | string,
+  select<
+    T,
+    a extends keyof T,
+    b extends keyof T[a],
+    c extends keyof T[a][b],
+    d extends keyof T[a][b][c],
+    e extends keyof T[a][b][c][d],
+    f extends keyof T[a][b][c][d][e],
+    K = any
+  >(
+    key1: a,
+    key2: b,
+    key3: c,
+    key4: d,
+    key5: e,
+    key6: f,
+    ...paths: string[]
+  ): Observable<K>;
+  select<T, Props = any>(
+    pathOrMapFn: ((state: T, props?: Props) => any) | keyof T,
     ...paths: string[]
   ): Observable<any> {
-    return select.call(null, pathOrMapFn, ...paths)(this);
+    return select.apply(null, arguments as any)(this);
   }
 
   lift<R>(operator: Operator<T, R>): Store<R> {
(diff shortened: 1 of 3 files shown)


modules/store/src/store.ts
```diff
@@ -62,15 +62,28 @@ export class Store<T> extends Observable<T> implements Observer<Action> {
     key5: e,
     key6: f
   ): Observable<T[a][b][c][d][e][f]>;
-  /**
-   * This overload is used to support spread operator with
-   * fixed length tuples type in typescript 2.7
-   */
-  select<K = any>(...paths: string[]): Observable<K>;
+  select<
```
macjohnny: I added this signature to allow invocation with an arbitrary number of path tokens, while avoiding invocations with less than 6 path tokens to fall back to the `select<K = any>(...paths: string[]): Observable<K>` signature if no valid key is passed.

alex-okrushko:
FYI, this PR is on my radar - will take a look whenever I get some free time.

macjohnny:
Let me know whether you agree with the idea, so I will adjust the router module accordingly to have all tests pass.

timdeschryver:
I'm getting an error with this implementation, but this is because `sub` can be `undefined`.


image

Another drawback is that we don't have intellisense:

image


I think this is fixable if we remove/modify the following typing:

This would be a breaking change if someone goes 6 levels deeps in user land.

```ts
export function select<T, Props = any, K = any>(
   propsOrPath: Props,
   ...paths: string[]
): (source$: Observable<T>) => Observable<K>;
```

Next, is to split up the `select` typing with props to one where `Props` is not defined, and another one where it is.

```diff
- export function select<T, Props, K>(
-   mapFn: (state: T, props: Props) => K,
-   props?: Props
- ): (source$: Observable<T>) => Observable<K>;


+ export function select<T, K>(
+  mapFn: (state: T) => K
+ ): (source$: Observable<T>) => Observable<K>;
+ export function select<T, Props, K>(
+   mapFn: (state: T, props: Props) => K,
+   props: Props
+ ): (source$: Observable<T>) => Observable<K>;

```

We can remove the props with string selectors, as this isn't supported:

```diff
export function select<T, a extends keyof T>(
  key: a,
- props: null
): (source$: Observable<T>) => Observable<T[a]>;
```

To be able to remove `Props` from the above typing, we have to modify the implementation's typing:

```diff
export function select<T, Props, K>(
  pathOrMapFn: ((state: T, props?: Props) => any) | string,
- propsOrPath: Props | string,
+ propsOrPath?: Props | string,
  ...paths: string[]
) 
```

This leads to working string selectors but still without intellisense:

image


But there's a problem with the select function on the store, and the only fix that I can see now is to cast it to `any` to make the error disappear..


image


Selectors with props are still working:

image
image

The same counts for function selectors:

image

macjohnny:
@timdeschryver how do you want me to proceed?

timdeschryver:
I think the changes I described are fine, or do you see any problems with them?
Happy to hear @alex-okrushko's thoughts about it.

alex-okrushko:
Sorry for the delay.
My current thinking is that selecting properties by strings is the legacy behavior which is not encouraged. I'm not sure it's worth increasing our code size with the fix.
Also, selecting 'by strings' will not work with some code optimizers that do property renaming.

macjohnny:
Ok, so then I close this PR and the corresponding issue, right?

timdeschryver:
I'm fine with closing this PR and the issue 👍
I think we all agree that "string selectors" are a bad practice, perhaps we should look into the path of removing this functionality?

alex-okrushko:
@macjohnny Thanks again for looking into this problem and creating a PR.

PR #2078, opened 2019-08-21
feat(store): add verbose error message for unprovided feature state in development mode

### LINKED ISSUE #1897: RFC: More verbose error message for unprovided feature state
**TL;DR: Add helpful error when the developer attempts to access feature state that has not been provided.**

As a newer NgRx developer, I occasionally had a bug where I was attempting to access state via a selector, but got a console error `ERROR TypeError: Cannot read property 'books' of undefined`.  The issue was, I was using the selector in a module in the app where the StoreModule had not yet provided with the feature state via `StoreModule.forFeature('books', reducers)`.  This module was usually accessed after navigating to another page first, who's module provided the correct feature state, so it was hard to track down.

My proposal is to add a more verbose error message at the `createFeatureSelector` level to provide some insight to the developer that they likely did not provide the feature State.

## Proposed solution

```typescript
export function createFeatureSelector<T>(
  featureName: string
): MemoizedSelector<object, T>;
export function createFeatureSelector<T, V>(
  featureName: keyof T
): MemoizedSelector<T, V>;
export function createFeatureSelector(
  featureName: any
): MemoizedSelector<any, any> {
  return createSelector(
    (state: any) => {
      const featureState = state[featureName];
      if (featureState === undefined) {
        console.error(`The feature name \"${featureName}\" does not exist in the state's root, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot() or StoreModule.forFeature().`);
      }
      return featureState;
    },
    (featureState: any) => featureState
  );
}
```

[ ] No

jtcrowson:
## PR Checklist

- [n/a] Tests for the changes have been added (for bug fixes / features)
- [n/a] Documentation has been added / updated (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1897 

## What is the new behavior?

Verbose error message when the developer attemps to access a selector without first registering the state in a module.

In the example-app, if we try to get something from books state on the login page:
```ts
export class LoginPageComponent implements OnInit {
  // ...
  constructor(private store: Store<fromAuth.State>, private bookStore: Store<fromBooks.State>) {
    bookStore.pipe(select(fromBooks.getAllBooks)).subscribe(val => console.log(val));
  }
  // ...
}
```
We get an error that accompanies the javascript "undefined" error that will help alert users to this common mistake:
Screen Shot 2019-08-21 at 12 24 31 AM

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -1,4 +1,5 @@
 import { Selector, SelectorWithProps } from './models';
+import { isDevMode } from '@angular/core';
 
 export type AnyFn = (...args: any[]) => any;
 
@@ -274,6 +275,7 @@ export function createSelector<State, S1, S2, S3, S4, S5, S6, Result>(
     Selector<State, S1>,
     Selector<State, S2>,
     Selector<State, S3>,
+
     Selector<State, S4>,
     Selector<State, S5>,
     Selector<State, S6>
@@ -601,8 +603,19 @@ export function createFeatureSelector<T, V>(
 export function createFeatureSelector(
   featureName: any
 ): MemoizedSelector<any, any> {
-  return createSelector(
-    (state: any) => state[featureName],
-    (featureState: any) => featureState
-  );
+  return createSelector((state: any) => {
+    const featureState = state[featureName];
+    if (isDevMode() && featureState === undefined) {
+      console.error(
+        `The feature name \"${featureName}\" does ` +
+          'not exist in the state, therefore createFeatureSelector ' +
+          'cannot access it.  Be sure it is imported in a loaded module ' +
+          `using StoreModule.forRoot('${featureName}', ...) or ` +
+          `StoreModule.forFeature('${featureName}', ...).  If the default ` +
+          'state is intended to be undefined, as is the case with router ' +
+          'state, this development-only error message can be ignored.'
+      );
+    }
+    return featureState;
+  }, (featureState: any) => featureState);
 }


modules/store/src/selector.ts
```diff
@@ -601,8 +603,19 @@ export function createFeatureSelector<T, V>(
...
   featureName: any
 ): MemoizedSelector<any, any> {
-  return createSelector(
-    (state: any) => state[featureName],
-    (featureState: any) => featureState
-  );
+  return createSelector((state: any) => {
+    const featureState = state[featureName];
+    if (isDevMode() && featureState === undefined) {
+      console.error(
```
timdeschryver: At other places we prefix the message to mention it's coming from NgRx (and the package name), and we usually use `console.warn`.

https://github.com/ngrx/platform/blob/8f05f1f0c24bdf84b65d18e243ccf62caf480d20/modules/entity/src/utils.ts#L8

jtcrowson: Changed to warning in 068ed28c14b03fe85bcdaac07610d20bbf77b46c.

brandonroberts:
@jtcrowson will you add a test for this?

jtcrowson:
@brandonroberts good call, added a unit test in fcabf12.

PR #2085, opened 2019-08-25
fix(data): set action input as Action type instead of EntityAction

Odonno:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

In a real-world application, you create effects and filter the `actions$` which is of type `Action` using `ofType` operator. But using the `ofEntityType` (and `ofEntityOp`), the `EntityAction` type is enforced so that using these operators with `actions$` is impossible... 

## What is the new behavior?

Updated the `OperatorFunction<EntityAction, T>` to `OperatorFunction<Action, T>` to fix the problem.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/data/src/actions/entity-action-operators.ts
+++ b/modules/data/src/actions/entity-action-operators.ts
@@ -4,6 +4,7 @@ import { filter } from 'rxjs/operators';
 import { EntityAction } from './entity-action';
 import { EntityOp } from './entity-op';
 import { flattenArgs } from '../utils/utilities';
+import { Action } from '@ngrx/store';
 
 /**
  * Select actions concerning one of the allowed Entity operations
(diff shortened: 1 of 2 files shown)


modules/data/src/actions/entity-action-operators.ts
```diff
@@ -18,29 +19,29 @@ import { flattenArgs } from '../utils/utilities';
...
 export function ofEntityOp<T extends EntityAction>(
   ...allowedEntityOps: any[]
-): OperatorFunction<EntityAction, T> {
+): OperatorFunction<Action, T> {
   const ops: string[] = flattenArgs(allowedEntityOps);
   switch (ops.length) {
     case 0:
       return filter(
-        (action: EntityAction): action is T =>
+        (action: any): action is T =>
```
brandonroberts: Why can't this remain an `EntityAction` as its internal to the function?

Odonno: Because the compiler gives an error if I do so. We now have an `Action` as parameter but this type does not contain the `payload` property. So, the only choice left is to use the `any` type.

brandonroberts:
Closing this for now as there is a workaround

modules/data/src/actions/entity-action-operators.ts
wardbell: I don't understand the use case. I would be extremely reluctant to endorse such a change.

PR #2086, opened 2019-08-25
docs: improve ngrx data guide

SerkanSipahi:
## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

### DIFF
--- a/projects/ngrx.io/content/guide/data/entity-metadata.md
+++ b/projects/ngrx.io/content/guide/data/entity-metadata.md
@@ -135,7 +135,7 @@ The demo uses this helper to create hero and villain filters. Here's how the app
  * matches the case-insensitive pattern.
  */
 export function nameAndSayingFilter(entities: Villain[], pattern: string) {
-  return PropsFilterFnFactory < Villain > ['name', 'saying'](guide/data/entities, pattern);
+  return PropsFilterFnFactory<Villain> ['name', 'saying'](entities, pattern);
 }
 ```
 
--- a/projects/ngrx.io/content/guide/data/entity-metadata.md
+++ b/projects/ngrx.io/content/guide/data/entity-metadata.md
@@ -227,6 +227,7 @@ If the property you want to add comes from `backend`, you will need some additio
 Create a new class `AdditionalPersistenceResultHandler` that `extends DefaultPersistenceResultHandler` and overwrite the [handleSuccess](https://github.com/ngrx/platform/blob/master/modules/data/src/dataservices/persistence-result-handler.service.ts) method, the purpose is to parse the data received from `DataService`, retrieve the additional property, and then save this to the `action.payload`. Here is an example.
 
 ```typescript
+@Injectable()
 export class AdditionalPropertyPersistenceResultHandler extends DefaultPersistenceResultHandler {
   handleSuccess(originalAction: EntityAction): (data: any) => Action {
     const actionHandler = super.handleSuccess(originalAction);
@@ -249,6 +250,7 @@ export class AdditionalPropertyPersistenceResultHandler extends DefaultPersisten
 Following the prior step, we have added the additional property to the `action.payload`. Up next we need to set it to the instance of EntityCollection in the `reducer`. In order to accomplish that, we need to create an `AdditionalEntityCollectionReducerMethods` that `extends EntityCollectionReducerMethods`. In addition, we will need to overwrite the method to match your `action`. For example, if the additional property `foo` is only available in `queryMany action(triggered by EntityCollectionService.getWithQuery)`, we can follow this approach.
 
 ```typescript
+@Injectable()
 export class AdditionalEntityCollectionReducerMethods<T> extends EntityCollectionReducerMethods<T> {
   constructor(public entityName: string, public definition: EntityDefinition<T>) {
     super(entityName, definition);
--- a/projects/ngrx.io/content/guide/data/entity-metadata.md
+++ b/projects/ngrx.io/content/guide/data/entity-metadata.md
@@ -276,7 +276,7 @@ Register `AdditionalPersistenceResultHandler` in `NgModule`,
 
 ```typescript
 @NgModule({
-  { provide: PersistenceResultHandler, useClass: PagePersistenceResultHandler },
+  { provide: PersistenceResultHandler, useClass: AdditionalPropertyPersistenceResultHandler },
 })
 ```
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,6 +82,11 @@ jobs:
             - ~/.cache/yarn
             - ~/.cache/Cypress
             - node_modules
+    # required since `publish-*` jobs have tag filters AND requires `test`
+    # https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag
+    filters:
+      tags:
+        only: /.*/
 
   schematics-core-check:
     <<: *run_in_browser
@@ -211,7 +216,9 @@ jobs:
       - run:
           name: Authenticate with registry
           command: echo "//registry.npmjs.org/:_authToken=$npm_TOKEN" > ~/repo/.npmrc
-      - run: yarn run publish:stable
+      - run:
+          name: Publish stable to npm
+          command: ./node_modules/.bin/ts-node ./build/publish-stable.ts
 
   publish-next:
     <<: *run_in_node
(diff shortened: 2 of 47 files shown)


SerkanSipahi:
Please don't merge this PR yet because I'm currently correcting docs bugs and want to change the documentation so it's easier to understand. I also want to add undocumented features to this guide.

projects/ngrx.io/content/guide/data/entity-metadata.md
```diff
@@ -227,6 +227,7 @@ If the property you want to add comes from `backend`, you will need some additio
 Create a new class `AdditionalPersistenceResultHandler` that `extends DefaultPersistenceResultHandler` and overwrite the [handleSuccess](https://github.com/ngrx/platform/blob/master/modules/data/src/dataservices/persistence-result-handler.service.ts) method, the purpose is to parse the data received from `DataService`, retrieve the additional property, and then save this to the `action.payload`. Here is an example.
 
 ```typescript
+@Injectable()
```
wesleygrimes: Should we add `providedIn: 'root'`?

```suggestion
@Injectable({ providedIn: 'root' })
```

projects/ngrx.io/content/guide/data/entity-metadata.md
```diff
@@ -249,6 +250,7 @@ export class AdditionalPropertyPersistenceResultHandler extends DefaultPersisten
 Following the prior step, we have added the additional property to the `action.payload`. Up next we need to set it to the instance of EntityCollection in the `reducer`. In order to accomplish that, we need to create an `AdditionalEntityCollectionReducerMethods` that `extends EntityCollectionReducerMethods`. In addition, we will need to overwrite the method to match your `action`. For example, if the additional property `foo` is only available in `queryMany action(triggered by EntityCollectionService.getWithQuery)`, we can follow this approach.
 
 ```typescript
+@Injectable()
```
wesleygrimes: Should we add `providedIn: 'root'`?

```suggestion
@Injectable({ providedIn: 'root' })
```

projects/ngrx.io/content/guide/data/entity-metadata.md
SerkanSipahi: No, it might be that the developer is using this service isolated in a lazy module! In this case you don't want it to be added to the application root. The developer should decide what to do with the service. Maybe we should add "providedIn" as an alternative and add a reference to the advantages and disadvantages of  providedIn: 'root'.

projects/ngrx.io/content/guide/data/entity-metadata.md
SerkanSipahi: See above!

wesleygrimes:
@SerkanSipahi Are you planning to add more to this PR? I will hold off on further review until you have pushed everything, if that is the case. Thanks,Wes

SerkanSipahi:
@wesleygrimes Yes, I'm still working on it. I already have a lot of information together. What will come next is the area `@ngrx/data -> Architecture -> Entity Selectors`. Unfortunately I have time for this after work or on weekends. The knowledge about ngrx/data I get from a private project.

yharaskrik:
Hey @SerkanSipahi, thank you for working on the documentation it is much needed! I would like to contribute as well, the largest gap I have found right now is the lack of explanation on how to use reducers with NGRX Data. From what I can see the documentation kind of goes in circles with "See <section>" as an example of how to do something. In `Extension Points` for `Custo Reducers` it says refer to the `Entity Reducer Guide` but then link to 3 other spots for how to override different reducers then leaves out how to `...extend a collection reducer with some additional reducer logic that runs before or after.`. That is specifically what I am looking for as I want to update the `additionalCollectionState` and cannot find a way to do that. 

I would be happy to write up the documentation myself and contribute it to your PR, so that @wardbell or @wesleygrimes can review it, once I know how to do it. Maybe you or someone else are aware of somewhere that has an example or more of an explanation on how to use Reducers with Data?

Thanks for the awesome work done already!

projects/ngrx.io/content/guide/data/entity-metadata.md
wardbell: I'm not clear on your reasoning. What do you mean by "isolation"?

Let me be clear about my thoughts on this:  if the service is a scalar singleton, it should almost always be `providedIn: 'root'`. By scalar singleton I mean that
* scalar: it is not multi-provided
* singleton: there will only ever be one instance of the service in the lifetime of the app.

It does not matter if the service is only used in a lazy loaded module. That is good conceptual scoping. But conceptual scoping is unrelated to `providedIn` which (in this case) is an implementation detail.

Of course there are exceptions. If this case is such an exception, we should add a code comment explaining why we're not using `providedIn` so that future readers don't try to add it.

projects/ngrx.io/content/guide/data/entity-metadata.md
```diff
@@ -249,6 +250,7 @@ export class AdditionalPropertyPersistenceResultHandler extends DefaultPersisten
...
 export class AdditionalEntityCollectionReducerMethods<T> extends EntityCollectionReducerMethods<T> {
```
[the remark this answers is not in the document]
wardbell: Not sure how this class got here (did I write it?).   It's not quite correct as it it mutates the collection (`ec`). That's OK if `ec` is new but, under some circumstances, it might be the original collection.  I'd want to look at that. I'd also like to get rid of the duplicative `any` "casting".  I'll weigh in soon.

wardbell:
We can use help. The matter of overriding reducers really does need attention and I've learned something about that in the year since this was drafted. Thank you for volunteering.

yharaskrik:
@wardbell do you happen to have some kind of basic example on exactly how to extend or override reducers for an ngrx data store? It is pretty easy for a normal ngrx store and the docs are very flushed out just unsure of how exactly it should be done when using ngrx data as you do not register reducers the same way (and don't have direct access to the actions since they are defined internally).

wardbell:
I'm in transit to a client right now so I must reply from memory.

I _might_ have an example in tests. In any case, I DO know that the intention was to override or replace the `EntityCollectionReducerMethods`.

This is designed to be customizable either for _all_ entity types or for a _specific_ type. For example, you might only have that `foo` property for a particular entity type/collection (IIRC, the example assumes a `foo` property for a specific entity collection.

The opportunity here is to demonstrate both approaches.

BTW, looking at the default method of the `queryManySuccess` method, I can see the example implementation is not safe because it WILL mutate the previous collection if the data returned from the server are empty. The default implementation delegates to `EntityChangeTrackerBase.mergeQueryResults` where you'll find:
```
   if (entities == null || entities.length === 0) { // entities are the data from the server
      return collection; // nothing to merge.
    }
```

The example  should be fixed.

projects/ngrx.io/content/guide/data/entity-metadata.md
SerkanSipahi: @wardbell
Lazy-Module:
This was not about lazy loading but about conceptual scoping as you described it. Depending on how (providedIn) and where (lazyModule) it is defined, the service (AdditionalPersistenceResultHandler) can be bundled in the main application or in the lazy-loaded file. Please correct me if I am wrong.

Here you describe how to do ngrx/data with lazy-loaded modules: https://ngrx.io/guide/data/entity-metadata#register-metadata

Isolated:
You're absolutely right about what you described being isolated! Exactly that's was what I meant but unfortunately I haven't described it very well. It was basically about good code organisation. I know the concept of Skalar and Singleton and know how it works in a NgModule.

projects/ngrx.io/content/guide/data/entity-metadata.md
```diff
@@ -227,6 +227,7 @@ If the property you want to add comes from `backend`, you will need some additio
 Create a new class `AdditionalPersistenceResultHandler` that `extends DefaultPersistenceResultHandler` and overwrite the [handleSuccess](https://github.com/ngrx/platform/blob/master/modules/data/src/dataservices/persistence-result-handler.service.ts) method, the purpose is to parse the data received from `DataService`, retrieve the additional property, and then save this to the `action.payload`. Here is an example.
 
 ```typescript
+@Injectable()
 export class AdditionalPropertyPersistenceResultHandler extends DefaultPersistenceResultHandler {
```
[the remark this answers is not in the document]
SerkanSipahi: Your argument makes sense. I will rename it to `CustomPersistenceResultHandler`

projects/ngrx.io/content/guide/data/entity-metadata.md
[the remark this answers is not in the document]
SerkanSipahi: No, that was created by @JiaLiPassion
>Not sure how this class got here (did I write it?)

brandonroberts:
What's the status of this PR? It seems like a relatively small change.

brandonroberts:
Just reopen this PR when you're ready to pick it back up. Thanks!

PR #2099, opened 2019-08-30
docs(store): add note to use resetSelectors to prevent cross-test state bleeding

### LINKED ISSUE #2044: bug(store): StoreModule.forRoot({}) and provideMockStore cross-test state bleeding
https://stackblitz.com/edit/angular-wgwg9m

https://angular-wgwg9m.stackblitz.io/?seed=96859 //success
https://angular-wgwg9m.stackblitz.io/?seed=64919 //failure

## Expected behavior:
Tests are run independently.

## Actual Behavior
Tests are order dependent.

```
Angular: 8.2.0
Ngrx: 8.2.0
```

[x] No

jtcrowson:
## PR Checklist

- [n/a] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2044

## What is the new behavior?

Show a workaround to prevent cross-test state bleeding using `afterAll` + `resetSelectors`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngrx.io/content/guide/store/testing.md
+++ b/projects/ngrx.io/content/guide/store/testing.md
@@ -74,6 +74,12 @@ Usage:
 
 In this example, we mock the `getLoggedIn` selector by using `overrideSelector`, passing in the `getLoggedIn` selector with a default mocked return value of `false`.  In the second test, we use `setResult()` to update the mock selector to return `true`.
 
+<div class="alert is-helpful">
+
+**Note:** You should reset overridden selectors using `resetSelectors()` to prevent other tests from using the overridden selector value.  This is shown in the above example.
+
+</div>
+
 Try the <live-example name="testing-store"></live-example>.
 
 ### Integration Testing
--- a/projects/ngrx.io/content/guide/store/testing.md
+++ b/projects/ngrx.io/content/guide/store/testing.md
@@ -76,7 +76,7 @@ In this example, we mock the `getLoggedIn` selector by using `overrideSelector`,
 
 <div class="alert is-helpful">
 
-**Note:** You should reset overridden selectors using `resetSelectors()` to prevent other tests from using the overridden selector value.  This is shown in the above example.
+**Note:** Reset overridden selectors using the `MockStore.resetSelectors()` method to prevent other tests from using the overridden selector value.  This is shown in the above example.
 
 </div>
 
--- a/projects/ngrx.io/content/examples/testing-store/src/app/auth-guard.service.spec.ts
+++ b/projects/ngrx.io/content/examples/testing-store/src/app/auth-guard.service.spec.ts
@@ -34,4 +34,6 @@ describe('Auth Guard', () => {
 
     expect(guard.canActivate()).toBeObservable(expected);
   });
+
+  afterAll(() => store.resetSelectors());
 });


projects/ngrx.io/content/guide/store/testing.md
```diff
@@ -74,6 +74,12 @@ Usage:
 
 In this example, we mock the `getLoggedIn` selector by using `overrideSelector`, passing in the `getLoggedIn` selector with a default mocked return value of `false`.  In the second test, we use `setResult()` to update the mock selector to return `true`.
 
+<div class="alert is-helpful">
+
+**Note:** You should reset overridden selectors using `resetSelectors()` to prevent other tests from using the overridden selector value.  This is shown in the above example.
```
brandonroberts:
```suggestion
**Note:** Reset overridden selectors using the `MockStore.resetSelectors()` method to prevent other tests from using the overridden selector value.  This is shown in the above example.
```

timdeschryver:
I haven't thought this through, but if we would assume that if you're using `mockSelectors` you shouldn't be using the state (`initialState` nor `mockStore.setState`). Does this make sense?

If it does, our selectors would all be called with `undefined` parameters because there is no state to pass through the selectors. If this is still correct, we could add an extra check based on this assumption to return the overridden result.

```diff
if (
  overrideResult !== undefined
+ && Array.from(arguments).every(arg => arg === undefined)
) {
  return overrideResult;
}
```

Just thinking out loud here because I think we'll all forget to cleanup, and we'll all try to figure out why our tests are randomly failing.
This would still cause for confusion, for example, test A depends on selector SA which is overridden, and test B also depends on SA but forgets to override it it will still have the result of test A in its cache.

Another wild idea is to make the static selectors globally available and to reset the selectors in the "real" store module... Which should solve this problem entirely but at what cost...

jtcrowson:
@timdeschryver I appreciate the discussion here!  My thoughts:

1. **Checking for undefined state to infer usage of mock selectors** - I think this would cover most cases, but there's nothing preventing a developer from using mock store + mock selectors.  If they do, their mock selectors won't work at all.
2. **Making static selectors globally available** - Since the mock selectors map is a property of `MockStore` instead of `Store`, would the StoreModule be able to do that?  If so, the obvious cost is adding more code to the StoreModule to support mock selectors, which is one of the drawbacks of the current mock selectors implementation, IMO.

In general, I doubt many developers are writing integration tests and unit tests for the same component, which I think might favor adding less code to the NgRx code base.

timdeschryver:
@jtcrowson I agree with everything.
I just find a pitty.

jtcrowson:
@timdeschryver are you comfortable approving this change or would you rather consider other options?

timdeschryver:
@jtcrowson yes, I'm good with this PR 👍
I was just thinking out loud

timdeschryver:
What about doing the following in the `MockStore` file 🤔?

```ts
  afterEach(() => {
    try {
      const mockStore = TestBed.get(MockStore) as MockStore<any>;
      if (mockStore) {
        mockStore.resetSelectors();
      }
    } catch {}
  });
}
```

jtcrowson:
@timdeschryver I'm confused, where would that code go?

timdeschryver:
Directly in the `mock_store` file, because the mockstore get's imported it will run the code. (inspiration from React Testing Library - https://github.com/testing-library/react-testing-library/blob/master/src/index.js)

brandonroberts:
I like this idea of providing an auto-cleanup internally if available.

brandonroberts:
We can look into adding the auto-cleanup in a separate PR.

PR #2101, opened 2019-08-31
fix(store): improve consistency of memoized selector result when projection fails

### LINKED ISSUE #2100: Memoized selectors: Inconsistent behavior when projection fails
## Description

The current behavior of memoization as implemented in `defaultMemoize`
can be very surprising in the case of a failed evaluation of the projection.

An error may be thrown in a projector either due to a bug in the projector logic itself (unanticipated cases when the projector is very complex) or when a precondition for the projector is not fulfilled (i.e. inconsistent store state due to a bug elsewhere).   

Let `selector` be any memoizing selector, `GS` a "good state" from which `R` is selected, and `FS` a "bad state" for which the projection fails. The current behavior for three consecutive calls to the selector is (pseudocode): 

```
selector(GS) => returns R
selector(FS) => throws
selector(FS) => returns R
```    

I would expect the third call to the selector to throw as well. There seems to be no good argument why `R` should suddenly be selected from `FS`.
Therefore I suggest a fix to the memoization logic such that:

```
selector(GS) => returns R
selector(FS) => throws
selector(FS) => throws
```

The following minimal (currently failing) testcase reproduces the issue:
```typescript
      const firstState = { ok: 'a' };
      const secondState = { faulty: 'b' };
      const fail = () => { throw new Error(); };
      const selector = createSelector(
        identity,
        (s: any) => s.ok ? s.ok : fail()
      );

      selector(firstState);

      expect(() => selector(secondState)).toThrow(new Error());
      expect(() => selector(secondState)).toThrow(new Error()); // fails here, returns 'a'
```

## Expected behavior:

As described above, I would expect more consistent behavior when the selector is triggered a second time with the same "faulty" state. 

In simple scenarios, the current behavior may often be overlooked because the selector is usually not evaluated again with the same state (due to `distinctUntilChanged` in the store). However, if the selector is nested and only a part of the state changes it may happen of course. Further, if a select fails due to an error in a selector, the observable usually errors out. But when chained with operators like `retry`, the current behavior can obfuscate problems and may be very hard to debug.

Current NgRx master

[ ] No

REPLicated:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

See description in #2100

## What is the new behavior?
Closes #2100

See description in #2100

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/selector.spec.ts
+++ b/modules/store/spec/selector.spec.ts
@@ -41,11 +41,9 @@ describe('Selectors', () => {
     it('should deliver the value of selectors to the projection function', () => {
       const projectFn = jasmine.createSpy('projectionFn');
 
-      const selector = createSelector(
-        incrementOne,
-        incrementTwo,
-        projectFn
-      )({});
+      const selector = createSelector(incrementOne, incrementTwo, projectFn)(
+        {}
+      );
 
       expect(projectFn).toHaveBeenCalledWith(countOne, countTwo);
     });
@@ -53,11 +51,7 @@ describe('Selectors', () => {
     it('should allow an override of the selector return', () => {
       const projectFn = jasmine.createSpy('projectionFn').and.returnValue(2);
 
-      const selector = createSelector(
-        incrementOne,
-        incrementTwo,
-        projectFn
-      );
+      const selector = createSelector(incrementOne, incrementTwo, projectFn);
 
       expect(selector.projector()).toBe(2);
 
@@ -70,11 +64,7 @@ describe('Selectors', () => {
 
     it('should be possible to test a projector fn independent from the selectors it is composed of', () => {
       const projectFn = jasmine.createSpy('projectionFn');
-      const selector = createSelector(
-        incrementOne,
-        incrementTwo,
-        projectFn
-      );
+      const selector = createSelector(incrementOne, incrementTwo, projectFn);
 
       selector.projector('', '');
 
@@ -92,10 +82,7 @@ describe('Selectors', () => {
           return state.unchanged;
         });
       const projectFn = jasmine.createSpy('projectionFn');
-      const selector = createSelector(
-        neverChangingSelector,
-        projectFn
-      );
+      const selector = createSelector(neverChangingSelector, projectFn);
 
       selector(firstState);
       selector(secondState);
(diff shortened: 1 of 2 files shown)


modules/store/spec/selector.spec.ts
```diff
@@ -126,13 +114,27 @@ describe('Selectors', () => {
...
+      };
+      const selector = createSelector(
+        identity,
+        (s: any) => (s.ok ? s.ok : fail())
+      );
+
+      selector(firstState);
+
+      expect(() => selector(secondState)).toThrow(new Error());
+      expect(() => selector(secondState)).toThrow(new Error()); // fails here, returns 'a'
```
timdeschryver: Comment can be left out, as you've fixed it

```suggestion
      expect(() => selector(secondState)).toThrow(new Error());
```

modules/store/spec/selector.spec.ts
```diff
@@ -126,13 +114,27 @@ describe('Selectors', () => {
...
     });
 
+    it('should not memoize last successful projection result in case of error', () => {
+      const firstState = { ok: 'a' };
+      const secondState = { faulty: 'b' };
+      const fail = () => {
+        throw new Error();
+      };
+      const selector = createSelector(
+        identity,
```
timdeschryver: Instead of using `identity`, let's just return state

```suggestion
        state => state,
```

modules/store/spec/selector.spec.ts
```diff
@@ -126,13 +114,27 @@ describe('Selectors', () => {
       expect(projectFn).toHaveBeenCalledTimes(2);
     });
 
+    it('should not memoize last successful projection result in case of error', () => {
+      const firstState = { ok: 'a' };
```
timdeschryver: I think it would be easier to understand the logic if we would just assign `true` or `false` to `ok` here.

```suggestion
      const firstState = { ok: true };
```

modules/store/spec/selector.spec.ts
```diff
@@ -126,13 +114,27 @@ describe('Selectors', () => {
...
+      const secondState = { faulty: 'b' };
```
timdeschryver:
```suggestion
      const secondState = { ok: false };
```

modules/store/spec/selector.spec.ts
timdeschryver: Can we also expect the selector to be called 3 times here? To make sure it does re-execute the projector function.

To be completely sure we should also re-execute the selector with `firstState`, to make sure it still has the previous result memoized.

modules/store/spec/selector.spec.ts
REPLicated: Indeed, thanks.

modules/store/spec/selector.spec.ts
REPLicated: Not sure if I understood you correctly and if it is any better now, please check the updated commit. To verify how many times the selector and its projector are called, I had to introduce some spying.

It now verifies that the selector is called 4 times (this should be obvious as those calls happen directly in the test case - probably not required). The projector is called 3 times: 1x for `firstState`, and 2x for `secondState`. Because the originally memoized value for `firstState` is still valid after 2 failures, the projector is not evaluated again in the end when selecting from that state.

timdeschryver: This is exactly what I meant, thanks 👍 
The test to verify the selector has been called 4 times, may seem redundant but this makes it clear imho.
Thanks!

Could you in the future not force push your new commits and just create a new commit? This makes it easier for us to verify the changes to the original commit. We'll squash the commits into one when we merge to master.

REPLicated: Sorry, I will make sure to preserve a proper commit history for future PRs.

By the way I was surprised that some (untouched, unrelated) code got reformatted. But I believe this was done by a pre-commit hook and not my IDE.

PR #2103, opened 2019-09-01
fix/feat(store): createReducer: on should not be overwritten

### LINKED ISSUE #1956: Docs: Add more explanation that you can only have one distinct action type per on function with createReducer
It is possible to have 2 effects looking at one action. But it seems impossible to have 2 `ons` looking at one action.

https://github.com/ngrx/platform/blob/b07ae4ee835cbc2e9ac9a923c46ca64966c75270/modules/store/src/reducer_creator.ts#L189

It is not obvious and it will override my previous action reducer.

After thinking about it, a selector combining 2 variables will do the job, but it's just not obvious and it does fail quite silently!

zhaosiyang:
createReducer should allow ons to take actions of same type and excute all ons to match the action dispatched

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

If the following is configured and initial state is 0, if `increase` action is dispatched, the next state is 2, because the the second `on` has same action type as the first `on`, and the first `on` is overwritten.
```
const increase = createAction('[COUNTER] INCREASE');
createReducer(0, on(increase, state = state + 1), on(increase, state = state + 2))
```

(Indirectly) Closes #1956 

## What is the new behavior?
In `createReducer`, `on` should not overwrite previous `on`s if they have the same action types.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/reducer_creator.ts
+++ b/modules/store/src/reducer_creator.ts
@@ -229,25 +229,22 @@ export function createReducer<S, A extends Action = Action>(
   initialState: S,
   ...ons: On<S>[]
 ): ActionReducer<S, A> {
-  const map = new Map<string, ActionReducer<S, A>[]>();
+  const map = new Map<string, ActionReducer<S, A>>();
   for (let on of ons) {
     for (let type of on.types) {
       if (map.has(type)) {
-        const existingReducers = map.get(type) as ActionReducer<S, A>[];
-        map.set(type, [...existingReducers, on.reducer]);
+        const existingReducer = map.get(type) as ActionReducer<S, A>;
+        const newReducer: ActionReducer<S, A> = (state, action) =>
+          on.reducer(existingReducer(state, action), action);
+        map.set(type, newReducer);
       } else {
-        map.set(type, [on.reducer]);
+        map.set(type, on.reducer);
       }
     }
   }
 
   return function(state: S = initialState, action: A): S {
-    const reducers = map.get(action.type);
-    return reducers && reducers.length > 0
-      ? reducers.reduce(
-          (previousState, reducer) => reducer(previousState, action),
-          state
-        )
-      : state;
+    const reducer = map.get(action.type);
+    return reducer ? reducer(state, action) : state;
   };
 }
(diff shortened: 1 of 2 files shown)


zhaosiyang:
I have added an example of the necessity of this behaviour here (https://github.com/ngrx/platform/issues/1956#issuecomment-526720340)

timdeschryver:
I'm not sure if this is a "bug".
I think both options are viable, and we'll have to decide which behavior makes the most sense and is safest to use.

zhaosiyang:
> I'm not sure if this is a "bug".
> I think both options are viable, and we'll have to decide which behavior makes the most sense and is safest to use.

Yeah, it could be either feature or bug. 

In my opinion, the `on` should simply mean "execute this function when this action is dispatched" regardless of any other settings. We thus decoupled `on`s in this way.

Some usage case for justifying this feature could be:
1. Update some state (that I defined) whenever some library-defined action is dispatched, for example (`@ngrx/store/init`, `'@ngrx/store/update-reducers` )
2. Create higher order reducers which inject some global `on`s: please see https://github.com/ngrx/platform/issues/1956#issuecomment-526720340 for more detail

However, I can't think of any case that need to overwrite previous `on` if action type is the same.

modules/store/src/reducer_creator.ts
```diff
@@ -229,15 +229,25 @@ export function createReducer<S, A extends Action = Action>(
   initialState: S,
   ...ons: On<S>[]
 ): ActionReducer<S, A> {
-  const map = new Map<string, ActionReducer<S, A>>();
+  const map = new Map<string, ActionReducer<S, A>[]>();
   for (let on of ons) {
     for (let type of on.types) {
-      map.set(type, on.reducer);
+      if (map.has(type)) {
```
MikeRyanDev: You could do the composition on the fly here instead of storing arrays for a small perf bump in the generated reducer:

```ts
if (map.has(type)) {
  const reducer = map.get(type) as ActionReducer<S, A>;
  map.set(type, (state, action) => on.reducer(reducer(state, action), action);
}
else {
  map.set(type, on.reducer);
}
```

zhaosiyang: Thanks @MikeRyanDev for the great suggestion. I have pushed a commit. :)

brandonroberts:
@timdeschryver you have anymore thoughts on this?

PR #2119, opened 2019-09-15
feat(selectors):  deep-cached selectors

### LINKED ISSUE #2104: RFC: feat(selectors): SelectorFactoryWithCache - allow to re-use selectors for props
Memoized Selectors are a powerful way to wrap computed state logic in a simple yet performant functional casing. We love them. 
Selectors with props solve the problem of computing (perhaps expensive) state that relies on runtime supplied values, Howvever when put to test one finds (and can be confused by) the fact that props are checked for referential equality, if the prop type is not scalar (eg object) each new value thrown at the selector will cause all the projection functions to be re-invoked.

Even team members advise to use object props:
https://github.com/ngrx/platform/pull/1233#issuecomment-415301148
which will cause further invocations to loose memoization

And community issues like this arise
https://github.com/ngrx/platform/issues/1415

However sometimes the set of runtime values (props) relevant to a computed state is limited, caching selectors for each prop value allows us to keep the memorization and work done in the past.

For example in our medical imaging app we have a grid of canvases, each one has its own state model which is computed by merging few other states. we have a selector that needs the prop of layoutID (identifying that canvas) which is a runtime value, but has limited range. When a user switches layouts or view modes a lot of scripting is happening, we creating the concept of CachedSelectorFactories to reduce some of that scripting time.

The use of the selector from the component/service side is same as selectors without props 
`this.store.select(selectImageboxModelByLayoutID(layoutID))`

but the selector is created by calling this factory function is cached.
````
export const selectImageboxModelByLayoutID = createSelectorFactoryWithCache((layoutID: string) =>
	createSelector(
		selectFirstModelForImageboxModelLayoutID(layoutID),
        ...furtherSelectorsByLayoutID(layoutID),
		(...projectedModels) => {
			return {...projectedModels}; // further expensive projection code here...
		}
	)
);
````

the signature of that function is this
````
function createSelectorFactoryWithCache<State, Props, Result>(
  selectorFactory: SelectorFactoryByParams<State,Props,Result>
): SelectorFactoryWithParam<State, Props, Result>
````
whilst that types are:
````
interface SelectorFactoryByParams<State,Props,Result> {
  (props: Props): MemoizedSelector<State, Result>;
}
interface SelectorFactoryWithParam<State, Props, Result> extends SelectorFactoryByParams<State,Props,Result>{
  release(): void;

  setResult(value: any): void;
}
````
SelectorFactoryByParams is a factory function creating a selector in the closure of the param.
we only invoke this function if we don't have it cached. I used a Map -> Map<Props, MemoizedSelector<State, Result>> to keep selectors and the release function allows to clear memory when the expected props are not relevant anymore.

The full code is available [here](https://github.com/Algotec/platform/commit/04f1c4af914a107b556c9bde3fd732998a56a1d3) , I will open a PR if/when I get positive input on this.

We have this functionality in an internal lib. It works well for us and allowed us to improve application performance.

**[X]** Yes (Assistance is provided if you need help submitting a pull request)

Will need guidance with regards to tests you'd like to add and perhaps improve typing.

### PR DESCRIPTION
**Pull request by nadavsinai**
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

selectors with props are only cached when the props are equal by reference.
all underlying selectors are "dropped" if the prop is changed.
"deep-caching" allows selector function to cache its previous calls to underlaying selectors/projection when the props are re-used.
 
Closes #2104

## What is the new behavior?
2 proposed APIs are added here which provide the above functionality.

```
[ ] Yes
[x] No
```

## Other information

* *I will rebase and update this PR by comments and directions - 
 I am still looking for feedback and suggestions on how to improving the typing/tests and obviously still need to write docs**

### DIFF
--- a/modules/store/spec/selector.spec.ts
+++ b/modules/store/spec/selector.spec.ts
@@ -7,6 +7,8 @@ import {
   createSelectorFactory,
   resultMemoize,
   MemoizedProjection,
+  createSelectorFactoryWithCache,
+  SelectorFactoryWithParam,
 } from '@ngrx/store';
 import { map, distinctUntilChanged } from 'rxjs/operators';
 
@@ -41,11 +43,9 @@ describe('Selectors', () => {
     it('should deliver the value of selectors to the projection function', () => {
       const projectFn = jasmine.createSpy('projectionFn');
 
-      const selector = createSelector(
-        incrementOne,
-        incrementTwo,
-        projectFn
-      )({});
+      const selector = createSelector(incrementOne, incrementTwo, projectFn)(
+        {}
+      );
 
       expect(projectFn).toHaveBeenCalledWith(countOne, countTwo);
     });
@@ -53,11 +53,7 @@ describe('Selectors', () => {
     it('should allow an override of the selector return', () => {
       const projectFn = jasmine.createSpy('projectionFn').and.returnValue(2);
 
-      const selector = createSelector(
-        incrementOne,
-        incrementTwo,
-        projectFn
-      );
+      const selector = createSelector(incrementOne, incrementTwo, projectFn);
 
       expect(selector.projector()).toBe(2);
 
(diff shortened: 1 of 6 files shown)


brandonroberts:
Going to close this along with the issue for now. We may revisit it at some point.

PR #2133, opened 2019-09-28
feat(data): add entity config in App module declaration

norato:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

On ng-add @ngrx/data we do not create the entity-config file

Closes #

## What is the new behavior?

Create and declare entity-config file to AppModule.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/data/schematics/ng-add/files/entity-metadata.ts.template
+++ b/modules/data/schematics/ng-add/files/entity-metadata.ts.template
@@ -1,0 +1,10 @@
+import { EntityMetadataMap } from '@ngrx/data';
+
+const entityMetadata: EntityMetadataMap = {};
+
+const pluralNames = {  };
+
+export const entityConfig = {
+  entityMetadata,
+  pluralNames
+};
--- a/modules/data/schematics/ng-add/files/entity-metadata.ts.template
+++ b/modules/data/schematics/ng-add/files/entity-metadata.ts.template
@@ -1,10 +1,10 @@
-import { EntityMetadataMap } from '@ngrx/data';
+import { EntityMetadataMap, EntityDataModuleConfig } from '@ngrx/data';
 
 const entityMetadata: EntityMetadataMap = {};
 
 const pluralNames = {  };
 
-export const entityConfig = {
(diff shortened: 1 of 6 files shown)


brandonroberts:
Thanks for the PR @norato! Isn't this pretty much a required step when adding `@ngrx/data`? I don't think it needs to be behind a flag.

norato:
Hello, @brandonroberts.
The entity-config is a required configuration to `@ngrx/data` works, as at the [documentation](https://next.ngrx.io/guide/data#registering-the-entity-store) we need to declare it to include in the store.

The `ng-add` only includes the `EntityDataModule` or `EntityDataModuleWithouEffects` at AppModule, but do not declare the entityConfig.

modules/data/schematics/ng-add/files/entity-metadata.ts.template
```diff
@@ -0,0 +1,10 @@
+import { EntityMetadataMap } from '@ngrx/data';
+
+const entityMetadata: EntityMetadataMap = {};
+
+const pluralNames = {  };
+
+export const entityConfig = {
```
timdeschryver: I think it would be better to provide a type for this.

```suggestion
export const entityConfig: EntityDataModuleConfig = {
```

timdeschryver:
As the `config` is required I would make this the default and not provide a flag for it, as @brandonroberts mentioned.

modules/data/schematics/ng-add/index.ts
```diff
@@ -260,6 +293,8 @@ export default function(options: EntityDataOptions): Rule {
     const parsedPath = parseName(options.path, '');
     options.path = parsedPath.path;
 
+    const templateSource = createEntityConfigFile(options, parsedPath.path);
```
timdeschryver: I would rather stay consistent with the rest of the schematic and "inline" this as:

```ts
options.entityConfig
  ? createEntityConfigFile(options, parsedPath.path)
  : noop();
```

PR #2145, opened 2019-10-03
fix(schematics): improve error message when no project can be determined

### LINKED ISSUE #2004: Schematics: Schematic fails with cryptic error when a project is not specified
Copied from: https://github.com/nrwl/nx/issues/1545

_[Please make sure you have read the submission guidelines before posting an issue](https://github.com/nrwl/nx/blob/master/CONTRIBUTING.md#-submitting-issue)_

## Expected Behavior
Please describe the behavior you are expecting

When running the below command without a `--project` flag, I expect the error message to reflect the missing flag, instead of the message `Cannot read property 'root' of undefined`, as it is non-descriptive. The command I ran:

```sh
$ ng generate @ngrx/schematics:store State --root --module app.module.ts
```

The message could be improved by something like:

```
Could not determine the Angular project. Please provide a '--project' flag.
```

## Current Behavior

What is the current behavior?

The current behavior is that when running the above command, it **does not** generate the expected files. Instead, it presents the non-descriptive error message `Cannot read property 'root' of undefined`. When running the below command, it **does** generate the expected files.

```sh
$ ng generate @ngrx/schematics:store State --root --module app.module.ts --project my-angular-project
```

Note that `my-angular-project` should obviously be defined in you `angular.json` file.

## Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Create a new Nx Workspace, say `my-workspace` and be sure to include an Angular project
2. Step 1 generates the Angular files. Now, delete the full Angular app and create one with a different name, say `my-different-angular-project`.
3. Now, generate the NgRx root store using the command:

```sh
$ ng generate @ngrx/schematics:store State --root --module app.module.ts
```

4. Now, you should see the error.

### Context

Please provide any relevant information about your setup:

- version of Nx used
**"@nrwl/workspace": "8.2.0"**
- version of Angular CLI used
**@angular/cli@8.0.1**
- `angular.json` configuration
**Can't share**
- version of Angular DevKit used
**"@angular-devkit/build-angular": "^0.800.0",**
- 3rd-party libraries and their versions
**@ngrx/schematics": "^8.0.1"**
- and most importantly - a use-case that fails
**See the above reproduction steps.**

**A minimal reproduce scenario using allows us to quickly confirm a bug (or point out coding problem) as well as confirm that we are fixing the right problem.**

### Failure Logs

Please include any relevant log snippets or files here.
image

## Other

Any other relevant information that will help us help you.
According to [this issue](https://github.com/nrwl/nx/issues/641), it could be caused by a mismatch between the Angular Project name in `angular.json` and the name used in `package.json`, but I'm not sure about that.

chrstnbrn:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

See description in #2004 

## What is the new behavior?

Closes #2004

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/data/schematics-core/utility/project.ts
+++ b/modules/data/schematics-core/utility/project.ts
@@ -1,4 +1,4 @@
-import { getWorkspace } from './config';
+import { getWorkspace, WorkspaceSchema } from './config';
 import { Tree } from '@angular-devkit/schematics';
 
 export interface WorkspaceProject {
@@ -6,12 +6,10 @@ export interface WorkspaceProject {
   projectType: string;
 }
 
-export function getProject(
-  host: Tree,
+function getProjectName(
+  workspace: WorkspaceSchema,
   options: { project?: string | undefined; path?: string | undefined }
-): WorkspaceProject {
-  const workspace = getWorkspace(host);
-
+): string {
   if (!options.project) {
     options.project =
       workspace.defaultProject !== undefined
@@ -19,7 +17,24 @@ export function getProject(
         : Object.keys(workspace.projects)[0];
   }
 
-  return workspace.projects[options.project];
+  return options.project;
+}
+
+export function getProject(
+  host: Tree,
+  options: { project?: string | undefined; path?: string | undefined }
+): WorkspaceProject {
+  const workspace = getWorkspace(host);
+  const projectName = getProjectName(workspace, options);
+  const project = workspace.projects[projectName];
+
+  if (!project) {
+    throw new Error(
+      `Project ${projectName} does not exist. Please provide a valid project.`
+    );
+  }
+
+  return project;
 }
 
 export function getProjectPath(
--- a/modules/effects/schematics-core/utility/project.ts
+++ b/modules/effects/schematics-core/utility/project.ts
@@ -1,4 +1,4 @@
-import { getWorkspace } from './config';
+import { getWorkspace, WorkspaceSchema } from './config';
 import { Tree } from '@angular-devkit/schematics';
 
 export interface WorkspaceProject {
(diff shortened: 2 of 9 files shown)


timdeschryver:
Would it be possible to mention which project could not be found? 
It would give a direction on where to look at.

chrstnbrn:
@timdeschryver Yes, I agree that displaying the project name would be helpful for the user. I changed the code accordingly.

brandonroberts:
I'd like to add a check here to see if they have provided `options.path`. I have some custom schematics that don't provide a project, but provide the full path to the location, and those shouldn't break.

chrstnbrn:
@brandonroberts If I understand you correctly your schematics are using the `getProject` function. So the problem is that the function now throws an error instead of returning `undefined` if the project does not exist. We can check if `options.path` was provided before throwing the error but then `getProjectPath` and `isLib` may break anyway because right now they access `project` without checking if it exists (leading to the `Cannot read property x` of undefined error from the original issue). Maybe we shouldn't change the behavior of `getProject` at all. Instead `getProjectPath` and `isLib` could call another function that throws the error. What do you think?

timdeschryver:
Correct me if I'm wrong, but when you do `ng generate store --path=./foo/bar/baz` it should not throw an error because the project was not defined.

I think this proposal in the PR will throw in this case.
We could simply return the provided if it's provided inside `getProjectPath` - if it's not provided we build up the path with the added check introduced in this PR.

brandonroberts:
Just reopen this PR when you're ready to pick it back up. Thanks!

PR #2148, opened 2019-10-04
feat(store): add refreshState method to mock store

### LINKED ISSUE #2121: Re-triggering mock selectors

If we override the selector's result during a test, we have to use `mockStore.setState({ random: 'data'})` (maybe we could add this into `MockStore`, @tomastrajan coined `mockStore.triggerSelectors()`?). Because of this we get a new state reference and our selectors will re-execute.

It seems like component only gets the new data if we use the `select` pipe and not the `select` method on the `store`.

```
// fails
this.todos = this.store.select(fromTodos.getAllTodos);

// succeeds
this.todos = this.store.pipe(select(fromTodos.getAllTodos));
```

```ts
it('should render todo items', done => {
    mockSelectorGetAllTodos.setResult([{ name: 'bbb', done: true }]);
    mockStore.setState({ dummy: true });

    fixture.detectChanges();
    setTimeout(() => {
      expect(getTodoItems().length).toBe(1);
      expect(getTodoItems()[0].nativeElement.textContent.trim()).toBe('bbb'); // fails here
      done();
    });
  });
```

A reproduction can be found in @tomastrajan's repo:
[Component](
https://github.com/tomastrajan/ngrx-test/blob/ab345628aee25af304e424cc326ef21db38672b8/src/app/features/todos/todos.component.spec.ts#L50) and [Test](https://github.com/tomastrajan/ngrx-test/blob/ab345628aee25af304e424cc326ef21db38672b8/src/app/features/todos/todos.component.spec.ts#L50)

## Expected behavior:

It should work in both cases.

8.2

brandonroberts:
The MockStore.refreshState() method emits a clone of the existing
state on the state observable, that will cause the selectors to recompute.
This enables multiple calls of setResult for mocked selectors during testing.
Also fixes a bug with store.select not refreshing mocked selectors

Closes #2121

## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2121 

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/testing/src/mock_store.ts
+++ b/modules/store/testing/src/mock_store.ts
@@ -36,6 +36,7 @@ export class MockStore<T> extends Store<T> {
   >();
 
   public scannedActions$: Observable<Action>;
+  private lastState: T | undefined;
 
   constructor(
     private state$: MockState<T>,
@@ -47,6 +48,7 @@ export class MockStore<T> extends Store<T> {
     super(state$, actionsObserver, reducerManager);
     this.resetSelectors();
     this.state$.next(this.initialState);
+    this.lastState = this.initialState;
     this.scannedActions$ = actionsObserver.asObservable();
     if (mockSelectors) {
       mockSelectors.forEach(mockSelector => {
@@ -62,6 +64,7 @@ export class MockStore<T> extends Store<T> {
 
   setState(nextState: T): void {
     this.state$.next(nextState);
+    this.lastState = nextState;
   }
 
   overrideSelector<T, Result>(
@@ -108,7 +111,7 @@ export class MockStore<T> extends Store<T> {
   }
 
   select(selector: any, prop?: any) {
-    if (MockStore.selectors.has(selector)) {
+    if (typeof selector === 'string' && MockStore.selectors.has(selector)) {
       return new BehaviorSubject<any>(
         MockStore.selectors.get(selector)
       ).asObservable();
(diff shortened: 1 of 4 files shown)


brandonroberts:
cc: @tomastrajan

tomastrajan:
@brandonroberts thx mate, sounds amazing !

modules/store/testing/src/mock_store.ts
```diff
@@ -47,6 +48,7 @@ export class MockStore<T> extends Store<T> {
     super(state$, actionsObserver, reducerManager);
     this.resetSelectors();
     this.state$.next(this.initialState);
+    this.lastState = this.initialState;
```
timdeschryver: what about calling  `setState(this.initialState)` here?

modules/store/testing/src/mock_store.ts
```diff
@@ -36,6 +36,7 @@ export class MockStore<T> extends Store<T> {
   >();
 
   public scannedActions$: Observable<Action>;
+  private lastState: T | undefined;
```
timdeschryver:
```suggestion
  private lastState: T;
```

I think `lastState` will always be provided (or is an empty object)?
Because of `config.initialState || {}`

brandonroberts: Yep. I initially didn't have the `|| {}` there.

modules/store/testing/src/mock_store.ts
brandonroberts: Yep

PR #2151, opened 2019-10-06
Rename selectors example app

### LINKED ISSUE #2142: Selector naming convention
In the example app, I have noticed that the selectors are sometime starting with `select` and sometimes starting with `get`. And in this file we find a mix of both: 

https://github.com/ngrx/platform/blob/master/projects/example-app/src/app/auth/reducers/index.ts

So what is the way going forward, can you point me to a documentation for it, or reasonings for starting with `get` over `select` or vice versa.

jordanpowell88:
## PR Checklist

- [X ] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ X] Documentation content changes
[ ] Other... Please describe:
```

names selectors with `get` naming convention
Closes #2142 

## What is the new behavior?
names selectors with `select` naming convention

```
[ ] Yes
[X ] No
```


wesleygrimes:
Thanks for the PR Jordan! I see where you are going, but if we renamed these selectors to use `select` instead of `get` we would need to rename all of the other selectors in the application. `Get` is used throughout. Personally I have always followed the convention of `get*` for private helpers and then `select*` for anything that is an actual selector.

I think it would be a good change to start introducing best practices. If so we would need to change it everywhere, and also change it in the `ngrx.io` docs and readmes.

@brandonroberts thoughts?

brandonroberts:
I don't see an issue with using `get*` for private helpers, but I use `select*` across the board for selectors, whether they be getter functions, or the exported selectors. The example app could use some refactoring in this regard just for consistency.

wesleygrimes:
@jordanpowell88 do you want to proceed with renaming the remaining selectors in the example app and docs to use the select prefix?

jordanpowell88:
@wesleygrimes Yeah I can rename the rest of the selectors to keep it consistent.

brandonroberts:
@jordanpowell88 you will need to update the example snapshots where necessary. Run `yarn example:test`to run them

PR #2160, opened 2019-10-10
feat(schematics): selector schematics

### LINKED ISSUE #2140: Schematics: Generate selectors file as part of feature schematic
It's common practice to have a separate `{feature-name}.selectors.ts` file alongside your actions, reducers, etc. The `feature` schematic would add a flag to generate this file and it's tests in addition to the current set, and eventually become the default.

We could optionally do this for the `entity` schematic also.

At a minimum, the selectors file would:

- Import the usual `createSelector` functions
- Import the reducer exports
- Include a selector that uses `createFeatureSelector` to select that feature key

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2140 

## What is the new behavior?

This PR adds selector schematics with:

```bash
ng g selector selectorName
ng g s selectorName
ng g feature featureName
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/schematics/collection.json
+++ b/modules/schematics/collection.json
@@ -58,7 +58,7 @@
     },
 
     "selector": {
-      "aliases": ["s"],
+      "aliases": ["se"],
       "factory": "./src/selector",
       "schema": "./src/selector/schema.json",
       "description": "Add selectors"
--- a/modules/schematics/collection.json
+++ b/modules/schematics/collection.json
@@ -57,6 +57,13 @@
       "description": "Adds initial setup for state management"
     },
 
+    "selector": {
+      "aliases": ["s"],
+      "factory": "./src/selector",
+      "schema": "./src/selector/schema.json",
+      "description": "Add selectors"
+    },
+
     "ng-add": {
       "aliases": ["init"],
       "factory": "./src/ng-add",
--- a/modules/schematics/src/selector/files/__name@dasherize@if-flat__/__name@dasherize__.selectors.spec.ts.template
+++ b/modules/schematics/src/selector/files/__name@dasherize@if-flat__/__name@dasherize__.selectors.spec.ts.template
@@ -1,0 +1,12 @@
+<% if(feature) { %>import { <%= camelize(name) %>FeatureKey } from '<%= reducerPath %>';
+import { select<%= classify(name) %>State } from './<%= dasherize(name) %>.selectors';<% } %>
+
+describe('<%= classify(name) %> Selectors', () => {
+  it('should select the feature state', () => {
+    <% if(feature) { %>const result = select<%= classify(name) %>State({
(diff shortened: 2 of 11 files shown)


modules/schematics/src/selector/files/__name@dasherize@if-flat__/__name@dasherize__.selectors.spec.ts.template
```diff
@@ -0,0 +1,12 @@
+<% if(feature) { %>import { <%= camelize(name) %>FeatureKey } from '<%= reducerPath %>';
```
brandonroberts: This should come out to `import * as fromFoo from 'fooReducerPath';`

modules/schematics/src/selector/files/__name@dasherize@if-flat__/__name@dasherize__.selectors.ts.template
```diff
@@ -0,0 +1,6 @@
+import { createFeatureSelector, createSelector } from '@ngrx/store';
+<% if(feature) { %>import { State, <%= camelize(name) %>FeatureKey } from '<%= reducerPath %>';
```
brandonroberts: Same as above `import * as fromFoo ...`

modules/schematics/src/selector/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    );
+  };
+}
+
+function relativePath(options: SelectorOptions) {
+  if (options.feature) {
+    return stringUtils.featurePath(
+      options.group,
+      options.flat,
+      'reducers',
```
brandonroberts:
```suggestion
      'selectors',
```

timdeschryver: This should stay as `reducers`, we're building up the path to the reducers file here.
The selector spec file will always live next to the selector file so this import will always be `./selector-featureName`.
For reducers we have to navigate to the correct folder because this may vary based on the `group` and `flat` flags.

If we would change this, we would import the reducer as:
```
import * as fromFoo from '../../selectors/foo/foo.reducer'
```

modules/schematics/collection.json
```diff
@@ -57,6 +57,13 @@
       "description": "Adds initial setup for state management"
     },
 
+    "selector": {
+      "aliases": ["s"],
```
brandonroberts: Let's change the alias to `se`, the `s` alias is already being used for services. https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/collection.json#L87

timdeschryver: Good catch, will do 👍

PR #2163, opened 2019-10-10
fix(store): improve createFeatureSelector warning

### LINKED ISSUE #2116: New warning about unprovided feature state wrecks unit test output
We have unit tests where we use the state without mocking. With the new warning about undefined feature state, that is given while it shouldn't in case of the router state, our unit test are now littered with this warning. Also it shows up in the debugger console when we run the app.

Warning:  

> The feature name "router" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('router', ...) or StoreModule.forFeature('router', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.

Also see the discussion here: [Issue 1897](https://github.com/ngrx/platform/issues/1897)

image

## Expected behavior:
I think it is bad practice to give warnings that should be ignored in some cases. This leads to warning/error fatigue. Also unit tests that have hundreds or even thousands of warnings quickly lose their usefulness.

So please find a way to not warn when there is nothing to warn about, like in the case of the router state. Or add the ability to suppress the warnings for a user provided key, or even remove this warning completely as it does seem to do more harm than good.

NgRx 8.3
Ng 8.2.5

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2116 

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -605,7 +605,7 @@ export function createFeatureSelector(
 ): MemoizedSelector<any, any> {
   return createSelector((state: any) => {
     const featureState = state[featureName];
-    if (isDevMode() && featureName !== 'router' && !(featureName in state)) {
+    if (isDevMode() && !(featureName in state)) {
       console.warn(
         `@ngrx/store: The feature name \"${featureName}\" does ` +
           'not exist in the state, therefore createFeatureSelector ' +
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -605,9 +605,9 @@ export function createFeatureSelector(
 ): MemoizedSelector<any, any> {
   return createSelector((state: any) => {
     const featureState = state[featureName];
-    if (isDevMode() && featureState === undefined) {
+    if (isDevMode() && featureName !== 'router' && !(featureName in state)) {
       console.warn(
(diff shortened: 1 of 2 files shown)


modules/store/src/selector.ts
```diff
@@ -605,9 +605,9 @@ export function createFeatureSelector(
 ): MemoizedSelector<any, any> {
   return createSelector((state: any) => {
     const featureState = state[featureName];
-    if (isDevMode() && featureState === undefined) {
+    if (isDevMode() && featureName !== 'router' && !(featureName in state)) {
```
Halt001: May I ask what the added benefit is of having `featureName !== 'router'`? The ` !(featureName in state))` part already catches the initial state of undefined for the router.

timdeschryver: It's unneeded.
Good catch, thanks!

jtcrowson:
This is awesome @timdeschryver!  Thank you  🎉

timdeschryver:
It was a team effort @jtcrowson 💪

PR #2164, opened 2019-10-14
fix(effects): dispatch init action once

### LINKED ISSUE #2106: Lazy loaded effect initialized multiple times if used in multiple routes
## Expected behavior:

I have a lazy loaded module that imports EffectsModule.forFeature(...) and is accessible from multiple routes within the application.  In other words loadChildren is defined for this module multiple times in the app-routing.module.ts file:

/a/b/feature
/c/d/feature

When configured like this ngrxOnInitEffects is triggered multiple times when visiting each route. I would have expected it to only execute one time.

If this is intended what is the suggested approach to ensuring that init code would only run one time?

angular @ 8.1.1
ngrx @ 8.1.0

timdeschryver:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2106

## What is the new behavior?

See the breaking change section below

```
[x] Yes
[ ] No
```

BREAKING CHANGE

Before
When the effect class was registered, the init action would be dispatched.
If the effect was provided in multiple lazy loaded modules, the init action would be dispatched for every module.

After
The init action is only dispatched once
The init action is now dispatched based on the identifier of the effect (via ngrxOnIdentifyEffects)

### DIFF
--- a/modules/effects/spec/effect_sources.spec.ts
+++ b/modules/effects/spec/effect_sources.spec.ts
@@ -1,8 +1,16 @@
 import { ErrorHandler } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { cold, hot, getTestScheduler } from 'jasmine-marbles';
-import { concat, NEVER, Observable, of, throwError, timer } from 'rxjs';
-import { map } from 'rxjs/operators';
+import {
+  concat,
+  NEVER,
+  Observable,
+  of,
+  throwError,
+  timer,
+  Subject,
+} from 'rxjs';
+import { map, mapTo } from 'rxjs/operators';
 
 import {
   Effect,
@@ -10,9 +18,11 @@ import {
   OnIdentifyEffects,
   OnInitEffects,
   createEffect,
+  Actions,
 } from '../';
 import { EffectsRunner } from '../src/effects_runner';
 import { Store } from '@ngrx/store';
+import { ofType } from '../src';
 
 describe('EffectSources', () => {
   let mockErrorReporter: ErrorHandler;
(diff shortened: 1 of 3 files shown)


modules/effects/src/effect_sources.ts
```diff
@@ -47,7 +41,21 @@ export class EffectSources extends Subject<any> {
...
     return this.pipe(
       groupBy(getSourceForInstance),
-      mergeMap(source$ => source$.pipe(groupBy(effectsInstance))),
+      mergeMap(source$ => {
+        return source$.pipe(
+          groupBy(effectsInstance),
+          tap(() => {
+            if (
+              onInitEffects in source$.key &&
+              typeof source$.key[onInitEffects] === 'function'
```
alex-okrushko: I think we should push this logic back into lifecycle_hooks.ts and expose it has `isOnInitEffects(key): key is OnInitEffects` type guard

modules/effects/src/effect_sources.ts
```diff
@@ -47,7 +41,21 @@ export class EffectSources extends Subject<any> {
...
+            ) {
+              this.store.dispatch(
+                source$.key[onInitEffects].bind(source$.key)()
```
alex-okrushko: this would be the same as `((key: OnInitEffects) => key.ngrxOnInitEffects())(source$.key)`, right?

Do we need that much complexity here? maybe just `this.store.dispatch(source$.key.ngrxOnInitEffects())`? (after we get it through type-guard)

modules/effects/spec/effect_sources.spec.ts
```diff
@@ -51,15 +56,52 @@ describe('EffectSources', () => {
...
+        return { type: '[EffectWithInitAction] Init' };
+      }
+    }
+    const store = TestBed.get(Store);
+
+    effectSources.addEffects(new EffectWithInitAction());
+    effectSources.addEffects(new EffectWithInitAction());
+
+    expect(store.dispatch).toHaveBeenCalledTimes(1);
+  });
```
alex-okrushko: Can we please add a similar test but where effects class actually has 2 effects? (and should be called once)

modules/effects/src/effect_sources.ts
timdeschryver: I'll do this in another PR if that's OK.
Currently, all checks on lifecycle hooks are done in a similar way, in the next PR we can refactor all of them to type guards.

modules/effects/src/effect_sources.ts
timdeschryver: Great catch, I did it to bind `this` to the effect instance inside the `isOnInitEffects` function of the effect.

But this won't be possible AFAIK because we're using `Object.getPrototypeOf` to group the effect instances.

modules/effects/spec/effect_sources.spec.ts
timdeschryver: I addressed the feedback - thanks!

PR #2165, opened 2019-10-14
fix(effects): resubscribe every time an error occurs

timdeschryver:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

## Other information

Seems like with the original PR we forgot to modify the spec for the usage with `createEffect`. This PR uses the same implementation as the original PR.

History:

- #2026 - original PR with the fix
- #2037  - revert caused by inconsistant CI

### DIFF
--- a/modules/effects/src/effects_resolver.ts
+++ b/modules/effects/src/effects_resolver.ts
@@ -24,14 +24,19 @@ export function mergeEffects(
           ? sourceInstance[propertyName]()
           : sourceInstance[propertyName];
 
+      const resubscribeInCaseOfError = (
+        observable$: Observable<any>
+      ): Observable<any> =>
+        observable$.pipe(
+          catchError(error => {
+            if (errorHandler) errorHandler.handleError(error);
+            // Return observable that produces this particular effect
+            return resubscribeInCaseOfError(observable$);
+          })
+        );
+
       const resubscribable$ = resubscribeOnError
-        ? observable$.pipe(
-            catchError(error => {
-              if (errorHandler) errorHandler.handleError(error);
-              // Return observable that produces this particular effect
-              return observable$;
-            })
-          )
+        ? resubscribeInCaseOfError(observable$)
         : observable$;
 
       if (dispatch === false) {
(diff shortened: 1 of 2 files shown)


modules/effects/src/effects_resolver.ts
```diff
@@ -24,14 +24,19 @@ export function mergeEffects(
...
+      const resubscribeInCaseOfError = (
+        observable$: Observable<any>
+      ): Observable<any> =>
+        observable$.pipe(
+          catchError(error => {
+            if (errorHandler) errorHandler.handleError(error);
+            // Return observable that produces this particular effect
+            return resubscribeInCaseOfError(observable$);
+          })
+        );
```
alex-okrushko: Can you please make it as a package-private function declaration and put it outside of mergeEffects?

modules/effects/src/effects_resolver.ts
```diff
@@ -24,14 +24,19 @@ export function mergeEffects(
           ? sourceInstance[propertyName]()
           : sourceInstance[propertyName];
 
+      const resubscribeInCaseOfError = (
+        observable$: Observable<any>
```
alex-okrushko: both `any` should be replaced with `T` to make sure that both Observables are of the same kind.

modules/effects/src/effects_resolver.ts
```diff
@@ -61,3 +50,19 @@ export function mergeEffects(
 
   return merge(...observables$);
 }
+
+/**
+ * @internal
+ */
+export function resubscribeInCaseOfError<T extends Action = Action>(
```
alex-okrushko: No need to export it, right?

alex-okrushko: Is `= Action` needed? I think just extending Action is fine.

timdeschryver: Right, it also doesn't need the `@internal` comment

timdeschryver: No, it isn't needed. Thanks for the review 👍

PR #2172, opened 2019-10-20
feat(schematics): add createAction migration

evgenyfedorenko:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes # 2137

## What is the new behavior?
This PR adds a new schematic create-action-migration to migrate from class based actions to createAction() function.

```
[ ] Yes
[x ] No
```

## Other information
This is a rough first iteration of the migration. There are few things which can be improved - open for suggestions.


brandonroberts:
Thanks for the PR @evgenyfedorenko! I'm assuming this doesn't touch the actions in your reducers, or effects. How would running this migration not leave your app in a broken state?

evgenyfedorenko:
Hi @brandonroberts yeah it will. But if one decided to do the syntax update to use createAction syntax this schematics will still help to convert all action files. The rest can be done manually. I surely can look into expanding the scope of it and trying to convert all old actions usages across all the project - that may be a tricky job.

timdeschryver:
I was planning to giving this PR a look tomorrow, but I have some concerns.

I was (and still am) hesitant towards having this schematic, as it would leave your app in a broken state (as brandon mentioned). Even if we write the reducers and effect changes, we won't cover the whole migration to the `create` functions. I'm also afraid we might break an application at run-time which might be harder to debug if there are a lot of moving parts in this migration.

If we're going to have this schematic, I'm more in favor of having a partially upgrade to have less moving parts, and the possibility to make incremental changes in a code base. Mostly because this doesn't make NgRx better, it's syntactic sugar and it isn't needed to refactor a whole application. Our users can then decide the pace they want to follow if they want to use the new functions.
Therefore, I would like to propose that we can run this schematic on specific files of the repo.

Harpush:
I was looking for a migration schematic in order to make the create action change easier on our team and ended up here.
If i may add my input - i don't know much about schematics power but cant it update the actions currently declared as class to the createAction function and migrate the reducers and effects using those classes to use the new actions? What i mean is:
Reducers will switch on the type property on the action creator and effects will use the new ofType convention. That way it will update only the actions part and wont break the app. I believe it should look for .actions.ts files (either const or enum pattern) and through find references change the ofType and reducer as i wrote above. Does it sound reasonable? Might as well create another schematic for reducers migration once actions migration already took place on the code.

evgenyfedorenko:
@Harpush yeah you are mostly right. But it is not only effects and reducers - it is all other files where actions are dispatched from. Changing all occurrences of action class instantiations to new createAction syntax is likely going to break the app - it is just hard to cover all the bases. I will, however,  give it a look whether this is a realistic task to do at all.

evgenyfedorenko:
@timdeschryver @brandonroberts I've created a RFC before working on this PR which gave me a feeling that this schematic was needed. Should I spend more time on this trying to fix app's broken state - or you guys don't want it altogether? @timdeschryver will you review it before I start adding more code?

brandonroberts:
@evgenyfedorenko I think it would be useful to some people who want to refactor their app to use the new way, but it's not necessarily _needed_. If we're going to do it, it needs to refactor the actions to the style, update their references where used in components, effects, reducers, etc and not leave the app in a broken state after the migration is run. It's part of the reason we haven't done anything with it up until now. Its lots of edge cases to an opt-in feature for existing apps.

Harpush:
Just looked around a bit and found this library - https://github.com/dsherret/ts-morph and in its documentation there is a find all references functionality... Wont it help? https://ts-morph.com/navigation/finding-references

modules/schematics/src/create-action-migration/index.ts
```diff
@@ -0,0 +1,206 @@
...
+};
+
+const getActionClasses = (sourceFile: ts.SourceFile): ts.ClassDeclaration[] => {
+  return sourceFile.statements
+    .filter(ts.isClassDeclaration)
+    .filter(
+      statement => statement.heritageClauses && statement.heritageClauses.length
+    )
+    .filter(statement => {
+      return statement.heritageClauses!.filter(clause => {
```
timdeschryver: I think it would be easier to read if we would use `some` in favor of `filter([].length)`

modules/schematics/src/create-action-migration/index.ts
```diff
@@ -0,0 +1,206 @@
...
+          let typeProperty = actionClass.members
+            .filter(ts.isPropertyDeclaration)
+            .filter(member => (member.name as ts.Identifier).text === 'type')
+            .map(member => (member.initializer as ts.Identifier).text)[0];
+
+          if (!typeProperty) {
+            return undefined;
+          }
+
+          const typeVariable = actionTypeVariables.filter(
```
timdeschryver: I would rather see `actionTypeVariables` as a lookup, the key can be the variable name and the value can be the string value. That way we prevent unnecessary lookups here and would it also make more readable imho.

modules/schematics/src/create-action-migration/index.ts
```diff
@@ -0,0 +1,206 @@
...
+};
+
+const getConstructorPayloadType = (
+  host: Tree,
+  path: Path,
+  actionClass: ts.ClassDeclaration
+): string => {
+  let payloadType = '';
+
+  let payloadInConstructor = actionClass.members
```
timdeschryver: this would mean we would only convert one property, right?
e.g. below, we would only use `foo`?

```
constructor(foo: string, bar: string){}
```

modules/schematics/src/create-action-migration/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    .filter(ts.isConstructorDeclaration)
+    .filter(member => member.parameters.filter(param => param.type).length)
+    .map(member => member.parameters[0].type)[0];
+
+  if (payloadInConstructor) {
+    payloadType = host
+      .read(path)!
```
timdeschryver: I think there would be a better way to read the payload instance of reading the file.
Something like `node.getText()` should do it?
See https://astexplorer.net/#/gist/aeeece9f8bdd80fc585482846706efc2/2f7a868307f3137c05225bf10e67eca649b4c7f0 (open the console)

modules/schematics/src/create-action-migration/index.ts
```diff
@@ -0,0 +1,206 @@
...
+  payloadType: string,
+  actionClass: ts.ClassDeclaration
+): string => {
+  let actionName = camelize(actionClass.name!.text.replace('Action', ''));
+  let createAction = tags.stripIndent`
+        export const ${actionName} = createAction(
+          '${typeValue}'`;
+  if (payloadType) {
+    createAction += `,
+  props<{ payload: ${payloadType} }>()`;
```
timdeschryver: I think this is wrong, this would mean that even if the action doesn't have a `payload` but a `msg` for example, we would rename it to `payload`.

We should only migrate the ones with `payload`, or we should keep the naming (and the number of props).
I would prefer the latter.

modules/schematics/src/create-action-migration/index.ts
```diff
@@ -0,0 +1,206 @@
...
+            importProps = true;
+          }
+
+          const createActionDefinition = composeCreateAction(
+            typeValue,
+            payloadType,
+            actionClass
+          );
+
+          if (createActionDefinition.includes('props')) {
```
timdeschryver: I think this can be removed?
This would already be set to true with:

```
          if (payloadType) {
            importProps = true;
          }
```

modules/schematics/src/create-action-migration/index.ts
```diff
@@ -0,0 +1,206 @@
...
+
+const isActionImported = (sourceFile: ts.SourceFile): boolean => {
+  return !!sourceFile.statements
+    .filter(ts.isImportDeclaration)
+    .filter(({ moduleSpecifier }) =>
+      moduleSpecifier.getText(sourceFile).includes('@ngrx/store')
+    )
+    .filter(p => {
+      return (p.importClause!
+        .namedBindings! as ts.NamedImports).elements.filter(spec => {
```
timdeschryver: We need an extra check here, it burned us in the past. See https://github.com/ngrx/platform/pull/2095/files#diff-21e3d91ad9394c9035565f2b36790d17R58-R64

modules/schematics/src/create-action-migration/index.ts
```diff
@@ -0,0 +1,206 @@
...
+            return undefined;
+          }
+
+          const payloadType = getConstructorPayloadType(
+            host,
+            path,
+            actionClass
+          );
+          if (payloadType) {
+            importProps = true;
```
timdeschryver: Would this cause for problems, we set `importProps` to `false` outside of the loop.
But we never re-assign it back to `false`, thus once it's assigned to `true` it will always be `true`.

modules/schematics/src/create-action-migration/index.ts
evgenyfedorenko: Hey Tim, do you mean something like this?
```javascript
      return statement.heritageClauses!.some(clause => {
        return (clause.types || []).some(type => {
          return (
            type.expression && type.expression.getText(sourceFile) === 'Action'
          );
        });
      });
```

evgenyfedorenko:
I've spent another few days on this and it became messy. Needs significant effort to cover all bases. Closing for now.

PR #2191, opened 2019-10-30
refactor(effects): prepare for TS 3.5

alex-okrushko:
## PR Checklist

## PR Type

TS 3.5 becomes stricter with type, and generic `T` doesn't have `constructor` or `hasOwnProperty` methods.

for example, see https://github.com/microsoft/TypeScript/issues/31380

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

## What is the new behavior?
Fix issues that would come up with TS 3.5 upgrade

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/src/effect_creator.ts
+++ b/modules/effects/src/effect_creator.ts
@@ -61,7 +61,9 @@ export function createEffect<
   return effect;
 }
 
-export function getCreateEffectMetadata<T>(instance: T): EffectMetadata<T>[] {
+export function getCreateEffectMetadata<
+  T extends { [props in keyof T]: Object }
+>(instance: T): EffectMetadata<T>[] {
   const propertyNames = Object.getOwnPropertyNames(instance) as Extract<
     keyof T,
     string
--- a/modules/effects/src/effect_decorator.ts
+++ b/modules/effects/src/effect_decorator.ts
@@ -1,6 +1,6 @@
 import { compose } from '@ngrx/store';
 
-import { EffectConfig, EffectMetadata } from './models';
+import { EffectConfig, EffectMetadata, EffectPropertyKey } from './models';
 import { getSourceForInstance } from './utils';
 
 const METADATA_KEY = '__@ngrx/effects__';
@@ -8,8 +8,8 @@ const METADATA_KEY = '__@ngrx/effects__';
 export function Effect({
   dispatch = true,
   resubscribeOnError = true,
-}: EffectConfig = {}): PropertyDecorator {
-  return function<T extends Object, K extends Extract<keyof T, string>>(
+}: EffectConfig = {}) {
+  return function<T extends Object, K extends EffectPropertyKey<T>>(
     target: T,
     propertyName: K
   ) {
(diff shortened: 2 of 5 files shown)


modules/effects/src/effect_creator.ts
```diff
@@ -61,7 +61,9 @@ export function createEffect<
   return effect;
 }
 
-export function getCreateEffectMetadata<T>(instance: T): EffectMetadata<T>[] {
+export function getCreateEffectMetadata<
+  T extends { [props in keyof T]: Object }
```
L2jLiga: May be `T extends Record<keyof T, Object>` would be better?

modules/effects/src/effect_decorator.ts
```diff
@@ -21,7 +22,7 @@ export function Effect<T>({
       resubscribeOnError,
     };
     setEffectMetadataEntries<T>(target, [metadata]);
-  } as (target: {}, propertyName: string | symbol) => void;
+  };
```
L2jLiga: Woohoo! Love this change!

modules/effects/src/effect_notification.ts
```diff
@@ -46,7 +46,7 @@ function getEffectName({
 }: EffectNotification) {
   const isMethod = typeof sourceInstance[propertyName] === 'function';
 
-  return `"${sourceName}.${propertyName}${isMethod ? '()' : ''}"`;
+  return `"${sourceName}.${String(propertyName)}${isMethod ? '()' : ''}"`;
```
L2jLiga: Why this change required? I thought Typescript knew that `propertyName` will be transformer to string under the hood

L2jLiga:
I think it would be nice to adopt to TS 3.6 as far as Angular 9 rc0 published

alex-okrushko:
> I think it would be nice to adapt to TS 3.6 as far as Angular 9 rc0 published

I haven't check if 3.6 brings any problem yet. Let's take it one-by-one.

modules/effects/src/effect_decorator.ts
alex-okrushko: 😃

modules/effects/src/effect_creator.ts
alex-okrushko: Yes, it's the same thing. Don't think we used `Record` at the codebase yet.

modules/effects/src/effect_notification.ts
alex-okrushko: Good question @L2jLiga ! I'm bringing the `propertyName` back to the actual allowed property types, which are `string | number | Symbol`. 
`Symbol` cannot be implicitly converted to string, and needs some help.

I probably should rename `propertyName` to the `propertyKey` to be more descriptive. I'll leave it out for later. Thanks for asking!

modules/effects/src/effect_creator.ts
L2jLiga: Then fine :)

PR #2194, opened 2019-10-30
refactor(effects): Refactor types in effects, expose metadata key

alex-okrushko:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

PR is based off https://github.com/ngrx/platform/pull/2190

What kind of change does this PR introduce?

- Cleanup types within effects (remove `any`)
- expose `[CREATE_EFFECT_METADATA_KEY]` key on the `createEffect` and clean up the tests.

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

PR #2195, opened 2019-10-30
feat(effects): createEffect returns specific type for dispatch false

alex-okrushko:
## PR Checklist

## PR Type

based of https://github.com/ngrx/platform/pull/2194

`createEffect` with `{dispatch: false}` now returns whatever Observable type is actually returned (instead of `unknown`).

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`createEffect` with `{dispatch: false}` returns  `Observable<unknown>`.

## What is the new behavior?

`createEffect` with `{dispatch: false}` returns  `Observable<SpecificType>`.

```
[x] Yes
[ ] No
```

Somewhat. If anyone was relying that `Observable<unknown>` is returned. However, `Observable<SpecificType>` can be always widened to `Observable<unknown>` so it shouldn't really break anything.

### DIFF
--- a/modules/effects/src/effect_creator.ts
+++ b/modules/effects/src/effect_creator.ts
@@ -8,10 +8,8 @@ interface CreateEffectMetadata {
   [CREATE_EFFECT_METADATA_KEY]: EffectConfig;
 }
 
-type DispatchType<T> = T extends { dispatch: infer U } ? U : unknown;
-type ObservableReturnType<T> = T extends false
-  ? Observable<unknown>
-  : Observable<Action>;
+type DispatchType<T> = T extends { dispatch: infer U } ? U : true;
+type ObservableType<T, OriginalType> = T extends false ? OriginalType : Action;
 /**
  * @description
  * Creates an effect from an `Observable` and an `EffectConfig`.
@@ -46,9 +44,9 @@ type ObservableReturnType<T> = T extends false
  */
 export function createEffect<
   C extends EffectConfig,
-  T extends DispatchType<C>,
-  O extends ObservableReturnType<T>,
-  R extends O | ((...args: any[]) => O)
+  DT extends DispatchType<C>,
+  OT extends ObservableType<DT, OT>,
+  R extends Observable<OT> | ((...args: any[]) => Observable<OT>)
 >(source: () => R, config?: Partial<C>): R & CreateEffectMetadata {
   const effect = source();
   const value: EffectConfig = {
--- a/modules/effects/spec/types/effect_creator.spec.ts
+++ b/modules/effects/spec/types/effect_creator.spec.ts
@@ -47,7 +47,7 @@ describe('createEffect()', () => {
       expectSnippet(`
         const effect = createEffect(() => ({ foo: 'a' }), { dispatch: false });
       `).toFail(
-        /Type '{ foo: string; }' is not assignable to type 'Observable<unknown> | ((...args: any[]) => Observable<unknown>)'./
+        /Type '{ foo: string; }' is not assignable to type 'Observable<{}> | ((...args: any[]) => Observable<{}>)'./
       );
     });
   });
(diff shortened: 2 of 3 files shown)


alex-okrushko:
Rerunning the tests should fix the issue. (I do not have permissions at circle ci)

modules/effects/src/effect_creator.ts
```diff
@@ -42,30 +44,25 @@ type ObservableReturnType<T> = T extends false
...
 export function createEffect<
   C extends EffectConfig,
-  T extends DispatchType<C>,
-  O extends ObservableReturnType<T>,
-  R extends O | ((...args: any[]) => O)
->(source: () => R, config?: Partial<C>): R {
+  DT extends DispatchType<C>,
+  OT extends ObservableType<DT, OT>,
+  R extends Observable<OT> | ((...args: any[]) => Observable<OT>)
+>(source: () => R, config?: Partial<C>): R & CreateEffectMetadata {
```
timdeschryver: Neat! 🤩

timdeschryver: I'm wondering if `CreateEffectMetadata` might confuse people if they look at the typings.

modules/effects/src/effect_notification.ts
```diff
@@ -4,7 +4,7 @@ import { Notification, Observable } from 'rxjs';
 
 export interface EffectNotification {
   effect: Observable<any> | (() => Observable<any>);
-  propertyName: string;
+  propertyName: PropertyKey;
```
timdeschryver: Should this be `EffectPropertyKey`?

modules/effects/src/effect_creator.ts
alex-okrushko: That's what `createEffect` returns though. I think it's fine to provide that data.

modules/effects/src/effect_notification.ts
alex-okrushko: It probably should, but that would mean we'd have to re-type the effects_resolver and so on. Let's take it bit by bit.

modules/effects/src/effect_creator.ts
```diff
@@ -42,30 +44,25 @@ type ObservableReturnType<T> = T extends false
...
+  return effect as typeof effect & CreateEffectMetadata;
 }
 
-export function getCreateEffectMetadata<T>(instance: T): EffectMetadata<T>[] {
-  const propertyNames = Object.getOwnPropertyNames(instance) as Extract<
-    keyof T,
-    string
-  >[];
+export function getCreateEffectMetadata<
+  T extends { [props in keyof T]: Object }
```
timdeschryver: What's the difference between the following?

```ts
T extends { [props in keyof T]: Object }
```

VS

```ts
T extends Object 
```

alex-okrushko: That's a really good question, @timdeschryver.
We are accessing the property and checking `instance[propertyName].hasOwnProperty(...)`. With `T extends Object` there's no info about the what properties there are (and they could be `undefined` with TS 3.5 and thus not allowed to call `hasOwnProperty`).

At some point I put it behind the type guard, but that PR got way too big and without focus. I think I'll adjust it soon.

timdeschryver: Would the following cross the line? 

```
T extends {
    [props in keyof T]: { [CREATE_EFFECT_METADATA_KEY]: EffectMetadata<T> }
  }
```

Because, we're stilling casting to `any` to access the `metaData`:

```
const metaData = (instance[propertyName] as any)[
        CREATE_EFFECT_METADATA_KEY
];
```

timdeschryver: Got it, thanks for the info Alex!

Would the following cross the line? 

```
T extends {
    [props in keyof T]: { [CREATE_EFFECT_METADATA_KEY]: EffectMetadata<T> }
  }
```

Because, we're stilling casting to `any` to access the `metaData`:

```
const metaData = (instance[propertyName] as any)[
        CREATE_EFFECT_METADATA_KEY
];
```

brandonroberts:
@alex-okrushko this one needs another pass on the merge conflicts.

Issue #1432, opened 2018-11-16
Allow EffectModule.forFeature() to accept angular injection token


j-thebault:
Hello

I'm trying to create a reusable effect. This effect use some constant and a service :

```typescript
@Injectable()
export class TreeNodeEffects {

constructor(private pagePrefix : string, private componentId : string , 
private actions$: Actions, private service: TreeService) {}
...
@Effect()
openNode$ = this.actions$.pipe( 
   ofType(TreeNodeActionTypes.OpenTreeNode)
   filter(((action: OpenTreeNode) => 
       action.pagePrefix == this.pagePrefix && action.componentId ==  this.componentId)),
   switchMap((action: OpenTreeNode) => {
      return this.service.loadNodeContent(action.payload.id)....
```

The problem is that TreeService is an abstract class and i would like to provide an implementation provided directly by my module... However, even if my module provide something like this :

```typescript
@NgModule({
  imports: [
   ...
    TreeModule,
    StoreModule.forFeature(fromStore.STORE_FEATURE_NAME, fromStore.reducers, {metaReducers: fromStore.metaReducers}),
    EffectsModule.forFeature([ TreeNodeEffects , ...fromStore.effects]),
  ],

 providers : [
...
{
      provide: TreeService,
      useClass: TestCaseWorkspaceService
    }, {
      provide : TreeNodeEffects,
      useFactory : (actions$, service) => {
        return [new TreeNodeEffects(TEST_CASE_WORKSPACE_ROOT_URL, MAIN_TREE_ID, actions$, service)];
      },
      deps : [Actions, TreeService],
    }
```

The instanciation of the TreeNodeEffects fail on injection of the abstract service, angular is not able to provide the concrete implementation. The others dependencies are correctly injected. However, if i change the constructor of the effect to the concrete class, the injection is done but it make my effect hard bound to the concrete type...

## Describe any alternatives/workarounds you're currently using

I managed to realize the injection by doing this :

```typescript
import {ɵngrx_modules_effects_effects_e} from '@ngrx/effects';

imports: [
    EffectsModule.forFeature([...fromStore.effects]), //here i import my other effects
  ],
 providers: [
    {
      provide: TreeService,
      useClass: TestCaseWorkspaceService
    },
    {
      provide : ɵngrx_modules_effects_effects_e,
      useFactory : (actions$, service) => {
[...]

PR #2231, opened 2019-11-08
feat(effects): add user provided effects to EffectsModule.forFeature

### LINKED ISSUE #2232: RFC: Accept InjectionToken in EffectsModule.forFeature
It is currently not possible to register effects (`EffectsModule.forFeature`) with an `InjectionToken` like it's possible in `StoreModule.forFeature`. Adding this feature could open some possiblities of reducing boilerplate in applications.
Pull Request #2231 (Help needed for how to document and test)

An alternative is using `InjectionTokens` for everything but the effects, but that leads to some complications, an example from my current project:
```
EffectsModule.forFeature([
  FeatureStore.createEffects(
    ACTIONS_TOKEN,
    FeatureService,
  ),
])
```

## Other information:

Hey there, this is my first attempt at contributing and this is my first feature request :D

[ ] No

leon-marzahn:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

It is currently not possible to register effects with an `InjectionToken` like it's possible in `StoreModule.forFeature`
Closes #2232

## What is the new behavior?

It's now possible to pass an `InjectionToken` to either `USER_PROVIDED_EFFECTS` to register Effects. E.g.
```

@NgModule({
  imports: [
    EffectsModule.forFeature(),
  ],
  providers: [
    SomeEffect,
    SomeOtherEffect,
    {
      provide: USER_PROVIDED_EFFECTS,
      multi: true,
      useValue: [SomeEffect, SomeOtherEffect],
    },
  ],
})
class FeatureModuleUsingStore {
}
```
This is useful to create consistency with `StoreModule.forFeature` and helps in making it possible to reducer boilerplate for NGRX. You could load effects by providers, which creates possibilities, e.g. putting actions, reducers and effects into a class, that is loaded via dependency injection.

```
[ ] Yes
[x] No
```

## Other information
This is my first attempt at contributing to any project, so please be easy on me :D
I couldn't write documentation yet, as i don't really know if this needs any documentation. I would appreciate some feedback and am willing for change in this.


PsychoPflanze:
This also fixes #1432 (Although i don't know if he found a solution yet)

PsychoPflanze:
Now that i think about it, it would be even better if you can pass instances of effects as well. So the token would be of type `InjectionToken<(Type<any> | any)[]>`. And it would only inject them if they are of type `Type`. Otherwise it can just use the instance.

modules/effects/src/effects_module.ts
```diff
@@ -68,3 +97,22 @@ export function _provideForRootGuard(runner: EffectsRunner): any {
...
   return 'guarded';
 }
+
+export function _createFeatureEffects(
+  injector: Injector,
+  effectCollections: Type<any>[][] | InjectionToken<Type<any>[]>[]
+): any[] {
+  const effects: Type<any>[] = [];
+
+  effectCollections.forEach(
```
timdeschryver: Imho it's more readable if we'd use a `map` here - see https://github.com/ngrx/platform/blob/master/modules/store/src/store_module.ts#L278 for an example

PsychoPflanze: > I'm not sure which use cases this feature has, in the issue it mentions that it reduces boilerplate but I'm not seeing how this can reduce it.
> 
> But it got me thinking about a different API, instead of providing the token in the `forFeature` function, what about having a similar API to `USER_PROVIDED_META_REDUCERS`? This would look as follows:
> 
> ```ts
> providers: [
>   {
>     provide: USER_PROVIDED_EFFECTS,
>     useValue: [FooEffect, BarEffect]
>   }
> ]
> ```
> 
> The above only if it makes sense to have a similar feature...
> 
> If we implement this we should also add the feature to use a token in the `EffectsModule.forRoot()` function for consistency.

That would also work

PsychoPflanze:
@timdeschryver Because you said you don't know how this can reduce boilerplate. Essentially you would be able to create effects, reducers and etc. at runtime, which allows you to do that in any way, shape or form. In my example i would use a base class `BaseStore` and some module that creates effects/reducers from this class. The class would have functions like `createReducer` and `createEffects`.

PsychoPflanze:
Something like this: https://stackblitz.com/edit/angular-rm3eiu
Obviously this doesn't work and is very quick and dirty. But this should let you get the idea

brandonroberts:
What's the status on this?

PsychoPflanze:
I couldn't work on this over the holidays, i would continue today with the changes suggested

PsychoPflanze:
@brandonroberts & @timdeschryver I've made changes to this and implemented it as suggested by @timdeschryver. Also, i don't know how to fix the conflict, as every time i fix it, the githooks break it again :D

brandonroberts:
@PsychoPflanze are you rebased on latest master?

PsychoPflanze:
> @PsychoPflanze are you rebased on latest master?

@brandonroberts I am yes, but the githooks correct the formatting of a specific import, which causes a conflict

timdeschryver:
@PsychoPflanze  can you try to rebase again? 
You should pick the export with the most imports:

```doff
- export { EffectsFeatureModule } from './effects_feature_module';
+ export {
  ROOT_EFFECTS_INIT,
  rootEffectsInit,
  EffectsRootModule,
} from './effects_root_module';
```

modules/effects/src/index.ts
```diff
@@ -11,10 +11,15 @@ export { EffectNotification } from './effect_notification';
 export {
   ROOT_EFFECTS_INIT,
   rootEffectsInit,
+  EffectsRootModule,
```
PsychoPflanze: @brandonroberts The github merge conflict thing is definitely weird. As now apparently i've added this, even though it said this is in master.

PsychoPflanze:
@timdeschryver & @alex-okrushko Any updates?

PsychoPflanze:
I've figured out how to use forks, and the pull request is now up to date with master

timdeschryver:
I will have a look at the PR later this week.

modules/effects/src/effects_module.ts
```diff
@@ -46,16 +64,44 @@ export class EffectsModule {
...
     };
   }
 }
 
+export function createEffects(
+  injector: Injector,
+  effects: Type<any>[],
+  userProvidedEffectGroups: Type<any>[][]
+): any[] {
+  const mergedEffects: Type<any>[] = effects;
```
alex-okrushko: How about the following
```ts
const mergedEffects: Set<Type<any>> = new Set([...userProvidedEffectGroups.flat(), ...effects]);
```
this is all remove the duplicates, if there are any.

modules/effects/src/effects_module.ts
```diff
@@ -46,16 +64,44 @@ export class EffectsModule {
...
+  userProvidedEffectGroups.forEach(group => mergedEffects.push(...group));
+  return createSourceInstances(
+    ...createEffectInstances(injector, ...mergedEffects)
+  );
+}
+
+export function createEffectInstances(
+  injector: Injector,
+  ...effects: Type<any>[]
```
alex-okrushko: no need to spread here as well

modules/effects/src/effects_module.ts
PsychoPflanze: I think the 'flat' method is not yet supported by typescript (So it tells you something like this)
Screenshot_2

That's also the reason i had to spread so much

alex-okrushko: I see... `flat` is not supported on edge, so I many we should avoid it as well for now.

Still, let's not use `forEach` of the Array, let's do it with `for ... of` loop.

alex-okrushko: Spreads are fine :)

As for the `forEach`, here is the explanation why not to use it:
> Do not use Array.prototype.forEach, it makes code harder to debug and defeats some useful compiler checks (e.g. reachability).

PsychoPflanze:
> I would like to change the token to have a single value, for example
> 
> ```ts
> {
>     provide: USER_PROVIDED_FEATURE_EFFECTS,
>     multi: true,
>     useValue: FeatUserProvidedEffect,
> },
> ```
> 
> This is more in line with the meta-reducers, and I also think most of the features are registered with a single Effect. Thoughts?
> 
> We'll also need some tests for the root effects and perhaps also for the order of the effect registrations (effects as parameter first before the tokens)?

1. I don't know really, it might lead to the case where someone wants to pass multiple effects and has to provide the same token twice. But i'm fine with doing that or best-case both.
2. Yes, i'll write tests for the Root Effects.
3. I don't know if order of effects matters, as they have to be unique anyway, right?

PsychoPflanze:
Also, just as a question, what is the workflow on writing documentation?

modules/effects/src/effects_module.ts
PsychoPflanze: Used a `for ... of` loop 👍

modules/effects/src/effects_module.ts
PsychoPflanze: Removed the spread operator

modules/effects/src/effects_module.ts
```diff
@@ -46,16 +64,44 @@ export class EffectsModule {
...
 
+export function createEffects(
+  injector: Injector,
+  effects: Type<any>[],
+  userProvidedEffectGroups: Type<any>[][]
+): any[] {
+  const mergedEffects: Type<any>[] = effects;
+  userProvidedEffectGroups.forEach(group => mergedEffects.push(...group));
+  return createSourceInstances(
+    ...createEffectInstances(injector, ...mergedEffects)
```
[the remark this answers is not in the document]
PsychoPflanze: Removed `createSourceInstances` in general, as i am already injecting the effects at this point

PsychoPflanze:
@timdeschryver I wrote tests for `RootEffects` and wrote some more cases for `FeatureEffects` (Such as providing multiple times)

@alex-okrushko I use a `for ... of` loop now, and i removed some uneccessary stuff (E.g the spread operators and `createSourceInstances`)

PsychoPflanze:
I've also added documentation. But i don't know if it is correct at this position, as i'd much rather put it in like an "Advanced" section

timdeschryver:
@PsychoPflanze the docs changes are fine at first sight - you can see them at https://previews.ngrx.io/pr2231-f717336/

modules/effects/spec/integration.spec.ts
```diff
@@ -90,6 +92,56 @@ describe('NgRx Effects Integration spec', () => {
     });
   });
 
+  it('should execute user provided root effects', (done: DoneFn) => {
```
timdeschryver: Instead of writing these tests, what about adding the providers in the `should dispatch init actions in the correct order` spec? This way, we also test the order of effects registration

modules/effects/spec/integration.spec.ts
```diff
@@ -90,6 +92,56 @@ describe('NgRx Effects Integration spec', () => {
...
+    const store = TestBed.get(Store) as Store<any>;
+    dispatch = store.dispatch as jasmine.Spy;
+
+    expect(dispatch).toHaveBeenCalledWith({
+      type: '[RootUserProvidedEffect]: INIT',
+    });
+    done();
+  });
+
+  it('should execute user provided feature effects', (done: DoneFn) => {
```
timdeschryver: Same here

modules/effects/spec/integration.spec.ts
timdeschryver: Could we also create a simple unit test to verify the effect is added?
Something as:

```ts
describe('when registered with the USER_PROVIDED_FEATURE_EFFECTS token', () => {
    class SourceA {}
    const sourceA = new SourceA();

    let mockEffectSources: { addEffects: jasmine.Spy };

    beforeEach(() => {
        imports: [EffectsModule.forFeature()],
        providers: [
          SourceA,
          {
            provide: EffectsRootModule,
            useValue: {
              addEffects: jasmine.createSpy('addEffects'),
            },
          },
          {
            provide: USER_PROVIDED_FEATURE_EFFECTS,
            useValue: [SourceA],
            multi: true,
          },
        ],
      });
      mockEffectSources = TestBed.get(EffectsRootModule);
    });

    it('should add effects when provided', () => {
      TestBed.get(EffectsFeatureModule);
      expect(mockEffectSources.addEffects).toHaveBeenCalledWith(sourceA);
    });
  });
```

zakhenry:
I have a question/concern about the approach here with using injection tokens - what is the expectation of the lifetime of an effect when it is provided in a component?

To illustrate an example, imagine we have a component that provides an effects class directly: 

```ts
@Component({
  selector: 'app-child',
  templateUrl: './child.component.html',
  styleUrls: ['./child.component.css'],
  providers: [{
      provide: USER_PROVIDED_FEATURE_EFFECTS,
      useValue: [SomeEffect],
    }]
})
export class ChildComponent implements OnInit {

  constructor() { }

  ngOnInit() {
  }

}
```

If I provide an `@Injectable` class directly to a component, I have the expectation that the class will at some point have `ngOnDestroy` called ([example](https://stackblitz.com/edit/angular-hybffa)). Should it be the same for an effect provided in a similar way? Or should ngrx automatically unsubscribe from the effects when the component is destroyed?

I *think* the actual answer is that with this PR, ngrx will actually do nothing if I provide the symbol in a component decorator, but the API is close enough that it feels like I should be able to do that. Especially because effects that last the lifetime of a component is a pretty useful feature.

leon-marzahn:
@zakhenry 
> I have a question/concern about the approach here with using injection tokens - what is the expectation of the lifetime of an effect when it is provided in a component?
> 
> To illustrate an example, imagine we have a component that provides an effects class directly:
> 
> ```ts
> @Component({
>   selector: 'app-child',
>   templateUrl: './child.component.html',
>   styleUrls: ['./child.component.css'],
>   providers: [{
>       provide: USER_PROVIDED_FEATURE_EFFECTS,
>       useValue: [SomeEffect],
>     }]
> })
> export class ChildComponent implements OnInit {
> 
>   constructor() { }
> 
>   ngOnInit() {
>   }
> 
> }
> ```
> 
> If I provide an `@Injectable` class directly to a component, I have the expectation that the class will at some point have `ngOnDestroy` called ([example](https://stackblitz.com/edit/angular-hybffa)). Should it be the same for an effect provided in a similar way? Or should ngrx automatically unsubscribe from the effects when the component is destroyed?
> 
> I _think_ the actual answer is that with this PR, ngrx will actually do nothing if I provide the symbol in a component decorator, but the API is close enough that it feels like I should be able to do that. Especially because effects that last the lifetime of a component is a pretty useful feature.

That is sadly not how that would or could work. Essentially, you're not just providing a service, you're providing a class, which's instance is passed to a module, which in turn passes the instances to the root module. That is how the system is currently written. If you'd like to have this as a feature, you have to submit a feature request. I don't think this is related to this feature

leon-marzahn:
I was actually thinking about merging `USER_PROVIDED_ROOT_EFFECTS` and `USER_PROVIDED_FEATURE_EFFECTS` as they are essentially the same and in the end the effects are passed to the root module anyway. Thoughts?

alex-okrushko:
@leon-marzahn I like that idea! The ever increasing number of InjectionTokens bothers me, bth.

alex-okrushko:
I see that the tests are failing, was that addressed?

leon-marzahn:
> I see that the tests are failing, was that addressed?

It's probably just out of date, when i sync it, it's probably fine

modules/effects/src/effects_module.ts
```diff
@@ -7,6 +8,16 @@ import {
 } from '@angular/core';
 import { Actions } from './actions';
 import { EffectSources } from './effect_sources';
+import { Actions } from './actions';
```
brandonroberts: Duplicate import

leon-marzahn: Removed it

leon-marzahn:
How do i restart the tests? They failed because of a js heap error

timdeschryver:
@leon-marzahn I restarted the CI

timdeschryver:
Before this gets merged, do we agree that providing an array of effects is what we want, or do we want to provide a single effect instance? It's my feeling that providing a single effect is cleaner, it's just a gut feeling - but it's similar to what we do with the metareducers.


```ts
{
    provide: USER_PROVIDED_FEATURE_EFFECTS,
    multi: true,
    useValue: [ FeatUserProvidedEffect ],
},
```

VS


```ts
{
    provide: USER_PROVIDED_FEATURE_EFFECTS,
    multi: true,
    useValue: FeatUserProvidedEffect,
},
```

leon-marzahn:
> Before this gets merged, do we agree that providing an array of effects is what we want, or do we want to provide a single effect instance? It's my feeling that providing a single effect is cleaner, it's just a gut feeling - but it's similar to what we do with the metareducers.
> 
> ```ts
> {
>     provide: USER_PROVIDED_FEATURE_EFFECTS,
>     multi: true,
>     useValue: [ FeatUserProvidedEffect ],
> },
> ```
> 
> VS
> 
> ```ts
> {
>     provide: USER_PROVIDED_FEATURE_EFFECTS,
>     multi: true,
>     useValue: FeatUserProvidedEffect,
> },
> ```

I really don't know, it might be cleaner if you assume the end user is going to have one Effect per module, which may or may not be the case.

leon-marzahn:
What's open:
- [ ] Merge `USER_PROVIDED_ROOT_EFFECTS` and `USER_PROVIDED_FEATURE_EFFECTS` to `USER_PROVIDED_EFFECTS`
- [ ] Test order of effects

timdeschryver:
If we want this in the v9 release we'll have to tackle the open TODOs.
@leon-marzahn do you think you can take them up, or can we help you in a way?

leon-marzahn:
> If we want this in the v9 release we'll have to tackle the open TODOs.
> @leon-marzahn do you think you can take them up, or can we help you in a way?

Uhm, yes, thank you i've been busy lately. I can do this later today, in terms of help i would appreciate some help with testing the order, if possible

leon-marzahn:
Alright, i've fixed the tests and merged the tokens to `USER_PROVIDED_EFFECTS`. Also i managed to test the order better now, thanks for the change, it made it easier to test :D

leon-marzahn:
There is a test failing i haven't touched, what should i do about it?

leon-marzahn:
@timdeschryver or @brandonroberts or @alex-okrushko

timdeschryver:
I think the test fails because of the changes made in the PR.
I'll have a look at it later today.

timdeschryver:
The test fails because in the module, multiple `forFeatures` are called and they all override eachother.
As a fix, make the token a `multi` token (and reflect this change in the `createEffects` method):

```ts
{
          provide: _FEATURE_EFFECTS,
          useValue: featureEffects,
          multi: true,
},
```

leon-marzahn:
> The test fails because in the module, multiple `forFeatures` are called and they all override eachother.
> As a fix, make the token a `multi` token (and reflect this change in the `createEffects` method):
> 
> ```ts
> {
>           provide: _FEATURE_EFFECTS,
>           useValue: featureEffects,
>           multi: true,
> },
> ```
> 
> @leon-marzahn

Alright, thank you, i didn't notice 😄

leon-marzahn:
I just remembered i will have to adjust the documentation

modules/effects/src/effects_module.ts
```diff
@@ -56,18 +70,49 @@ export class EffectsModule {
         EffectSources,
         Actions,
         rootEffects,
+        {
+          provide: _ROOT_EFFECTS,
+          multi: true,
```
timdeschryver: For root effects, we don't need `multi` as you can only call `forRoot` once.

modules/effects/spec/integration.spec.ts
```diff
@@ -215,6 +216,60 @@ describe('NgRx Effects Integration spec', () => {
...
+            dispatched: createDispatchedReducer(dispatchedActionsLog),
+          }),
+          EffectsModule.forRoot([
+            EffectLoggerWithOnRunEffects,
+            RootEffectWithInitAction,
+          ]),
+          RouterTestingModule.withRoutes([]),
+        ],
+        providers: [
+          UserProvidedEffect1,
```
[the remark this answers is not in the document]
leon-marzahn: Yes actually, because they are not provided in the module anymore. Do you have an idea how we could provide the value of a token in the module?

modules/effects/src/effects_module.ts
leon-marzahn: Alright, removed it

modules/effects/spec/integration.spec.ts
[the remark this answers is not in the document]
timdeschryver: Not that I know of 🤔
I would add the provider to the docs tho, instead of only the provider.

```diff
+ providers: [
+ MovieEffects,
  {
    provide: USER_PROVIDED_EFFECTS,
    multi: true,
    useValue: [MovieEffects],
  }
+ }
```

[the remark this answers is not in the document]
leon-marzahn: Added it

projects/ngrx.io/content/guide/effects/index.md
```diff
@@ -225,6 +225,27 @@ export class MovieModule {}
 
 </div>
 
+## Alternative way of registering effects
+
+You can provide root-/feature-level effects with the provider `USER_PROVIDED_EFFECTS`.
+
+<code-example header="admin.module.ts">
```
timdeschryver:
```suggestion
<code-example header="movies.module.ts">
```

projects/ngrx.io/content/guide/effects/index.md
```diff
@@ -225,6 +225,27 @@ export class MovieModule {}
...
+    provide: USER_PROVIDED_EFFECTS,
+    multi: true,
+    useValue: [MovieEffects],
+  },
+]
+</code-example>
+
+<div class="alert is-critical">
+
+The `EffectsModule.forChild()` method must be added to the module imports even if you only provide effects over token, and don't pass them via parameters. (Same goes for `EffectsModule.forRoot()`)
```
timdeschryver: The last one 😅

```suggestion
The `EffectsModule.forFeature()` method must be added to the module imports even if you only provide effects over token, and don't pass them via parameters. (Same goes for `EffectsModule.forRoot()`)
```

leon-marzahn: 😆

brandonroberts:
Thanks @leon-marzahn!

vash72:
When there will be a chance to make a release? or may you point me out how to build the package myself from master branch

brandonroberts:
Builds from the master branch are here https://ngrx.io/guide/nightlies#effects

PR #2272, opened 2019-11-20
feat(router-store):  add action creator for root router actions

### LINKED ISSUE #2206: feat(router): router actions as ActionCreators
If we're using the creator functions for reducers and effects we aren't able to use the router actions.
I'm suggesting that we'd also create these router actions as `ActionCreators`, for example:

```ts
export const routerRequestAction = createAction(ROUTER_REQUEST, props<{ payload: RouterRequestPayload<BaseRouterStoreState> }>())
```

We can create the `ActionCreators` in our project.

[ ] No

Jefiozie:
## PR Checklist

## PR Type

we  added the action creator for root router actions

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2206 

```
[ ] Yes
[ ] No
```


Jefiozie:
@timdeschryver, I made the changes to the doc. I also made a small diagram for the route-store flow. I think a picture tells more than some text. Looking forward on your reply.

projects/ngrx.io/content/guide/router-store/actions.md
```diff
@@ -1,50 +1,45 @@
...
 
 The action contains the store state before the navigation. You can use it to restore the consistency of the store.
+
+<div class="alert is-important">
+
+**Note:** You can also still use the action type, which was the previously defined way before action creators were introduced in NgRx. If you are looking for examples of the action types, visit the documentation for [versions 7.x and prior](https://v7.ngrx.io/guide/router-store/actions).
+
+</div>
+
+## Order of actions
```
timdeschryver: Can we keep this for a separate PR?

projects/ngrx.io/content/guide/router-store/actions.md
```diff
@@ -1,50 +1,45 @@
...
-
-- `ROUTER_REQUEST`
-- `ROUTER_NAVIGATION`
-- `ROUTER_CANCEL` / `ROUTER_ERROR`
-
-Error / Cancel case (with late Navigation Action Timing):
-
-- `ROUTER_REQUEST`
-- `ROUTER_CANCEL` / `ROUTER_ERROR`
+Router Store provides five navigation actions which are dispatched in a specific order. The `routerReducer` provided by Router Store updates its state with the latest router state given by the actions. By default we recommend to use the creator functions we already provide for your.
```
timdeschryver: I believe `your` is a typo here.

```suggestion
Router Store provides five navigation actions which are dispatched in a specific order. The `routerReducer` provided by Router Store updates its state with the latest router state given by the actions. By default we recommend to use the creator functions we provide.
```

modules/router-store/src/actions.ts
```diff
@@ -34,6 +35,10 @@ export type RouterRequestAction<
   payload: RouterRequestPayload<T>;
 };
 
+export const routerRequestAction = createAction(
+  ROUTER_REQUEST,
+  props<{ payload: RouterRequestPayload<BaseRouterStoreState> }>()
```
timdeschryver: These payloads should be from the generic `SerializedRouterSnapshot` (my bad, I was wrong in the example).

```suggestion
  props<{ payload: RouterRequestPayload<SerializedRouterSnapshot> }>()
```

projects/ngrx.io/content/guide/router-store/actions.md
Jefiozie: Sure, glad you like it, I will make a separate PR for this (when this PR is finished)

modules/router-store/src/actions.ts
Jefiozie: I think you mean `SerializedRouterStateSnapshot` instead of `SerializedRouterSnapshot`

Jefiozie:
> Tbh, I have no idea how these actions are used in the wild.
I'm wondering if we also should provide a router action creator, to keep them generic.
This because a const can't have a generic AFAIK.
This would look as follows:
// allow the flexibility to create a custom router action with a generic
export const createRouterCancelAction = <
  T,
  V extends BaseRouterStoreState = SerializedRouterStateSnapshot
>() => createAction(ROUTER_CANCEL, props<{ payload: RouterCancelPayload<T, V> }>());
// the default implementation with the default router store state
export const routerCancelAction = createRouterCancelAction<SerializedRouterStateSnapshot>();
Does this make sense?

So the main reason I think we should do this is to have a common `Actions API` for all of the packages of `@ngrx`. I believe this will help people to adopt it easier as they understand the way it is working. 

Do you agree with this?

I don't think we should change the default router reducer in a similar way as in the `@ngrx/store`. Probably most projects use the default reducer and use the actions for specific needs. What are you thoughts around this?

timdeschryver:
> I don't think we should change the default router reducer in a similar way as in the @ngrx/store.

I agree, the reason I mentioned it was for our end-users.

The only reason I see to create the router action creators, is for cases where a custom router serializer is used. In these serializers it's possible to add extra meta-data to the router state, which is added to the dispatched action. 

In the current implementation (without creators), it's possible to define this extra meta-data with the generic on the Action. This can be helpful if the Action is used in, for example, an effect because it will provide a type-safe Action with the extra meta-data, for example you be able to do `action.routerState.foo`. 

With the current implementation with the creator functions, this isn't possible anymore and the `routerState` on the action will always be of the type `SerializedRouterStateSnapshot`. 

That being said, I think this PR is good. We can always add the router action creators later if needed, because I'm uncertain if this is use-case is being used.

modules/router-store/src/actions.ts
timdeschryver: Correct, thanks!

Jefiozie:
Okay, just to be sure, this PR is good for now? 😊

timdeschryver:
For me it is! 👍

Jefiozie:
Great, thanks for your help will make a separate PR for the image that explains the action order.

PR #2284, opened 2019-11-29
fix(effects): support resubscribe if throw in error handler

lydemann:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

If an error is thrown in the `ErrorHandler` it will not resubscribe the effects observable stream on error.

Closes #

## What is the new behavior?
Now it is possible to throw an error in `ErrorHandler` and still resubscribe the effects observable stream on error. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/src/effect_notification.ts
+++ b/modules/effects/src/effect_notification.ts
@@ -19,13 +19,16 @@ export function reportInvalidActions(
     const isInvalidAction = !isAction(action);
 
     if (isInvalidAction) {
-      reporter.handleError(
-        new Error(
-          `Effect ${getEffectName(
-            output
-          )} dispatched an invalid action: ${stringify(action)}`
-        )
-      );
+      try {
+        // handleError can throw errors
+        reporter.handleError(
+          new Error(
+            `Effect ${getEffectName(
+              output
+            )} dispatched an invalid action: ${stringify(action)}`
+          )
+        );
+      } catch {}
     }
   }
 }
--- a/modules/effects/src/effects_resolver.ts
+++ b/modules/effects/src/effects_resolver.ts
@@ -57,7 +57,13 @@ function resubscribeInCaseOfError<T extends Action>(
 ): Observable<T> {
   return observable$.pipe(
     catchError(error => {
-      if (errorHandler) errorHandler.handleError(error);
+      if (errorHandler) {
+        try {
+          errorHandler.handleError(error);
+        } catch {
+          // Handle throw in error handler
+        }
+      }
       // Return observable that produces this particular effect
       return resubscribeInCaseOfError(observable$, errorHandler);
     })
--- a/modules/effects/spec/effect_sources.spec.ts
+++ b/modules/effects/spec/effect_sources.spec.ts
@@ -282,6 +282,28 @@ describe('EffectSources', () => {
         expect(toActions(sources$)).toBeObservable(expected);
       });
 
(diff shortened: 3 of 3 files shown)


modules/effects/src/effect_notification.ts
```diff
@@ -19,13 +19,16 @@ export function reportInvalidActions(
...
+      try {
+        // handleError can throw errors
+        reporter.handleError(
+          new Error(
+            `Effect ${getEffectName(
+              output
+            )} dispatched an invalid action: ${stringify(action)}`
+          )
+        );
+      } catch (error) {}
```
Edouardbozon: Hi, I noticed a minor detail, this can be shortened without creating an unused binding.

```suggestion
      } catch {}
```

modules/effects/src/effects_resolver.ts
```diff
@@ -57,7 +57,13 @@ function resubscribeInCaseOfError<T extends Action>(
 ): Observable<T> {
   return observable$.pipe(
     catchError(error => {
-      if (errorHandler) errorHandler.handleError(error);
+      if (errorHandler) {
+        try {
+          errorHandler.handleError(error);
+        } catch (error) {
```
Edouardbozon: And also here.

modules/effects/src/effect_notification.ts
alex-okrushko: This just swallows the error. I think error handler should not be throwing errors to begin with.

lydemann: I think it is a common practice, that developers throw in errorHandler, to stop the js from executing. This is also the default behavior without having implemented a custom error handler:
https://medium.com/@amcdnl/global-error-handling-with-angular2-6b992bdfb59c

alex-okrushko: I see. What if it's done unintentionally? Debugging swallowed exceptions would be really hard, right?

lydemann: Yes, we should log it at least.
We could do a `console.error` in the catch?

alex-okrushko: That's what default error handler does: https://github.com/angular/angular/blob/0cd843169884a57d92c1ec114f7e778720b5ef24/packages/core/src/errors.ts#L30

alex-okrushko: btw, the default `ErrorHandler` from angular does not throw error: https://github.com/angular/angular/blob/master/packages/core/src/error_handler.ts#L44

lydemann: Ok I see.

lydemann: The question is if this should still be handled in case error handler throws?

timdeschryver: I would not expect that the error handler throws an error.
If it's an unexpected exception, my opinion is that it must be handled in the error handler itself. But if we want to resubscribe on error, we can't make this assumption, and expect that it's possible that the error handler throws.

lydemann: Ok, I will close the PR. I just had some problems with this but I am not throwing in the error handler anymore.

zakhenry:
@lydemann fyi I’ve just opened https://github.com/ngrx/platform/pull/2295 which will allow for external customisation of the error handling behaviour. Hopefully if it is accepted it will work for your use case too 😄

PR #2295, opened 2019-12-18
feat(effects): make resubscription handler overridable

### LINKED ISSUE #2294: RFC: Ability to customise resubscribeInCaseOfError functionality to allow for user interaction with errors
Currently the `resubscribeInCaseOfError` function is not configurable, and it simply calls `errorHandler.handleError(error)` and resubscribes. This behavior is not sufficient in the case that the application has a custom error handler that allows users to decide whether or not to retry the failure.

The cleanest solution would likely be to `provide` the `resubscribeInCaseOfError` function using an injected symbol, which would allow downstream consumers to do something along the lines of the following:

```ts
{
  provide: RESUBSCRIBE_ON_ERROR_HANDLER,
  deps: [MyCustomErrorHandler],
  useFactory: (errorHandler: ErrorHander) => {
    return (observable$) => observable$.pipe(
      retryWhen(errors =>
        errors.pipe(
          switchMap((e: IsRetryableError) => {

            if (e.isRetryable) {
              return errorHandler.handleRecoverableError(e)
            }

            errorHandler.handleError(e);
            return throwError(e);
          }),
        ),
      ),
    );
  },
},
```

Using a custom rxjs operator for every effect

[ ] No

zakhenry:
## PR Checklist

Please check if your PR fulfils the following requirements:

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Behavior of resubscription to closed observables is not configurable

Closes #2294

## What is the new behavior?
Behavior of resubscription to closed observables is now configurable by injecting a substitute implementation.

```
[x] Yes
[ ] No
```

Impact is relatively low - it is a simple rename of a field that was introduced in v8. See @alex-okrushko's [comment](https://github.com/ngrx/platform/issues/2294#issuecomment-573445745) - he found nowhere in google that is using this feature yet.

Migration path is a simple rename, an automated migration with a schematic is theoretically possible (I've not done this before but willing to learn).


brandonroberts:
Closing this along with #2294

zakhenry:
@brandonroberts would you mind reopening this PR please given #2294 was reopened?

CHANGELOG.md
```diff
@@ -199,7 +199,7 @@ login$ = createEffect(
       )
       // Errors are handled and it is safe to disable resubscription
     ),
-  { resubscribeOnError: false }
+  { useEffectsErrorHandler: false }
```
zakhenry: revert

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+        ...providers,
+      ],
+    });
+
+    TestBed.get(EffectSources);
+    globalErrorHandler = TestBed.get(ErrorHandler).handleError;
+    const store = TestBed.get(Store);
+    dispatch = store.dispatch;
+    storeNext = store.next;
+  }
```
alex-okrushko: I think these are not used in your tests

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    TestBed.configureTestingModule({
+      imports: [EffectsModule.forRoot([ErrorEffect])],
+      providers: [
+        {
+          provide: Store,
+          useValue: {
+            dispatch: jasmine.createSpy('dispatch'),
+            next: jasmine.createSpy('storeNext'),
+          },
+        },
```
alex-okrushko: I think you can just `provideMockStore()` here

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+          provide: ErrorHandler,
+          useValue: {
+            handleError: jasmine.createSpy('globalErrorHandler'),
+          },
+        },
+        ...providers,
+      ],
+    });
+
+    TestBed.get(EffectSources);
```
alex-okrushko: What does this do?

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    effect$ = createEffect(errorFirstSubscriber, {
+      useEffectsErrorHandler: true,
+    });
+  }
+
+  function errorFirstSubscriber(): Observable<Action> {
+    return new Observable(observer => {
+      subscriptionCount++;
+
+      if (subscriptionCount === 1) {
```
alex-okrushko: This condition is met right away, right?

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    globalErrorHandler = TestBed.get(ErrorHandler).handleError;
+    const store = TestBed.get(Store);
+    dispatch = store.dispatch;
+    storeNext = store.next;
+  }
+
+  it('should retry and notify error handler when effect error handler is not provided', () => {
+    makeEffectTestBed();
+
+    expect(subscriptionCount).toBe(2);
```
alex-okrushko: Why is it `2`?

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  EffectSources,
+} from '..';
+
+describe('NgRx Effects Error Handler spec', () => {
+  let subscriptionCount: number;
+  let globalErrorHandler: jasmine.Spy;
+  let dispatch: jasmine.Spy;
+  let storeNext: jasmine.Spy;
+
+  function makeEffectTestBed(...providers: any[]) {
```
alex-okrushko: types please :)

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  it('should retry and notify error handler when effect error handler is not provided', () => {
+    makeEffectTestBed();
+
+    expect(subscriptionCount).toBe(2);
+    expect(globalErrorHandler).toHaveBeenCalledWith(new Error('effectError'));
+  });
+
+  it('should use custom error behavior when EFFECTS_ERROR_HANDLER is provided', () => {
+    const effectsErrorHandlerSpy = jasmine.createSpy();
+    effectsErrorHandlerSpy.and.callFake((_$, errorHandler: ErrorHandler) => {
```
alex-okrushko: you could continue that chain `const effectsErrorHandlerSpy = jasmine.createSpy().and.callFake(...);`

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    expect(effectsErrorHandlerSpy).toHaveBeenCalledWith(
+      jasmine.any(Observable),
+      TestBed.get(ErrorHandler)
+    );
+    expect(globalErrorHandler).toHaveBeenCalledWith(
+      new Error('inside custom handler')
+    );
+    expect(subscriptionCount).toBe(0);
```
alex-okrushko: this one is confusing. Why is it zero?

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    expect(storeNext).toHaveBeenCalledWith({ type: 'custom action' });
```
alex-okrushko: I find this assertion to be irrelevant for what you are putting under test

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: I don't believe that I can because I need to later assert on the `Store.next` method being called, which is immediately called when the test bed initialises, not allowing me to use the MockStore's public `scannedActions$` to assert on. I guess I could change the effect under test to only error when I tell it to by controlling a subject that it returns?

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: I think they are? or at least they are now :)

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: 1. Initial subscription to the effect (this will error)
2. Resubscription to the effect after error (this will not error)

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: Should I break it into another `it()` block? I think it is important to test that the value returned from the custom error handler is what ends up being passed to the store as an action stream.

I've updated the error handler substitution to make this clearer

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: Correct, but on subsequent subscribers to this observable it will not error (it won't emit anything either)

zakhenry:
> Correct me if I'm wrong, if we set `EFFECTS_ERROR_HANDLER` in any of the modules, it would override the default effects error handle for all the effects classes.
> 
> I think we need something that we can pass to a particular `EffectsModule` either `forRoot` or `forChild` and it would apply only to that group of Effects. WDYT?

@alex-okrushko Hmmm this is an interesting thought - I'm not sure that I'm convinced that the effects grouping is the level that developers would want to customise the behaviour, because at that point they would have limited ability to customise the behavior on a per-class basis.

Perhaps the better solution to address that case would be to (maybe later) introduce a new hook, and if developers desire grouping within a module it could be done through inheritance or a custom decorator if they so choose.

e.g.

```ts
class MyEffects implements EffectErrorHandler {

   ngrxEffectErrorHandler(action$: Observable<Action>, errorHandler: ErrorHandler) {
     // ...
   }

   effect$ = createEffect(...);

}
```

This would be more expected IMO considering similar control behavior with `OnRunEffects` being class hook based.

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: I've refactored this to assert count is `1`. It was actually correct before too, as I was not subscribing to the effect stream at all, instead I was replacing the effect stream entirely. This is not a very realistic test, so I've updated the test to make it actually subscribe to the stream and handle the error, this now means there is actually one subscription.

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: I attempted to do the above, but it ended up being even more complicated and difficult to read; I think what I have may be the simpler option unless I've missed something obvious.

alex-okrushko:
This is still on my radar - will try to get some time to review it tomorrow.

zakhenry:
@alex-okrushko yea no problem, whenever is fine

modules/effects/spec/effects_error_handler.spec.ts
alex-okrushko: Can you please add that comment to the code (so it's clearer when reading the test)?

modules/effects/spec/effects_error_handler.spec.ts
alex-okrushko: would be nice to see that comment in the code as well. I missed that at first.

modules/effects/src/effect_sources.ts
```diff
@@ -14,18 +14,25 @@ import {
...
 import { getSourceForInstance } from './utils';
 
 @Injectable()
 export class EffectSources extends Subject<any> {
-  constructor(private errorHandler: ErrorHandler, private store: Store<any>) {
+  constructor(
+    private errorHandler: ErrorHandler,
+    private store: Store<any>,
+    @Optional()
+    @Inject(EFFECTS_ERROR_HANDLER)
```
alex-okrushko: [Optional]: just a thought
I wonder if instead of making it `@Optional()` here, maybe it's worth providing it in the EffectsRootModule?
WDYT? @timdeschryver WDYT as well?

modules/effects/src/effects_resolver.ts
```diff
@@ -1,38 +1,50 @@
...
+} from 'rxjs/operators';
 
 import { EffectNotification } from './effect_notification';
 import { getSourceMetadata } from './effects_metadata';
 import { getSourceForInstance } from './utils';
 import { ErrorHandler } from '@angular/core';
 
+export type EffectsErrorHandler = <T extends Action>(
+  observable$: Observable<T>,
+  errorHandler?: ErrorHandler
```
alex-okrushko: we should probably make it not-optional:
- it would be harder to add addition _required_ arguments (should our custom error handling functions have those)
- it's present Angular and passed into `mergeEffects` anyway

modules/effects/src/effect_sources.ts
zakhenry: Hmm I'm on the fence as it doesn't really change much functionality, the choice may be more based on future API changes. In some sense it would be tidier as most of the other symbols are provided there too. Tell you what I'll make the change in a separate commit that I can easily drop or squash based on feedback from there.

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: done

modules/effects/spec/effects_error_handler.spec.ts
zakhenry: done

modules/effects/src/effects_resolver.ts
zakhenry: agree, done

modules/effects/src/effect_sources.ts
zakhenry: OK new commit added to PR with just this change. I also had to rebase hence the noise on the PR pushes.

timdeschryver: Sorry for the late response, but I agree here with @alex-okrushko 
But I see it's already resolved 👍👍

timdeschryver:
I like the API this PR provides, good job 👍

> Migration path is a simple rename, an automated migration with a schematic is theoretically possible (I've not done this before but willing to learn).

Yes, this is possible and should be a simpler one.
If you want I can grab this one after this PR lands, or assist you in making a PR - the choice is yours 🙂

modules/effects/spec/effects_error_handler.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Observable, of } from 'rxjs';
+import { catchError } from 'rxjs/operators';
+import { createEffect, EFFECTS_ERROR_HANDLER, EffectsModule } from '..';
+
+describe('NgRx Effects Error Handler spec', () => {
+  let subscriptionCount: number;
+  let globalErrorHandler: jasmine.Spy;
+  let storeNext: jasmine.Spy;
+
+  function makeEffectTestBed(...providers: Provider[]) {
```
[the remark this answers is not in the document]
zakhenry: So I tried to do this refactor and ended up having to create a bunch of duplicate code, which is a step backwards in my opinion. This test is somewhat unusual compared to other specs in that all the behavior happens on startup (the only difference in the two tests is the providers, which can only be set when initialising the whole test bed.
To use beforeEach, I would have to completely re-declare the `TestBed.configureTestingModule` in separate beforeEach blocks.

I think it is safer to have the spec focussed on what is different between each test rather than following the pattern used elsewhere.

modules/effects/src/tokens.ts
```diff
@@ -12,3 +13,6 @@ export const ROOT_EFFECTS = new InjectionToken<Type<any>[]>(
 export const FEATURE_EFFECTS = new InjectionToken<any[][]>(
   'ngrx/effects: Feature Effects'
 );
+export const EFFECTS_ERROR_HANDLER = new InjectionToken<EffectsErrorHandler>(
```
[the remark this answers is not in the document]
zakhenry: this symbol is exported directly from index.ts, along with most other type exports. Should I move it here?

modules/effects/src/effects_resolver.ts
```diff
@@ -51,9 +57,9 @@ export function mergeEffects(
   return merge(...observables$);
 }
 
-function resubscribeInCaseOfError<T extends Action>(
+export function resubscribeInCaseOfError<T extends Action>(
```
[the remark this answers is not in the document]
zakhenry: Yep agreed, I've done a little more that that though that needs confirmation. In order to both ensure type safety and describe intent I renamed that function to `defaultEffectsErrorHandler` and made it a constant so that it can be strongly typed with the exported type symbol. wdyt?

projects/ngrx.io/content/guide/effects/lifecycle.md
```diff
@@ -91,6 +91,57 @@ export class AuthEffects {
...
+
+@NgModule({
+  imports: [EffectsModule.forRoot([MovieEffects])],
+  providers: [
+    {
+      provide: EFFECTS_ERROR_HANDLER,
+      useValue: effectResubscriptionHandler,
+    },
+    {
+      provide: ErrorHandle,
```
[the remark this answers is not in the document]
zakhenry: Oops! This is when I should plug my library [embedme](https://github.com/zakhenry/embedme) which allows for markdown files to import typescript files directly, ensuring mistakes like this can't slip through, plus ensures breaking changes in documentation code is caught. 😁

modules/effects/spec/effects_error_handler.spec.ts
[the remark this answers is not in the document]
timdeschryver: I missed the `providers` argument. 
👍

modules/effects/src/tokens.ts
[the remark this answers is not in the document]
timdeschryver: If it's in the `index.ts` file it's good

zakhenry:
@timdeschryver thanks, I'm going to take a crack at the migration PR, I'll let you know if/when I need a hand. I've got the basics working already, just need to narrow the visitor so it's not just doing a dangerous global find/replace. I'll open the PR this weekend to discuss over. I'll keep this PR for the actual change 👍

modules/effects/src/effects_error_handler.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Action } from '@ngrx/store';
+import { Observable } from 'rxjs';
+import { catchError } from 'rxjs/operators';
+
+export type EffectsErrorHandler = <T extends Action>(
+  observable$: Observable<T>,
+  errorHandler: ErrorHandler
+) => Observable<T>;
+
+// this is the default error handler provided in EffectsModule.forRoot
```
timdeschryver: Nit, I think this comment can be left out

modules/effects/src/effects_resolver.ts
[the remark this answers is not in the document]
timdeschryver: I agree, the name makes it more clear

modules/effects/src/effects_error_handler.ts
zakhenry: dropped

PR #2313, opened 2020-01-10
fix(store): add not allowed check to action creator config

timdeschryver:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The last commit broke the build, this PR fixes the build.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/types/utils.ts
+++ b/modules/store/spec/types/utils.ts
@@ -2,6 +2,7 @@ export const compilerOptions = () => ({
   moduleResolution: 'node',
   target: 'es2015',
   baseUrl: '.',
+  experimentalDecorators: true,
   paths: {
     '@ngrx/store': ['./modules/store'],
   },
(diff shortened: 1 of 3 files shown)


modules/store/spec/types/utils.ts
```diff
@@ -2,6 +2,7 @@ export const compilerOptions = () => ({
   moduleResolution: 'node',
   target: 'es2015',
   baseUrl: '.',
+  experimentalDecorators: true,
```
timdeschryver: Added for debug purposes, because `expectSnippet().toSucceed()` gives the error on the `@Injectable` decorators.

alex-okrushko: this comment should be in the file itself :)

timdeschryver: 🤯 I usually just putted the comment in the PR, but it now that I read this comment it's indeed better to add it in the file. Thanks for the tip!

PR #2319, opened 2020-01-17
fix(store): improve type-safety for selector's projector func parameters

### LINKED ISSUE #2317: No type-safety for selector's projector function
image

Is there a reason why projector function does not have a type-safety for its parameters yet?
This quickly becomes a nightmare when having a lot of unit tests for complicated projectors.
I've looked into internals and found that for some reason the [`MemoizedSelector`](https://github.com/ngrx/platform/blob/2cc88857421b74e01b816f99fe28234bbe77ff1f/modules/store/src/selector.ts#L34) is using a default type [`DefaultProjectorFn`](https://github.com/ngrx/platform/blob/2cc88857421b74e01b816f99fe28234bbe77ff1f/modules/store/src/selector.ts#L17) for its projector which does not know anything about parameter types but only `Result`:

```ts
export function createSelector<State, S1, Result>(
  s1: Selector<State, S1>,
  projector: (s1: S1) => Result
): MemoizedSelector<State, Result>; // <-- third type is a `DefaultProjectorFn`
```

This could be easily fixed by explicitly providing a type of projector function (for every `createSelector` of course):
```ts
export function createSelector<State, S1, Result>(
  s1: Selector<State, S1>,
  projector: (s1: S1) => Result
): MemoizedSelector<State, Result, (s1: S1) => Result>;
```

image

Maybe this was already discussed? Are there any hidden issues that I'm not aware of? 
Otherwise it's strange why this was not escalated yet.

[x] Yes
[ ] No

I have all the changes ready to be reviewed - just let me know if the solution here is appropriate.

dzonatan:
## PR Checklist

## What is the current/new behavior?

Closes #2317

Kind-of. See https://github.com/ngrx/platform/issues/2317#issuecomment-574485374.

### DIFF
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -120,58 +120,58 @@ export function defaultMemoize(
 export function createSelector<State, S1, Result>(
   s1: Selector<State, S1>,
   projector: (s1: S1) => Result
-): MemoizedSelector<State, Result>;
+): MemoizedSelector<State, Result, typeof projector>;
 export function createSelector<State, Props, S1, Result>(
   s1: SelectorWithProps<State, Props, S1>,
   projector: (s1: S1, props: Props) => Result
-): MemoizedSelectorWithProps<State, Props, Result>;
+): MemoizedSelectorWithProps<State, Props, Result, typeof projector>;
 export function createSelector<State, S1, Result>(
   selectors: [Selector<State, S1>],
   projector: (s1: S1) => Result
-): MemoizedSelector<State, Result>;
+): MemoizedSelector<State, Result, typeof projector>;
 export function createSelector<State, Props, S1, Result>(
   selectors: [SelectorWithProps<State, Props, S1>],
   projector: (s1: S1, props: Props) => Result
-): MemoizedSelectorWithProps<State, Props, Result>;
+): MemoizedSelectorWithProps<State, Props, Result, typeof projector>;
 
 export function createSelector<State, S1, S2, Result>(
   s1: Selector<State, S1>,
   s2: Selector<State, S2>,
   projector: (s1: S1, s2: S2) => Result
-): MemoizedSelector<State, Result>;
+): MemoizedSelector<State, Result, typeof projector>;
 export function createSelector<State, Props, S1, S2, Result>(
   s1: SelectorWithProps<State, Props, S1>,
   s2: SelectorWithProps<State, Props, S2>,
   projector: (s1: S1, s2: S2, props: Props) => Result
-): MemoizedSelectorWithProps<State, Props, Result>;
+): MemoizedSelectorWithProps<State, Props, Result, typeof projector>;
 export function createSelector<State, S1, S2, Result>(
   selectors: [Selector<State, S1>, Selector<State, S2>],
   projector: (s1: S1, s2: S2) => Result
-): MemoizedSelector<State, Result>;
+): MemoizedSelector<State, Result, typeof projector>;
 export function createSelector<State, Props, S1, S2, Result>(
   selectors: [
     SelectorWithProps<State, Props, S1>,
     SelectorWithProps<State, Props, S2>
   ],
   projector: (s1: S1, s2: S2, props: Props) => Result
-): MemoizedSelectorWithProps<State, Props, Result>;
+): MemoizedSelectorWithProps<State, Props, Result, typeof projector>;
 
 export function createSelector<State, S1, S2, S3, Result>(
   s1: Selector<State, S1>,
   s2: Selector<State, S2>,
   s3: Selector<State, S3>,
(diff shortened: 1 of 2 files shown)


modules/store/src/selector.ts
```diff
@@ -120,81 +120,122 @@ export function defaultMemoize(
 export function createSelector<State, S1, Result>(
   s1: Selector<State, S1>,
   projector: (s1: S1) => Result
-): MemoizedSelector<State, Result>;
+): MemoizedSelector<State, Result, (s1: S1) => Result>;
```
alex-okrushko: I think this can be replaced with `typeof projector` in all of the function definitions.

modules/store/src/selector.ts
```diff
@@ -120,81 +120,122 @@ export function defaultMemoize(
...
+  (s1: S1, props: Props) => Result
+>;
 
-export function createSelector<State, S1, S2, Result>(
+export function createSelector<
+  State,
+  S1,
+  S2,
+  Result,
+  Projector extends (s1: S1, s2: S2) => Result
```
alex-okrushko: ?

modules/store/src/selector.ts
```diff
@@ -120,81 +120,122 @@ export function defaultMemoize(
...
+): MemoizedSelector<State, Result, (s1: S1) => Result>;
 export function createSelector<State, Props, S1, Result>(
   s1: SelectorWithProps<State, Props, S1>,
   projector: (s1: S1, props: Props) => Result
-): MemoizedSelectorWithProps<State, Props, Result>;
+): MemoizedSelectorWithProps<
+  State,
+  Props,
+  Result,
+  (s1: S1, props: Props) => Result
```
alex-okrushko: what does this do here?

modules/store/src/selector.ts
dzonatan: Sorry, it was a leftover from my playing around to come up with something more wise but you nailed it with `typeof projector`. 👏

modules/store/src/selector.ts
dzonatan: Nice! Thanks.

modules/store/src/selector.ts
dzonatan: Not sure what isn't clear here. 🤔

alex-okrushko:
Some tests are failing: https://circleci.com/gh/ngrx/platform/21244

dzonatan:
> Some tests are failing: https://circleci.com/gh/ngrx/platform/21244

Looks like not related to my changes. Rebased on master - should pass now.

modules/store/src/selector.ts
alex-okrushko: Well, you removed it 😜

alex-okrushko:
Let me run this through global tests and see what's the damage

modules/store/src/selector.ts
dzonatan: I've changed it to `typeof projector`.

alex-okrushko:
So I ran the deltas:
currently with `DefaultProjectorFn` it breaks 54 test targets (that's not the number of tests - which is a lot more, to give some perspective - within ngrx/platfrom [ngrx/effects/spec:test](https://github.com/ngrx/platform/blob/7b3aea3b662e03848b4e74110a6ab7d6b7131c38/modules/effects/spec/BUILD#L21))

With this PR on top, that breakage number goes to 159.
Failures are typically like this:

```ts
error TS2741: Property 'id' is missing in type '{ count: number; }' but required in type 'Syncing'.

14         expect(selectors.isSyncing().projector({'makeFoo': {count: 1}}, {
```

Don't know if it's really worth it. Might be for the new codebase, but upgrading would be annoying (these are not real potential prod issues that we are catching - just stricter tests).

brandonroberts:
Going to close this one for now. We will revisit to see if we want to make this change for v10

david-shortman:
Since v10 is out, I'm wondering if anyone has thoughts on revisiting this for v11. Adding types is always gonna be a breaking change I suppose, so it seems it's a merely a matter of committing to a certain release.

NicBright:
Citing @timdeschryver :
> LGTM, but as @alex-okrushko pointed out this will potentially break a lot of tests.

A first shot at trying to mitigate this:

Maybe a workaround can be suggested to those that have only little time for upgrading:

- do a Find and Replace in all your Spec Files:
- Find (by regex): `(\.projector\()`
- Replace by: `\n// @ts-ignore\n$1`

Admittedly: the code will look very ugly but this should handle most of the problems. It works as long as all object-literal arguments of a projector start on the same line. And in case it doesn't work, it's not so much work to manually add another `// @ts-ignore` where necessary.

Example:
- Before: 
image
- After (before prettier formatting): 
image
- After (after prettier formatting), now everything's ok: 
image



What do you think @alex-okrushko @timdeschryver ?

PR #2320, opened 2020-01-17
refactor(effects): lifecycle methods guards

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/src/effect_sources.ts
+++ b/modules/effects/src/effect_sources.ts
@@ -21,6 +21,9 @@ import {
   onRunEffectsKey,
   OnRunEffects,
   onInitEffects,
+  isOnIdentifyEffects,
+  isOnRunEffects,
+  isOnInitEffects,
 } from './lifecycle_hooks';
 import { getSourceForInstance } from './utils';
 
@@ -44,10 +47,7 @@ export class EffectSources extends Subject<any> {
         return source$.pipe(
           groupBy(effectsInstance),
           tap(() => {
-            if (
-              onInitEffects in source$.key &&
-              typeof source$.key[onInitEffects] === 'function'
-            ) {
+            if (isOnInitEffects(source$.key)) {
               this.store.dispatch(source$.key.ngrxOnInitEffects());
             }
           })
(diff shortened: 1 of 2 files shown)


modules/effects/src/effect_sources.ts
```diff
@@ -72,11 +72,8 @@ export class EffectSources extends Subject<any> {
 }
 
 function effectsInstance(sourceInstance: any) {
-  if (
-    onIdentifyEffectsKey in sourceInstance &&
-    typeof sourceInstance[onIdentifyEffectsKey] === 'function'
-  ) {
-    return sourceInstance[onIdentifyEffectsKey]();
+  if (isOnIdentifyEffects(sourceInstance)) {
+    return sourceInstance.ngrxOnIdentifyEffects();
```
alex-okrushko: could you please add `declare` to all interfaces (`OnIdentifyEffects`, `OnRunEffects ` and `OnInitEffects `), so they are not property-renamed and we are able to call these methods on the instance.

timdeschryver: @alex-okrushko I can, and I did 🙂

tonivj5: @alex-okrushko out of curiosity, I don't end up of understand why `declare` keyword is needed with interfaces :thinking:

alex-okrushko: Using `declare` keyword is helping TS mark an interface as external. This will help some JS Compilers (for example like [Closure Compiler](https://developers.google.com/closure/compiler)) that do property renaming, **not to rename these fields**. 
e.g. object with prop rename would turn `{ngrxOnIdentifyEffects: () => doStuff()}` to `{bds: () => aef()}` which helps reduce the final bundle size.

Why would we care if these are renamed in this case?
We are [passing function names](https://github.com/ngrx/platform/pull/2320/files#diff-7f170bfaeb28b7144e3545dbe130a918R127) as strings, and string are not renamed, so we would attempt to check/call `obj['ngrxOnIdentifyEffects']` while the prop was renamed to `bds` and so it won't be found, causing a bug (or unexpected behavior).

PR #2323, opened 2020-01-19
refactor(example): use featureKey for layout selector

erhise:
## PR Checklist

## PR Type

Consistently use feature keys exported from reducer file, e.g. where selectors for underlying state are being created.

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Both feature keys and hard-corded values are used.

## What is the new behavior?

Declared feature keys are being used in all places where they should be referenced which scope potential name changes to one place.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/example-app/src/app/auth/reducers/index.ts
+++ b/projects/example-app/src/app/auth/reducers/index.ts
@@ -32,7 +32,7 @@ export const selectAuthState = createFeatureSelector<State, AuthState>(
 
 export const selectAuthStatusState = createSelector(
   selectAuthState,
-  (state: AuthState) => state.status
+  state => state.status
 );
 export const selectUser = createSelector(
   selectAuthStatusState,
@@ -42,7 +42,7 @@ export const selectLoggedIn = createSelector(selectUser, user => !!user);
 
 export const selectLoginPageState = createSelector(
   selectAuthState,
-  (state: AuthState) => state.loginPage
+  state => state.loginPage
 );
 export const selectLoginPageError = createSelector(
   selectLoginPageState,
--- a/projects/example-app/src/app/reducers/index.ts
+++ b/projects/example-app/src/app/reducers/index.ts
@@ -69,7 +69,7 @@ export const metaReducers: MetaReducer<State>[] = !environment.production
  * Layout Reducers
  */
 export const selectLayoutState = createFeatureSelector<State, fromLayout.State>(
-  'layout'
+  fromLayout.layoutFeatureKey
 );
 
 export const selecthowSidenav = createSelector(
--- a/projects/example-app/src/app/books/reducers/index.ts
+++ b/projects/example-app/src/app/books/reducers/index.ts
@@ -103,7 +103,7 @@ export const selectSelectedBook = createSelector(
  */
 export const selectSearchState = createSelector(
   selectBooksState,
-  (state: BooksState) => state.search
+  state => state.search
 );
 
 export const selectSearchBookIds = createSelector(
@@ -139,7 +139,7 @@ export const selectSearchResults = createSelector(
 
 export const selectCollectionState = createSelector(
   selectBooksState,
-  (state: BooksState) => state.collection
+  state => state.collection
 );
 
 export const selectCollectionLoaded = createSelector(


projects/example-app/src/app/auth/reducers/index.ts
```diff
@@ -32,7 +32,7 @@ export const selectAuthState = createFeatureSelector<State, AuthState>(
 
 export const selectAuthStatusState = createSelector(
   selectAuthState,
-  (state: AuthState) => state.status
+  (state: AuthState) => state[fromAuth.statusFeatureKey]
```
timdeschryver: Personally, I'm unsure if this adds value.
The difference between this selector and a `createFeatureSelector`, is that this selector is type-checked.

projects/example-app/src/app/reducers/index.ts
```diff
@@ -69,7 +69,7 @@ export const metaReducers: MetaReducer<State>[] = !environment.production
  * Layout Reducers
  */
 export const selectLayoutState = createFeatureSelector<State, fromLayout.State>(
-  'layout'
+  fromLayout.layoutFeatureKey
```
timdeschryver: 👍

projects/example-app/src/app/auth/reducers/index.ts
```diff
@@ -42,7 +42,7 @@ export const selectLoggedIn = createSelector(selectUser, user => !!user);
 
 export const selectLoginPageState = createSelector(
   selectAuthState,
-  (state: AuthState) => state.loginPage
+  (state: AuthState) => state[fromLoginPage.loginPageFeatureKey]
```
timdeschryver: By seeing this PR, I think we can/should remove the type in these selectors?
```suggestion
  state => state[fromLoginPage.loginPageFeatureKey]
```

projects/example-app/src/app/auth/reducers/index.ts
erhise: Do you mean to remove the `selectAuthStatusState` selector completely and then change `selectUser` selector?

timdeschryver: No, not exactly. I only see a benefit of using feature keys for top-level slices (with `createFeatureSelector`). For selectors with the `createSelector` function, we have type safety in place and there are no magic strings because you can access the property (e.g. `state.status`). But this is my personal opinion 🙂

projects/example-app/src/app/auth/reducers/index.ts
timdeschryver: Thanks for making this change!

projects/example-app/src/app/auth/reducers/index.ts
erhise: Okey, now I see what you mean :smile: . I kind of agree 95% agree. One thing to add before dismissal would be,

* `state.status` would indeed be typechecked and give error for e.g. `state.notStatus`
* but for the other example,
```
// statusFeatureKey = 'status';

const randomValue = 'notStatus';

export interface AuthState {
  [fromAuth.statusFeatureKey]: fromAuth.State;
}

export const selectAuthStatusState = createSelector(
  selectAuthState,
  state => state[randomValue] <-- also gives type error 'notStatus' does not exist
);

// if randomValue instead has matching value
const randomValue = 'status';

state => state[randomValue] <-- this would be fine from type checking perspective
```

So the only thing here would be that if someone changes e.g. `statusFeatureKey` within the auth.reducer.ts there would be no need for direct changes if `state => state[fromAuth.statusFeatureKey]`.

However, I could see this as an invalid argue since the name of the selector presumably should be changed (selectAuth~~Status~~NewNameState).

I let you decide :relaxed:

erhise: I preventively removed all the unnecessary things and squashed all the changes to one commit.

timdeschryver: Yes, you're right. 
(Also, the main IDEs can do a refactor to rename all occurrences of the property name)

We can wait for someone else his/her opinion before deciding.

alex-okrushko:
> "use featurekey in all places" 
I see only one place that changed :)

projects/example-app/src/app/reducers/index.ts
```diff
@@ -69,7 +69,7 @@ export const metaReducers: MetaReducer<State>[] = !environment.production
  * Layout Reducers
  */
 export const selectLayoutState = createFeatureSelector<State, fromLayout.State>(
```
alex-okrushko: just one generic is fine `createFeatureSelector<fromLayout.State>(...)`
Typically there is no global `State` and many features are lazy-loaded, so there cannot be a single interface that connects them all.

timdeschryver:
@alex-okrushko  see https://github.com/ngrx/platform/commit/691189f01afa39240e05236f5905bbec82a4debb for the original commit.

The changes that were made were (but are now reverted):

```diff
export const selectAuthState = createFeatureSelector<State, AuthState>(
  authFeatureKey
);

export const selectAuthStatusState = createSelector(
  selectAuthState,
-  (state: AuthState) => state.status
+  (state: AuthState) => state[fromAuth.statusFeatureKey]
);
```

But I'm not a big fan of using the feature keys for `createSelectors`. See the conversation we had at https://github.com/ngrx/platform/pull/2323#discussion_r369780018.
Pinged you and Brandon to hear your thoughts on it.

alex-okrushko:
> But I'm not a big fan of using the feature keys for createSelectors

100% agree. Feature keys only for `createFeatureSelector` which sets the type of the rest of selectors for a feature.

projects/example-app/src/app/reducers/index.ts
erhise: If this is to be added I guess it would imply changes to all other places where FeatureSelectors are being created e.g. https://github.com/ngrx/platform/blob/4b48e51f95/projects/example-app/src/app/auth/reducers/index.ts#L29 ??

alex-okrushko: I think so, yes. Maybe in the follow-up PRs.
Also, after PR https://github.com/ngrx/platform/pull/2325 lands there would be some more changes
@timdeschryver @brandonroberts hope you don't mind.

timdeschryver: The 2 generics are there to provide type safety, but you have to extend the Feature state with the Root State:

```ts
export interface State extends fromRoot.State {
  [booksFeatureKey]: BooksState;
}
```

But this was before the usage of feature keys.

I think we could indeed refactor this, but I don't got a strong opinion on this. 
Sometimes it will be needed to extend from the root, if you're combining selectors from the feature with the root.

We can take this one in a later PR and merge this PR if everything is OK with it.

alex-okrushko: > Sometimes it will still be needed to extend from the root, this is when you're combining selectors from the feature with the root.

There are workarounds for this :P

brandonroberts:
I agree also. Let the types flow from the top

timdeschryver:
Thanks @erhise

PR #2335, opened 2020-01-25
Schematic for customisable error handling migration

zakhenry:
## PR Checklist

<strike>- [ ] Documentation has been added / updated (for bug fixes / features)</strike>

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: Migration schematic to migrate breaking change introduced in effects.
```

https://github.com/ngrx/platform/pull/2295 introduces a breaking change, this PR adds a migration to ease the transition to ngrx@9 for developers.

Closes #

## What is the new behavior?

On upgrade to ngrx 9 developers can run the migration to automatically fix the renamed configuration variable

```
[ ] Yes
[x] No
```

## Other information
~This PR is pending the merge of https://github.com/ngrx/platform/pull/2295, thus it currently contains those commits.~

### DIFF
--- a/modules/effects/migrations/9_0_0/index.spec.ts
+++ b/modules/effects/migrations/9_0_0/index.spec.ts
@@ -1,0 +1,146 @@
+import { Tree } from '@angular-devkit/schematics';
+import {
+  SchematicTestRunner,
+  UnitTestTree,
+} from '@angular-devkit/schematics/testing';
+import * as path from 'path';
+import { createPackageJson } from '../../../schematics-core/testing/create-package';
+
+describe('Effects Migration 9_0_0', () => {
+  let appTree: UnitTestTree;
+  const collectionPath = path.join(__dirname, '../migration.json');
+  const pkgName = 'effects';
+
+  beforeEach(() => {
+    appTree = new UnitTestTree(Tree.empty());
+    appTree.create(
+      '/tsconfig.json',
+      `
+        {
+          "include": [**./*.ts"]
+        }
+       `
+    );
+    createPackageJson('', pkgName, appTree);
+  });
+
+  describe('Replaces resubscribeOnError with useEffectsErrorHandler in effect options', () => {
+    describe('should replace resubscribeOnError configuration key with useEffectsErrorHandler', () => {
+      it('in createEffect() effect creator', () => {
+        const input = `
+  import { Injectable } from '@angular/core';
+  import { Actions, ofType, createEffect } from '@ngrx/effects';
+  import { tap } from 'rxjs/operators';
+   
+  @Injectable()
+  export class LogEffects {
+    constructor(private actions$: Actions) {}
+    
+    logActions$ = createEffect(() =>
+      this.actions$.pipe(
+        tap(action => console.log(action))
+      ), { resubscribeOnError: false });
+  }
+        `;
+
+        const expected = `
+  import { Injectable } from '@angular/core';
+  import { Actions, ofType, createEffect } from '@ngrx/effects';
+  import { tap } from 'rxjs/operators';
+   
+  @Injectable()
+  export class LogEffects {
+    constructor(private actions$: Actions) {}
+    
+    logActions$ = createEffect(() =>
+      this.actions$.pipe(
+        tap(action => console.log(action))
+      ), { useEffectsErrorHandler: false });
+  }
+        `;
+
+        test(input, expected);
+      });
+
+      it('in @Effect() decorator', () => {
+        const input = `
+  import { Injectable } from '@angular/core';
+  import { Actions, Effect, ofType } from '@ngrx/effects';
+  import { tap } from 'rxjs/operators';
+   
+  @Injectable()
+  export class LogEffects {
+    constructor(private actions$: Actions) {}
+    
+    @Effect({ resubscribeOnError: false })
+    logActions$ = this.actions$.pipe(
+      tap(action => console.log(action))
+    )
+  }
+        `;
+
+        const expected = `
+  import { Injectable } from '@angular/core';
+  import { Actions, Effect, ofType } from '@ngrx/effects';
+  import { tap } from 'rxjs/operators';
+   
+  @Injectable()
+  export class LogEffects {
+    constructor(private actions$: Actions) {}
+    
+    @Effect({ useEffectsErrorHandler: false })
+    logActions$ = this.actions$.pipe(
+      tap(action => console.log(action))
+    )
+  }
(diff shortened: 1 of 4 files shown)


projects/ngrx.io/content/guide/effects/lifecycle.md
```diff
@@ -24,7 +24,7 @@ Usage:
 
 <code-example header="log.effects.ts">
 import { Injectable } from '@angular/core';
-import { Actions, Effect, ofType } from '@ngrx/effects';
+import { Actions, createEffect, ofType } from '@ngrx/effects';
```
zakhenry: this was just a driveby change that I noticed the import was wrong

modules/effects/migrations/9_0_0/index.ts
```diff
@@ -0,0 +1,103 @@
...
+) {
+  const changes: ReplaceChange[] = [];
+  ts.forEachChild(sourceFile, node => findKey(node, changes));
+  return changes;
+
+  function findKey(node: ts.Node, changes: ReplaceChange[]) {
+    if (
+      ts.isIdentifier(node) &&
+      node.text === OLD_TEXT &&
+      isEffectDecorator(node.parent.parent.parent.parent)
```
timdeschryver: I'm worried that this (`parent.parent.parent`) might throw an error in some cases.
What about finding the `createEffect` first, and then looking for `resubscribeOnError`?

modules/effects/migrations/9_0_0/index.ts
```diff
@@ -0,0 +1,103 @@
...
+) {
+  const changes: ReplaceChange[] = [];
+  ts.forEachChild(sourceFile, node => findKey(node, changes));
+  return changes;
+
+  function findKey(node: ts.Node, changes: ReplaceChange[]) {
+    if (
+      ts.isIdentifier(node) &&
+      node.text === OLD_TEXT &&
+      isCreateEffectFunction(node.parent.parent.parent)
```
timdeschryver: Same here

modules/effects/migrations/9_0_0/index.ts
zakhenry: Hm I worried about this too, so tested it. Interestingly node.parent is never undefined, I think it just becomes recursive at the top level somehow. I agree though that it would be less confusing to find the effect context first and then descend into children so I'll change that 👍 

It will be a little more complex I suspect as I will need to make sure not to descend into the effect declaration as that could plausibly contain identifiers that I shouldn't be changing.

zakhenry: Ok refactor done, I actually went with a different strategy (visitor pattern) which ended up being a lot cleaner

modules/effects/migrations/9_0_0/index.ts
zakhenry: done

zakhenry:
Rebased on master now that the dependent PR has merged.

modules/effects/migrations/9_0_0/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  sourceFile: ts.SourceFile,
+  oldText: string,
+  newText: string
+): ReplaceChange[] {
+  const changes: ReplaceChange[] = [];
+
+  ts.forEachChild(sourceFile, node => {
+    visitCreateEffectFunctionCreator(node, createEffectNode => {
+      // only search the second arg to createEffect() (the config object)
+      if (createEffectNode.arguments[1]) {
```
timdeschryver: Personally I would rather see this as:

```ts
const [configNode] = createEffectNode.arguments
```

This also makes the comment obsolete

modules/effects/migrations/9_0_0/index.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+        {
+          "include": [**./*.ts"]
+        }
+       `
+    );
+    createPackageJson('', pkgName, appTree);
+  });
+
+  describe('Replaces resubscribeOnError with useEffectsErrorHandler in effect options', () => {
+    describe('should replace resubscribeOnError configuration key with useEffectsErrorHandler', () => {
```
timdeschryver: this describe can be left out? It has more or less the same description as the parent

modules/effects/migrations/9_0_0/index.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+      this.actions$.pipe(
+        tap(action => console.log(action))
+      ), { useEffectsErrorHandler: false });
+  }
+        `;
+
+        test(input, expected);
+      });
+
+      it('in @Effect() effect creator', () => {
```
timdeschryver:
```suggestion
      it('in @Effect() decorator', () => {
```

projects/ngrx.io/content/guide/effects/lifecycle.md
timdeschryver: Could you create a separate PR for this change please?

timdeschryver:
I like this approach more in comparison to what we first had 👏

projects/ngrx.io/content/guide/effects/lifecycle.md
zakhenry: sure

modules/effects/migrations/9_0_0/index.ts
zakhenry: yep done

modules/effects/migrations/9_0_0/index.spec.ts
zakhenry: ah actually you've picked up a mistake in my describe nesting. the structure should be

```
Effects Migration 9_0_0
└── Replaces resubscribeOnError with useEffectsErrorHandler in effect options
    ├── should replace resubscribeOnError configuration key with useEffectsErrorHandler
    |   ├── in createEffect() effect creator
    |   ├── in @Effect() decorator
    └── should not replace non-ngrx identifiers
        ├── in module scope
        └── within create effect callback
```

fixed.

modules/effects/migrations/9_0_0/index.spec.ts
zakhenry: done

projects/ngrx.io/content/guide/effects/lifecycle.md
zakhenry: moved to https://github.com/ngrx/platform/pull/2336

PR #2361, opened 2020-02-10
feat(store): testing - clean up mock store a bit

alex-okrushko:
## PR Checklist

## PR Type

- Add tests that would consistently catch `resetSelectors` in `afterEach`
- Make `selectors` private, instead of static
- cleanup constructor
- remove function overloads for `overrideSelector` - can we typed all at once

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

`MockStore.selectors` is not longer static or visible. I don't think that's the API that we meant to support to begin with.

```
[x] Yes
[ ] No
```

### DIFF
--- a/modules/store/testing/spec/mock_store.spec.ts
+++ b/modules/store/testing/spec/mock_store.spec.ts
@@ -344,6 +344,17 @@ describe('Refreshing state', () => {
     expect(getTodoItems('p').length).toBe(1);
     expect(getTodoItems('p')[0].nativeElement.textContent.trim()).toBe('bbb');
   });
+
+  it('should work with overrideSelector twice', () => {
+    const newTodos = [{ name: 'bbb', done: true }];
+    mockStore.overrideSelector(todos, newTodos);
+    mockStore.refreshState();
+
+    fixture.detectChanges();
+
+    expect(getTodoItems('li').length).toBe(1);
+    expect(getTodoItems('li')[0].nativeElement.textContent.trim()).toBe('bbb');
+  });
 });
 
 describe('Cleans up after each test', () => {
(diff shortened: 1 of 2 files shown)


modules/store/testing/src/mock_store.ts
```diff
@@ -18,46 +18,43 @@ import { MOCK_SELECTORS } from './tokens';
 if (typeof afterEach === 'function') {
```
timdeschryver: Do we still need this (because the static selectors aren't static anymore)?

alex-okrushko: We still need it, because we don't know what test framework is using `MockStore` - it might not have `afterEach` teardown callback  🤷‍♂️
I'd still keep it.

modules/store/testing/spec/mock_store.spec.ts
```diff
@@ -383,3 +394,62 @@ describe('Cleans up after each test', () => {
...
 });
+
+describe('Resets selectors after each test', () => {
+  const selectorUnderTest = createSelector(
+    (state: any) => state,
+    state => state.value
+  );
+  let shouldSetMockStore = true;
+
+  function setupModules(shouldSetMockStore: boolean) {
```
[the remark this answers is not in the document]
alex-okrushko: It's used - I pass it at the invocation `setupModules(shouldSetMockStore);` I probably should've named the property to something else.

modules/store/testing/src/mock_store.ts
timdeschryver: Sorry, I highlighted the wrong line 😅
It was more the question if we still have to cleanup, since we don't using static selectors anymore.

alex-okrushko: Ah, yes, we still need to cleanup the reducers, because we set values on them (we keep them in MockStore as well, but we do set values on Memoized selectors that live outside).
I added the test that would fail if those are not cleaned up.

brandonroberts:
@alex-okrushko will you take a look at why the example app tests are failing?

alex-okrushko:
> @alex-okrushko will you take a look at why the example app tests are failing? I'd like to land this before we cut another beta

Yes. Looking into it right now

modules/store/testing/src/mock_store.ts
timdeschryver: Thanks!

PR #2376, opened 2020-02-13
feat(effects): limit retries to 10 by default

### LINKED ISSUE #2303: Effects: resubscribeOnError for "broken" effect results in infinite loop and ends with "RangeError: Maximum call stack size exceeded"

This one ended up being very hard to debug. (Resulted from the recent merge to Google attempt).

There is a test that imports a Module that imports many other Modules. Some of them import `EffectsModule.forFeature(...)`.
The author of the test tried hard to substitute all the Effects Classes with empty objects, so that effects are not running (with many `{provide: FooEffects, useValue: {}}`), however missed one. 
That one missed effect was producing an error immediately (due to the fact that it was using `store.select(selector)` and it was erroring within a `selector`)
Anyhow, the effect was immediately resubscribing and resubscribing, eventually ending up with _"RangeError: Maximum call stack size exceeded"_ error.

Nothing in the stacktrace was pointing to the resubscribing effect (and led to significant debugging time).

## Expected behavior:

Maybe after a number of failed attempts we should actually fail the effect? 100? And make it configurable. 
I wonder if that would still be better that a potentially frozen prod app.

[ ] No

alex-okrushko:
## PR Checklist

## PR Type

Limit the number of retries to 10 by default

Removes infinite loop if the effect is producing errors on every retry

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2303

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/spec/effects_error_handler.spec.ts
+++ b/modules/effects/spec/effects_error_handler.spec.ts
@@ -1,20 +1,21 @@
-import { ErrorHandler, Provider } from '@angular/core';
+import { ErrorHandler, Provider, Type } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { Action, Store } from '@ngrx/store';
-import { Observable, of } from 'rxjs';
+import { Observable, of, throwError } from 'rxjs';
 import { catchError } from 'rxjs/operators';
 import { createEffect, EFFECTS_ERROR_HANDLER, EffectsModule } from '..';
+import * as effectsSrc from '../src/effects_error_handler';
 
 describe('Effects Error Handler', () => {
   let subscriptionCount: number;
   let globalErrorHandler: jasmine.Spy;
   let storeNext: jasmine.Spy;
 
-  function makeEffectTestBed(...providers: Provider[]) {
+  function makeEffectTestBed(effect: Type<any>, ...providers: Provider[]) {
     subscriptionCount = 0;
 
     TestBed.configureTestingModule({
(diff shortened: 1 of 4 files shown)


brandonroberts:
The changes LGTM. Is it correct that if you want to only change the number of retries, you still have to re-implement the default behavior plus that change?

alex-okrushko:
> Is it correct that if you want to only change the number of retries, you still have to re-implement the default behavior plus that change?

One can provide the custom effects error handler like this:
```ts
@NgModule({
  imports: [EffectsModule.forRoot([MovieEffects])],
  providers: [
    {
      provide: EFFECTS_ERROR_HANDLER,
      useValue: customEffectResubscriptionHandler,
    },
    
  ],
})

// adjust the number of retries
const CUSTOM_NUMBER_OF_RETRIES = 100; 

export function customEffectResubscriptionHandler<T extends Action>(
  observable$: Observable<T>,
  errorHandler: CustomErrorHandler
): Observable<T> {
  return defaultEffectsErrorHandler(observable$, errorHandler, CUSTOM_NUMBER_OF_RETRIES);
}
```

Thank means that there's no need to reimplement the behavior if one wants to increase the retries number.

timdeschryver:
I find that the above snippet can be added to the docs.
If you agree, I can open up a PR for it 🙂

alex-okrushko:
Good idea @timdeschryver

Issue #1641, opened 2019-03-19
[entity] updateOne, updateAll and map methods on EntityStore clear entity prototypes


mtaran-google:
## Minimal reproduction of the bug/regression with instructions:
1. Put class instances as entities inside an ngrx/entity store
2. Try to use the above EntityAdapter operations on them
3. The operations work, but the objects turn into plain objects (prototype is just Object)

## Expected behavior:
The prototype chain on the entities is preserved.

## Versions of NgRx, Angular, Node, affected browser(s) and operating system(s):
all (pretty sure)

## I would be willing to submit a PR to fix this issue

Maybe, if nobody else is interested :o

Seems like you'd need to do a get/setPrototypeOf near https://github.com/ngrx/platform/blob/master/modules/entity/src/unsorted_state_adapter.ts#L100 & the sorted equivalent

brandonroberts:
@mtaran-google this is intentional. More context is provided here https://github.com/ngrx/platform/issues/1376#issuecomment-431691078

mtaran-google:
I was aware that it's a best practice to use plain objects with ngrx/redux, but also that most things generally work fine/as expected if you don't.

Is this constraint documented somewhere on https://ngrx.io/docs? If not, this bug can be refocused on just spelling out that this isn't a well-supported use case.

brandonroberts:
It is documented here currently https://ngrx.io/guide/entity#entity-and-class-instances

We can add more if it's not explicit enough

mtaran-google:
I see. That is detailed, though it's not actually accurate. You only run into problems when you use the three helpers listed -- otherwise the objects are kept as they are. In fact if other helpers started messing with our objects, that would be a significant breaking change and our project would have a hard time dealing with it, and I suspect the same is true for others. It would be nice if this was more explicit.

alex-okrushko:
@mtaran-google Yes, we ran into this issue as well (due to the generated classes). 
I think we could provide custom clone function (instead of `Object.assign({}, original, update.changes)` which looses getters and setters)  that would solve the problem.

The current solution is to serialize the class before passing it to those entity methods.

mtaran-google:
That would be nice! For now we're just avoiding those methods, which is not too terrible, but still not the best.

alex-okrushko:
@mtaran-google  I sent you the link with the serialization workaround - not sure if you saw it before it was deleted. Ping me if you missed it :)

ankemp:
[...]

PR #2397, opened 2020-02-19
fix(entity): support primitive type entities

antoinebrault:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

It is not currently possible to update a list of string values.
Object.assign({}, ...) transforms a string value (ex: Book) into an array['B', 'o', 'o', 'k'] on update

My use case: backend has multiple string enums that might change between deployments or context. I really didn't want to hack the response to transform those strings into objects with ids.

## What is the new behavior?
check if the update value is a string, and return value as is.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/entity/spec/sorted_state_adapter.spec.ts
+++ b/modules/entity/spec/sorted_state_adapter.spec.ts
@@ -1,4 +1,4 @@
-import { EntityStateAdapter, EntityState, Update } from '../src/models';
+import { EntityStateAdapter, EntityState } from '../src/models';
 import { createEntityAdapter } from '../src/create_adapter';
 import {
   BookModel,
@@ -6,6 +6,7 @@ import {
   AClockworkOrange,
   AnimalFarm,
 } from './fixtures/book';
+import { BookType, MovieType, ShowType, TypeModel } from './fixtures/type';
 
 describe('Sorted State Adapter', () => {
   let adapter: EntityStateAdapter<BookModel>;
@@ -89,7 +90,7 @@ describe('Sorted State Adapter', () => {
   it('should remove existing and add new ones on addAll (deprecated)', () => {
     const withOneEntity = adapter.addOne(TheGreatGatsby, state);
 
-    const withAll = adapter.setAll(
+    const withAll = adapter.addAll(
       [AClockworkOrange, AnimalFarm],
       withOneEntity
     );
@@ -337,8 +338,8 @@ describe('Sorted State Adapter', () => {
         book.title === TheGreatGatsby.title
           ? firstChange
           : book.title === AClockworkOrange.title
-            ? secondChange
-            : book,
+          ? secondChange
+          : book,
       withMany
     );
 
@@ -404,6 +405,223 @@ describe('Sorted State Adapter', () => {
           ...firstChange,
         },
         [AClockworkOrange.id]: AClockworkOrange,
+      },
+    });
+  });
+});
+
+describe('Sorted State Adapter with string entities', () => {
+  let adapter: EntityStateAdapter<TypeModel>;
+  let state: EntityState<TypeModel>;
+
+  beforeEach(() => {
+    adapter = createEntityAdapter({
+      selectId: (type: TypeModel) => type,
+      sortComparer: (a, b) => a.localeCompare(b),
+    });
+
+    state = { ids: [], entities: {} };
+  });
+
+  it('should let you add one entity to the state', () => {
+    const withOneEntity = adapter.addOne(BookType, state);
+
+    expect(withOneEntity).toEqual({
+      ids: [BookType],
+      entities: {
+        [BookType]: BookType,
+      },
+    });
+  });
+
+  it('should not change state if you attempt to re-add an entity', () => {
+    const withOneEntity = adapter.addOne(BookType, state);
+
+    const readded = adapter.addOne(BookType, withOneEntity);
+
+    expect(readded).toBe(withOneEntity);
+  });
+
+  it('should let you add many entities to the state', () => {
+    const withOneEntity = adapter.addOne(BookType, state);
+
+    const withManyMore = adapter.addMany([MovieType, ShowType], withOneEntity);
+
+    expect(withManyMore).toEqual({
+      ids: [BookType, MovieType, ShowType],
+      entities: {
+        [BookType]: BookType,
+        [MovieType]: MovieType,
+        [ShowType]: ShowType,
+      },
+    });
+  });
+
+  it('should remove existing and add new ones on setAll', () => {
+    const withOneEntity = adapter.addOne(BookType, state);
+
+    const withAll = adapter.setAll([MovieType, ShowType], withOneEntity);
+
+    expect(withAll).toEqual({
+      ids: [MovieType, ShowType],
+      entities: {
+        [MovieType]: MovieType,
+        [ShowType]: ShowType,
+      },
+    });
+  });
+
+  it('should remove existing and add new ones on addAll (deprecated)', () => {
+    const withOneEntity = adapter.addOne(BookType, state);
+
+    const withAll = adapter.addAll([MovieType, ShowType], withOneEntity);
+
+    expect(withAll).toEqual({
+      ids: [MovieType, ShowType],
+      entities: {
+        [MovieType]: MovieType,
+        [ShowType]: ShowType,
+      },
+    });
+  });
+
+  it('should let you add remove an entity from the state', () => {
+    adapter.addOne(BookType, state);
+
+    const withoutOne = adapter.removeOne(BookType, state);
+
+    expect(withoutOne).toEqual({
+      ids: [],
+      entities: {},
+    });
+  });
+
+  it('should let you remove many entities by id from the state', () => {
+    const withAll = adapter.setAll([BookType, MovieType, ShowType], state);
+
+    const withoutMany = adapter.removeMany([BookType, MovieType], withAll);
+
+    expect(withoutMany).toEqual({
+      ids: [ShowType],
+      entities: {
+        [ShowType]: ShowType,
+      },
+    });
+  });
+
+  it('should let you remove many entities by a predicate from the state', () => {
+    const withAll = adapter.setAll([BookType, MovieType, ShowType], state);
+
+    const withoutMany = adapter.removeMany(
+      p => p === BookType || p === MovieType,
+      withAll
+    );
+
+    expect(withoutMany).toEqual({
+      ids: [ShowType],
+      entities: {
+        [ShowType]: ShowType,
+      },
+    });
+  });
+
+  it('should let you remove all entities from the state', () => {
+    const withAll = adapter.setAll([BookType, MovieType, ShowType], state);
+
+    const withoutAll = adapter.removeAll(withAll);
+
+    expect(withoutAll).toEqual({
+      ids: [],
+      entities: {},
+    });
+  });
+
+  it('should let you update an entity in the state', () => {
+    const withOne = adapter.addOne(BookType, state);
+
+    const withUpdates = adapter.updateOne(
+      {
+        id: BookType,
+        changes: BookType,
+      },
+      withOne
+    );
+
+    expect(withUpdates).toEqual({
+      ids: [BookType],
+      entities: {
+        [BookType]: BookType,
+      },
+    });
+  });
+
+  it('should not change state if you attempt to update an entity that has not been added', () => {
+    const withUpdates = adapter.updateOne(
+      {
+        id: BookType,
+        changes: BookType,
+      },
+      state
+    );
+
+    expect(withUpdates).toBe(state);
+  });
+
+  it('should not change ids state if you attempt to update an entity that does not impact sorting', () => {
+    const withAll = adapter.setAll([BookType, MovieType, ShowType], state);
+
+    const withUpdates = adapter.updateOne(
+      {
+        id: BookType,
+        changes: BookType,
+      },
+      withAll
+    );
+
+    expect(withAll.ids).toBe(withUpdates.ids);
+  });
+
+  it('should let you update the id of entity', () => {
+    const withOne = adapter.addOne(BookType, state);
+
+    const withUpdates = adapter.updateOne(
+      {
+        id: BookType,
+        changes: BookType,
+      },
+      withOne
+    );
+
+    expect(withUpdates).toEqual({
+      ids: [BookType],
+      entities: {
+        [BookType]: BookType,
+      },
+    });
+  });
+
+  it('should let you add one entity to the state with upsert()', () => {
+    const withOneEntity = adapter.upsertOne(BookType, state);
+
+    expect(withOneEntity).toEqual({
+      ids: [BookType],
+      entities: {
+        [BookType]: BookType,
+      },
+    });
+  });
+
+  it('should let you upsert many entities in the state', () => {
+    const withMany = adapter.setAll([BookType], state);
+
+    const withUpserts = adapter.upsertMany([MovieType, ShowType], withMany);
+
+    expect(withUpserts).toEqual({
+      ids: [BookType, MovieType, ShowType],
+      entities: {
+        [BookType]: BookType,
+        [MovieType]: MovieType,
+        [ShowType]: ShowType,
       },
     });
   });
--- a/modules/entity/spec/unsorted_state_adapter.spec.ts
+++ b/modules/entity/spec/unsorted_state_adapter.spec.ts
@@ -6,6 +6,7 @@ import {
   AClockworkOrange,
   AnimalFarm,
 } from './fixtures/book';
+import { BookType, MovieType, ShowType, TypeModel } from './fixtures/type';
 
 describe('Unsorted State Adapter', () => {
   let adapter: EntityStateAdapter<BookModel>;
@@ -277,8 +278,8 @@ describe('Unsorted State Adapter', () => {
         book.title === TheGreatGatsby.title
           ? firstChange
           : book.title === AClockworkOrange.title
-            ? secondChange
-            : book,
+          ? secondChange
+          : book,
       withMany
     );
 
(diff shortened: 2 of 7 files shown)


brandonroberts:
Why not flip it so if it's an object, do the assignment, otherwise use the changes?

timdeschryver:
I'm not a big fan of support primitives, but the change seems small.
I'll take a look later today.

modules/entity/spec/sorted_state_adapter.spec.ts
```diff
@@ -89,7 +90,7 @@ describe('Sorted State Adapter', () => {
   it('should remove existing and add new ones on addAll (deprecated)', () => {
     const withOneEntity = adapter.addOne(TheGreatGatsby, state);
 
-    const withAll = adapter.setAll(
+    const withAll = adapter.addAll(
```
alex-okrushko: wrong rebase? `addAll` is deprecated in favor of `setAll` (this also applies for the rest of the tests).

modules/entity/spec/unsorted_state_adapter.spec.ts
```diff
@@ -348,3 +349,198 @@ describe('Unsorted State Adapter', () => {
...
+
+  it('should let you add many entities to the state', () => {
+    const withOneEntity = adapter.addOne(MovieType, state);
+
+    const withManyMore = adapter.addMany([BookType, ShowType], withOneEntity);
+
+    expect(withManyMore).toEqual({
+      ids: [MovieType, BookType, ShowType],
+      entities: {
+        [MovieType]: MovieType,
```
alex-okrushko: What's the purpose of it? From what I see you'd need to pass something, to get the same thing.

modules/entity/src/utils.ts
```diff
@@ -17,3 +17,11 @@ export function selectIdValue<T>(entity: T, selectId: IdSelector<T>) {
 
   return key;
 }
+
+export function isObjectLike(target: any): target is object {
+  return typeof target === 'object' && target !== null;
+}
+
+export function isObject(target: any): target is object {
```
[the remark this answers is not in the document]
antoinebrault: it's the same function that is already in /store. I wasn't sure if you guys want to cross import functions between projects.
https://github.com/ngrx/platform/blob/master/modules/store/src/meta-reducers/utils.ts

modules/entity/spec/sorted_state_adapter.spec.ts
antoinebrault: look at the test title. It was supposed to test addAll() not setAll(). There is already another test on setAll()

modules/entity/spec/unsorted_state_adapter.spec.ts
antoinebrault: testing if the addMany() works with primitives objects, or doesn't break in the future.

modules/entity/src/utils.ts
```diff
@@ -17,3 +17,11 @@ export function selectIdValue<T>(entity: T, selectId: IdSelector<T>) {
 
   return key;
 }
+
+export function isObjectLike(target: any): target is object {
```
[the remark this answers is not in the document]
antoinebrault: ok, I can remove the export

antoinebrault:
> I don't understand the use case. Could you please give the example (stackblitz?) of what's broken?

I'm using ngrx-data on all entitites in my projet. A couple of them are string entities.

**/orders** = list of orders { id: number, state: string, ... } 
**/orderStates** = list of string states

**orderStates** could be updated between backend deployments.

If you revert the line of code that I updated and keep the tests, you can see that updateOne(), upsertMany() fails because the code assumes that the entity is an object.

I could provide a stackblitz, but it will show you exactly what the tests are showing (that the string entity is transformed into an object because of Object.assign())


>6) Sorted State Adapter with string entities should let you update an entity in the state
>  Message:
>   Expected $.ids[0] = Object({ 0: 'B', 1: 'O', 2: 'O', 3: 'K' }) to equal 'BOOK'.
>   Expected $.entities to have properties
>      BOOK: 'BOOK'
>  Expected $.entities not to have properties
>     [object Object]: Object({ 0: 'B', 1: 'O', 2: 'O', 3: 'K' })

antoinebrault:
There you go https://stackblitz.com/edit/ngrx-data-primitive-entities

modules/entity/spec/sorted_state_adapter.spec.ts
alex-okrushko: Ah, I see. Yeah, it's not necessary to test them both, just use `setAll`. The `setAll` is tested separately to make sure it's the same as `addAll`.

alex-okrushko:
> /orderStates = list of string states

ngrx/entity is meant to work with collections of entities. e.g. when you can read all of them as an array or get one by one through Dictionary. Array of primitives doesn't make much sense to store as EntityState structure
e.g.
```ts
{
  ids: ['item1', 'item1', 'item1'],
  entities: {
    'item1': 'item1',
    'item2': 'item2',
    'item3': 'item3',
  }
}
```
vs
```ts
{
  items: ['item1', 'item1', 'item1'],
}
```

You'd never be accessing the entity structure by index.

antoinebrault:
> > /orderStates = list of string states
> 
> ngrx/entity is meant to work with collections of entities. e.g. when you can read all of them as an array or get one by one through Dictionary. Array of primitives doesn't make much sense to store as EntityState structure
> e.g.
> 
> ```ts
> {
>   ids: ['item1', 'item1', 'item1'],
>   entities: {
>     'item1': 'item1',
>     'item2': 'item2',
>     'item3': 'item3',
>   }
> }
> ```
> 
> vs
> 
> ```ts
> {
>   items: ['item1', 'item1', 'item1'],
> }
> ```
> 
> You'd never be accessing the entity structure by index.

I know that, but it doesn't make sense to me to code all that by hand (state, reducers, etc. or service 
 + obs, data service) to support those endpoints. I just want to use ngrx-data for all endpoints. 
If that entity is changed in the future for an object, code will still work.  All my code is generic, doesn't matter if it's a string list or an object list.

It's a really small change to support that use case. (and btw, it works well with addAll. It just doesn't work once it's updated)

nasreddineskandrani:
image

this is totally wrong... 

@alex-okrushko 
I do understand your point about:
{
  items: ['item1', 'item1', 'item1'],
}
but since `ngrx-data` is broken for this usecase it's should be fixed

The way the dev use `ngrx-data` is decided by the team itself not the library creator.

alex-okrushko:
Thanks @nasreddineskandrani and @antoinebrault 

I understand that the behavior from 'ngrx' as a whole is broken. What I'm trying to say is that it's not `ngrx/entity` that is broken, but `ngrx/data` is using it incorrectly, so the fix should go there instead. Would you agree?

nasreddineskandrani:
I agree with you `ngrx-data` can probably handle this usecase of primitive to create a none entity state but the structure you propose.

In the other side as @antoinebrault stated:
> it works well with addAll. It just doesn't work once it's updated
I think he still has a good point and his fix is legit to merge because adding and updating don't give the same result today.

antoinebrault:
This PR was about fixing a broken behavior in the current code.

I think the point to have cleaner state is right, but should be a seperate issue. 

As for myself, I really don't mind if the state structure is not 100% as it should be. It works for my use case and saves me a lot of time/testing. I'm able to have generic code for all my entities and to update them dynamically with websockets.

alex-okrushko:
Since @timdeschryver and @brandonroberts didn't have strong objections, I'm going to just dismiss my review and let them decide if they want to merge it in.

I do understand the frustration when something is broken and one just wants to get it fixed.
> It works for my use case
This is something that I don't quite welcome. It works for one's case, which is great, but it could become harder to maintain and brings the ability to be misused (I'm not saying that this is the case).

We do have the issue open (https://github.com/ngrx/platform/issues/1641) that could address this one as well - basically if we has the function that we can pass to `createEntityAdapter` guiding how we "update" then it might solve the case for primitives as well.

timdeschryver:
I don't have a strong objection, but I don't see primitives as entities. 
ngrx/entity provides a clean api to manage entities efficiently, I think it isn't really needed for primitives.
 
The only reason why I'm not strongly against it, is that it's a small change.
On the other hand, it might limit us in the future.

AdditionAddict:
One of the things I like about ngrx has always been a strong ethos of what’s good practice and acceptable architecture. Similar to the zen of python - _There should be one-- and preferably only one --obvious way to do it._  Hence, I personally object to placing constraints on `@ngrx/entity` (not only that but allowing a pathway to bad state that should be discouraged) in order to make `@ngrx/data` work in one use case. I think this should be dealt with within `@ngrx/data`. I defer to better minds however

antoinebrault:
> Hence, I personally object to placing constraints on `@ngrx/entity`

It's literally the opposite of that. I'm not adding constraints.

> ngrx has always been a strong ethos of what’s good practice and acceptable architecture

If it was that strong, why are all generics in entity and data of **any** type. If it was supposed to be only objects, then it should enforce **T extends object** everywhere or throw an exception on primitives.

It's honestly frustrating. I shouldn't have wasted that much time for a 2 lines fix.
People talk about architecture when they haven't seen my code. I can guarantee you that my project with 100+ entities is gonna be a lot cleaner without resorting to hacks or boilerplate to support primitives, and easier to maintain.

wardbell:
>It is not currently possible to update a list of string values. `Object.assign({}, ...)` transforms a string value (ex: Book) into an array `['B', 'o', 'o', 'k']` on update.
>My use case: backend has multiple string enums that might change between deployments or context. 

I'm guessing that you mean that you have something like a collection called `Colors` with members such as `["red", "green", "blue"]`.

Fantastic. That's good data to keep in store. But let's be precise: those are **value types** not **entity types**.  There are all kinds of value types in this world. The set of all integers, for example. All of the primitive types in JS are value types. But NgRx Data is not intended for management of value types, persisted or not.

>since ngrx-data is broken for this use case

I strongly disagree that either NgRx Data or Entity is "broken" for this use case. Rather, I contend (with respect) that NgRx Data is not intended for that use case and, further, should not be "twisted" for non-entity use cases.

NgRx Data and Entity are dedicated to the management of _entity_ data.  An entity is "defined ... by a thread of continuity and identity."

See https://lostechies.com/jimmybogard/2008/05/21/entities-value-objects-aggregates-and-roots/

An entity's **primary key** is the representation of that  "thread of identity".  An **entity must have a primary key** that is unique within its collection.  It might be a multi-part key but it is still a primary key.

NgRx Data manages **collections of entities**. Not collections of just _any_ data;  collections of _entity data_ .  There is no ambiguity about that.

While we may want to store and retrieve all kinds of data that are value types, that's not what NgRx Data is for.  And a desire to manage other kinds of data easily is not a license to  warp Ngrx Data or Entity to a purpose for which it is specifically unsuited.

>It's a really small change to support that use case.

It may be a small change in code but it's a sea change in semantics. Much of NgRx Data is dedicated to maintaining entity identity. That's sacred to its purpose. We simply can't break that fundamental characteristic of NgRx Data itself.

>I really don't mind if the state structure is not 100% as it should be. It works for my use case and saves me a lot of time/testing. I'm able to have generic code for all my entities and to update them dynamically with websockets.

I am sympathetic to your desire to have a consistent and simple way to fetch and maintain data from a server. You are welcome to have your own _private_ interpretation of what we should do.

But I'm afraid I will have to insist that we will not join you in that interpretation. 

You also have simple workarounds to transform your enumeration into an entity collection, as described earlier in this thread. If you are enjoying what NgRx Data brings to your project, accommodating it with an occasional workaround is a small price, no?

antoinebrault:
> I strongly disagree that either NgRx Data or Entity is "broken" for this use case. Rather, I contend (with respect) that NgRx Data is not intended for that use case and, further, should not be "twisted" for non-entity use cases.

This is your interpretation of what my use case is. You know in the real world when you work with "average" backend developers, you don't control how the endpoints are built, what is returned. 

I have to deal with like 60+ endpoints ```{ id: number, name: string }[]``` and around 10 ```string[]```. They are managed, fetched, updated the same way, and used the same way in data tables and dynamic forms. You're gonna tell me that because "theoretically" it shouldn't be entities, I should manage them differently.

> You also have simple workarounds to transform your enumeration into an entity collection, as described earlier in this thread.

In a TODO list app maybe, in a real world app with 30+ entities using those other 70 "value" entities (think of an object with 10-15 "value" fields), it's not possible to "simply" transform data. It's gonna result in a big unmaintainable mess.

> It may be a small change in code but it's a sea change in semantics. Much of NgRx Data is dedicated to maintaining entity identity. That's sacred to its purpose. We simply can't break that fundamental characteristic of NgRx Data itself.

Then you should enforce ```T extends object``` everywhere. Because how it stands, your sacred 
principle is not clear at all in your api and your documentation.

Forgive my frustration, I'm probably gonna waste another week or 2 to do a major refactoring.

wardbell:
I am sympathetic to your overall challenge. And I even acknowledge that the name of the library ("Data") implies a much broader interpretation than just "Entity Data". 

>We had a lively discussion about the name before sticking with this one.

But it is what it is and this is not a merely "theoretical" distinction that I'm making. It is fundamental ... as fundamental as the "immutability" principle in NgRx itself.  These principles can be inconvenient. But when so much of the guidance and implementation depends upon these principles, relaxing them can be catastrophic.

So let's find another way to meet your need.

I can think of several rather easy approaches to dealing with your 10 cases.  My recommendation would depend upon how you use them.  A few questions:

* Are they read-only from the client perspective? My guess is "Yes"; that they do not change during the lifetime of a given user session.

* How do you consume these values within the app. Are they like enumerations? For example, do you have something like an array of `StatusCodes` that you're populating from server data and then using this "enumeration" to constrain the status values of some other entity types?

* Do you read them all at once (say at application start) or as needed?

* Is there anything distinctive about the endpoints for these types? The answer is not crucial but if there were, it makes an implementation either ultra-easy or merely easy.

I assure you this is not a big deal.

nasreddineskandrani:
This command `this.statesService.upsertManyInCache(['INITIAL', 'NEW', 'COMPLETED']);`
Endup with: 'I', 'N', 'I', 'T', 'A', 'L'... in the store
so **it is broken**. If you want keep saying it's not :) it changes nothing about the fact it is... You can post a poll on twitter. We will see :) what more developers think about it.

Now you are free to keep the broken behavior and explain it in the documentation that it's not meant for primitives. It is ok.
But you can probably take it to another level to help developers by handling it by either improving the typing to fail at compile time or exception at run time.

~~~
Now about `ngrx-entity` meant to be for `object` that has IDs... you are focusing so much on what is meant to be and forget that a dev had a challenge => you close the PR and just told him to do conversion... bad bad bad... 

you need something to handle primitives if `ngrx-data` is meant to be `ngrx-data-entity` 
since it is a real use case. You can have `ngrx-primitives` and that's it... the guy don't have to complexify his code x 10 to handle this after that.

wardbell:
You say
>This command this.statesService.upsertManyInCache(['INITIAL', 'NEW', 'COMPLETED']);
Endup with: 'I', 'N', 'I', 'T', 'A', 'L'... in the store

I don't see how that is possible. I tried to reproduce it. I imagined that you defined your `StatesService` to be of type `string` something like this
```
import { EntityCollectionServiceBase, EntityCollectionServiceElementsFactory} from '@ngrx/data';

@Injectable({providedIn: 'root'})
export class StatesService  extends EntityCollectionServiceBase<string>{
  constructor(factory: EntityCollectionServiceElementsFactory) {
    super('State', factory);
  }
}
```
Later I call `this.statesService.upsertManyInCache(['INITIAL', 'NEW', 'COMPLETED']);`

I do NOT endup with: 'I', 'N', 'I', 'T', 'A', 'L'... in the store. 

I get the following error message.
```
Error: State EntityAction guard for "[State] @ngrx/data/upsert-many": payload , item 1, does not have a valid entity key (id)
```
That is exactly the error I would expect.

You must have done something to define a primary key in an unexpected way that you're not telling us.

In any case, you're well off the intended path and already working hard to make NgRx Data do what it is not supposed to do. 

If you keep insisting that I change the semantics, this discussion is over.

If you want to consider how you might achieve your intention in some other way without to much effort, I'm interested. Your call.

nasreddineskandrani:
the author of the issue posted an example about the problem:
https://github.com/ngrx/platform/pull/2397#issuecomment-589992772
I don't see any error..

wardbell:
Ah yes. There it is as I suspected, a creative definition of the function that determines the primary key
```
export function selectObj(o: any) {
  return o;
}

const entityMetadata: EntityMetadataMap = {
  states: { selectId: selectObj }
};
```
So the object is the primary key of itself.  Nothing in our documentation suggested that this would be a valid primary key. It adheres to no notion of "primary key" known to me.

I am out of patience. Come back when you want to solve the problem without changing the semantics..

nasreddineskandrani:
Let's recap.
Basically 3 years of devs saying boilerplate everywhere... mike was joking brandon about it in many conf. 
You come with `ngrx-data` that john papa in this video say it's godlike move about boilerplate: https://www.youtube.com/watch?v=dHptnyroFNA no boilerplate...

And now you tell me you created `ngrx-data` for entities only to save the boilerplate but those that deal with array of primitives (string, number, etc...) we don't care about them => they need to craft a reducer, effect etc... to handle it... so boilerplate again.

I love all what ngrx team does but answering without checking the example and closing the PR.
It's too much for me... i am out.

AdditionAddict:
@nasreddineskandrani 

> Meanwhile, let's find another way to meet your need. You won't be the only one with this challenge. Maybe we can find something satisfactory and teach it.

☝️ This is the reaction from a busy, experienced, but patience person providing free software.

> And now you tell me you created `ngrx-data` for entities only to save the boilerplate but those that deal with array of primitives (string, number, etc...) **we don't care about them** => they need to craft a reducer, effect to handle it... OR convert to object etc...

Your reaction. ¯\\_(ツ)_/¯

Whilst ngrx-data doesn't support primitives the code base offers a roadmap as to abstract the usual round trip for value types using factories and facades:

component → action → effect → action → reducer → selector → component

Roadmap

- Define `ValueTypeOp` of operations to support
- Create a `ValueTypeActionFactory` that creates actions with these ops
- Create a `ValueTypeDispatcher` that creates and returns actions against a CRUD api
- Create a `ValueTypeService` a facade over the dispatcher and include selectors
- Create a `ValueTypeEffects` which calls service and maps to success / fail actions
- Create a `ValueTypeCollectionReducerMethodsFactory` + `ValueTypeCollectionReducerMethods` to implement reducers
- Create a `ValueTypeSelectorsFactory`

image

This would be separate to ngrx-data. If you look for example at https://github.com/ngrx/platform/blob/master/modules/data/src/reducers/entity-change-tracker-base.ts you would see how much effort goes into the fact the objects stored are in fact entities which can undergo change but with a thread of identity.

PR #2410, opened 2020-02-24
feat(entity): add 'setOne' method

### LINKED ISSUE #2369: @ngrx/entity add setOne (replaceOne?) method
Currently, the only way to replace an item is by using upsertOne. That does not work well with 'undefined' though (which is the case when the backend omits nulls when serializing to JSON). For example, i have an entity in the collection: {id: 1, name: 'foo', note: 'lorem ipsum'}. I want to replace it with {id: 1, name: 'foo'}. Obviously, upsertOne does not do the job well. Since the object  is already there, it attempts to perform an update, thus treating the new object as a patch and leaving 'note' value unchanged. So, a new method 'setOne' is needed here.

removeOne/addOne does the trick, but is not very efficient.

[X ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

dmytro-gokun:
## PR Checklist

## PR Type

```
[] Bugfix
[X] Feature
[] Code style update (formatting, local variables)
[] Refactoring (no functional changes, no api changes)
[] Build related changes
[] CI related changes
[] Documentation content changes
[] Other... Please describe:
```

Closes #2369

```
[] Yes
[X] No
```

### DIFF
--- a/docs/entity/adapter.md
+++ b/docs/entity/adapter.md
@@ -87,7 +87,6 @@ state if no changes were made.
 - `addMany`: Add multiple entities to the collection
 - ~~`addAll`~~: (Deprecated and renamed to `setAll`). ~~Replace current collection with provided collection~~
 - `setAll`: Replace current collection with provided collection
-- `setOne`: Add or Replace on entity in the collection
 - `removeOne`: Remove one entity from the collection
 - `removeMany`: Remove multiple entities from the collection
 - `removeAll`: Clear entity collection
@@ -118,7 +117,6 @@ import { User } from './user.model';
 export enum UserActionTypes {
   LOAD_USERS = '[User] Load Users',
   ADD_USER = '[User] Add User',
-  SET_USER = '[User] Set User',
   UPSERT_USER = '[User] Upsert User',
   ADD_USERS = '[User] Add Users',
   UPSERT_USERS = '[User] Upsert Users',
@@ -137,12 +135,6 @@ export class LoadUsers implements Action {
 
 export class AddUser implements Action {
   readonly type = UserActionTypes.ADD_USER;
-
-  constructor(public payload: { user: User }) {}
-}
-
-export class SetUser implements Action {
-  readonly type = UserActionTypes.SET_USER;
 
   constructor(public payload: { user: User }) {}
 }
@@ -196,7 +188,6 @@ export class ClearUsers implements Action {
 export type UserActionsUnion =
   | LoadUsers
   | AddUser
-  | SetUser
   | UpsertUser
   | AddUsers
   | UpsertUsers
@@ -230,10 +221,6 @@ export function reducer(state = initialState, action: UserActionsUnion): State {
   switch (action.type) {
     case UserActionTypes.ADD_USER: {
       return adapter.addOne(action.payload.user, state);
-    }
-
-    case UserActionTypes.SET_USER: {
-      return adapter.setOne(action.payload.user, state);
     }
 
     case UserActionTypes.UPSERT_USER: {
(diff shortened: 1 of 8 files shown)


modules/entity/spec/sorted_state_adapter.spec.ts
```diff
@@ -407,4 +407,36 @@ describe('Sorted State Adapter', () => {
...
+
+  it('should let you replace an entity in the state with setOne()', () => {
+    const withOne = adapter.addOne(
+      TheGreatGatsby,
+      adapter.addOne(AnimalFarm, adapter.addOne(AClockworkOrange, state))
+    );
+    const updatedBook = {
+      id: TheGreatGatsby.id,
+      title: 'A New Hope',
+      description: undefined,
```
alex-okrushko: Could you please remove the description altogether? With `description: undefined` even `upsertOne` would behave the same way.

dmytro-gokun: I can remove it, that is not a problem. But that is the only way to test if 'setOne' actually behaves in a way different from 'upsertOne'. I mean, without this 'undefined' the test does not make much sense. Does it?

modules/entity/spec/sorted_state_adapter.spec.ts
```diff
@@ -407,4 +407,36 @@ describe('Sorted State Adapter', () => {
...
+      entities: {
+        [TheGreatGatsby.id]: TheGreatGatsby,
+      },
+    });
+  });
+
+  it('should let you replace an entity in the state with setOne()', () => {
+    const withOne = adapter.addOne(
+      TheGreatGatsby,
+      adapter.addOne(AnimalFarm, adapter.addOne(AClockworkOrange, state))
```
timdeschryver:
```suggestion
      state
```

docs/entity/adapter.md
```diff
@@ -87,6 +87,7 @@ state if no changes were made.
 - `addMany`: Add multiple entities to the collection
 - ~~`addAll`~~: (Deprecated and renamed to `setAll`). ~~Replace current collection with provided collection~~
 - `setAll`: Replace current collection with provided collection
+- `setOne`: Add or Replace on entity in the collection
```
timdeschryver: Could you move these changes to https://github.com/ngrx/platform/blob/master/projects/ngrx.io/content/guide/entity/adapter.md so they become visible on the website 🙂?
These docs are for older versions of NgRx

modules/entity/spec/sorted_state_adapter.spec.ts
dmytro-gokun: I'm not sure if I understand this. The reason I'm adding three books to the collection here is to test that sorted adapter preserves the correct sorting order after a 'setOne' call. If I only have one book in the collection that is not checked anymore. Do you think that is okay?

docs/entity/adapter.md
dmytro-gokun: Done

modules/entity/spec/sorted_state_adapter.spec.ts
timdeschryver: Oh yea, that makes sense! Perhaps rename the `withOne` variable then tho

dmytro-gokun: Oops. That's a copy&paste error :(. Should be fixed now.

modules/entity/spec/sorted_state_adapter.spec.ts
alex-okrushko: It actually does make more sense without that property. You can make a comment to be very explicit that we are updating the the entity that doesn't have that property and `toEqual` does the deep equality check.
With the test as-is even `adapter.upsertOne(updatedBook, withOne)` would pass.

I see that this PR is merge, I'd suggest to have a follow-up PR to adjust this test behavoir.

dmytro-gokun: Okay, we might have a misunderstanding here... Do you think we need to:
1) remove line 429;
or
2) remove 'description' property from the Book interface altogether?

alex-okrushko: remove the line from 429 and replace with `/* description property is not provided */`

PR #2421, opened 2020-03-02
fix(entity): fix a 'setOne' test case

dmytro-gokun:
This is a small fix suggested by @alex-okrushko here: https://github.com/ngrx/platform/pull/2410#discussion_r386036013


alex-okrushko:
Thanks for the follow-up PR 👍

PR #2457, opened 2020-03-28
test(data): refactor testing setup

BioPhoton:
@wardbell 

I noticed the scheduler in the data package.

If it is just for testing I suggest remove the scheduler in the code and 
use e.g. rxjs-marbles.
IMHO scheduling can be changed without providers in the tests.

Here a quick example. If you take a look at the tests I could help in refactoring. :)


brandonroberts:
We don't need to refactor these tests at the moment, so let's revisit this later.

BioPhoton:
Perfect!

Issue #2458, opened 2020-03-28
@ngrx/component implement strategies 


BioPhoton:
**UPDATE the coalescing stuff is irrelevant**

# Problem

In the alpha release, the optimization should be opt-in. 
Also, the implementation needs a way to change the strategy in `CdAware`.

**Actual**
```typescript
export function createCdAware<U>(cfg: {
  work: () => void;
  resetContextObserver: NextObserver<unknown>;
  configurableBehaviour: (
    o: Observable<Observable<U | null | undefined>>
  ) => Observable<Observable<U | null | undefined>>;
  updateViewContextObserver: PartialObserver<U | null | undefined>;
}): CdAware<U | undefined | null>
``` 

**Usage**

```typescript
constructor(cdRef: ChangeDetectorRef, ngZone: NgZone) {
    this.cdAware = createCdAware<S>({
      work: setUpWork({
        ngZone,
        cdRef,
        context: (cdRef as EmbeddedViewRef<Type<any>>).context,
      }),
      updateViewContextObserver: this.updateViewContextObserver,
      resetContextObserver: this.resetContextObserver,
      configurableBehaviour: this.configurableBehaviour,
    });
    this.subscription = this.cdAware.subscribe();
  }
```

The problem here is `work` and `configurableBehaviour` depend on each other, the compositin of work is done outside they rely on the environment.
`resetContextObserver` and `updateViewContextObserver` is related to the class it is used in.

Based on this implementation configurable strategies are not possible.

# Solution

**Strategy Utils**

```typescript
export const DEFAULT_STRATEGY_NAME = 'idle';

export interface CdStrategy<T> {
    behaviour: (cfg?: any) => MonoTypeOperatorFunction<Observable<T>>;
    render: () => void;
    name: string;
}

export interface StrategySelection<U> {
    idle: CdStrategy<U>;
    [key: string]: CdStrategy<U>;
}

export function getStrategies<T>(cfg: StrategyFactoryConfig): StrategySelection<T> {
    return {
        idle: createIdleStrategy<T>(cfg)
    };
}
```

**Idle Strategy** 

 This strategy is the drop-in replacement for Angular's built-in `async` pipe.
 This is the only strategy that **does not also work in zone-less environments**.

 - \>=ViewEngine  

 |          | Render Method      | Coalescing | Coalesce Scope |
 | -------- | -------------------| ---------- | -------------- |
 | ZoneFull | cdRef.markForCheck | ❌         | None           |
[...]

PR #2456, opened 2020-03-28
feat(component): add coalesce and generateFrames operators

BioPhoton:
implement the operators, added tests, added docs, added diagrams, added rxjs-marbles to dev dependencies.

**Demonstrations for the need of these operators:**
- ~[generateFrames](https://stackblitz.com/edit/ng-zone-af-tests)~
- [coalesce](https://stackblitz.com/edit/ng-af-coalescing-tests)

closes ~#2446~ #2448 related to #2441

PR #2473, opened 2020-04-04
feat(component): cd aware and strategies 2443 2458 

### LINKED ISSUE #2443: @ngrc/component Refactor cdAware factory
Make sure it is aligned with the current way of creating factories. 
It would be good to consult @alex-okrushko for detailed requirements.

**Todo**

-  [ ] Refactored out things related to ngZone or optimizations in Ivy 
   - [ ] Change pipe config argument to a string
  - [ ] Adopt pipe and let
- [ ] Updte tests
  - [ ] Remove ivy and zone related tests

Related: #2441

[ ] No

BioPhoton:
Implement better architecture, Implement strategies. Details here: #2458

Closes #2443, #2458


BioPhoton:
@alex-okrushko 

The new architecture is ready for review.  (Tests and docs follow you agree on the implementation)

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Also custom behaviour is something you need to implement in the extending class
+ */
+export function createCdAware<U>(cfg: {
+  strategies: StrategySelection<U>;
+  resetContextObserver: NextObserver<any>;
+  updateViewContextObserver: PartialObserver<U> & NextObserver<U>;
+}): CdAware<U | undefined | null> {
+  const configSubject = new Subject<string>();
+  const config$: Observable<CdStrategy<U>> = configSubject.pipe(
+    distinctUntilChanged(),
```
benlesh, line 47: May want to move this after `startWith` to ensure that it doesn't fire the same thing twice if the first value from `configSubject` happens to be the same as `startWith`.

Also consider using a `BehaviorSubject` instead of `startWith`? Depends on your goal. It seems like you might end up with a situation where some code is changed or adjusted to subscribe in a more lazy way, and `renderSideEffect$` could be subscribed to *after* the initial value is emitted from here, and it may be confusing to debug.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -0,0 +1,98 @@
...
+          ? cfg.strategies[strategy]
+          : cfg.strategies.idle
+    )
+  );
+
+  const observablesSubject = new Subject<Observable<U>>();
+  const observables$$ = observablesSubject.pipe(distinctUntilChanged());
+
+  let prevObservable: Observable<U>;
+  const renderSideEffect$ = combineLatest([observables$$, config$]).pipe(
```
benlesh, line 61: I would definitely add a lot of comments about when, where, and _why_ this observable exists and is subscribed to.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -0,0 +1,98 @@
...
+    map(
+      (strategy: string): CdStrategy<U> =>
+        cfg.strategies[strategy]
+          ? cfg.strategies[strategy]
+          : cfg.strategies.idle
+    )
+  );
+
+  const observablesSubject = new Subject<Observable<U>>();
+  const observables$$ = observablesSubject.pipe(distinctUntilChanged());
```
benlesh, line 58: names are hard... perhaps change this to `observablesFrom?????$` so people get the idea where these are coming from? Also add more comments.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  let prevObservable: Observable<U>;
+  const renderSideEffect$ = combineLatest([observables$$, config$]).pipe(
+    switchMap(([observable$, strategy]) => {
+      if (prevObservable === observable$) {
```
benlesh, line 63: The `distinctUntilChanged` should take care of this. This should never happen.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -0,0 +1,98 @@
...
+        return NEVER;
+      }
+
+      if (observable$ == null) {
+        cfg.updateViewContextObserver.next(observable$);
+        strategy.render();
+        return NEVER;
```
benlesh, line 70: Favor `EMPTY` instead. The `swtichMap` cannot complete if the outer observable is still active, so the net result is the same, with the exception that it will clean up resources earlier with `EMPTY`.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -0,0 +1,98 @@
...
+      prevObservable = observable$;
+      cfg.resetContextObserver.next(observable$);
+      strategy.render();
+
+      return observable$.pipe(
+        distinctUntilChanged(),
+        tap(cfg.updateViewContextObserver),
+        strategy.behaviour(),
+        tap(() => strategy.render()),
+        catchError(e => EMPTY)
```
benlesh, line 82: Recommend at least logging here before returning EMPTY.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -0,0 +1,98 @@
...
+      );
+    })
+  );
+
+  return {
```
benlesh, line 87: The curious thing about this API, it it could just be a `Subject<{ type: 'value', value: Observable<any> }| { type: 'config', value: config }>` or the like. I'd experiment with that and see if it ended up being cleaner or more useful.

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,330 @@
...
+  ChangeDetectorRef,
+  NgZone,
+  ɵdetectChanges as detectChanges,
+  ɵmarkDirty as markDirty,
+} from '@angular/core';
+import { hasZone, isIvy } from '../utils';
+import { mapTo } from 'rxjs/operators';
+
+function getSaveDurationSelector(ngZone: NgZone): () => Observable<number> {
+  return () => defer(() => from(Promise.resolve()).pipe(mapTo(1)));
```
benlesh, line 15: You can save on promise allocations here... but it's mildly tricky, because we have to wait for Zones to patch the Promise.... so change this like so:

```ts
/** A shared promise instance to cause a delay of one microtask */
let resolvedPromise: Promise<void>|null = null;

function getResolvedPromise() {
  resolvedPromise = resolvedPromise || Promise.resolve();
  return resolvedPromise;
}
function getSaveDurationSelector(ngZone: NgZone): () => Observable<number> {
  return () => defer(getResolvedPromise).pipe(mapTo(1));
```

Alternatively, this will do the same as above, just more tersely and with more promise allocations:

```ts
return () => defer(() => Promise.resolve(1));
```

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,330 @@
...
+    } else {
+      markDirty(cfg.component);
+    }
+  }
+
+  const behaviour = (o$: Observable<T>): Observable<T> => {
+    return !inZone && !inIvy
+      ? o$
+          .pipe
+          // @Notice this part of the code is in the coalescing PR
```
benlesh, line 156: I'm not sure what this means.... maybe add a link to the PR here?

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,330 @@
...
+      cfg.cdRef.detectChanges();
+    } else {
+      markDirty(cfg.component);
+    }
+  }
+
+  const behaviour = (o$: Observable<T>): Observable<T> => {
+    return !inZone && !inIvy
+      ? o$
+          .pipe
```
benlesh, line 210: Unsure of the goal here, again. `o$.pipe()` should be the same as `o$`.

modules/component/src/push/push.pipe.ts
```diff
@@ -67,63 +54,46 @@ import {
...
-  private readonly configSubject = new Subject<PushPipeConfig>();
-  private readonly config$ = this.configSubject
-    .asObservable()
-    .pipe(distinctUntilChanged());
+  private renderedValue: any | null | undefined;
 
   private readonly subscription: Unsubscribable;
   private readonly cdAware: CdAware<S | null | undefined>;
-  private readonly updateViewContextObserver: PartialObserver<
+  private readonly resetContextObserver: NextObserver<unknown> = {
```
benlesh, line 61: Honestly, I wouldn't use RxJS's types here, just let TS figure it out. We may remove `NextObserver` at some point

modules/component/src/push/push.pipe.ts
```diff
@@ -67,63 +54,46 @@ import {
...
-    this.cdAware.next(potentialObservable);
-    return this.renderedValue;
+  transform<T>(potentialObservable: null, config?: string): null;
+  transform<T>(potentialObservable: undefined, config?: string): undefined;
+  transform<T>(
+    potentialObservable: Observable<T> | Promise<T>,
+    config?: string
+  ): T;
+  transform<T>(
+    potentialObservable: Observable<T> | Promise<T> | null | undefined,
```
benlesh, line 91: Use `ObservableInput<T> | null | undefined` here.  `ObservableInput<T>` covers the gamut of what RxJS can handle.

modules/component/src/push/push.pipe.ts
```diff
@@ -67,63 +54,46 @@ import {
  */
 @Pipe({ name: 'ngrxPush', pure: false })
 export class PushPipe<S> implements PipeTransform, OnDestroy {
-  private renderedValue: S | null | undefined;
-
-  private readonly configSubject = new Subject<PushPipeConfig>();
-  private readonly config$ = this.configSubject
-    .asObservable()
-    .pipe(distinctUntilChanged());
+  private renderedValue: any | null | undefined;
```
benlesh, line 57: `any | null | undefined` is just `any`. Maybe be more specific?

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: True! Very good!
Will insert.

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: Excuse the question...
logging with `console.log()`?

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: https://github.com/ngrx/platform/pull/2456
Will also add it to the code.

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: the internals here are not finished as it needs the operators in PR https://github.com/ngrx/platform/pull/2456
I should have deleted it completely...

modules/component/src/push/push.pipe.ts
BioPhoton: I remember. There is the potential deprecation of callbacks.
I will definitely do it.

modules/component/src/push/push.pipe.ts
BioPhoton: Very nice!

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: Done.

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Done

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Done

modules/component/src/push/push.pipe.ts
BioPhoton: Done

modules/component/src/push/push.pipe.ts
BioPhoton: Done

modules/component/src/push/push.pipe.ts
BioPhoton: Ok, @benlesh I tried and I realized I don't know how... I can't not type it (implicit any).

I can just assign the function in the pipe without typing:
Before:
```typescript
 private readonly resetContextObserver: NextObserver<unknown> = {
    next: (value: unknown) => (this.renderedValue = undefined),
 };
```
After:
```typescript
 private readonly resetContextObserver = (value: unknown) => (this.renderedValue = undefined);
```

But in the CdAware creating function I have to type it, right??
```typescript
export function createCdAware<U>(cfg: {
  strategies: StrategySelection<U>;
  resetContextObserver: NextObserver<any>;
  updateViewContextObserver: PartialObserver<U> & NextObserver<U>;
})
```

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: AFAICT zone patched EVERYTHING before Angular bootstraps. 

I can get access to the unpatched version of a promise over:
```typescript
(window as any).__zone_symbol__Promise
```
for the `animationFrame` it is:

```typescript
 (window as any).__zone_symbol__requestAnimationFrame;
 (window as any).__zone_symbol__cancelAnimationFrame;
```

So I would check if we are in zone full env and return the patched version or if zoneless the original like this for Promises:
```typescript
/** A shared promise instance to cause a delay of one microtask */
let resolvedPromise: Promise<void>|null = null;

function getResolvedPromise() {
  resolvedPromise = resolvedPromise ||  hasZone(ngZone) ? (window as any).__zone_symbol__Promise.resolve() : Promise.resolve();
  return resolvedPromise;
}

function getSaveDurationSelector(ngZone: NgZone): () => Observable<number> {
    defer(getResolvedPromise).pipe(mapTo(1));
}
```

And for animation frame this:
```typescript
function getAnimationFrame() {
 return hasZone(ngZone)
      // custom operator similar to animationFrames, just let me pass the reference to the function to generate frames from (set/cancel interval, timeout, animationFrame)
      ? generateFrames(
          (window as any).__zone_symbol__requestAnimationFrame,
          (window as any).__zone_symbol__cancelAnimationFrame
        )
      : animationFrames();
}

function getSaveDurationSelector(ngZone: NgZone): () => Observable<number> {
    defer(getAnimationFrame);
}
```


Here the last question:

Promise vs AnimationFrame I had no differences in the tests but would go for AnimationFrame. At first, I thought I could miss values, but then I equalized angular is also using the AnimationFrame to coalesce.

Any technical debt here?

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: Done

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: I went with `observablesFromTemplate$`

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: > May want to move this after startWith to ensure that it doesn't fire the same thing twice if the first value from configSubject happens to be the same as startWith.

Done

BioPhoton: >Also consider using a BehaviorSubject instead of startWith? Depends on your goal. It seems like you might end up with a situation where some code is changed or adjusted to subscribe in a more lazy way, and renderSideEffect$ could be subscribed to after the initial value is emitted from here, and it may be confusing to debug.

Done

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: Done

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: Ah,  well... `console.error` would be better... haha

BioPhoton: Done

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: I don't understand how

How would I handle distinctUntilChange for observables and how can I provide both observable and config?

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: Done

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,339 @@
...
+  idle: CdStrategy<U>;
+
+  [key: string]: CdStrategy<U>;
+}
+
+export function getStrategies<T>(
+  cfg: StrategyFactoryConfig
+): StrategySelection<T> {
+  return {
+    idle: createIdleStrategy<T>(cfg),
```
BioPhoton, line 57: @SanderElias mentioned 'idle' is not a good name.

idle is the strategy that mirrors async pipe behavior.

Suggestions for new naming: 'async-like'.

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,339 @@
...
+    noop: createNoopStrategy<T>(),
+    pessimistic1: createPessimistic1Strategy<T>(cfg),
```
BioPhoton, line 59: @SanderElias suggest to change the numbers.

We have a scale of optimism that goes from most optimistic/performat to must pessimistic/performat behavior.
Also, the optimism assumes you know exactly what you do and your code is all correct.

**Optimism Scale**
- 1. `optimistic2`
- 2. `optimistic1`
- 3. `pessimistic2`
- 4. `pessimistic1`

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,339 @@
...
+// import { generateFrames } from '../projections/generateFrames';
+// import { coalesce } from '../operators/coalesce';
+import { defer, from, MonoTypeOperatorFunction, Observable } from 'rxjs';
+import {
+  ChangeDetectorRef,
+  NgZone,
+  ɵdetectChanges as detectChanges,
+  ɵmarkDirty as markDirty,
+} from '@angular/core';
+import { hasZone, isIvy } from '../utils';
```
BioPhoton, line 11: @SanderElias suggests to use the window object to determine if zone is present or not

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/**
+ * Idle Strategy
+ *
+ * This strategy is the drop-in replacement for Angular's built-in `async` pipe.
+ * This is the only strategy that **does not also work in zone-less environments**.
+ *
+ * - \>=ViewEngine
+ * |          | Render Method      | Coalescing | Coalesce Scope |
```
BioPhoton, line 73: @SanderElias suggests to change the naming form `Coalescing` and `Coalescing Scope`  to something developers can pick up fast. Coalescing is a word we all have to google so it will be hard to grasp.

Maybe `squash` or `squashing` like in `git squash` ?

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: In the code it looks like:

```html
<!-- push pipe -->
{{value1$ | ngrxPush: 'level1'}}
<!-- let directive-->
<ng-container *ngrxLet="value$ as v; let w" strategy="level1">
  Value: {{v}} {{w}}
</ng-container>
```

hoebbelsB: hey there, since I am very interested in the ngrx/component package and can see a huge benefit from those strategies, here are just my 2 cents.

Imho the current strategy API is way to complicated and unintuitive for the user. If I wouldn't know by heart what those strategy names do internally, I would have to look up the behavior pretty much every time I want to use it.

I like APIs being transparent and intuitive to use.
In a first step I would suggest to get rid of the internal `inZone` condition. A strategy should always behave the same independent of the zone. Implementing this gives users the confidence that the configured strategy will always behave the same.

Having the zone condition removed, the only difference between the strategies left is the `ChangeDetection` handling. I see 5 different approaches to implement:

- **markDirty** (or bottomUp): calls `markDirty` / `markForCheck`
- **detectChanges** (or topDown): uses `coalesce` on component scope and calls `detectChanges`
- **instant**: no coalescing, just call `detectChanges`
-**asyncLike**: mirror async pipe
-**noop**: don't do any changeDetection, just emit values

BioPhoton: I like the approach of ignoring the zone related thingy a bit. 
It helped me to think a bit more broad. 

Regarding **instant** as the coalescing happens as microtask over a Promise it is instant, menaing in the same tick.


There is the concept of local and global in my head...

So markForCheck and markDirty got up and render over AppRef.tick. 
DetectChanges doesn't. It only renders 'locally' from the current component down.

Also coalescing and local rendering should always happen together

**What about: **

- VE/I - Options for ViewEngine / Ivy. (* after the / means same for both )
- mFC - `cdRef.markForCheck`
- dC - `cdRef.detectChanges`
- ɵMD - `ɵmarkDirty` 
- ɵDC - `ɵdetectChanges`
- LV  - `LView`
- C - `Component`
  
| Name        | ZoneLess VE/I | Render Method VE/I  | Coalescing VE/I  |
|-------------| --------------- | ------------------------ | ----------------- | 
| `asyncLike` | ❌/❌            | mFC / mFC                 | ❌                      | 
| `global`      | ❌/✔️            | mFC  / ɵMD               | ❌                      |
| `local`         | ✔️/✔️            | dC / ɵDC                    | ✔️ + C/ LV         | 
| `noop`        | ❌/❌           | no rendering               | ❌                      |

modules/component/src/core/cd-aware/cd-aware_creator.ts
BioPhoton: I tried to get something working. No luck for me.

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: When looking at this today, I really like that there are only 2 real strategies. 
`global` and `local`. The 4 initial versions of strategies where based on a not existing set of test cases. As they work there is no need for this separation.

The naming should definitely be related to the rendering process.
detectChanges renders the very component down, markDirty/markForCheck the tree upwards.

`asyncLike` => `native` as it is the native behavior of Angulars `async` pipe?

BioPhoton: I still don't like global, local. Maybe something with component level?

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: As the strategy apply the coalescing when needed we can skip it completely

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: In a discussion in another comment

modules/component/src/core/cd-aware/strategy.ts
niklas-wortmann: Another naming option would be "root" & "branch" instead of global/local. I like that because it is kinda simular to the general tree structure of the component architecture.

modules/component/src/core/cd-aware/strategy.ts
LayZeeDK: Sometimes we pass the `'noop'` `ngZone` option but forget to remove `zone.js` imports from `polyfills.ts` 🤷‍♂️ In this case, we should be checking for NoopNgZone, not just global Zone.js objects and monkeypatches.

BioPhoton: Good Point!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Implemented.

modules/component/src/push/push.pipe.ts
BioPhoton: I was not able to have it working other than introducing my own typing for the same thing.

Keep it for now.

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * - Ivy
+ *
+ * |          | Render Method      | Coalescing | Coalesce Scope |
+ * | -------- | -------------------| ---------- | -------------- |
+ * | ZoneFull | cdRef.markForCheck | ❌         | None           |
+ * | ZoneLess | cdRef.markForCheck | ❌         | None           |
+ */
+export function createIdleStrategy<T>(
+  cfg: Pick<StrategyFactoryConfig, 'cdRef' | 'component'>
```
alex-okrushko, line 78: here the type Picks the only two properties that StrategyFactoryConfig has, so that's the same as `StrategyFactoryConfig`.

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,180 @@
...
+    },
+    behaviour: () => o => o,
+    name: 'asyncLike',
+  };
+}
+
+/**
+ * Noop Strategy
+ *
+ * This strategy is does nothing.
```
alex-okrushko, line 92: remove `is`

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * Global Strategy
+ *
+ * This strategy is leveraging the bottom up rendering approach.
+ *
+ */
+export function createGlobalStrategy<T>(
+  cfg: StrategyFactoryConfig
+): CdStrategy<T> {
+  const inIvy = isIvy();
```
alex-okrushko, line 125: Why this strategy got isIvy() saved into a variable? (the rest of them didn't)

modules/component/src/let/let.directive.ts
```diff
@@ -75,7 +52,7 @@ export interface LetViewContext<T> {
  * @usageNotes
  *
  * The `*ngrxLet` directive take over several things and makes it more convenient and save to work with streams in the template
- * `<ng-container *ngrxLet="observableNumber$ as c"></ng-container>`
+ * `<ng-container *let="observableNumber$ as c"></ng-container>`
```
alex-okrushko, line 55: where did `ngrx` go?

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -1,36 +1,30 @@
...
-  cdRef: ChangeDetectorRef;
-}
-
-export function setUpWork(cfg: WorkConfig): () => void {
-  const render: (component?: any) => void = getChangeDetectionHandler(
-    cfg.ngZone,
-    cfg.cdRef
-  );
-  return () => render(cfg.context);
+  nextPotentialObservable: (value: any) => void;
```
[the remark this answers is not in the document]
BioPhoton: Had it in some version. The idea is, I can't know what value ends up in the template and I convert ObservableInputs and null/undefined. The rest throws.

I can adapt to it. What do you say?

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -43,40 +37,69 @@ export function setUpWork(cfg: WorkConfig): () => void {
...
+        cfg.updateViewContextObserver.next(observable$);
+        // Render the view
+        strategy.render();
+        // Stop further processing
+        return EMPTY;
+      }
+
+      // If a new Observable arrives, reset the value to render
+      // We do this because we don't know when the next value arrives and want to get rid of the old value
+      cfg.resetContextObserver.next(undefined);
```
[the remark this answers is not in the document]
BioPhoton: Yes! Will update! 
Thanks!

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,180 @@
...
+export interface CdStrategy<T> {
+  behaviour: (cfg?: any) => MonoTypeOperatorFunction<T>;
+  render: () => void;
+  name: string;
+}
+
+export const DEFAULT_STRATEGY_NAME = 'asyncLike';
+
+export interface StrategySelection<U> {
+  asyncLike: CdStrategy<U>;
```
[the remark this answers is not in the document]
BioPhoton: I didn't implement the strategy logic because I wait for another PR to get merged.
Is this what you mean?

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Yes!
The strategies need more love. It's an artifact from refactoring.
Will fix it.

modules/component/src/let/let.directive.ts
BioPhoton: Goooood catch!
I had some problem merging master. I made a mistake.

Will fix!

modules/component/src/core/cd-aware/strategy.ts
hoebbelsB: `asyncLike` => `native` is quite cool. But why would you invent new names instead of just using `markForCheck` and `detectChanges` as names?

BioPhoton: Because the render methods use differ based on the platform you ran.

the concept is the same `global` or `bottom-up`, the technical implementation is different and is also related to different conditions. e.g. `global` uses `markForCheck` in ViewEngine and markDirty in ivy.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -43,40 +37,69 @@ export function setUpWork(cfg: WorkConfig): () => void {
...
-    map(value$ =>
-      value$.pipe(distinctUntilChanged(), tap(cfg.updateViewContextObserver))
-    ),
-    cfg.configurableBehaviour,
-    switchAll(),
-    tap(() => cfg.work())
+  const strategyNameSubject = new BehaviorSubject<string>(
+    DEFAULT_STRATEGY_NAME
+  );
+  const strategy$: Observable<CdStrategy<U>> = strategyNameSubject.pipe(
```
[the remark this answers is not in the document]
BioPhoton: I implemented it based on the fact that it is technically possible to change it as these are input bindings. Therefore, some users will do it.

This situation should not throw errors nor work unexpectedly. (even if it might happen rarely)

This was the reason for implementing it in that way. If it can change I have to handle it correctly.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -43,40 +37,69 @@ export function setUpWork(cfg: WorkConfig): () => void {
...
 export function createCdAware<U>(cfg: {
-  work: () => void;
-  resetContextObserver: NextObserver<unknown>;
-  configurableBehaviour: (
-    o: Observable<Observable<U | null | undefined>>
-  ) => Observable<Observable<U | null | undefined>>;
-  updateViewContextObserver: PartialObserver<U | null | undefined>;
+  strategies: StrategySelection<U>;
+  resetContextObserver: NextObserver<undefined>;
+  updateViewContextObserver: PartialObserver<U> & NextObserver<U>;
```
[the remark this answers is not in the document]
BioPhoton: Ah, good catch I guess I searched for partial Observer with next required.

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: @LayZeeDK 

So there are 2 things:
- zone.js is used and patches the global API, or not
- Angulars NgZone  service is of type `NgZone` or `NgNoopZone`

I thought if there are valid cases where `zone.js` patches the global API and Angular is NOT using `NgZone` but `NgNoopZone`. 

I found Angular elements as a valid example where you can write the element with `NgNoopZone` and the consumer with `NgZone`. 

**So it requires me to always check for the ng zone services instance. Right?**

As I changed this logic 2 times now I thought I make it nice first this time.

I pushed 2 new files:
- zone-checks.ts
- view-engine-checks.ts

The detailed documentation is present now for all the different cases and checks.

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: I double-checked. 
PartialObserver<T> = NextObserver<T> | ErrorObserver<T> | CompletionObserver<T>;

acts prette much as 
```typescript
{
next: (v) => void;
error?: (v) => void;
complete?: () => void;
}
```

modules/component/src/core/cd-aware/strategy.ts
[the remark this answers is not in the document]
BioPhoton: Ah!! It is the default strategy name. I have to use the constant here.

I really like your reviews! The things you spot are pretty nice! Thanks for your time!

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,180 @@
...
+  return () => defer(getResolvedPromise).pipe(mapTo(1));
+}
+
+export interface StrategyFactoryConfig {
+  component: any;
+  cdRef: ChangeDetectorRef;
+}
+
+export interface CdStrategy<T> {
+  behaviour: (cfg?: any) => MonoTypeOperatorFunction<T>;
```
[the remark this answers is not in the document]
BioPhoton: At the moment it is not used.I guess it is best to just do put ie as long as it is not used.

Good catch!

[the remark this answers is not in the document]
BioPhoton: Fixed.

modules/component/src/core/cd-aware/strategy.ts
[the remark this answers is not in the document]
BioPhoton: Fixed!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Fixed!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Fixed!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Right! As isIvy is evaluated once only, I move it to a constant.

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Fixed!

modules/component/src/let/let.directive.ts
BioPhoton: Fixed!

modules/component/src/core/cd-aware/strategy.ts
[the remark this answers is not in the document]
BioPhoton: I replaced all usages with DEFAULT_STRATEGY_NAME.

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -43,40 +37,69 @@ export function setUpWork(cfg: WorkConfig): () => void {
...
-      value$.pipe(distinctUntilChanged(), tap(cfg.updateViewContextObserver))
-    ),
-    cfg.configurableBehaviour,
-    switchAll(),
-    tap(() => cfg.work())
+  const strategyNameSubject = new BehaviorSubject<string>(
+    DEFAULT_STRATEGY_NAME
+  );
+  const strategy$: Observable<CdStrategy<U>> = strategyNameSubject.pipe(
+    nameToStrategy(cfg.strategies)
```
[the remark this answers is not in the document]
BioPhoton: Good point! I will use string.literals.

I found this issue;
>... replaces the NotiticationKind const enum that was introduced in 6.4.0 with a string-literal union type. The latter is type-safe and compatible with single-module transpilation. Const enums are not, as the transpiler cannot know what should be used as a replacement for a reference to a const enum without using type information.

here the link: https://github.com/ReactiveX/rxjs/pull/4556

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: fixed!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Fixed!

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: I asked some peoples that experimented with the different rendering approaches. 
AFAICT they had the use-case to chance the rendering based on some used dynamic components. I will analyze that one and get back with infos and solutions.

modules/component/src/core/cd-aware/nameToStrategy.ts
```diff
@@ -0,0 +1,22 @@
...
+  CdStrategy,
+  DEFAULT_STRATEGY_NAME,
+  StrategySelection,
+} from './strategy';
+
+export function nameToStrategy<U>(strategies: StrategySelection<U>) {
+  return (o$: Observable<string>): Observable<CdStrategy<U>> => {
+    return o$.pipe(
+      distinctUntilChanged(),
+      startWith(DEFAULT_STRATEGY_NAME),
```
timdeschryver, line 13: After going through Ben's comments, should the `startWith` go before the `distinct`?

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: @LayZeeDK
I rethought the whole thing. After different experiments and tests I concluded a really nice thing!

The really only thing I have to check is if the PROMISE API is untouched or not (If we evaluate a safe usage of animationFrame that it would be that API).

The rest can now be used in both ZoneLess and ZoneFull without reflecting it in the code. I basically kicked out ALL zone related checks everywhere but the creation of Promise that is here for coalescing.

This refactoring also included:
- get rid of passing the component
- most probable get ird of _lView (if _lView  === cdRef.context across class and pipe/directive in the template)
- getting rid if generateFrages RxJS operator
- get rid of All zone related tests but the one for the Promise

Thanks a lot for making me think the 10th time on that problem !!!

modules/component/src/core/cd-aware/nameToStrategy.ts
```diff
@@ -0,0 +1,22 @@
...
+      map(
+        (strategy: string): CdStrategy<U> =>
+          strategies[strategy]
+            ? strategies[strategy]
+            : strategies[DEFAULT_STRATEGY_NAME]
```
timdeschryver, line 18: Reads a bit easier imho

```suggestion
          strategies[strategy] || strategies[DEFAULT_STRATEGY_NAME]
```

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Conclusion:
The only thing where I have to deal with `zone.js` is in the coalescing logic and the only API interesting to check is the `Promise` API.

BioPhoton: Let me know your thoughts

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,207 @@
...
+  render: () => void;
+  name: string;
+}
+
+export const DEFAULT_STRATEGY_NAME = 'native';
+export const IS_VIEW_ENGINE_IVY = isViewEngineIvy();
+
+export interface StrategySelection<U> {
+  [DEFAULT_STRATEGY_NAME]: CdStrategy<U>;
+  [key: string]: CdStrategy<U>;
```
timdeschryver, line 29: If we only have default, noop, global, and local, why don't we explicitly set these properties instead of using `[key: string]`?

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Ok, the current naming is: `native`, `noop`, `local`, `global`.

The team has to decide what to name `global` and `local`

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,207 @@
...
+export function createGlobalStrategy<T>(
+  config: StrategyFactoryConfig
+): CdStrategy<T> {
+  function render() {
+    if (!IS_VIEW_ENGINE_IVY) {
+      config.cdRef.markForCheck();
+    } else {
+      markDirty((config.cdRef as any).context);
+    }
+  }
```
timdeschryver, line 134: Nit, I find a negated if statement harder to read
```suggestion
    if (IS_VIEW_ENGINE_IVY) {
      markDirty((config.cdRef as any).context);
    } else {
      config.cdRef.markForCheck();
    }
  }
```

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * Noop Strategy
+ *
+ * This strategy is does nothing. It serves for debugging only
+ *
+ * | Name        | ZoneLess VE/I | Render Method VE/I  | Coalescing VE/I  |
+ * |-------------| --------------| ------------ ------ | ---------------- |
+ * | `noop`      | ❌/❌         | no rendering        | ❌               |
+ *
+ * @param config { StrategyFactoryConfig } - The values this strategy needs to get calculated.
```
timdeschryver, line 97: Seems like `createNoopStrategy` doesn't expect a config 

```suggestion
 
```

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,207 @@
...
+    name: 'noop',
+  };
+}
+
+/**
+ *
+ * Global Strategy
+ *
+ * This strategy is rendering the application root and
+ * all it's children that are on a path
```
timdeschryver, line 114:
```suggestion
 * all its children that are on a path
```

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,207 @@
...
+    render,
+    name: 'global',
+  };
+}
+
+/**
+ *  Local Strategy
+ *
+ * This strategy is rendering the actual component and
+ * all it's children that are on a path
```
timdeschryver, line 149:
```suggestion
 * all its children that are on a path
```

modules/component/src/core/cd-aware/strategy.ts
```diff
@@ -0,0 +1,207 @@
...
+  };
+
+  function render() {
+    // @TODO ensure that detectChanges is behaves identical to ɵdetectChanges
+    // If yes, kick out ɵdetectChanges
+    if (!IS_VIEW_ENGINE_IVY) {
+      config.cdRef.detectChanges();
+    } else {
+      detectChanges((config.cdRef as any).context);
+    }
```
timdeschryver, line 191: Same nit here
```suggestion
    if (IS_VIEW_ENGINE_IVY) {
      detectChanges((config.cdRef as any).context);
    } else {
      config.cdRef.detectChanges();
    }
```

modules/component/src/core/utils/view-engine-checks.ts
```diff
@@ -0,0 +1,33 @@
+import { getGlobalThis } from './get-global-this';
+
+/**
+ * envRunsIvy
```
timdeschryver, line 4: This can be left out (it's enough to have a description)? 
Or should it say more?

```suggestion

```

modules/component/src/core/utils/zone-checks.ts
```diff
@@ -0,0 +1,16 @@
+import { getGlobalThis } from './get-global-this';
+
+/**
+ * apiZonePatched
```
timdeschryver, line 4: Same here
```suggestion
```

modules/component/src/core/utils/zone-checks.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @description
+ *
+ * This function checks the a specific Browser API is patched by `zone.js`.
```
timdeschryver, line 8:
```suggestion
 * This function checks if a specific Browser API is patched by `zone.js`.
```

modules/component/src/rxjs/operators/renderChanges.ts
```diff
@@ -0,0 +1,25 @@
+import { MonoTypeOperatorFunction, Observable } from 'rxjs';
+import { CdStrategy } from '../../core';
+import { tap } from 'rxjs/operators';
+
+/**
+ * renderChanges
```
timdeschryver, line 6: Here too
```suggestion
```

modules/component/src/rxjs/operators/renderChanges.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This operator is useful in situations where the developer can't code reactively everywhere, but wants the component to rerender on emission.
+ *
+ * A good example would be assigning the value arrived over a Observable to a class property and rerender the component.
```
timdeschryver, line 10:
```suggestion
 * A good example would be assigning the value arrived over an Observable to a class property and rerender the component.
```

modules/component/src/core/cd-aware/nameToStrategy.ts
BioPhoton: Fixed!

modules/component/src/core/cd-aware/nameToStrategy.ts
BioPhoton: Very nice!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: fixed!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: I thought of custom strategies. But I will set it directly for now!

Thanks!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Thanks Fixed!

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Fixed!

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
alex-okrushko: Of course no one knows what values ends up in the template, but with generics you say _"let's say the passed in value here would be of type `T` then here it would be `T` as well and over here it would be `Observable<T>`"_ and so on.

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
alex-okrushko: Nicholas used  literal union of string instead there. Not just strings.
Non-const enum would work great as well.

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
alex-okrushko: Which is just a `NextObserver<U>`, right?

https://github.com/ReactiveX/rxjs/blob/49304ffef8d7a0663c57fe8e673359a602e9d3e1/src/internal/types.ts#L85

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: Wil fix it

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: will use an enum

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: It helps a lot in testing.

But I kick it out!

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: yes. will fix it

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -43,40 +37,69 @@ export function setUpWork(cfg: WorkConfig): () => void {
...
+  );
+
+  const potentialObservablesSubject = new Subject<Observable<U>>();
+  const observablesFromTemplate$ = potentialObservablesSubject.pipe(
```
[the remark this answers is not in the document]
BioPhoton: will fix it

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: @alex-okrushko as we don't include the performance optimizations part I could delete the other strategies. so we have the to from the beginning. native is identical to `async`,
And uses `ChangeSDetectorRef#markForCheck` always 
and the other uses:
-  `ChangeSDetectorRef#markForCheck` in ViewEngine/Zoneful
-  `ChangeSDetectorRef#detectChanges` in ViewEngine/ZoneLess
-  `emarkForCheck` in Ivy/Zoneful
-   `ChangeSDetectorRef#detectChanges` in Ivy/ZoneLess
Now I could go back to your earlier suggestion 
```typescript
nextStrategy(nextConfig: string): void {
      strategy =
        cfg.strategies[nextConfig] || cfg.strategies[DEFAULT_STRATEGY_NAME];
    }
```

Are you fine with that?

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: Done

modules/component/src/core/cd-aware/strategy.ts
BioPhoton: Fixed by implementing the suggestion of Alex

modules/component/src/core/cd-aware/cd-aware_creator.ts
```diff
@@ -43,40 +37,69 @@ export function setUpWork(cfg: WorkConfig): () => void {
...
+          return EMPTY;
+        })
+      );
+    })
   );
 
   return {
-    next(value: any): void {
-      observablesSubject.next(value);
+    nextPotentialObservable(value: any): void {
```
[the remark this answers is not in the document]
BioPhoton: Fixed

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: Fixed

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: Fixed

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: Fixed

brandonroberts:
Cleanup work is still being done on this PR. Will unblock when its ready for another review

BioPhoton:
@brandonroberts 
ATM I have a problem that e2e tests failing.
I remember this happened already and got fixed in circle ci?

I would need a bit of help to proceed.

timdeschryver:
@BioPhoton I believe that if you pull in the latest version of master that the e2e tests will succeed.

modules/component/src/core/cd-aware/cd-aware_creator.ts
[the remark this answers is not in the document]
BioPhoton: I aligned the whole process now to your wishes.
The reference is here: https://github.com/BioPhoton/platform/pull/1/files

After this is merged we can go even less reactive and align better with the ideas in the team.

brandonroberts:
@BioPhoton is this ready for a review?

BioPhoton:
Hi Sry for the delay! I want to do a last check in a more real life scenario.

I hope I can squeeze in time tomorrow.

BioPhoton:
I gave it a try. Not a complex project but the basics.

I would focus on small PRs to improve typing and tests and investigate in #2488 

It this approach fine?

modules/component/spec/core/cd-aware/render_creator.ts
```diff
@@ -0,0 +1,31 @@
+import { createRender } from '../../../src/core/cd-aware';
```
brandonroberts, line 1: nit: add `.spec` to the filename

modules/component/src/core/utils/zone-checks.ts
```diff
@@ -0,0 +1,22 @@
...
+ * if the constructor name is equal to `NgZone`.
+ * This means the Angular application that instantiated this service assumes it runs in a ZuneLess environment,
+ * and therefor it's change detection will not be triggered by zone related logic.
+ *
+ * However, keep in mind this does not mean `zone.js` is not present.
+ * The environment could still run in ZoneFull mode even if Angular turned it off.
+ * Consider the situation of a Angular element configured for ZoneLess
+ * environments is used in an Angular application relining on the zone mechanism.
+ *
+ * @param instance {Class Instance} - The instance to check for constructor name of `NgZone`.
```
brandonroberts, line 16: nit: remove the `{types}` from the comments

modules/component/src/core/utils/zone-checks.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @return {boolean} - true if instance is of type `NgZone`.
```
brandonroberts, line 17: nit: remove this as it causes issues with the annotations

modules/component/spec/core/cd-aware/render_creator.ts
BioPhoton: done

modules/component/src/core/utils/zone-checks.ts
BioPhoton: done

modules/component/src/core/utils/zone-checks.ts
BioPhoton: done

BioPhoton:
I need some time to fix the conflicts

brandonroberts:
Closing in favor of #2566

PR #2492, opened 2020-04-15
fix(router-store): selectors should return MemoizedSelector

### LINKED ISSUE #2486: Built-in router-store selectors can't be mocked

It seems impossible to properly mock the built-in router-store selectors with `mockStore.overrideSelector`. Look at the `user-greeting.component.spec.ts` in the StackBlitz below. One of the tests doesn't compile because the types are not compatible on line 22:

`mockStore.overrideSelector(selectRouteParams, { id: 5 });`

The other test attempts to use the string version of `mockStore.overrideSelector`, but that doesn't work either. Could be user error.

https://stackblitz.com/edit/angular-7ntszo?file=src%2Fapp%2Fuser-greeting.component.spec.ts

## Expected behavior:

I expect to be able to mock the built-in router-store selectors like any other selector created with `createSelector`. Instead, I get a type error. It becomes clear why this is happening if you take a look at how the selectors are defined:

https://github.com/ngrx/platform/blob/master/modules/router-store/src/models.ts

The types used in the public API do not define the selectors as `MemoizedSelector`, as the `overrideSelector` method expects.

Maybe there's a good reason for this. If there is, and there's no appetite to change the types, I recommend this be considered a documentation issue, and documentation be added to clarify how to mock these selectors. I believe the selectors in the `entity` module will have the same problem for the same reason, but I haven't tested them.

One workaround I have used is to tell TypeScript that `selectRouteParams` is an `any`:

`mockStore.overrideSelector((selectRouteParams as any), { id: 5 });`

Far from ideal, of course.

Angular 9.1.1
NgRx 9.1
Chrome

[ ] No

timdeschryver:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2486 

Could be a breaking change, if these types were being used

```
[ ] Yes
[ ] No
```

### DIFF
--- a/modules/router-store/src/models.ts
+++ b/modules/router-store/src/models.ts
@@ -1,11 +1,14 @@
 import { Data, Params } from '@angular/router';
+import { MemoizedSelector } from '@ngrx/store';
 
 export interface RouterStateSelectors<V> {
-  selectCurrentRoute: (state: V) => any;
-  selectQueryParams: (state: V) => Params;
-  selectQueryParam: (param: string) => (state: V) => string | undefined;
-  selectRouteParams: (state: V) => Params;
-  selectRouteParam: (param: string) => (state: V) => string | undefined;
-  selectRouteData: (state: V) => Data;
-  selectUrl: (state: V) => string;
+  selectCurrentRoute: MemoizedSelector<V, any>;
+  selectQueryParams: MemoizedSelector<V, Params>;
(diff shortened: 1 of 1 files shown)


joshuakeel:
Thanks for working on this, @timdeschryver. 🤓

PR #2502, opened 2020-04-20
fix(entity): selectors should return MemoizedSelector

twittwer:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The entity selectors can not be mocked via overrideSelector - like the [bug in the router store](https://github.com/ngrx/platform/pull/2492)

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -1,3 +1,5 @@
+import { MemoizedSelector } from '@ngrx/store';
+
 export type ComparerStr<T> = (a: T, b: T) => string;
 export type ComparerNum<T> = (a: T, b: T) => number;
 
@@ -71,10 +73,10 @@ export interface EntityStateAdapter<T> {
 }
 
 export interface EntitySelectors<T, V> {
-  selectIds: (state: V) => string[] | number[];
-  selectEntities: (state: V) => Dictionary<T>;
-  selectAll: (state: V) => T[];
-  selectTotal: (state: V) => number;
+  selectIds: MemoizedSelector<V, string[] | number[]>;
+  selectEntities: MemoizedSelector<V, Dictionary<T>>;
+  selectAll: MemoizedSelector<V, T[]>;
+  selectTotal: MemoizedSelector<V, number>;
 }
 
 export interface EntityAdapter<T> extends EntityStateAdapter<T> {
--- a/modules/entity/src/state_selectors.ts
+++ b/modules/entity/src/state_selectors.ts
@@ -9,8 +9,11 @@ export function createSelectorsFactory<T>() {
   function getSelectors(
     selectState?: (state: any) => EntityState<T>
   ): EntitySelectors<T, any> {
-    const selectIds = (state: any) => state.ids;
-    const selectEntities = (state: EntityState<T>) => state.entities;
+    const selectIds = createSelector((state: any) => state, state => state.ids);
+    const selectEntities = createSelector(
+      (state: EntityState<T>) => state,
+      state => state.entities
+    );
     const selectAll = createSelector(
       selectIds,
       selectEntities,


twittwer:
Yes, sorry - I'll check it today

twittwer:
@alex-okrushko Currently the `selectIds` and `selectEntities` are MemoizedSelectors only in case a `selectState` parameter is passed ([code](https://github.com/ngrx/platform/blob/master/modules/entity/src/state_selectors.ts#L23)).
Would you prefer to adapt the typing accordingly or to ensure there are always MemoizedSelectors?

In my opinion option A sounds better regarding consistency and the ability  to override all selectors.

### Option A (Ensure MemoizedSelectors)
* current
```ts
    const selectIds = (state: any) => state.ids;
    const selectEntities = (state: EntityState<T>) => state.entities;
```
* proposal
```ts
    const selectIds = createSelector((state: any) => state, state => state.ids);
    const selectEntities = createSelector((state: EntityState<T>) => state, state => state.entities);
```

### Option B (Conditional MemoizedSelectors)
* current
```ts
export interface EntitySelectors<T, V> {
  selectIds: (state: V) => string[] | number[];
  selectEntities: (state: V) => Dictionary<T>;
  selectAll: (state: V) => T[];
  selectTotal: (state: V) => number;
}
```
```ts
  function getSelectors(): EntitySelectors<T, EntityState<T>>;
  function getSelectors<V>(
    selectState: (state: V) => EntityState<T>
  ): EntitySelectors<T, V>;
  function getSelectors(
    selectState?: (state: any) => EntityState<T>
  ): EntitySelectors<T, any> {
```
* proposal
```ts
export interface EntitySelectors<T, V> {
  selectIds: (state: V) => string[] | number[];
  selectEntities: (state: V) => Dictionary<T>;
  selectAll: MemoizedSelector<V, T[]>;
  selectTotal: MemoizedSelector<V, number>;
}
export interface MemoizedEntitySelectors<T, V> {
  selectIds: MemoizedSelector<V, string[] | number[]>;
  selectEntities: MemoizedSelector<V, Dictionary<T>>;
  selectAll: MemoizedSelector<V, T[]>;
  selectTotal: MemoizedSelector<V, number>;
}
```
```ts
  function getSelectors(): EntitySelectors<T, EntityState<T>>;
  function getSelectors<V>(
    selectState: (state: V) => EntityState<T>
  ): MemoizedEntitySelectors<T, V>;
  function getSelectors(
    selectState?: (state: any) => EntityState<T>
  ): EntitySelectors<T, any> | MemoizedEntitySelectors<T, any> {
```

criskrzysiu:
OK, the question is should they be memoized? I think the answer is no, as it will not increase performance, only increase complexity. With selectIds, and selectEntities we are just returning array and object. There is no way to speed this up. Memorizing this, would actually make it slower.

twittwer:
My current use case is the to mock the selectors during tests, currently I get a TS error when using `mockStore.overrideSelector(selectEntities)`.
As we are passing the `selectState` parameter to `getSelectors` it already is a memoized selector, even if not typed so.

@criskrzysiu Thx for your input - I just thought it would be the better option reagarding testing, but there may be no use case to override the basic selectors if not used in combination with `selectState`.

@alex-okrushko In case option B is the preferred one I'm gonna update the PR accordingly.

criskrzysiu:
Ok, so why not change the behavior of the MockStore.ovverideSelector? 
internally it just store a dictionary of values and can be extended to handle simple functions too. This would be better solution, and it would fix the problem for testing for other cases not only this specific one.
[Mock Store](https://github.com/ngrx/platform/blob/master/modules/store/testing/src/mock_store.ts)

alex-okrushko:
Thanks @twittwer and @criskrzysiu for the inputs.
I agree with @criskrzysiu on both cases:
- I don't think these should be memoized (maybe even regardless if `selectState` is passed or not)
- adjusting `overrideSelector` to take a simple [`Selector`](https://github.com/ngrx/platform/blob/470b78868d0bbb21a64c3454fb0802766b22433f/modules/store/src/models.ts#L49)  would help other cases as well

brandonroberts:
@twittwer Are you still interested in landing this change? If its potentially breaking, we'd like to get it in for v10. Thanks

brandonroberts:
Ok cool. @alex-okrushko can help you with that part. The overrideSelector code is here https://github.com/ngrx/platform/blob/master/modules/store/testing/src/mock_store.ts

brandonroberts:
@twittwer if you want to pick this back up, reopen the PR. Closing for now

PR #2420, opened 2020-03-02
No logs about empty state for tests with mocked store

evgenyfedorenko:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

This code adds an internal setting and a setter for mock store environment.
When set it suppresses the logs about the state which does not exist.

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Console warn logs numerous logs for mocked store which should not happen

Closes #2363

## What is the new behavior?

Logs will be suppressed

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

## Other information
Please take a look at this draft. I made some assumptions here which may not be true. Please comment.

PR #2513, opened 2020-05-06
No logs about empty state for tests with mocked store

### LINKED ISSUE #2363: MockStore selectors resulting in warning logs

Currently all the tests that use a mockstore with mocked selectors, [will log the warning that the state does not exists](https://github.com/ngrx/platform/blob/master/modules/store/src/selector.ts#L618).

This is normal because when we define a mocks store, we don't define the state object.

## Expected behavior:

It should not be logged while running tests.

Angular 9
NgRx 9-beta.1

## Other information:

AFAIK, it isn't possible to detect when we run in a test environment.
But we could set a property on the state (`__ngrxMockEnvironment__`), and the `createFeatureSelector` could use this property to check if it the warning has to logged or not.

[ ] No

evgenyfedorenko:
## PR Checklist

## PR Type

This code adds an internal setting and a setter for mock store environment.
When set it suppresses the logs about the state which does not exist.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Console warn logs numerous logs for mocked store which should not happen

Closes #2363

## What is the new behavior?

Logs will be suppressed

```
[ ] Yes
[x] No
```

## Other information
This is a copy of https://github.com/ngrx/platform/pull/2420

### DIFF
--- a/modules/store/testing/src/testing.ts
+++ b/modules/store/testing/src/testing.ts
@@ -6,6 +6,7 @@ import {
   ReducerManager,
   StateObservable,
   Store,
+  setNgrxMockEnvironment,
 } from '@ngrx/store';
 import { MockStore } from './mock_store';
 import { MockReducerManager } from './mock_reducer_manager';
@@ -20,6 +21,7 @@ export interface MockStoreConfig<T> {
 export function provideMockStore<T = any>(
   config: MockStoreConfig<T> = {}
 ): Provider[] {
+  setNgrxMockEnvironment(true);
   return [
     ActionsSubject,
     MockState,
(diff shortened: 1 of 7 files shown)


modules/store/testing/src/testing.ts
```diff
@@ -20,6 +21,7 @@ export interface MockStoreConfig<T> {
 export function provideMockStore<T = any>(
   config: MockStoreConfig<T> = {}
 ): Provider[] {
+  setNgrxMockEnvironment(true);
```
timdeschryver, line 24: Can there also be a test to verify that `setNgrxMockEnvironment` is called here?

PR #2520, opened 2020-05-09
feat(store): add runtime check for action type uniqueness

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently there is no way to warn devs when there are multiple actions with the same action type.
I've seen that this might lead to some confusion and some debugging time to spot the cause of a bug. 

The most recent one, https://twitter.com/FabianGosebrink/status/1259043401724542978

The ngrx-tslint-rules has a TSLint rule to spot these, but not many people know that this exists.

What if we create a runtime check for it?

## What is the new behavior?

In the PR, you see that action types are registered in a global array while the action is instantiated with `createAction`. When a store module is loaded, we can perform a check to see if there are duplicates in this array, and throw an error if there are.

This new feature is by default disabled, but can be enabled with the `strictActionTypeUniqueness` runtime check flag.

Thoughts?

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/runtime_checks.spec.ts
+++ b/modules/store/spec/runtime_checks.spec.ts
@@ -423,19 +423,6 @@ describe('ActionType uniqueness', () => {
     }).not.toThrowError();
   });
 
-  it('should not register action types if devMode is false', () => {
-    spyOn(ngCore, 'isDevMode').and.returnValue(false);
-
-    createAction('action 1');
-    createAction('action 1');
-
-    expect(REGISTERED_ACTION_TYPES).toEqual({});
-
-    expect(() => {
-      setupStore({ strictActionTypeUniqueness: false });
-    }).not.toThrowError();
-  });
-
   it('should not throw when disabled', () => {
     createAction('action 1');
     createAction('action 1');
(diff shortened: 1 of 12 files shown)


brandonroberts:
What do you think about using a dictionary of `ActionType: boolean` instead of array for lookups?

timdeschryver:
@brandonroberts that works! 
I will make the change later and with it, I will add a new section to the docs.

modules/store/spec/runtime_checks.spec.ts
```diff
@@ -384,6 +400,39 @@ describe('Runtime checks:', () => {
   });
 });
 
+describe('ActionType uniqueness', () => {
+  beforeEach(() => {
+    resetRegisteredActionTypes();
+  });
+
+  it('should throw when having no duplicate action types', () => {
```
jsaguet, line 408: shouldn't this be `'should throw when having duplicate action types'` ?

timdeschryver:
🤔 now why does it format differently ...

modules/store/spec/runtime_checks.spec.ts
timdeschryver: Thanks!

PR #2528, opened 2020-05-16
feat(component-store): initialization + updater/setState

alex-okrushko:
## PR Checklist

## PR Type

Initial implementation of ComponentStore - adding the initialization and state updates via `setState` and `updater`.

Detailed overview of this PR can be watched here: https://www.youtube.com/watch?v=-hp72aA0wBw
😁

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component-store/src/component-store.ts
+++ b/modules/component-store/src/component-store.ts
@@ -1,0 +1,113 @@
+import {
+  isObservable,
+  Observable,
+  of,
+  ReplaySubject,
+  Subscription,
+  throwError,
+} from 'rxjs';
+import { concatMap, takeUntil, withLatestFrom } from 'rxjs/operators';
+
+export class ComponentStore<T extends object> {
+  private readonly stateSubject$ = new ReplaySubject<T>(1);
+  private isInitialized = false;
+  readonly state$: Observable<T> = this.stateSubject$.asObservable();
+
+  // Should be used only in ngOnDestroy.
+  private readonly destroySubject$ = new ReplaySubject<void>(1);
+  // Exposed to any extending Store to be used for the teardowns.
+  readonly destroy$ = this.destroySubject$.asObservable();
+
+  constructor(defaultState?: T) {
+    // State can be initialized either through constructor, or initState or
+    // setState.
+    if (defaultState) {
+      this.initState(defaultState);
+    }
+  }
+
+  /** Completes all relevant Observable streams. */
+  ngOnDestroy() {
+    this.stateSubject$.complete();
+    this.destroySubject$.next();
+  }
+
+  /**
+   * Creates an updater.
+   *
+   * Throws an error if updater is called with synchronous values (either
+   * imperative value or Observable that is synchronous) before ComponentStore
+   * is initialized. If called with async Observable before initialization then
+   * state will not be updated and subscription would be closed.
+   *
+   * @param updaterFn A static updater function that takes 2 parameters (the
+   * current state and an argument object) and returns a new instance of the
+   * state.
+   * @return A function that accepts one argument which is forwarded as the
+   *     second argument to `updaterFn`. Everytime this function is called
+   *     subscribers will be notified of the state change.
+   */
+  updater<V>(
+    updaterFn: (state: T, value: V) => T
+  ): unknown extends V ? () => void : (t: V | Observable<V>) => Subscription {
+    return ((observableOrValue?: V | Observable<V>): Subscription => {
+      let initializationError: Error | undefined;
+      // We can receive either the value or an observable. In case it's a
+      // simple value, we'll wrap it with `of` operator to turn it into
+      // Observable.
+      const observable$ = isObservable(observableOrValue)
+        ? observableOrValue
+        : of(observableOrValue);
+      const subscription = observable$
+        .pipe(
+          concatMap(
+            value =>
+              this.isInitialized
+                ? of(value).pipe(withLatestFrom(this.stateSubject$))
+                : // If state was not initialized, we'll throw an error.
+                  throwError(
+                    Error(`${this.constructor.name} has not been initialized`)
+                  )
+          ),
+          takeUntil(this.destroy$)
+        )
+        .subscribe({
+          next: ([value, currentState]) => {
+            this.stateSubject$.next(updaterFn(currentState, value!));
+          },
+          error: error => {
+            initializationError = error;
+            this.stateSubject$.error(error);
+          },
+        });
+
+      if (initializationError) {
+        throw initializationError;
+      }
+      return subscription;
+    }) as unknown extends V
+      ? () => void
+      : (t: V | Observable<V>) => Subscription;
+  }
+
+  /**
+   * Initializes state. If it was already initialized then it resets the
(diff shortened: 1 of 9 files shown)


modules/component-store/src/component-store.ts
```diff
@@ -0,0 +1,113 @@
...
+    this.stateSubject$.next(state);
+  }
+
+  /**
+   * Sets the state specific value.
+   * @param stateOrUpdaterFn
+   */
+  setState(stateOrUpdaterFn: T | ((state: T) => T)): void {
+    if (typeof stateOrUpdaterFn !== 'function') {
+      this.initState(stateOrUpdaterFn);
```
brandonroberts, line 108: I may have missed this in the video review 😉, but why isn't `initState` just inlined in here? I don't think we need 2 ways to initialize state.

projects/ngrx.io/content/guide/component-store/index.md
```diff
@@ -1,3 +1,48 @@
...
+- Users of ComponentStore can read the state through `select` or a top-level `state$`. Selectors are extremely performant.
+- Users of ComponentStore may start side-effects with `effect`, both sync and async, and feed the data both imperatively or reactively.
+
+## Installation
+
+// TODO(alex-okrushko): fill-up the installation, including pros/cons of extending the service vs Component-scoped providers
+
+## Initialization
+
+ComponentStore can be intialized in 3 ways:
```
brandonroberts, line 23:
```suggestion
ComponentStore can be initialized in 3 ways:
```

projects/ngrx.io/content/guide/component-store/index.md
```diff
@@ -1,3 +1,48 @@
...
+    &lt;li *ngFor="let movie of (movies$ | async)"&gt;
+      {{ movie.name }}
+    &lt;/li&gt;
+  `,
+  providers: [ComponentStore],
+})
+export class MoviesPageComponent {
+  movies$: this.componentStore.state$;
+
+  constructor(private readonly componentStore: ComponentStore<Movie[]>) {}
```
brandonroberts, line 40: Should be an object

```suggestion
  constructor(private readonly componentStore: ComponentStore<{ movies: Movie[] }>) {}
```

projects/ngrx.io/content/guide/component-store/index.md
```diff
@@ -1,3 +1,48 @@
...
+@Component({
+  template: `
+    &lt;li *ngFor="let movie of (movies$ | async)"&gt;
+      {{ movie.name }}
+    &lt;/li&gt;
+  `,
+  providers: [ComponentStore],
+})
+export class MoviesPageComponent {
+  movies$: this.componentStore.state$;
```
brandonroberts, line 38: Should map to a `movies` property

```suggestion
  movies$: this.componentStore.state$.pipe(map(state => state.movies));
```

brandonroberts, line 38: Should map to a `movies` property

```suggestion
  movies$ = this.componentStore.state$.pipe(map(state => state.movies));
```

projects/ngrx.io/content/guide/component-store/index.md
```diff
@@ -1,3 +1,48 @@
...
+
+  constructor(private readonly componentStore: ComponentStore<Movie[]>) {}
+
+  ngOnInit() {
+    this.componentStore.initState([]);
```
brandonroberts, line 43:
```suggestion
    this.componentStore.setState({ movies: [] });
```

projects/ngrx.io/content/guide/component-store/index.md
alex-okrushko: Technically, Array is also an object :)
But, yeah, I followed through with your suggestion to replace it with `{movies: Movie[]}` interface.

modules/component-store/src/component-store.ts
alex-okrushko: Made `initState` private. I think it's ok to have `setState` call either function, depending on what's passed - also `initState` is called in the constructor.

projects/ngrx.io/content/guide/component-store/index.md
alex-okrushko: Thanks!

modules/component-store/src/component-store.ts
```diff
@@ -0,0 +1,114 @@
...
+        ? observableOrValue
+        : of(observableOrValue);
+      const subscription = observable$
+        .pipe(
+          concatMap(
+            value =>
+              this.isInitialized
+                ? of(value).pipe(withLatestFrom(this.stateSubject$))
+                : // If state was not initialized, we'll throw an error.
+                  throwError(
```
timdeschryver, line 68: Why are we throwing an error here?
If the state has to initialized, could make the initial value in the constructor required?

modules/component-store/src/component-store.ts
```diff
@@ -0,0 +1,114 @@
...
+    }) as unknown extends V
+      ? () => void
+      : (t: V | Observable<V>) => Subscription;
+  }
+
+  /**
+   * Initializes state. If it was already initialized then it resets the
+   * state.
+   */
+  private initState(state: T): void {
```
timdeschryver, line 97: Nit, I find the name `initState` a bit missleading because it can also be invoked via `setState`.
Should we rename this function to `updateState`/`resetState`?

projects/ngrx.io/content/guide/component-store/index.md
```diff
@@ -1,3 +1,51 @@
 # @ngrx/component-store
 
-Placeholder
+ComponentStore is a standalone library that helps managing local/component state. It's an alternative to push-based "Service with a Subject".
```
timdeschryver, line 3:
```suggestion
ComponentStore is a standalone library that helps to manage local/component state. It's an alternative to push-based "Service with a Subject".
```

modules/component-store/src/component-store.ts
alex-okrushko: The state has to be initialized before we can make updates to it using the state info, e.g.:
`setState(currentState => ({...currentState, value: newValue}))`

The idea behind the deferred initialization is to avoid selecting "default" state, if there is no need for one. E.g. I might want to have some state only when `@Input`s were passed into the component (or those inputs trigger effect which fetches the data and only then we get or first state 🙂)

modules/component-store/src/component-store.ts
alex-okrushko: it could be `resetState`, however it's also the only part that can set `this.isInitialized = true`, which might be more important piece of the info.

I'm fine either way.
WDYT?

timdeschryver: I'm fine with keeping it, the comment was mostly based of the previous commit.

alex-okrushko: I see.
Since it's also private now, we can revisit this later if necessary 🙂 (should we decide to open this method up...)

PR #2535, opened 2020-05-21
build: add Nx workspace support

brandonroberts:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[x] CI related changes
[ ] Other... Please describe:
```

- Adds support for using an Nx workspace to build/test the NgRx libraries.
- We will still have Bazel for build and deployment of npm packages, as this is considered a potential breaking change, but we will use Nx for all other tasks.
- The Bazel BUILD files will remain. 
- Further tooling will be added for better release support

Closes #

```
[ ] Yes
[x] No
```


modules/store/jest.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
timdeschryver, line 1: I'm just fiddling with the PR, but this should have `store` as name and coverage directory.

Also as a fyi, I recently discovered we can add a name (with displayName) to the [config](https://jestjs.io/docs/en/configuration#displayname-string-object) to make it more clear which module is running - do we want to add it here? 

This would result in:

image

brandonroberts: Ahh yes, I'll update that. And yes that's a good idea

timdeschryver:
Forgot about the example app... 😅
It seems like the tests don't run locally (could be me) while running `yarn example:test`

```
An unhandled exception occurred: Project 'example-app' does not support the 'test' target.
See "C:\Users\TDESCH~1\AppData\Local\Temp\ng-hYm7L5\angular-errors.log" for further details.
```

And I also think that the example tests aren't run during the CI.

angular.json
```diff
@@ -76,13 +73,506 @@
...
+            "parallel": false,
+            "commands": [
+              {
+                "command": "ng build schematics"
+              },
+              {
+                "command": "yarn tsc -p modules/schematics/tsconfig.schematics.json"
+              },
+              {
+                "command": "cp modules/schematics/collection.json dist/modules/schematics/collection.json"
```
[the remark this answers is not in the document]
brandonroberts: Yep. Maybe [cpx](https://www.npmjs.com/package/cpx) or another suggestion?

brandonroberts:
I'll get the copy commands cleaned up, and the example app test target. The Bazel build also broke while upgrading the packages 😞 so I'll need to either fix that, or we just leave it and switch over officially for V10.

angular.json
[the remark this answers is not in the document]
timdeschryver: cpx is good for me, thanks 👍

angular.json
```diff
@@ -81,8 +78,524 @@
...
+                "command": "yarn tsc -p modules/schematics/tsconfig.schematics.json"
+              },
+              {
+                "command": "cpy modules/schematics/collection.json dist/modules/schematics"
+              },
+              {
+                "command": "cpy modules/schematics/migrations/migration.json dist/modules/schematics/migrations"
+              },
+              {
+                "command": "cpy modules/schematics/src/**/files/**/* dist/ --parents"
```
timdeschryver, line 472: One more Windows nit... it seems like we also have to provide a file extension here 
```suggestion
                "command": "cpy modules/schematics/src/**/files/**/*.* dist/ --parents"
```

brandonroberts: Got it

PR #2306, opened 2020-01-04
feature(store): catch action creators being passed to dispatch without () even sooner

alex-okrushko:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

During the recent merge of NgRx to Google I discovered a number of dispatch calls of new action creators without actual calls, e.g. `store.dispatch(fooAction)` instead of `store.dispatch(fooAction())`.

In NgRx we already added the runtime check, but I think we can do even better and catch it at compile time (IDEs will catch this too).

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Errors (warnings) at runtime

## What is the new behavior?

Type Error: Functions are not allowed to be dispatched. Did you forget to call action creator?

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

PR #2536, opened 2020-05-22
feat(effect): catch action creators being returned in effect without ()

alex-okrushko:
## PR Checklist

## PR Type

In https://github.com/ngrx/platform/pull/2306 we are catching actionCreators that are not called in `store.dispatch(actionCreatorNotCalled)` at compile time, however it was still possible to map to `ActionCreator` in the effects by accident: https://twitter.com/DeborahKurata/status/1263612553067819008

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

We are catching at runtime, which is nice, but we can do better.

Closes #

## What is the new behavior?

Catch `ActionCreators` that are not called at compile time:

effect-catching-action-creator

```
[ ] Yes
[x] No
```

Well... somewhat. I needed to adjust the error message and `throwError` operators (or any other one that returns `never`) would not be taken any more.

### DIFF
--- a/modules/effects/spec/types/effect_creator.spec.ts
+++ b/modules/effects/spec/types/effect_creator.spec.ts
@@ -6,6 +6,7 @@ describe('createEffect()', () => {
     code => `
       import { Action } from '@ngrx/store';
       import { createEffect } from '@ngrx/effects';
+      import { createAction } from '@ngrx/store';
       import { of } from 'rxjs';
 
       ${code}`,
@@ -21,7 +22,22 @@ describe('createEffect()', () => {
       expectSnippet(`
         const effect = createEffect(() => of({ foo: 'a' }));
       `).toFail(
-        /Type 'Observable<{ foo: string; }>' is not assignable to type 'Observable<Action> | ((...args: any[]) => Observable<Action>)'./
+        /Type 'Observable<{ foo: string; }>' is not assignable to type 'EffectResult<Action>'./
+      );
+    });
+
+    it('should help with action creator that is not called', () => {
+      // action creator is called with parentheses
+      expectSnippet(`
+      const action = createAction('action without props');
+      const effect = createEffect(() => of(action()));
+      `).toSucceed();
+
+      expectSnippet(`
+      const action = createAction('action without props');
+      const effect = createEffect(() => of(action));
+      `).toFail(
+        /ActionCreator cannot be dispatched. Did you forget to call action creator function/
       );
     });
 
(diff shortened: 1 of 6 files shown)


modules/effects/src/effect_creator.ts
```diff
@@ -10,6 +10,15 @@ import {
 
 type DispatchType<T> = T extends { dispatch: infer U } ? U : true;
 type ObservableType<T, OriginalType> = T extends false ? OriginalType : Action;
+type EffectResult<OT> = Observable<OT> | ((...args: any[]) => Observable<OT>);
+type ConditionallyDisallowActionCreator<DT, Result> = DT extends false
+  ? unknown // If DT (DispatchType is false, then we don't enforce any return types)
+  : Result extends EffectResult<infer OT>
+    ? OT extends ActionCreator
+      ? 'ActionCreator cannot be dispatched. Did you forget to call action creator function?'
```
timdeschryver, line 18:
```suggestion
      ? 'ActionCreator cannot be dispatched. Did you forget to call the action creator function?'
```

modules/effects/spec/types/effect_creator.spec.ts
```diff
@@ -21,7 +22,22 @@ describe('createEffect()', () => {
...
+      expectSnippet(`
+      const action = createAction('action without props');
+      const effect = createEffect(() => of(action()));
+      `).toSucceed();
+
+      expectSnippet(`
+      const action = createAction('action without props');
+      const effect = createEffect(() => of(action));
+      `).toFail(
+        /ActionCreator cannot be dispatched. Did you forget to call action creator function/
```
timdeschryver, line 40:
```suggestion
        /ActionCreator cannot be dispatched. Did you forget to call the action creator function/
```

modules/effects/src/effect_creator.ts
alex-okrushko: I was going for the similar message as we have here: https://github.com/ngrx/platform/blob/6caae70677166e24468f9d69807e3efe0d95bee3/modules/store/src/store.ts#L101

Should I adjust `the` there as well?

timdeschryver: I think so, I think that's how it's written?  🤷‍♂️

alex-okrushko: Thanks Tim! Adjusted in both places.

PR #2547, opened 2020-05-29
fix(component): detect zone.js using instanceof comparison

chrisguttandin:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently the detection of zone.js doesn't work when building a production build with the Angular CLI. The classname of the `NoopNgZone` class doesn't survive the minification.

## What is the new behavior?

The new detection introduced in this PR uses the fact that `NgZone` calls `apply()` on a function that is passed to `runOutsideAngular()`. The `NoopNgZone` doesn't do that.

https://github.com/angular/angular/blob/master/packages/core/src/zone/ng_zone.ts#L226
https://github.com/angular/angular/blob/master/packages/core/src/zone/ng_zone.ts#L381

```
[ ] Yes
[x] No
```

## Other information

Another approach that I tried is to use `runGuarded()`. The `NgZone` will catch errors while the `NoopNgZone` doesn't do that. Therefore something like this will also work.

```ts
try {
    z.runGuarded(() => { throw new Error() });
} catch {
    return false;
}

return true;
```

However, when using zone.js the error will still be logged which is probably not wanted.

I'm not sure if this new test will trigger a change detection run when using zone.js. It might be beneficial to memoize the result. Please let me know if you want me to add that functionality.


brandonroberts:
@chrisguttandin will you rebase on master, and fix the merge conflicts?

modules/component/spec/core/utils/has-zone.spec.ts
```diff
@@ -1,14 +1,20 @@
...
+          fn.apply(null);
+        },
+      } as NgZone)
+    ).toBe(false);
   });
 
   it('should return true if a noop zone is passed', () => {
-    expect(!hasZone(new NoopNgZone() as NgZone)).toBe(true);
+    expect(!hasZone(new ɵNoopNgZone())).toBe(true);
   });
```
alex-okrushko, line 19: I think this one can benefit from refactoring. 
"should return true if a noon zone is passed" - it returns `true` because it's negated with `!hasZone`.

I think this was a result of a change from `isZoneLess` to `hasZone`, which was a good change, however tests were not adjusted accordingly.

modules/component/spec/core/cd-aware/get-change-detection-handling.spec.ts
```diff
@@ -2,8 +2,16 @@ import { getGlobalThis } from '../../../src/core/utils';
 import { Injector } from '@angular/core';
 import { getChangeDetectionHandler } from '../../../src/core/cd-aware';
 
-class NgZone {}
-class NoopNgZone {}
+class NgZone {
+  public runOutsideAngular(fn: Function) {
```
alex-okrushko, line 6: `public` not needed, they are public by default

modules/component/src/core/utils/has-zone.ts
```diff
@@ -14,5 +17,12 @@ import { NgZone } from '@angular/core';
...
 export function hasZone(z: NgZone): boolean {
-  return z.constructor.name !== 'NoopNgZone';
+  let isNgZone = false;
+
+  function fn() {}
+  fn.apply = () => (isNgZone = true);
+
+  z.runOutsideAngular(fn);
+
+  return isNgZone;
```
alex-okrushko, line 27: This feels overly complex.

Would the check for `instanceof` work? Or are none of those classes exported?

Also, `Zone?.current?.name === 'angular'` might work. Needs a bit of testing, since this is just guessing right now.

alex-okrushko:
@chrisguttandin 
Hey Christoph. So you actually ended up dragging me into this 😆

So this is a good enough check:
```typescript
export function hasZone(z: NgZone): boolean {
  return z instanceof NgZone;
}
```

With a little caveat. User might provide their own `Zone` instance 😳 and then this would fail the check.

Alternatively this can be checked:
```typescript
export function hasZone(z: NgZone): boolean {
  return !(z instanceof ɵNoopNgZone);
}
```

However this uses "private" API. 🤷‍♂️ So maybe the first one is good enough.

Here are the actual proper tests for `hasZone`, because otherwise we are just creating fake change-detector really, which I have zero confidence in - we weren't really checking if our function was able to identify it correctly.
```ts
describe('hasZone', () => {
  async function setup({ defaultZone }: { defaultZone: boolean }) {
    @Component({
      selector: 'body',
      template: '<div></div>',
    })
    class NgZoneTestComponent {
      checkNgZone = hasZone(this.ngZone);
      constructor(readonly ngZone: NgZone) {}
    }

    @NgModule({
      declarations: [NgZoneTestComponent],
      exports: [NgZoneTestComponent],
      bootstrap: [NgZoneTestComponent],
      imports: [NoopAnimationsModule],
    })
    class MyAppModule {}

    const platform = getTestBed().platform;
    const moduleRef = defaultZone
      ? await platform.bootstrapModule(MyAppModule)
      : await platform.bootstrapModule(MyAppModule, { ngZone: 'noop' });
    const appRef = moduleRef.injector.get(ApplicationRef);
    const testComp = appRef.components[0].instance;

    return { hasZone: testComp.checkNgZone };
  }

  it('returns false when default zone is used', async () => {
    expect(await setup({ defaultZone: true })).toEqual({ hasZone: true });
  });

  it('returns true when noop zone is chosen', async () => {
    expect(await setup({ defaultZone: false })).toEqual({ hasZone: false });
  });
});
```

chrisguttandin:
Hi @brandonroberts, hi @alex-okrushko,

thanks for your feedback. I just rebased my changes on top of the master branch. That did reduce the number of changes made by this PR as some of the original changes are obsolete now.

@alex-okrushko I chose not to use `ɵNoopNgZone` since it is private, as you mentioned, and I thought it's better to not use it. However my solution is using an implementation detail which is probably equally bad. At least I'm convinced now that `runOutsideAngular()` does not trigger change detection since that is the use case it is designed for.

The only argument I have for checking `runOutsideAngular()` instead of using `instance of NgZone` is that the latter introduces a runtime dependency on `NgZone` even for apps which are not using it.

Anyway besides that I don't have a strong preference for either solution. Should I still make the changes you suggested, Alex? What about using an `instanceof` check with `ɵNoopNgZone`? It's not perfect but maybe the best solution out of the ones we have.

alex-okrushko:
> The only argument I have for checking runOutsideAngular() instead of using instance of NgZone is that the latter introduces a runtime dependency on NgZone even for apps which are not using it.

We are still pulling it in, so `NgZone` check is what I'd prefer ATM.

Thanks for working on it @chrisguttandin 👍

chrisguttandin:
Now that I updated `hasZone()` and added and tests for it, I realized that it isn't used anywhere anymore. 🤷‍♂️

Should I delete `hasZone()` and update `isNgZone()` instead?

alex-okrushko:
Having "dead code" at this point of the library that should be fairly light-height is surprising.

I'd suggest removing `isNgZone` instead.

`hasZone` is in the state that I'm comfortable with ATM - including being properly tested.

BioPhoton:
Hi @AlexOkrushko I really like the test of ngZone over the component! 👍 
This is definitely the best way I saw so far to test zone!

My 2 cent on importing zone to check it is i would also go for a method check. Imho the less dependencies on zone related stuff the better.

alex-okrushko:
Angular is importing `NgZone` regardless, so it will be part of the bundle - if that's the concern.

chrisguttandin:
I updated the code to replace any occurrence of `isNgZone()` with `hasZone()`. I also removed the mocked `NgZone` since it didn't pass the `instanceof` test and the tests seem to run with the real `NgZone` as well.

BioPhoton:
Corner case:
Yes, I was thinking of that situation. I liked the if being zone independent.

But both is fine.

modules/component/src/core/utils/has-zone.ts
```diff
@@ -5,6 +5,9 @@ import { NgZone } from '@angular/core';
  *
  * Determines if the application uses `NgZone` or `NgNoopZone` as ngZone service instance.
  *
+ * It works by detecting if `apply()` is called on a function passed to `runOutsideAngular()`.
+ * The `NgZone` does that while the `NgNoopZone` doesn't do it.
+ *
```
JiaLiPassion, line 10: I am not quite sure about the meaning of this one, just want to say about this comment, since there is a open PR to add `runOutsideAngular()` to `NgNooZone` too https://github.com/angular/angular/pull/37581/files#diff-9d549aa403494fd043a146b8db64f503R410

chrisguttandin: Hi @JiaLiPassion, thanks for making me aware of this. That comment is outdated. I removed it.

This was my initial implementation which used the different internal implementation of `runOutsideAngular()`: https://github.com/ngrx/platform/pull/2547/commits/89ffe67693c5ddd42a6dcadca52b2f4979bd0261#diff-cf11b5203f02c04c7a9bfeacc85c39c8R20-R27

As far as I understand your PR that would still keep working to detect the `NoopNgZone`, right?

JiaLiPassion: So my PR is just make the signature consistent, and sure, this PR will still work!

alex-okrushko: Thanks Jia!

PR #2550, opened 2020-05-31
fix(example-app): fixed the conditionals in the auth system of exampl…

NareshAtnPLUS:
The auth conditionals used in the auth service mistakenly checks the username twice for username and password so i have changed the username variable to password allowing the system to function without any error

### DIFF
--- a/projects/example-app/src/app/auth/services/auth.service.ts
+++ b/projects/example-app/src/app/auth/services/auth.service.ts
@@ -14,7 +14,7 @@ export class AuthService {
      * Simulate a failed login to display the error
      * message for the login form.
      */
-    if (username !== 'test' && username !== 'ngrx') {
+    if (password !== 'test' && username !== 'ngrx') {
       return throwError('Invalid username or password');
     }
 


projects/example-app/src/app/auth/services/auth.service.ts
```diff
@@ -14,7 +14,7 @@ export class AuthService {
      * Simulate a failed login to display the error
      * message for the login form.
      */
-    if (username !== 'test' && username !== 'ngrx') {
+    if (password !== 'test' && username !== 'ngrx') {
```
timdeschryver, line 17: The reason why we don't check for a password is to make the example app easier to run.
If we ask for a password, people need to know it to browse the application app.

NareshAtnPLUS: I got it,that's a good one

timdeschryver: I'm going to close this PR for now, we might revisit later.

PR #2579, opened 2020-06-14
fix(component): removed ivy checks as obsolete

### LINKED ISSUE #2575: Component: fix globalThis handling
**Update**
We can delete it as unused.

As a next improvement, I would start to fix the `globalThis` handling in the component package.

- The easiest way would be we to dot care about SSR
- The second best way would be to just tell them if they want to support IE11 they have to provide a polyfill.
- The third best way would be to include that case in the code.

What is the preferred way I should approach that?

BioPhoton:
Removed unused checks and tests
closes #2575


brandonroberts:
There are merge conflicts that need to be fixed up here also.

PR #2580, opened 2020-06-14
feat(component-store): make CS compatible with ViewEngine

alex-okrushko:
## PR Checklist

## PR Type

Making ComponentStore compatible with ViewEngine. This unfortunately means that we are no longer platform-agnostic.

Also adding integration tests for all kinds of usages that I expect from ComponentStore.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component-store/src/component-store.ts
+++ b/modules/component-store/src/component-store.ts
@@ -17,6 +17,13 @@ import {
   shareReplay,
 } from 'rxjs/operators';
 import { debounceSync } from './debounceSync';
+import {
+  Injectable,
+  OnDestroy,
+  Optional,
+  InjectionToken,
+  Inject,
+} from '@angular/core';
 
 /**
  * Return type of the effect, that behaves differently based on whether the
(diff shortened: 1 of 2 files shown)


modules/component-store/src/component-store.ts
```diff
@@ -38,7 +48,7 @@ export class ComponentStore<T extends object> {
   // Needs to be after destroy$ is declared because it's used in select.
   readonly state$: Observable<T> = this.select((s) => s);
 
-  constructor(defaultState?: T) {
+  constructor(@Optional() @Inject(initialStateToken) defaultState?: T) {
```
brandonroberts, line 51: There's a chance you get an initial state here you're not expecting. Maybe we should look at exposing a provider function that let's you provide the token, and the initial state. Separate from this PR though

alex-okrushko: TBH, I was hesitant to even export this token. The only reason why it's there, it's because providing of the `ComponentStore` would fail due to "_Can't resolve all parameters for ComponentStore (?)_" error. Even with `@Optional()` Angular still tries to resolve it to something.
This token is help with that search for injectable.

It does open up the possibility to provide both the ComponentStore AND the token in Component's `providers` but I'd rather not recommend that - because it could lead to another `ComponentStore` in the downstream Components to be provided with a wrong token. 
The solutions for that are complicated and unnecessary, IMHO.

I'm not exporting this token for the `public_api`.

PR #2584, opened 2020-06-16
fix(entity): remove ComparerStr

LMFinney:
The compare function is used in two places, neither of which expect it to be able to return a string:
The [first caller](https://github.com/ngrx/platform/blob/master/modules/entity/src/sorted_state_adapter.ts#L174) is the [Array prototype sort function](https://www.w3schools.com/js/js_array_sort.asp), and there it "should return a negative, zero, or positive value, depending on the arguments".
The [second caller](https://github.com/ngrx/platform/blob/master/modules/entity/src/sorted_state_adapter.ts#L187) does a numerical comparison with the result.

Even though an id can be a string, the result of a comparison shouldn't be.

This might be a breaking change.

## PR Checklist

- [x ] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[x ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

According to the typing, a sortComparer can return a string, but it would be a mistake to allow that. Even though an id can be a string, the result of a comparison shouldn't be.

Closes #

## What is the new behavior?

Remove the ComparerStr option for Comparar.

```
[x ] Yes
[ ] No
```

Functions that return a string instead of a number would not compile. The user would need to fix the comparer.

### DIFF
--- a/modules/data/spec/entity-metadata/entity-definition.spec.ts
+++ b/modules/data/spec/entity-metadata/entity-definition.spec.ts
@@ -10,7 +10,7 @@ interface NonIdClass {
   something: any;
 }
 
-const sorter = <T>(a: T, b: T) => 'foo';
+const sorter = <T>(a: T, b: T) => 1;
 
 const filter = <T>(entities: T[], pattern?: any) => entities;
 
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -1,7 +1,4 @@
-export type ComparerStr<T> = (a: T, b: T) => string;
-export type ComparerNum<T> = (a: T, b: T) => number;
-
-export type Comparer<T> = ComparerNum<T> | ComparerStr<T>;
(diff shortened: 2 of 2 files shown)


LMFinney:
I can't get build-affected to fail locally, and I can't imagine how my change causes the compilation error that's shown. Could there be another issue?

brandonroberts:
Yea, there is a separate issue. Store should have been built before Entity with the affected commands. I can reproduce it locally by deleting the @ngrx packages under node_modules before running `nx affected --target=build-release` with your branch.

brandonroberts:
Will you squash down to a single commit, and add the breaking change to the commit message?

LMFinney:
I squashed the commits and tweaked the commit message, as requested.

Issue #2572, opened 2020-06-09
Component store race condition issue


artaommahe:
## Minimal reproduction of the bug/regression with instructions:

https://stackblitz.com/edit/ngrx-component-store-race-condition?file=index.ts

in console
```
> field2 false
> state {field1: false, field2: false}
> field2 true
> state {field1: true, field2: true}
> field2 false
> state {field1: true, field2: false}
```

## Expected behavior:
```
> field2 false
> state {field1: false, field2: false}
> state {field1: true, field2: false}
> field2 true
> state {field1: true, field2: true}
```

Expecting having last state and field2 value equal to the store one.

## Versions of NgRx, Angular, Node, affected browser(s) and operating system(s):
Last version of https://github.com/ngrx/platform/blob/master/modules/component-store/src/component-store.ts

## Other information:

We have same issue in our small store lib. Will be glad to know how to fix it propely.
We've added optional `lock` param to update method and move any intermediate updates (that fires before all first update subscribers recieve new state value) to the queue. But with such solution there is an issue with wrong state value right after second update call, it's expecting to be updated but it's not cause update is in queue.

## I would be willing to submit a PR to fix this issue

[ ] Yes (Assistance is provided if you need help submitting a pull request)
[x] No

timdeschryver:
In `@ngrx/store` this is [resolved](https://github.com/ngrx/platform/blob/master/modules/store/src/state.ts#L32-L34) by using a `queueScheduler`.

I've [updated](https://stackblitz.com/edit/ngrx-component-store-race-condition-bmt1f4?file=component-store.ts) the stackblitz example and it seems to fix this issue.

cc @alex-okrushko 

```diff
updater<V>(
    updaterFn: (state: T, value: V) => T
  ): unknown extends V ? () => void : (t: V | Observable<V>) => Subscription {
    return ((observableOrValue?: V | Observable<V>): Subscription => {
      let initializationError: Error | undefined;
      // We can receive either the value or an observable. In case it's a
      // simple value, we'll wrap it with `of` operator to turn it into
      // Observable.
      const observable$ = isObservable(observableOrValue)
        ? observableOrValue
        : of(observableOrValue);
      const subscription = observable$
        .pipe(
+          observeOn(queueScheduler),
          concatMap((value) =>
            this.isInitialized
              ? of(value).pipe(withLatestFrom(this.stateSubject$))
              : // If state was not initialized, we'll throw an error.
                throwError(
[...]

PR #2586, opened 2020-06-17
perf(component-store): push updates to queueScheduler and single selectors to asapSchedulers

alex-okrushko:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes # https://github.com/ngrx/platform/issues/2572

## What is the new behavior?

Updates are now pushed onto queueScheduler and selectors that derive data from state (e.g.`select(state => state.prop)` ) are using `debounceSync()`.

```
[x] Yes
[ ] No
```


modules/component-store/src/component-store.ts
```diff
@@ -92,7 +93,10 @@ export class ComponentStore<T extends object> implements OnDestroy {
         .pipe(
           concatMap((value) =>
             this.isInitialized
-              ? of(value).pipe(withLatestFrom(this.stateSubject$))
+              ? // Push the value into queueScheduler
+                of(value, queueScheduler).pipe(
```
timdeschryver, line 97: Not sure if we should worry about it, but passing a scheduler to `of` is [deprecated](https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/of.ts) and will be removed in v8.

alex-okrushko: Thanks Tim. Didn't know about that.
As of right now - `scheduled` isn't typed correctly for single values. It's focused more on Array, promises and observables, so I had to wrap it with array.

Issue #2569, opened 2020-06-08
Component Store: add ng-add supoort


brandonroberts:
Component store doesn't have anything modules to install, but should be able to be installed using `ng add` similar to how `@ngrx/entity` is supported.

## If accepted, I would be willing to submit a PR for this feature

[ ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

PR #2590, opened 2020-06-22
chore: create a schematic to create a module

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

## Other information

I think having a schematic to create a new module came up in the past, and we all laughed with it.
But with having a look at #2569 I found myself copy pasting files, and replacing the module name - and lookup for stuff that I had missed. So I thought to myself... why not have a schematic to generate a new module.

Well it could look like this...

Is this something we want to have, because it will have to be maintained and that's why I'm hesitant to merge this in. On the other hand, we could just remove it if it causes more problems than it is intended to solve. Just wanted to hear more opinions on this, if we agree to move forward with it, I'll take another look to verify this covers most of the manual work.

WDYT?


brandonroberts:
I definitely like the automating of this. What I had in mind was to use a workspace schematic to do the scaffolding, that might be a little less to maintain, as we already have our own set of schematics for our packages.

timdeschryver:
That should be an easy migration I think. 
Let's look into doing that 🙂
(The reasoning why I didn't do it was to not be tied to the framework, and the upgrade paths)

brandonroberts:
I see. As long as its relatively easy to maintain, I'm good with the solution currently proposed.

timdeschryver:
Putting this on hold until we released v10.
I think there are some thing missing that I forgot to add.

What I meant with hard to maintain is keeping the "blueprint files" updated when we modify something in the existing modules (e.g. the move to NX).

timdeschryver:
😅 It's been a half year since this PR was opened, and I didn't continue to finish it...
I'm going to close this PR but keep the branch locally.

PR #2593, opened 2020-06-25
fix(router-store): also handle routerNavigated action in reducer

dhhyi:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

I wanted to use a Resolver to set a data field on a lazy-loaded route and then later on select the data (which we are mapping with a custom router serializer) from the store. I could trace that the custom serializer was called correctly two times (routerNavigation and routerNavigated actions), but the second serialization was not triggering a store modification. And only the second serialization was containing the data field I had on the lazy-loaded route.

tl;dr: the routerNavigated action has the serialized routerState as payload, but it is not persisted in the store.

## What is the new behavior?

routerNavigated actions are handled by the routerReducer.

```
[ ] Yes
[x] No?
```

## Other information

If you want, I can try to provide a minimal error example. Please tell.

### DIFF
--- a/modules/router-store/spec/integration.spec.ts
+++ b/modules/router-store/spec/integration.spec.ts
@@ -491,7 +491,7 @@ describe('integration spec', () => {
           { type: 'router', event: 'GuardsCheckEnd', url: '/next' },
           { type: 'router', event: 'ResolveStart', url: '/next' },
           { type: 'router', event: 'ResolveEnd', url: '/next' },
-          { type: 'store', state: null }, // ROUTER_NAVIGATED event in the store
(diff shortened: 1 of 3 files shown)


brandonroberts:
Thanks @dhhyi. This is already handled by adjusting the navigation action timing documented here. https://ngrx.io/guide/router-store/configuration#navigation-action-timing

dhhyi:
Hi @brandonroberts,
I was afraid, you were going to point me that way. If we switch to PostActivation timing, something else breaks, because all navigation actions arrive later. You know, enterprise software...

Shouldn't this also work for PreActivation timing?

PR #2606, opened 2020-07-04
feat(component-store): add config for debounce selectors

alex-okrushko:
## PR Checklist

## PR Type

~By removing `debounceSync` from a single-value selector, the selector value now becomes available immediately after the updates.~

~Longer versions:~
- ~updates/setState were always sync~
- ~selectors now fall into two buckets:~
  - ~state selectors like `selector(state => state.foo)` are back to being sync~
  - ~selectors that combine other selectors and/or Observables are still async, scheduled to run into the microtask.~

Introduce `SelectConfig` configuration for selector. Currently it allows to "debounce" selector values until the state is "settled".
By default it will be set to `debounce: false`, as it makes it more predictable (same behaviour as global store) and allows instant reads from the ComponentStore that would return the value synchronously.

### Why this change?
While having both types of selectors async/debounced by default is **more performant**, it was also:
- harder to write test for the ComponentStore(s) (on the user side)
- more confusing for the Developers to understand why the values were not emitted yet, 

This last point became more obvious when I saw effects like:

```ts
private readonly fooEffect = this.effect<void>(
      obs$ => obs$.pipe(
          withLatestFrom(this.selectedId$),
          switchMap(([, clientId]) => {
            return this.httpApiService.makeACall(clientId);
          }),
          ...
```

When `withLatestFrom` is called, the `selectedId$` is not ready yet.

The obvious alternative solution would be to provide an imperative way to read the data from the ComponentStore - that solution is also planned and will follow this PR.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[x] Yes
[ ] No
```

ComponentStore selectors are no longer debounced by default.

### DIFF
--- a/modules/component-store/src/component-store.ts
+++ b/modules/component-store/src/component-store.ts
@@ -36,6 +36,10 @@ export interface EffectReturnFn<T> {
   (t: T | Observable<T>): Subscription;
 }
 
+export interface SelectConfig {
+  debounce?: boolean;
+}
+
 export const initialStateToken = new InjectionToken('ComponentStore InitState');
 
 @Injectable()
@@ -99,7 +103,7 @@ export class ComponentStore<T extends object> implements OnDestroy {
                 )
               : // If state was not initialized, we'll throw an error.
                 throwError(
-                  Error(`${this.constructor.name} has not been initialized`)
+                  new Error(`${this.constructor.name} has not been initialized`)
                 )
           ),
           takeUntil(this.destroy$)
@@ -129,8 +133,10 @@ export class ComponentStore<T extends object> implements OnDestroy {
    * state.
    */
   private initState(state: T): void {
-    this.isInitialized = true;
-    this.stateSubject$.next(state);
+    scheduled([state], queueScheduler).subscribe((s) => {
+      this.isInitialized = true;
+      this.stateSubject$.next(s);
+    });
   }
 
   /**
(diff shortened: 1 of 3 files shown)


modules/component-store/src/component-store.ts
```diff
@@ -149,58 +155,71 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
-    const projector: (...args: any[]) => R = args.pop();
-    if (args.length === 0) {
-      // If projector was the only argument then we'll use map operator.
-      observable$ = this.stateSubject$.pipe(debounceSync(), map(projector));
+  select<
+    O extends Array<Observable<unknown> | SelectConfig | ProjectorFn>,
+    R,
+    ProjectorFn = (...a: unknown[]) => R
+  >(...args: O): Observable<R> {
+    const { observables, projector, config } = processSelectorArgs(args);
```
timdeschryver, line 198: I like that the logic of `processSelectorArgs` is extracted 👍

modules/component-store/src/component-store.ts
```diff
@@ -149,58 +155,71 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
-      // If there are multiple arguments, we're chaining selectors, so we need
+      // If there are multiple arguments, then we're aggregating selectors, so we need
       // to take the combineLatest of them before calling the map function.
-      observable$ = combineLatest(args).pipe(
-        // The most performant way to combine Observables avoiding unnecessary
-        // emissions and projector calls.
-        debounceSync(),
-        map((args: any[]) => projector(...args))
+      observable$ = combineLatest(observables).pipe(
+        config.debounce ? debounceSync() : (source$) => source$,
```
timdeschryver, line 211: What if we would extract this ternary statement?
Not sure if it's worth it tho, I'm thinking that when this line changes, the line in the `if` case would also be changed.

modules/component-store/src/component-store.ts
alex-okrushko: Thanks Tim 😊

modules/component-store/src/component-store.ts
alex-okrushko: Yeah, I couldn't think of a better way to extract it and make it readable/somewhat easier to understand at the same time.

PR #2607, opened 2020-07-06
docs (Testing): Update usage for provideMockActions

leggechr:
defer which is being called under the hood is being upgraded by rxjs and it only emits of the function passed to it returns something. So in order for this to work properly in tests the actions$ observable actually needs to be initialized. Updated documentation to reflect this.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[X] Documentation content changes
[ ] Other... Please describe:
```

After [this commit](https://github.com/ReactiveX/rxjs/commit/1ae937a8e594aef96b93313bb3c68ea910e6f528) the behavior of defer is corrected and it no longer allows factories to return void or undefined. In this case it will not emit. So the previous documentation for usage of provideMockActions will not work correctly. Tests will fail because no observable is provided. 

Closes #

## What is the new behavior?
The updated usage instantiates actions$ so that the observable emits correctly as expected.

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/ngrx.io/content/guide/effects/testing.md
+++ b/projects/ngrx.io/content/guide/effects/testing.md
@@ -10,7 +10,7 @@ An Effect subscribes to the `Actions` Observable to perform side effects.
 <code-example header="my.effects.spec.ts">
 import { provideMockActions } from '@ngrx/effects/testing';
 
-let actions$: Observable&lt;Action&gt;;
+let actions$ = new Observable&lt;Action&gt;();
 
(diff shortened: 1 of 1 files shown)


criskrzysiu:
Shouldn't this change also be included in migration guide and maybe migration schematics?

brandonroberts:
Thanks @leggechr!

PR #2728, opened 2020-09-28
fix(store): prevent unexpected behavior of {} as a props type

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```typescript
// Case 1
const foo = createAction('foo', props<{}>()); // no error
foo(1); // no error
foo('1'); // no error

// Case 2
const foo = createAction('foo', props<{ bar?: string }>()); // ok
foo({}); // ok
```

## What is the new behavior?

```typescript
// Case 1
const foo = createAction('foo', props<{}>()); // error
foo(1); // error
foo('1'); // error

// Case 2
const foo = createAction('foo', props<{ bar?: string }>()); // ok
foo({}); // ok
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/types/action_creator.spec.ts
+++ b/modules/store/spec/types/action_creator.spec.ts
@@ -43,11 +43,19 @@ describe('createAction()', () => {
       );
     });
 
-    it('should not allow ararys', () => {
+    it('should not allow arrays', () => {
       expectSnippet(`
         const foo = createAction('FOO', props<[]>());
       `).toFail(
         /Type 'Props<\[\]>' is not assignable to type '"arrays are not allowed in action creators"'/
+      );
+    });
+
+    it('should not allow empty objects', () => {
+      expectSnippet(`
+        const foo = createAction('FOO', props<{}>());
+      `).toFail(
+        /Type 'Props<\{\}>' is not assignable to type '"empty objects are not allowed in action creators"'/
       );
     });
   });
--- a/modules/store/src/models.ts
+++ b/modules/store/src/models.ts
@@ -61,6 +61,10 @@ export const typePropertyIsNotAllowedMsg =
   'type property is not allowed in action creators';
 type TypePropertyIsNotAllowed = typeof typePropertyIsNotAllowedMsg;
 
+export const emptyObjectsAreNotAllowedMsg =
+  'empty objects are not allowed in action creators';
+type EmptyObjectsAreNotAllowed = typeof emptyObjectsAreNotAllowedMsg;
+
 export type FunctionIsNotAllowed<
   T,
   ErrorMessage extends string
@@ -77,6 +81,8 @@ export type NotAllowedCheck<T extends object> = T extends any[]
   ? ArraysAreNotAllowed
   : T extends { type: any }
   ? TypePropertyIsNotAllowed
+  : keyof T extends never
+  ? EmptyObjectsAreNotAllowed
   : unknown;
 
 /**


timdeschryver (changes requested):
I think this is looking good! 
@alex-okrushko can you also give this a look please?

@markostanimirovic it is missing a test at https://github.com/markostanimirovic/platform/blob/fix-payload-type/modules/store/spec/types/action_creator.spec.ts#L14, could you add one please?

markostanimirovic:
> I think this is looking good!
> @alex-okrushko can you also give this a look please?
> 
> @markostanimirovic it is missing a test at https://github.com/markostanimirovic/platform/blob/fix-payload-type/modules/store/spec/types/action_creator.spec.ts#L14, could you add one please?

@timdeschryver sure

alex-okrushko:
Hey @markostanimirovic 
Didn't you want to switch it to the `Record<string, unknown>` or something similar?

markostanimirovic:
I think this solution is better for minor/patch release, because with `Record` I should change signatures of multiple methods. But, I could refactor it in this PR. What do you think @alex-okrushko ?

alex-okrushko:
> I should change signatures of multiple methods.
@markostanimirovic 
it shouldn't really change that much - it makes it a bit stricter, right?

markostanimirovic:
@alex-okrushko Yes, `Record` will make the props stricter. Array type will be denied by default and `ArraysAreNotAllowed` branch can be removed from `NotAllowedCheck` type guard, as we discussed.

I'll do that 👍

timdeschryver (approved):
LGTM, thanks!

alex-okrushko:
> So, it's not compatible with interfaces 👎

@markostanimirovic I'd argue that they shouldn't be compatible with the interfaces 🙂 Maybe that would be a good breaking (separate) change for the v11! (cc @brandonroberts and @timdeschryver ).

### Why?

`props<Foo>()` does **not** make sense, because that `Foo` would be appended with `type` property, e.g.
`const thingHappened = createAction('[source] a thing happened', props<Foo>())` would the `action` to reducer as `Foo & {type: [source] a thing happened}`.

How do you remove it in the reducer?

```ts
// One might even name it as foo 👇 while in fact it's still an action: foo obj + type prop
on(thingHappened, (state, foo) => ({ ... }))
```

if the `foo` is saved that way - you'll get more than you are expecting... and it would show in redux dev tools.
Ideally one would do this:

```ts
on(thingHappened, (state, {type, ...foo}) => ({ ... }))
```

That way type is detached and the actual `foo` is used further, but I doubt many people do that, and, on top that, it would be easier to wrap it with prop at that point in `props`: `props<{foo: Foo}()>` to avoid all of this:

```ts
on(thingHappened, (state, {foo}) => ({ ... }))
```

alex-okrushko (approved):

markostanimirovic:
> > So, it's not compatible with interfaces 👎
> 
> @markostanimirovic I'd argue that they shouldn't be compatible with the interfaces 🙂 Maybe that would be a good breaking (separate) change for the v11! (cc @brandonroberts and @timdeschryver ).
> 
> ### Why?
> `props<Foo>()` does **not** make sense, because that `Foo` would be appended with `type` property, e.g.
> `const thingHappened = createAction('[source] a thing happened', props<Foo>())` would the `action` to reducer as `Foo & {type: [source] a thing happened}`.
> 
> How do you remove it in the reducer?
> 
> ```ts
> // One might even name it as foo 👇 while in fact it's still an action: foo obj + type prop
> on(thingHappened, (state, foo) => ({ ... }))
> ```
> 
> if the `foo` is saved that way - you'll get more than you are expecting... and it would show in redux dev tools.
> Ideally one would do this:
> 
> ```ts
> on(thingHappened, (state, {type, ...foo}) => ({ ... }))
> ```
> 
> That way type is detached and the actual `foo` is used further, but I doubt many people do that, and, on top that, it would be easier to wrap it with prop at that point in `props`: `props<{foo: Foo}()>` to avoid all of this:
> 
> ```ts
> on(thingHappened, (state, {foo}) => ({ ... }))
> ```

@alex-okrushko Looks good. It will make NGRXers to use the library in the right way 🤘

alex-okrushko:
Btw, this is a breaking change (making it stricter).
cc @brandonroberts

PR #2733, opened 2020-10-07
docs(store): Add documentation for combineReducers

### LINKED ISSUE #1179: Add documentation on combineReducers
Why there is no mention in documentation about combineReducers function? 
Is it deprecated or what?
In redux world it seem to be quite popular approach to break complex reducers into smaller ones. 
I know that there is ActionReducerMap but, if I'm not mistaken it is for combining feature reducers into root reducer. What I want to do is to break my feature reducer into smaller reducers, that handle slice of feature state.

cmd-johnson:
Document the `combineReducers` function as it is not only used internally but can also be useful for
splitting reducers into multiple feature reducers.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `combineReducers` function is undocumented.

Closes #1179 

## What is the new behavior?

Added doc-comments to the `combineReducers` function, along with an example.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/utils.ts
+++ b/modules/store/src/utils.ts
@@ -26,61 +26,21 @@ export function combineReducers<T, V extends Action = Action>(
  *
  * **Example combining two feature reducers into one "root" reducer**
  *
- * Feature A:
  * ```ts
- * export const featureKey = 'featureA';
- * export interface State {
- *   counterA: number;
- * }
- * const initialState: State = {
- *   counterA: 0
- * };
- * export const featureAIncrement = createAction('[FeatureA] increment');
- * export const reducer = createReducer(
- *   initialState,
- *   on(featureAIncrement, state => ({ counterA: state.counterA + 1 }))
- * );
- * ```
- *
- * Feature B:
- * ```ts
- * export const featureKey = 'featureB';
- * export interface State {
- *   counterB: number;
- * }
- * const initialState: State = {
- *   counterB: 0
- * };
- * export const featureBIncrement = createAction('[FeatureB] increment');
- * export const reducer = createReducer(
- *   initialState,
- *   on(featureBIncrement, state => ({ counterB: state.counterB + 1 }))
- * );
- * ```
- *
- * Combining the feature reducers using `combineReducers`:
- * ```ts
- * import * as fromFeatureA from './featureA';
- * import * as fromFeatureB from './featureB';
- *
- * export interface State {
- *   [fromFeatureA.featureKey]: fromFeatureA.State;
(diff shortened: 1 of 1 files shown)


modules/store/src/utils.ts
```diff
@@ -11,6 +11,79 @@ export function combineReducers<T, V extends Action = Action>(
...
+ *
+ * You can also override the initial states of the sub-features:
+ * ```ts
+ * export const reducer = combineReducers({
+ *   [fromFeatureA.featureKey]: fromFeatureA.reducer,
+ *   [fromFeatureB.featureKey]: fromFeatureB.reducer
+ * }, {
+ *   [fromFeatureA.featureKey]: { counterA: 13 },
+ *   [fromFeatureB.featureKey]: { counterB: 37 }
+ * });
```
timdeschryver, line 84: Personally, I think this would be enough to add in the example notes.
Maybe also add in what state tree it results?

```json
{
   featureA: { counterA: 13 },
   featureB: { counterB: 37 }
}
```

What do you think?

timdeschryver (commented):
Hi, thanks for creating the PR, I added a small question.

modules/store/src/utils.ts
cmd-johnson: I think that could be enough if you think that it's obvious that `fromFeatureA` is imported as `import * as fromFeatureA from "./feature_a.ts";`.

I guess in that case you could also get rid of the computed object keys and just use `{ featureA: fromFeatureA.reducer /* , ... */ }` or even just `{ featureA: featureAReducer /* , ... */ }` and do the same for the initial state. That would also make your example state tree redundant.

timdeschryver: I like it, that's probably better.
The main idea is the just trim the example code to what is important for `combineReducers`.

timdeschryver: @cmd-johnson I'm just checking in to see if you want to make these changes.

cmd-johnson: @timdeschryver sorry, I totally forgot about this one. I've pushed the shortened example, what do you think?

timdeschryver (approved):

modules/store/src/utils.ts
timdeschryver: Np! I like the changes, thanks!

brandonroberts (approved):

brandonroberts:
Thanks @cmd-johnson!

PR #2759, opened 2020-10-22
feat(store): use provideMockStore outside of the TestBed

### LINKED ISSUE #2745: Store: add ability to create the Store/MockStore outside of the TestBed
Currently you have to use the TestBed  `provideMockStore()` to get an instance of the `MockStore`. This would declare static providers when using `provideMockStore()` so it can be used with `Injector.create()`

[ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`provideMockStore` can only be used with `TestBed.configureTestingModule`.

Closes #2745

## What is the new behavior?

`provideMockStore` can be used with `TestBed.configureTestingModule` and `Injector.create`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/testing/spec/mock_store.spec.ts
+++ b/modules/store/testing/spec/mock_store.spec.ts
@@ -1,6 +1,11 @@
 import { TestBed, ComponentFixture } from '@angular/core/testing';
 import { skip, take } from 'rxjs/operators';
-import { MockStore, provideMockStore } from '@ngrx/store/testing';
+import {
+  MockSelector,
+  MockState,
+  MockStore,
+  provideMockStore,
+} from '@ngrx/store/testing';
 import {
   Store,
   createSelector,
@@ -9,9 +14,14 @@ import {
   MemoizedSelector,
   createFeatureSelector,
   isNgrxMockEnvironment,
+  INITIAL_STATE,
+  ActionsSubject,
+  INIT,
+  StateObservable,
+  ReducerManager,
 } from '@ngrx/store';
 import { INCREMENT } from '../../spec/fixtures/counter';
-import { Component } from '@angular/core';
+import { Component, Injector } from '@angular/core';
 import { Observable } from 'rxjs';
 import { By } from '@angular/platform-browser';
 
@@ -22,7 +32,7 @@ interface TestAppSchema {
   counter4?: number;
 }
 
-describe('Mock Store', () => {
+describe('Mock Store with TestBed', () => {
   let mockStore: MockStore<TestAppSchema>;
   const initialState = { counter1: 0, counter2: 1, counter4: 3 };
   const stringSelector = 'counter4';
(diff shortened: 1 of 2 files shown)


alex-okrushko:
🤔
I like the idea @markostanimirovic 👍
Let me think about it a bit to see if it can be further iterated.

modules/store/testing/src/testing.ts
```diff
@@ -18,22 +22,115 @@ export interface MockStoreConfig<T> {
...
+ *       providers: [
+ *         provideMockStore({ initialState: { counter: 0 } }),
+ *       ],
+ *     });
+ *     store = injector.get(MockStore);
+ *   });
+ * });
+ * ```
+ */
 export function provideMockStore<T = any>(
```
alex-okrushko, line 76: How about the following?

This is close to what you had, but I want to make sure that `useExisting` is used, and it's a bit more compacted.

```suggestion
export function provideMockStore<T = any>(
  config: MockStoreConfig<T> = {}
): Provider[] {
  setNgrxMockEnvironment(true);
  return [
    { provide: ActionsSubject, useFactory: () => new ActionsSubject(), deps: []},
    { provide: MockState, useFactory: () => new MockState<T>(), deps: []},
    { provide: MockReducerManager, useFactory: () => new MockReducerManager(), deps: []},
    { provide: INITIAL_STATE, useValue: config.initialState || {} },
    { provide: MOCK_SELECTORS, useValue: config.selectors },
    { provide: StateObservable, useExisting: MockState },
    { provide: ReducerManager, useExisting: MockReducerManager },
    {
      provide: MockStore,
      useFactory: mockStoreFactory,
      deps: [
        MockState,
        ActionsSubject,
        ReducerManager,
        INITIAL_STATE,
        MOCK_SELECTORS,
      ],
    },
    { provide: Store, useExisting: MockStore },
  ];
}
```

modules/store/testing/src/testing.ts
```diff
@@ -18,22 +22,115 @@ export interface MockStoreConfig<T> {
...
+): MockStore<T> {
+  return new MockStore(
+    mockState,
+    actionsSubject,
+    reducerManager,
+    initialState,
+    mockSelectors
+  );
+}
+
```
alex-okrushko, line 133: Instead of going through the `Injector` every time, maybe let's create the `getMockStore` or `createMockStore` function:

```suggestion
export function getMockStore<T>(
  config: MockStoreConfig<T> = {}
): MockStore<T> {
  return Injector.create({ providers: [ provideMockStore(config) ]}).get(MockStore);
}
```

modules/store/testing/spec/mock_store.spec.ts
```diff
@@ -281,6 +291,101 @@ describe('Mock Store', () => {
   });
 });
 
+describe('Mock Store with Injector', () => {
+  let injector: Injector;
+  const initialState = { counter: 0 };
+  const mockSelector: MockSelector = { selector: 'counter', value: 10 };
```
alex-okrushko, line 297: You'll need to change to this one, if you'll be using `mockStore = getMockStore({ initialState, selectors: [mockSelector] });` instead.

```suggestion
   const mockSelector = { selector: 'counter', value: 10 } as const;
```

alex-okrushko (changes requested):
I really like where this is going!

modules/store/testing/src/testing.ts
markostanimirovic: @alex-okrushko 
Done. I assumed that `StateObservable` needs `useExisting: MockState`, but I added `useFactory: () => new MockState()` in order to keep the same behavior as previous `{ provide: StateObservable, useClass: MockState }`.

About return type, I kept `(ValueProvider | ExistingProvider | FactoryProvider)[]`, because if `Provider[]` is used, following code will not compile:
```typescript
injector = Injector.create({ providers: provideMockStore(...) });
``` 
because `Provider` is not compatible with `StaticProvider` and `(ValueProvider | ExistingProvider | FactoryProvider)` is compatible with both `Provider` and `StaticProvider`.

modules/store/testing/src/testing.ts
markostanimirovic: Done

modules/store/testing/spec/mock_store.spec.ts
markostanimirovic: done

modules/store/testing/src/testing.ts
```diff
@@ -18,22 +23,140 @@ export interface MockStoreConfig<T> {
...
-): Provider[] {
+): (ValueProvider | ExistingProvider | FactoryProvider)[] {
   setNgrxMockEnvironment(true);
   return [
-    ActionsSubject,
-    MockState,
-    MockStore,
+    {
+      provide: ActionsSubject,
+      useFactory: () => new ActionsSubject(),
```
brandonroberts, line 85: Why not use `useClass` here?

```suggestion
      useClass: ActionsSubject,
```

modules/store/testing/src/testing.ts
```diff
@@ -18,22 +23,140 @@ export interface MockStoreConfig<T> {
...
+      deps: [],
+    },
+    { provide: MockState, useFactory: () => new MockState<T>(), deps: [] },
```
brandonroberts, line 88: Same here with `useClass`

modules/store/testing/src/testing.ts
```diff
@@ -18,22 +23,140 @@ export interface MockStoreConfig<T> {
...
+    {
+      provide: MockReducerManager,
+      useFactory: () => new MockReducerManager(),
```
brandonroberts, line 91: Same with `useClass`

brandonroberts (changes requested):

brandonroberts:
This would be a separate effort, but what about creating a factory function for creating a `Store` also?  I did this as a POC for using NgRx Store in React a while back.

https://github.com/brandonroberts/react-ngrx-poc/blob/master/src/store.tsx

cc: @alex-okrushko

modules/store/testing/src/testing.ts
markostanimirovic: @brandonroberts Because class providers are different for `TestBed.configureTestingModule` and `Injector.create`. `TestBed.configureTestingModule` expects `ClassProvider`. On the other hand, `Injector.create` expects `StaticClassProvider`. However, they both accept `FactoryProvider`.

brandonroberts: Ok, what about just `{ provide: ActionsSubject, deps: [] }`? I know it's compatible with the Injector, and should be compatible with the TestBed also

alex-okrushko: For some reason the `ActionsSubject` is `undefined` in this case.

alex-okrushko (approved):
Thanks a lot @markostanimirovic 👍

brandonroberts (approved):

PR #2760, opened 2020-10-25
feat(effects): concatLatestFrom operator

### LINKED ISSUE #2222: RFC: Change behavior of ofType in Effects to be more consistent when combining observables

```ts
actions.pipe(
  ofType('SOME_ACTION')
)
```
and
```ts
actions.pipe(
  ofType('SOME_ACTION'),
)
```

```ts
actions.pipe(
  ofType('SOME_ACTION'),
  someMap(action =>
    of(action).pipe(
    )
)
```

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

An operator which does this for consumers of NgRx is nice to have, but does not currently exist.

Closes #2222 

## What is the new behavior?

```typescript
concatLatestFrom((action) =>
      ),
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/spec/concat_latest_from.spec.ts
+++ b/modules/effects/spec/concat_latest_from.spec.ts
@@ -1,0 +1,112 @@
+import { Observable, of } from 'rxjs';
+import { skipWhile } from 'rxjs/operators';
+import { hot } from 'jasmine-marbles';
+import { concatLatestFrom } from '../src/concat_latest_from';
+
+describe('concatLatestFrom', () => {
+  describe('no triggering value appears in source', () => {
+    it('should not evaluate the array', () => {
+      let evaluated = false;
+      const toBeLazilyEvaluated = () => {
+        evaluated = true;
+        return of(4);
+      };
+      const input$: Observable<number> = hot('-a-b-', { a: 1, b: 2 });
+      const numbers$: Observable<[number, number]> = input$.pipe(
+        skipWhile((value) => value < 3),
+        concatLatestFrom(() => [toBeLazilyEvaluated()])
+      );
+      expect(numbers$).toBeObservable(hot('----'));
+      expect(evaluated).toBe(false);
+    });
+    it('should not evaluate the observable', () => {
+      let evaluated = false;
+      const toBeLazilyEvaluated = () => {
+        evaluated = true;
+        return of(4);
+      };
+      const input$: Observable<number> = hot('-a-b-', { a: 1, b: 2 });
+      const numbers$: Observable<[number, number]> = input$.pipe(
+        skipWhile((value) => value < 3),
+        concatLatestFrom(() => toBeLazilyEvaluated())
+      );
+      expect(numbers$).toBeObservable(hot('----'));
+      expect(evaluated).toBe(false);
+    });
+  });
+  describe('a triggering value appears in source', () => {
+    it('should evaluate the array of observables', () => {
+      let evaluated = false;
+      const toBeLazilyEvaluated = () => {
+        evaluated = true;
+        return of(4);
+      };
+      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
+      const numbers$: Observable<[number, number]> = input$.pipe(
+        skipWhile((value) => value < 3),
+        concatLatestFrom(() => [toBeLazilyEvaluated()])
+      );
+      expect(numbers$).toBeObservable(hot('-----d', { d: [3, 4] }));
+      expect(evaluated).toBe(true);
+    });
+    it('should evaluate the observable', () => {
+      let evaluated = false;
+      const toBeLazilyEvaluated = () => {
+        evaluated = true;
+        return of(4);
+      };
+      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
+      const numbers$: Observable<[number, number]> = input$.pipe(
+        skipWhile((value) => value < 3),
+        concatLatestFrom(() => toBeLazilyEvaluated())
+      );
+      expect(numbers$).toBeObservable(hot('-----d', { d: [3, 4] }));
+      expect(evaluated).toBe(true);
+    });
+  });
+  describe('multiple triggering values appear in source', () => {
+    it('evaluates the array of observables', () => {
+      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
+      const numbers$: Observable<[number, string]> = input$.pipe(
+        concatLatestFrom(() => [of('eval')])
+      );
+      expect(numbers$).toBeObservable(
+        hot('-a-b-c', { a: [1, 'eval'], b: [2, 'eval'], c: [3, 'eval'] })
+      );
+    });
+    it('uses incoming value', () => {
+      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
+      const numbers$: Observable<[number, string]> = input$.pipe(
+        concatLatestFrom((num) => [of(num + ' eval')])
+      );
+      expect(numbers$).toBeObservable(
+        hot('-a-b-c', { a: [1, '1 eval'], b: [2, '2 eval'], c: [3, '3 eval'] })
+      );
+    });
+  });
+  describe('evaluates multiple observables', () => {
+    it('gets values from both observable in specific order', () => {
+      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
+      const numbers$: Observable<[number, string, string]> = input$.pipe(
+        skipWhile((value) => value < 3),
+        concatLatestFrom(() => [of('one'), of('two')])
+      );
+      expect(numbers$).toBeObservable(hot('-----d', { d: [3, 'one', 'two'] }));
+    });
+    it('can use the value passed through source observable', () => {
+      const input$: Observable<number> = hot('-a-b-c-d', {
+        a: 1,
+        b: 2,
+        c: 3,
+        d: 4,
+      });
+      const numbers$: Observable<[number, string, string]> = input$.pipe(
+        skipWhile((value) => value < 3),
+        concatLatestFrom((num) => [of(num + ' one'), of(num + ' two')])
+      );
+      expect(numbers$).toBeObservable(
+        hot('-----c-d', { c: [3, '3 one', '3 two'], d: [4, '4 one', '4 two'] })
+      );
+    });
+  });
+});
--- a/modules/effects/src/concat_latest_from.ts
+++ b/modules/effects/src/concat_latest_from.ts
@@ -1,0 +1,37 @@
+import { Observable, ObservedValueOf, of, OperatorFunction, pipe } from 'rxjs';
+import { concatMap, withLatestFrom } from 'rxjs/operators';
+
+// The array overload is needed first because we want to maintain the proper order in the resulting tuple
+export function concatLatestFrom<T extends Observable<unknown>[], V>(
+  observablesFactory: (value: V) => [...T]
+): OperatorFunction<V, [V, ...{ [i in keyof T]: ObservedValueOf<T[i]> }]>;
+export function concatLatestFrom<T extends Observable<unknown>, V>(
+  observableFactory: (value: V) => T
+): OperatorFunction<V, [V, ObservedValueOf<T>]>;
+/**
+ * 'concatLatestFrom' combines the source value
+ * and the last available value from a lazily evaluated Observable
+ * in a new array
+ */
+export function concatLatestFrom<
+  T extends Observable<unknown>[] | Observable<unknown>,
+  V,
+  R = [
+    V,
+    ...(T extends Observable<unknown>[]
+      ? { [i in keyof T]: ObservedValueOf<T[i]> }
+      : [ObservedValueOf<T>])
+  ]
+>(observablesFactory: (value: V) => T): OperatorFunction<V, R> {
+  return pipe(
+    concatMap((value) => {
+      const observables = observablesFactory(value);
+      const observablesAsArray = Array.isArray(observables)
+        ? observables
+        : [observables];
+      return of(value).pipe(
+        withLatestFrom(...observablesAsArray)
+      ) as Observable<R>;
+    })
+  );
+}
--- a/modules/effects/src/index.ts
+++ b/modules/effects/src/index.ts
@@ -27,3 +27,4 @@ export {
   OnInitEffects,
 } from './lifecycle_hooks';
 export { USER_PROVIDED_EFFECTS } from './tokens';
+export { concatLatestFrom } from './concat_latest_from';
--- a/projects/example-app/src/app/core/effects/router.effects.ts
+++ b/projects/example-app/src/app/core/effects/router.effects.ts
@@ -1,10 +1,9 @@
 import { Injectable } from '@angular/core';
 import { Title } from '@angular/platform-browser';
 
-import { of } from 'rxjs';
-import { concatMap, map, tap, withLatestFrom } from 'rxjs/operators';
+import { map, tap } from 'rxjs/operators';
 
-import { Actions, createEffect, ofType } from '@ngrx/effects';
+import { Actions, concatLatestFrom, createEffect, ofType } from '@ngrx/effects';
 import { Store } from '@ngrx/store';
 import { routerNavigatedAction } from '@ngrx/router-store';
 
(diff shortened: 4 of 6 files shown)


alex-okrushko (changes requested):
@david-shortman 
Thanks for the PR. 👍

- Could you also provide the support for a single select without wrapping it into tuple?
- as for the naming (typically the hardest part 🙂), what do you think about `withLazyLatestFrom`?

david-shortman:
@alex-okrushko Have achieved allowing single observable or array of observables to be returned by factory function.

I kept mulling the name over in my head, and I wonder if `concatLatestFrom` accurately represents what the operator does?

modules/effects/src/concat_latest_from.ts
```diff
@@ -0,0 +1,80 @@
...
+import { concatMap, withLatestFrom } from 'rxjs/operators';
+
+type TypeOfObservable<T> = T extends Observable<infer U> ? U : never;
+
+/**
+ * @description
+ * RxJS operator that lazily evaluates a list of Observables.
+ *
+ * @param observablesFactory A function which returns an `Observable[]`.
+ * @returns Returns an `OperatorFunction` with ??? (idk how to describe this)
```
david-shortman, line 11: deeply struggled with how to describe the return type lol. anyone, please advise

david-shortman:
@alex-okrushko benlesh over at RxJs gave a pretty good suggestion to use `defer` instead within `withLatestFrom`: https://github.com/ReactiveX/rxjs/issues/5845#issuecomment-720648178

Wondering if the better direction would be updating the docs to use this simpler way to delay running `store.select`?

srnec:
What about creating just a wrapper function?


```ts
export function withLazyLatestFrom(...observables: Observabkle): (source: Observable) => Observable {
    return (source: Observable) => {
        return source.pipe(concatMap((value) => of(value).pipe(withLatestFrom(...observables))));
    };
}
```

Typedefs omitted because I am not able to type it properly in order to infer all the types properly.

david-shortman:
@srnec your solution does not delay the evaluation of the input observable. The argument to the operator will be immediately evaluated unless you use a lambda and evaluate the lambda later.

alex-okrushko:
@david-shortman 
I don't think `defer` helps here.
It works same as without it. https://stackblitz.com/edit/angular-mpsnhs?file=src%2Fapp%2Feffects.ts

`defer` delays the evaluation until subscription, not until the value is pushed through the pipe - at least that's my understanding.

david-shortman:
@alex-okrushko ah, ok. I'll see what the RxJs folks think.

Side-note, what would be the best path forward for this PR be TS version-wise? Do I need to target it to pre-V4, or just let this hang around while?

alex-okrushko:
> @alex-okrushko ah, ok. I'll see what the RxJs folks think.
> 
> Side-note, what would be the best path forward for this PR be TS version-wise? Do I need to target it to pre-V4, or just let this hang around while?

TS 3.9, if possible.
TS 4.0 is fine. Angular will have 4.0.x for v11, and we can release this operator with v11 as well.

david-shortman:
Created the types for TS 3.9: https://github.com/ngrx/platform/pull/2798

david-shortman:
a

david-shortman:
This is now ready for review again @alex-okrushko. Not sure why unrelated test suites are failing...

brandonroberts:
@david-shortman the API docs aren't showing in the preview. Will you take a look?

modules/effects/src/concat_latest_from.ts
```diff
@@ -0,0 +1,38 @@
...
+import { concatMap, withLatestFrom } from 'rxjs/operators';
+
+type TypeOfObservable<T> = T extends Observable<infer U> ? U : never;
+
+/**
+ * 'concatLatestFrom' combines the source value
+ * and the last available value from a lazily evaluated Observable
+ * in a new array
+ */
+export function concatLatestFrom<T extends Observable<unknown>[], V>(
```
david-shortman, line 11: @brandonroberts can you advise on what would be causing this to not render in the preview?

not only does it not show the comment, but the other overrides are not listed.

Screen Shot 2020-12-29 at 20 25 23

david-shortman: or @alex-okrushko, if you know what could be causing this, am looking for help

brandonroberts:
@david-shortman I'll take a look this week

brandonroberts:
I'm not sure why this isn't showing in the PR preview for the API docs, but the changes LGTM and I think its ok to move forward with. I'll wait on @alex-okrushko or @timdeschryver to provide another review/approval before merging.

brandonroberts (approved):

modules/effects/src/concat_latest_from.ts
alex-okrushko: @david-shortman, you might what to add the JSDoc comment just before the actual implementation (instead of the first overload).

alex-okrushko, line 11: I'd add the comment here, explaining why overload `extends Observable<unknown>[]` is used first.

E.g. It took me some time to figure that out. I was combining everything into just a single function, but that was breaking the proper arguments types order.

```suggestion
// The array overload is needed first because we want to maintain the proper order in the resulting tuple
export function concatLatestFrom<T extends Observable<unknown>[], V>(
```

modules/effects/src/concat_latest_from.ts
```diff
@@ -0,0 +1,38 @@
+import { Observable, of, OperatorFunction } from 'rxjs';
+import { concatMap, withLatestFrom } from 'rxjs/operators';
+
+type TypeOfObservable<T> = T extends Observable<infer U> ? U : never;
```
alex-okrushko, line 4: you can use `ObservedValueOf` from `rxjs` instead.

```suggestion
```

modules/effects/src/concat_latest_from.ts
```diff
@@ -0,0 +1,38 @@
...
+  }) as OperatorFunction<
+    V,
+    [
+      V,
+      ...(T extends Observable<unknown>[]
+        ? { [i in keyof T]: TypeOfObservable<T[i]> }
+        : [TypeOfObservable<T>])
+    ]
+  >;
+}
```
alex-okrushko, line 38: That casting at the end looks very overwhelming 😬
How about specifying the return type of the function, and then do a gentler cast? 🙂
 
```suggestion
export function concatLatestFrom<
  T extends Observable<unknown>[] | Observable<unknown>,
  V,
  R = [
    V,
    ...(T extends Observable<unknown>[]
      ? { [i in keyof T]: ObservedValueOf<T[i]> }
      : [ObservedValueOf<T>])
  ]
>(observablesFactory: (value: V) => T): OperatorFunction<V, R> {
  return concatMap((value) => {
    const observables = observablesFactory(value);
    const observablesAsArray = Array.isArray(observables)
      ? observables
      : [observables];
    return of(value).pipe(
      withLatestFrom(...observablesAsArray)
    ) as Observable<R>;
  });
}
```

modules/effects/spec/concat_latest_from.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      const numbers$ = hot('-a-b-c-', { a: 1, b: 2, c: 3 }).pipe(
+        skipWhile((value) => value < 3),
+        concatLatestFrom(() => toBeLazilyEvaluated())
+      );
+
+      expect(numbers$).toBeObservable(hot('-----d', { d: [3, 4] }));
+      expect(evaluated).toBe(true);
+    });
+  });
+});
```
alex-okrushko, line 71: There weren't enough tests to validate a lot of scenarios. To make sure that the implementation works, I added a bunch of them, so that we can have a higher confidence in the implementation, and to make sure that it won't be broken in the future:

```suggestion
describe('concatLatestFrom', () => {
  describe('no triggering value appears in source', () => {
    it('should not evaluate the array', () => {
      let evaluated = false;
      const toBeLazilyEvaluated = () => {
        evaluated = true;
        return of(4);
      };

      const input$: Observable<number> = hot('-a-b-', { a: 1, b: 2 });
      const numbers$: Observable<[number, number]> = input$.pipe(
        skipWhile((value) => value < 3),
        concatLatestFrom(() => [toBeLazilyEvaluated()])
      );

      expect(numbers$).toBeObservable(hot('----'));
      expect(evaluated).toBe(false);
    });
    it('should not evaluate the observable', () => {
      let evaluated = false;
      const toBeLazilyEvaluated = () => {
        evaluated = true;
        return of(4);
      };

      const input$: Observable<number> = hot('-a-b-', { a: 1, b: 2 });
      const numbers$: Observable<[number, number]> = input$.pipe(
        skipWhile((value) => value < 3),
        concatLatestFrom(() => toBeLazilyEvaluated())
      );

      expect(numbers$).toBeObservable(hot('----'));
      expect(evaluated).toBe(false);
    });
  });
  describe('a triggering value appears in source', () => {
    it('should evaluate the array of observables', () => {
      let evaluated = false;
      const toBeLazilyEvaluated = () => {
        evaluated = true;
        return of(4);
      };

      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
      const numbers$: Observable<[number, number]> = input$.pipe(
        skipWhile((value) => value < 3),
        concatLatestFrom(() => [toBeLazilyEvaluated()])
      );

      expect(numbers$).toBeObservable(hot('-----d', { d: [3, 4] }));
      expect(evaluated).toBe(true);
    });
    it('should evaluate the observable', () => {
      let evaluated = false;
      const toBeLazilyEvaluated = () => {
        evaluated = true;
        return of(4);
      };

      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
      const numbers$: Observable<[number, number]> = input$.pipe(
        skipWhile((value) => value < 3),
        concatLatestFrom(() => toBeLazilyEvaluated())
      );

      expect(numbers$).toBeObservable(hot('-----d', { d: [3, 4] }));
      expect(evaluated).toBe(true);
    });
  });

  describe('multiple triggering values appear in source', () => {
    it('evaluates the array of observables', () => {
      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
      const numbers$: Observable<[number, string]> = input$.pipe(
        concatLatestFrom(() => [of('eval')])
      );

      expect(numbers$).toBeObservable(
        hot('-a-b-c', { a: [1, 'eval'], b: [2, 'eval'], c: [3, 'eval'] })
      );
    });

    it('uses incoming value', () => {
      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
      const numbers$: Observable<[number, string]> = input$.pipe(
        concatLatestFrom((num) => [of(num + ' eval')])
      );

      expect(numbers$).toBeObservable(
        hot('-a-b-c', { a: [1, '1 eval'], b: [2, '2 eval'], c: [3, '3 eval'] })
      );
    });
  });

  describe('evaluates multiple observables', () => {
    it('gets values from both observable in specific order', () => {
      const input$: Observable<number> = hot('-a-b-c-', { a: 1, b: 2, c: 3 });
      const numbers$: Observable<[number, string, string]> = input$.pipe(
        skipWhile((value) => value < 3),
        concatLatestFrom(() => [of('one'), of('two')])
      );

      expect(numbers$).toBeObservable(hot('-----d', { d: [3, 'one', 'two'] }));
    });
    it('can use the value passed through source observable', () => {
      const input$: Observable<number> = hot('-a-b-c-d', {
        a: 1,
        b: 2,
        c: 3,
        d: 4,
      });

      const numbers$: Observable<[number, string, string]> = input$.pipe(
        skipWhile((value) => value < 3),
        concatLatestFrom((num) => [of(num + ' one'), of(num + ' two')])
      );

      expect(numbers$).toBeObservable(
        hot('-----c-d', { c: [3, '3 one', '3 two'], d: [4, '4 one', '4 two'] })
      );
    });
  });
});

```

alex-okrushko (changes requested):
@david-shortman 

Fantastic work, David! Thank you so much!

Sorry that it took a bit longer than expected.
Let's address these comments and we'd be able to get it into v11, which hopefully can be release within next week or two.

@brandonroberts  😉

modules/effects/src/concat_latest_from.ts
```diff
@@ -0,0 +1,38 @@
...
+export function concatLatestFrom<T extends Observable<unknown>, V>(
+  observableFactory: (value: V) => T
+): OperatorFunction<V, [V, TypeOfObservable<T>]>;
+export function concatLatestFrom<
+  T extends Observable<unknown>[] | Observable<unknown>,
+  V
+>(
+  observablesFactory: (value: V) => T extends Observable<unknown>[] ? [...T] : T
+) {
+  return concatMap((value) => {
```
alex-okrushko, line 23: Btw, when I write custom operators, I typically do 
```ts
function myOperator(foo: string): (source: Observable<A>) => Observable<b> {
   return source.pipe(
   .... // here goes the implementation
   )
}
```

I don't quite remember why that's the way to do it. E.g. @cartant also uses it in his article: https://ncjamieson.com/managing-operator-state/

Might be a good idea to do it this way anyway. (in the meantime, I'll try to find out why it's done so).

e.g. here's what I also arrived at for this operator: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgeQEYGcCmUBuBDVAG0wBoUNt8jMBJAOzAFcYy0tdMATANT0MczIAZmQhC4AXzhCoEEHADkUAB4ArdAoDcAKFCRYiOAGMIdI3hgBZPGDIB3YDAAWAGQuZ0MAGKz5UmXKKKuoA9BBg2BbQGjraISFwAIJQUHgAnnAQONiEEHiccMDocHSYXFzSwFCeAHTamMr68EKMZjDApsam5jBuMB7evgA8ACpwDf10nMVslATEQ60A1nQQdnQAfADaALpk3BsAFNpwmRS48x5eeEYw0GkAXHCH+PyYT9wAlHAAvBtwWxqQJGOzg8Tg6DAUEw+WAdAA5hCIHA7JhjHg6HB4Zh4M40TBGGBiJkoJxsJlxJhiCBMHQYOhtJ8nsdTugIIwoEZ3uR2FQFgdGb9-rMLtQhltuGQgTUkFsANKFTFLTBpMRwEFPEXZHh8ATCUbynb-CRGnQNJrSVq3DqYkxmCx9AY+OSjcbKSbTHlzMXLVbrDb7I4nM68y7oa63e7M14CD7fP7qxnMtkcrma858zBDA7x4UZy7iyVejg6t76kYbU1xABU1ZO1cUdp6js8zpACi6IFQcI8cDxSNTaJjmHrcAxBX7hDwnjH+GAU+ocGH0l8Y7gU4AXvOMpgY+4ClrLqO4WvSnYxyl0vWQvVGtBmlb2p0mw73K3hsGxhNaZ7Dz66CsaybHAAA+xaZosAF+psuwkMGkrBgASr8ALBqcCGnKc0qHF+7o-sUySpGkkGAf6aGYQA-IY8qKnAyqquIGrgVwvBlkIBpykakjkacTxbCCnzBjs2hBqcED5tQ4Y3HcUCPM8w5xkKiZMs8Kactyf78hsubgQWiH-AgwbQgSUCYocalprpYo5kK5EWZgNRgMAEQsphpwvlYNiHC8uqYDphluW5dozuJoaSShoXesQUmRrJPlvJ8OiBZhwXwJForRYk6CEekKE5WkNRFPlhzpZm6CCclblUSVEnRWOMy1bQDDMCR0HbDsFWVbxAKlWGOxJZVxkcpiYjxQInyOc5mCuV1DjOC2gxyIc0q9ZJWX5Z1yXfNOVkLPpA1uRIm1wIl2gSEAA

david-shortman: Since writing this PR, I have spent more time on the RxJs docs site and [found that they give the same recommendation](https://rxjs.dev/guide/operators). So I'll swap it over.
Thanks for the playground link 👍.

modules/effects/src/concat_latest_from.ts
david-shortman: Yeah I was definitely at the beginning of understanding these more complex typings when I wrote this. Thanks!

alex-okrushko: Yeah, there are some interesting types here 🙂, e.g. `{ [i in keyof T]: ObservedValueOf<T[i]> }`.
What sucks is that TS doesn't let you assign it to a type and spread it elsewhere, even though it knows that it's still an `Array`-like type - very frustrating 😐

modules/effects/src/concat_latest_from.ts
david-shortman: Revising my statement- RxJs now provides the `pipe` function to independently and more quickly create custom operators. No need to spend the extra characters referring to a source and writing details that expose a deeper understanding than necessary of the nature of observables.

modules/effects/spec/concat_latest_from.spec.ts
david-shortman: added 👍

alex-okrushko (approved):

alex-okrushko:
Awesome work, @david-shortman 👍

Issue #1224, opened 2018-08-02
Custom RxJS operator for safe calling the service in the Effect


alex-okrushko:
Many Effects are used to call services, and most of the time they should result in `SUCCESS` or `ERROR` actions. 
From the code that I review it's quite often that error case is either:
- not handled
- handled incorrectly:
  - error handling is put onto main Observable, instead of piped from service call
  - error is handled, but without wrapping Error action into observable

Even people who are familiar with all of these mistakes still make them sometimes.

Custom operator might help here, e.g.:
```ts
const effect$: Observable<FooSuccess|FooError> = actions$.pipe(
  ofType<Foo>(FOO),
  concatMap(a => this.service.fetch(a.bar).pipe(
    map(resp => new FooSuccess(resp)),
    catchError(e => of(new FooError()))
  )),
);
```

could turn into
```ts
const effecOp$: Observable<FooSuccess|FooError> = actions$.pipe(
  ofType<Foo>(FOO),
  safeCall(
    action => this.service.fetch(a.bar).pipe(map(resp => new FooSuccess(resp))),
    error => new FooError()),
)
```

where the `safeCall` (not crazy about the naming) would be that safety operator.

the operator itself could be implemented like this:
```ts
function safeCall<A extends Action, T extends Action, V extends Action>(
  callFn: (action: A) => Observable<T>,
  errorHandlingFn: (error: any) => V
): (source: Observable<A>) => Observable<T | V> {
  return (source: Observable<A>) =>
    source.pipe(
      concatMap(a =>
        callFn(a).pipe(catchError(err => of(errorHandlingFn(err))))
      )
    );
}
```

Thoughts?
Should we add this to ngrx?

## If accepted, I would be willing to submit a PR for this feature

[ x ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

alex-okrushko:
Ah, here is the stackblitz: https://stackblitz.com/edit/typescript-snyfkg

brandonroberts:
@MikeRyanDev had a similar idea for a pipeable operator like this one. I think it has some value for repetitive mapping and error handling.

MikeRyanDev:
Here's the proposal I submitted to redux-observable: https://github.com/redux-observable/redux-observable/issues/457

They accepted it, I just haven't gotten around to implementing it in both locations. I welcome anyone to make the PR.

alex-okrushko:
[...]

PR #2763, opened 2020-10-27
feat(component-store): add tapResponse operator

alex-okrushko:
## PR Checklist

## PR Type

Adds a simple wrapper around `tap` that handles success and error + `catchError(() => EMPTY)`.

The was some confusion as to whether `catchError(() => EMPTY)` was needed when `tap({error: (error) => handleError(error)})` was used.

To avoid such confusion and remove a bit of boilerplate around not letting effect collapse, I suggest to have this very thin operator.

If not used, this operator is 100% tree-shakable from the bundle.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component-store/src/handle-response.ts
+++ b/modules/component-store/src/handle-response.ts
@@ -1,0 +1,39 @@
+import { EMPTY, Observable } from 'rxjs';
+
+import { catchError, tap } from 'rxjs/operators';
+
+/**
+ * Handles the response in ComponentStore effects in a safe way, without
+ * additional boilerplate.
+ * It enforces that error case is handled and that the effect would still be
+ * running should an error occur.
+ *
+ * Takes an optional third argument for a `complete` callback.
+ *
+ * ```typescript
+ * readonly dismissedAlerts = this.effect<Alert>(alert$ => {
+ *  return alert$.pipe(
+ *      concatMap(
+ *          (alert) => this.alertsService.dismissAlert(alert).pipe(
+ *              handleResponse(
+ *                 (dismissedAlert) => this.alertDismissed(dismissedAlert),
+ *                 (error) => this.logError(error),
+ *              ))));
+ *   });
+ * ```
+ */
+export function handleResponse<T>(
+  nextFn: (next: T) => void,
+  errorFn: (error: unknown) => void,
+  completeFn?: () => void
+): (source: Observable<T>) => Observable<T> {
+  return (source) =>
+    source.pipe(
+      tap({
+        next: nextFn,
+        error: errorFn,
+        complete: completeFn,
+      }),
+      catchError(() => EMPTY)
+    );
+}
--- a/modules/component-store/src/map-response.ts
+++ b/modules/component-store/src/map-response.ts
@@ -3,9 +3,9 @@ import { EMPTY, Observable } from 'rxjs';
 import { catchError, tap } from 'rxjs/operators';
 
 /**
- * Handles the response in ComponentStore effects in a safe way, without
+ * Maps the response in ComponentStore effects in a safe way, without
  * additional boilerplate.
- * It enforces that error case is handled and that the effect would still be
+ * It enforces that the error case is handled and that the effect would still be
  * running should an error occur.
  *
  * Takes an optional third argument for a `complete` callback.
@@ -15,14 +15,14 @@ import { catchError, tap } from 'rxjs/operators';
  *  return alert$.pipe(
  *      concatMap(
  *          (alert) => this.alertsService.dismissAlert(alert).pipe(
- *              handleResponse(
+ *              mapResponse(
  *                 (dismissedAlert) => this.alertDismissed(dismissedAlert),
  *                 (error) => this.logError(error),
  *              ))));
  *   });
  * ```
  */
-export function handleResponse<T>(
+export function mapResponse<T>(
   nextFn: (next: T) => void,
   errorFn: (error: unknown) => void,
   completeFn?: () => void
--- a/modules/component-store/src/index.ts
+++ b/modules/component-store/src/index.ts
@@ -1,1 +1,2 @@
 export * from './component-store';
+export * from './handle-response';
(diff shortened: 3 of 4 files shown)


alex-okrushko:
FYI, we've been using this operator within Firebase Console for some time already 🙂 so this is pushing it upstream.

Also, if there are any other preferences around the naming - I'm all ears 🙂

modules/component-store/src/handle-response.ts
```diff
@@ -0,0 +1,39 @@
+import { EMPTY, Observable } from 'rxjs';
+
+import { catchError, tap } from 'rxjs/operators';
+
+/**
+ * Handles the response in ComponentStore effects in a safe way, without
+ * additional boilerplate.
+ * It enforces that error case is handled and that the effect would still be
```
timdeschryver, line 8:
```suggestion
 * It enforces that the error case is handled and that the effect would still be
```

timdeschryver (commented):
About the naming, what about `mapResponse`?
Reasoning, what is `handle`, and `map` feels RxJS-y.

timdeschryver (approved):

markostanimirovic:
> About the naming, what about `mapResponse`?
> Reasoning, what is `handle`, and `map` feels RxJS-y.

@timdeschryver 
All RxJS `__map` operators are used to transform emitted value and return a new one, but `handleResponse` is used to perform side effects 🤔

timdeschryver:
That's a good point @markostanimirovic

modules/component-store/src/map-response.ts
```diff
@@ -0,0 +1,39 @@
...
+ *                 (error) => this.logError(error),
+ *              ))));
+ *   });
+ * ```
+ */
+export function mapResponse<T>(
+  nextFn: (next: T) => void,
+  errorFn: (error: unknown) => void,
+  completeFn?: () => void
+): (source: Observable<T>) => Observable<T> {
```
markostanimirovic, line 29: I suggest to add two signatures for this function, similar to `tap` operator in order to make it more flexible.

```typescript
export function tap<T>(next?: (x: T) => void, error?: (e: any) => void, complete?: () => void): MonoTypeOperatorFunction<T>;
export function tap<T>(observer: PartialObserver<T>): MonoTypeOperatorFunction<T>;
```

alex-okrushko: The goal is to not make it flexible. I want to force to handle the error.

alex-okrushko:
> That's a good point @markostanimirovic

@timdeschryver 

so `tapResponse`? 😅

timdeschryver:
I was thinking the same thing @alex-okrushko 
On the other hand we're still in an `effect` so everything is a side effect.
I would find `mapResponse`  easier to remember and explain, because you would probably say you map the response to effect X or Y 😅

Sorry for not helping, naming things is hard... 
I'm fine with the 3 options (handle, map, tap) we have, if you have a preference I'm happy to use that terminology.
If you don't have a preference we might have a problem, what about a quick poll?

beeman:
My 2 cents, I think I like `tapResponse` best. With `mapResponse` I would expect that something would be returned, like with `map()` or `mapTo()`.

modules/component-store/src/map-response.ts
markostanimirovic: Okay, I suggested that based on the implementation. According to the operator implementation too, I suggest the name **safeTap** or something similar. 🙂

alex-okrushko: Thanks @markostanimirovic that's a good suggestion.
We had similar discussions about `safeCall` operator for `@ngrx/effect` and eventually called it `mapToAction` and then renamed to `act`: https://github.com/ngrx/platform/issues/1224

Looking at the poll `tapResponse` appears to be the clear winner.

@beeman, technically this operator still `maps`, but only the error case :)

markostanimirovic: @alex-okrushko 
Naming - the hardest thing 😄

I would also choose `tapResponse` between `tapResponse`, `mapResponse` and `handleResponse`.

`Response` as a part of name confused me a little bit, because it associates me with HTTP response 🙂

timdeschryver (approved):

brandonroberts (approved):

Ash-kosakyan:
Hi, can you observe this solution for tapResponse operator
`typescript
export function tapResponse<T>(
  nextFn: (next: T) => void,
  errorFn: (error: unknown) => void,
  completeFn?: () => void
): (source: Observable<T>) => Observable<T> {
  return (source) =>
    source.pipe(
      tap(nextFn),
      catchError((error) => {
        errorFn(error);
        return EMPTY;
      }),
      finalize(completeFn || noop)
    );
}
`

MichaliEverson:
Return error for the action is being observed as watcher to react to choices

alex-okrushko:
Hi @Ash-kosakyan 
What does your solution improve?

MichaliEverson:
yes nd no binary security with key words as in shortcut or jump like a skip
expect if skip then jus means your tring to enter prohibited area

Ash-kosakyan:
Hi @alex-okrushko 
The difference: "finalize" is working every time, although observable can be errored
```typescript
 /**
  * Effects
  * */
 readonly loadAll = this.effect((origin$: Observable<void>) =>
   origin$.pipe(
     tap(() => this.setLoading(true)),
     concatMap(() =>
       this.modelService.getAll().pipe(
         tap((models) => {
           this.setAll(models);
         }),
         catchError((error: HttpErrorResponse) => {
           this.setError(error.error);
           return EMPTY;
         }),
         finalize(() => {
           this.setLoading(false);
         })
       )
     )
   )
 );
```

alex-okrushko:
@Ash-kosakyan - that's true. That also means that `complete` could bring the unexpected behavior to `tapResponse`.

Thanks for your input! 
I think for this one `tapReponse` should be as close to `tap` as possible, so we'll keep the current implementation.

Thanks again.

PR #2764, opened 2020-10-27
perf(router-store): optimize selectQueryParams, selectQueryParam and selectFragment selectors

### LINKED ISSUE #2758: Router Store: Add selector to retrieve param from anywhere in route tree
If you have a route parameter in the route tree, it could be at different levels in the hierarchy based on how the parameter is used.

Example:

```ts
[
  {
    path: 'myRoute',
    children: [
      {
        path: ':accountId',
        component: AnotherComponent,
      }
    ]
  },
  { path: ':accountId', component: MyComponent }
]
```

This selector would traverse the route tree, and return the first occurrence of the route param, instead of the bottom-most route in the tree.

Selector projector
```ts
let route = router.state.root;
let param = null;
while (route.firstChild) {
  route = route.firstChild;
  if (route.params[paramName]) {
    param = route.params[paramName];
    break;
  }
}
return param;
```

This would allow you to reuse the selector and not have to worry about where the parameter is.

[ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: Performance improvement
```

`selectQueryParams` - returns query params from the last router state node
`selectQueryParam` - returns a query param from the last router state node
`selectFragment` - returns the fragment from the last router state node

## What is the new behavior?

Optional chaining is introduced for all selectors.

`selectQueryParams` - returns query params from `routerState.root`
`selectQueryParam` - returns a query param from `routerState.root`
`selectFragment` - returns the fragment from `routerState.root`

```
[x] Yes
[ ] No
```

Actually, this can be a breaking change in test files (for `selectQueryParams`, `selectQueryParam` and `selectFragment` selectors) if mock router state is defined in the wrong way. But when it is used with Angular router, there are no breaking changes.

### DIFF
--- a/modules/router-store/src/router_selectors.ts
+++ b/modules/router-store/src/router_selectors.ts
@@ -4,33 +4,31 @@ import { RouterReducerState } from './reducer';
 
 export function getSelectors<V>(
   selectState: (state: V) => RouterReducerState<any>
-): RouterStateSelectors<V>;
-export function getSelectors<V>(
-  selectState: (state: V) => RouterReducerState<any>
 ): RouterStateSelectors<V> {
   const selectRouterState = createSelector(
     selectState,
     (router) => router && router.state
   );
-  const selectCurrentRoute = createSelector(
+  const selectRootRoute = createSelector(
     selectRouterState,
-    (routerState) => {
-      if (!routerState) {
-        return undefined;
-      }
-      let route = routerState.root;
-      while (route.firstChild) {
-        route = route.firstChild;
-      }
-      return route;
-    }
+    (routerState) => routerState && routerState.root
   );
+  const selectCurrentRoute = createSelector(selectRootRoute, (rootRoute) => {
+    if (!rootRoute) {
+      return undefined;
+    }
+    let route = rootRoute;
+    while (route.firstChild) {
+      route = route.firstChild;
+    }
+    return route;
+  });
   const selectFragment = createSelector(
-    selectCurrentRoute,
+    selectRootRoute,
     (route) => route && route.fragment
   );
   const selectQueryParams = createSelector(
-    selectCurrentRoute,
+    selectRootRoute,
     (route) => route && route.queryParams
   );
   const selectQueryParam = (param: string) =>
--- a/modules/router-store/spec/router_selectors.spec.ts
+++ b/modules/router-store/spec/router_selectors.spec.ts
@@ -12,11 +12,15 @@ const mockData = {
       url: [],
       outlet: 'primary',
       routeConfig: null,
-      queryParams: {},
-      queryParamMap: {
-        params: {},
+      queryParams: {
+        ref: 'ngrx.io',
       },
-      fragment: null,
+      queryParamMap: {
+        params: {
+          ref: 'ngrx.io',
+        },
+      },
+      fragment: 'test-fragment',
       firstChild: {
         params: {},
         paramMap: {
@@ -34,10 +38,12 @@ const mockData = {
           path: 'login',
         },
         queryParams: {
-          id: 3,
+          ref: 'ngrx.io',
         },
         queryParamMap: {
-          params: {},
+          params: {
+            ref: 'ngrx.io',
+          },
         },
         firstChild: {
           params: {
@@ -72,7 +78,7 @@ const mockData = {
           fragment: 'test-fragment',
           children: [],
         },
-        fragment: null,
+        fragment: 'test-fragment',
         children: [],
       },
       children: [
@@ -92,10 +98,15 @@ const mockData = {
           routeConfig: {
             path: 'login',
           },
-          queryParams: {},
-          queryParamMap: {
-            params: {},
+          queryParams: {
+            ref: 'ngrx.io',
           },
+          queryParamMap: {
+            params: {
+              ref: 'ngrx.io',
+            },
+          },
+          fragment: 'test-fragment',
           children: [],
         },
       ],
@@ -104,6 +115,7 @@ const mockData = {
   },
   navigationId: 1,
 };
+
 describe('Router State Selectors', () => {
   describe('Composed Selectors', () => {
     interface State {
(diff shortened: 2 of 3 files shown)


markostanimirovic:
@alex-okrushko @brandonroberts @timdeschryver 
I observed the same issue with build on master branch. Steps:
1. `rm -rf node_modules/.cache`
2. `yarn ng build router-store`

brandonroberts:
> @alex-okrushko @brandonroberts @timdeschryver
> I observed the same issue with build on master branch. Steps:
> 
> 1. `rm -rf node_modules/`
> 2. `yarn`
> 3. `yarn ng build router-store`

@markostanimirovic you have to use the `--with-deps` option to build router-store, as it depends on the `store` package being built first. There may be an issue with `build-affected` in CI when using `--parallel`

modules/router-store/src/router_selectors.ts
```diff
@@ -2,52 +2,60 @@ import { createSelector } from '@ngrx/store';
...
   );
   const selectQueryParam = (param: string) =>
-    createSelector(selectQueryParams, (params) => params && params[param]);
+    createSelector(selectQueryParams, (params) => params?.[param]);
   const selectRouteParams = createSelector(
     selectCurrentRoute,
-    (route) => route && route.params
+    (route) => route?.params
   );
   const selectRouteParam = (param: string) =>
```
brandonroberts, line 37: This shouldn't modify the existing selector, but add a new selector, such as `selectParamFromRouteTree` or something similar, so there's a difference between selecting the current one, and the first occurrence of the parameter.

I think this should be in a separate PR from the perf changes also

brandonroberts (changes requested):

modules/router-store/src/router_selectors.ts
markostanimirovic: Okay, I'll create a separate PR for new selector.

timdeschryver (commented):
Out of curiosity, I see this was first a perf change but how does this optimize the router selectors?
Doesn't the Optional Chaining operator compile to the same code?

Since this is a refactor, I would also revert the changes made to the `spec` to be sure we didn't break something. Or perhaps create a second test case for the root with the added fragment and params?

markostanimirovic:
> Out of curiosity, I see this was first a perf change but how does this optimize the router selectors?
> Doesn't the Optional Chaining operator compile to the same code?

> Since this is a refactor, I would also revert the changes made to the `spec` to be sure we didn't break something. Or perhaps create a second test case for the root with the added fragment and params?

@timdeschryver 
Of course, optional chaining is not optimization, this is just an improvement to use new features :)

Optimized are three selectors: `selectFragment`, `selectQueryParams` and `selectQueryParam` to return the fragment/queryParam(s) from `routerState.root` instead of looping and returning fragment/queryParam(s) from the last router state tree node. I wrote this in PR description :)

About the breaking changes: this should not be a breaking change with Angular router, because fragment and query parameters are the same on root and any other child router tree node.

markostanimirovic:
I can add tests with `RouterTestingModule` to be sure that everything is working fine.

timdeschryver:
> Optimized are three selectors: selectFragment, selectQueryParams and selectQueryParam to return the fragment/queryParam(s) from routerState.root instead of looping and returning fragment/queryParam(s) from the last router state tree node. I wrote this in PR description :)

My bad, I didn't read the description 😅.

But this change would also mean that the behavior is different (root vs current route), which you also pointed out.

So imho, we should either mark this as a breaking change or leave the implementation as is and create root selectors.
My opinion would be the second one, because these selectors are probably used in projects.
Thoughts?

 
> I can add tests with RouterTestingModule to be sure that everything is working fine.

My comment about the spec was that originally it was verifying the child route, while it now validates the root route.
I was expecting the current tests to fail, but since I didn't read the PR's description I didn't know that this was intended.

markostanimirovic:
> So imho, we should either mark this as a breaking change or leave the implementation as is and create root selectors.
> My opinion would be the second one, because these selectors are probably used in projects.
> Thoughts?

Sure, let's wait for the opinion of others, and when the final decision is made, I'll fix this PR :)

alex-okrushko:
So it is a breaking change, right?
For non-breaking changes the tests shouldn't change - just added.

> Of course, optional chaining is not optimization, this is just an improvement to use new features :)

Optional chaining actually increases the generated code size. Unless we are reducing both sources code size AND generated code size, I wouldn't suggest to change it.

markostanimirovic:
@alex-okrushko 

> So it is a breaking change, right?
> For non-breaking changes the tests shouldn't change - just added.

`mockData` object from `router_selectors.spec.ts` should represent Angular Router state. However, it is wrongly defined, because `queryParams` and `fragment` properties should be the same in `root` as well as in `firstChild`s down to the router state tree. That's why I changed it, to be the real mock of Angular Router state. I can revert the tests. Old ones will work with new `mockData` object too.

Anyway, I can mark this PR as breaking change and change its title to `feature` or `performance`.
Or simply close it, if we don't want `selectQueryParams` and `selectFragment` optimization. What do you think? 🙂
 
> Optional chaining actually increases the generated code size. Unless we are reducing both sources code size AND generated code size, I wouldn't suggest to change it.

Okay, I will revert optional chaining 👍

brandonroberts:
@markostanimirovic What's left on this PR? Removal of the optional chaining?

markostanimirovic:
@brandonroberts 

Tim commented:

> So imho, we should either mark this as a breaking change or leave the implementation as is and create root selectors.
> My opinion would be the second one, because these selectors are probably used in projects.
> Thoughts?

And I wanted to hear other thoughts on this and update this PR accordingly 🙂

My opinion is that we don't need new selectors for selecting fragment and query params from root, because with Angular router and default serializer `root.fragment` is equal to `root.firstChild.fragment` (the same is for query params).

Btw, optional chaining is removed now 🙂

cc @timdeschryver

brandonroberts:
I don't think we need new selectors either, but this shouldn't be a breaking change functionally because they are equivalent as you stated. I'd still mark it as a breaking change just so people are aware of it.

brandonroberts (approved):
Changes LGTM. Just needs breaking changes to be added to commit message when merging

timdeschryver:
We should also add a note to the [migration docs](https://github.com/ngrx/platform/blob/master/projects/ngrx.io/content/guide/migration/v10.md)

markostanimirovic:
@brandonroberts @timdeschryver
I changed PR type from `refactor` to `perf`. Also, note is added to the migration guide 🙂

timdeschryver (approved):
Thanks for adding it to the migration guide

PR #2798, opened 2020-11-25
feat(effects): add lazy concatLatestFrom operator for TS 3.9

### LINKED ISSUE #2222: RFC: Change behavior of ofType in Effects to be more consistent when combining observables

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

An operator which does this for consumers of NgRx would be nice to have, but does not currently exist.

Closes #2222 

## What is the new behavior?

```typescript
concatLatestFrom((action) =>
      ),
```

```
[ ] Yes
[x] No
```

## Other information

- I have opened a [separate PR](https://github.com/ngrx/platform/pull/2760) with much simpler types for TS 4.X

### DIFF
--- a/modules/effects/spec/concat_latest_from_operator.spec.ts
+++ b/modules/effects/spec/concat_latest_from_operator.spec.ts
@@ -1,0 +1,71 @@
+import { of } from 'rxjs';
+import { skipWhile } from 'rxjs/operators';
+import { hot } from 'jasmine-marbles';
+import { concatLatestFrom } from '../src/concat_latest_from_operator';
+
+describe('concatLatestFrom', () => {
+  describe('no triggering value appears in source', () => {
+    it('should not evaluate the array', () => {
+      let evaluated = false;
+      const toBeLazilyEvaluated = () => {
+        evaluated = true;
+        return of(4);
+      };
+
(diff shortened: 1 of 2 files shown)


david-shortman:
Fun(?) note:

In the series of function overload declarations, the declaration that comes last MUST come last.

I was losing my sanity trying to figure out why the types weren't working when it was put at the top. Apparently, when TS attempts to check if a function call matches a signature, it would refuse to believe an array could be narrowed to any of the tuples in the signatures.
Moving the function signature that uses no array for the observable factory to the bottom fixed this but I don't know why.

david-shortman:
I'm seeing test failures that aren't for the two brand new files, but it might be bc I was out of date with master? Seems unlikely but closing this one and opening https://github.com/ngrx/platform/pull/2799

PR #2794, opened 2020-11-22
chore: update to Angular 11

timdeschryver:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Closes #2787
Closes #2742

## Does this PR introduce a breaking change?

```
[ ] Yes
[ ] No
```

## Other information

About done with `ng-update`

PR #2803, opened 2020-12-01
refactor(store): variadic tuple types for on

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

Not that I know of 

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/reducer_creator.ts
+++ b/modules/store/src/reducer_creator.ts
@@ -11,163 +11,6 @@ export interface OnReducer<S, C extends ActionCreator[]> {
   (state: S, action: ActionType<C[number]>): S;
 }
 
-export function on<C1 extends ActionCreator, S>(
-  creator1: C1,
-  reducer: OnReducer<S, [C1]>
-): On<S>;
-export function on<C1 extends ActionCreator, C2 extends ActionCreator, S>(
-  creator1: C1,
-  creator2: C2,
-  reducer: OnReducer<S, [C1, C2]>
-): On<S>;
-export function on<
-  C1 extends ActionCreator,
-  C2 extends ActionCreator,
-  C3 extends ActionCreator,
-  S
->(
-  creator1: C1,
-  creator2: C2,
-  creator3: C3,
-  reducer: OnReducer<S, [C1, C2, C3]>
-): On<S>;
-export function on<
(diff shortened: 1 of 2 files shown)


timdeschryver:
Oh right, first #2794 has to be merged in 😅

modules/store/src/reducer_creator.ts
```diff
@@ -11,177 +11,25 @@ export interface OnReducer<S, C extends ActionCreator[]> {
...
-  ...args: (ActionCreator | Function)[]
-): { reducer: Function; types: string[] } {
-  const reducer = args.pop() as Function;
+export function on<
+  C extends ActionCreator[],
+  S,
+  OR extends OnReducer<S, C> = OnReducer<S, C>,
+  AR extends ActionReducer<S> = ActionReducer<S>
+>(...args: [...C, OR]): On<S> {
+  const reducer = (args.pop() as Function) as AR;
```
markostanimirovic, line 32:
```suggestion
export function on<
  Creators extends ActionCreator[],
  State,
  Reducer extends OnReducer<State, Creators>
>(...args: [...creators: Creators, reducer: Reducer]): On<State> {
  const reducer = (args.pop() as Function) as ActionReducer<State>;
```

markostanimirovic: @timdeschryver
Btw, Prettier should be updated to [v2.1.0](https://prettier.io/blog/2020/08/24/2.1.0.html#support-for-typescript-40) or higher to support new TypeScript 4 features 🙂

timdeschryver: Good suggestion, thanks!
I'll update prettier in #2794

alex-okrushko (approved):
Nice!

brandonroberts (approved):

bbaia:
👍 
Can we do the same for `createSelector`?

timdeschryver:
@bbaia Yes, we're looking into it, but we're hitting against some restrictions of TS.
It seems like TS will resolve this in one of the upcoming versions, once that's done we can revisit `createSelector`.

PR #2821, opened 2020-12-11
feat(store): add object-style StoreModule.forFeature overload

### LINKED ISSUE #2809: Add overload support for StoreModule.forFeature({ name, reducer })
### Proposal

I propose a new overload signature of `StoreModule.forFeature()`. 

```ts
StoreModule.forFeature({ name, reducer });
```

### Motivation

The interface `{name, reducer}` is compatible with the [`Slice`](https://redux-toolkit.js.org/api/createSlice#return-value) type from [Redux Toolkit](https://redux-toolkit.js.org/) package.

Redux Toolkit is a set of best practices for using Redux. I think it's better to have interoperability with that.
With Redux Toolkit, a typical small example will be like below:

```ts
import { createSlice } from '@reduxjs/tooklit';
import { StoreModule } from '@ngrx/store';

const counterSlice = createSlice({
  name: 'counter',
  initialState: { count: 0 },
  reducers: {
    increment: (state) => { state.count++; }
  }
});

@NgModule({
  imports: [
    StoreModule.forFeature(counterSlice.name, counterSlice.reducer)
  ]
})
export class CounterFeatureStoreModule {}
```

A slice has its `name` and `reducer`. Its concept is similar to NgRx's "feature". So I want to link them seamlessly. The following is an example when `forFeature()`supports `{name, reducer}` option.

```ts
import { createSlice } from '@reduxjs/tooklit';
import { StoreModule } from '@ngrx/store';

const counterSlice = createSlice({
  name: 'counter',
  initialState: { count: 0 },
  reducers: {
    increment: (state) => { state.count++; }
  }
});

@NgModule({
  imports: [
    StoreModule.forFeature(counterSlice)
  ]
})
export class CounterFeatureStoreModule {}
```

[ ] No

lacolaco:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2809 

## What is the new behavior?

Adding a new overload for `StoreModule.forFeature()`;

```ts
  static forFeature<T, V extends Action = Action>(
    slice: FeatureSlice<T, V>,
    config?: StoreConfig<T, V> | InjectionToken<StoreConfig<T, V>>
  ): ModuleWithProviders<StoreFeatureModule>;
```

```
[ ] Yes
[x] No
```

## Other information

I think my code is not elegant because added overload has a very different shape. I'd like to be commented on codes.

### DIFF
--- a/modules/store/spec/modules.spec.ts
+++ b/modules/store/spec/modules.spec.ts
@@ -216,4 +216,34 @@ describe(`Store Modules`, () => {
       });
     });
   });
+
+  describe(`: Slice-Like paramter`, () => {
+    @NgModule({
+      imports: [
+        StoreModule.forFeature({ name: 'a', reducer: featureAReducer }),
+      ],
+    })
+    class FeatureAModule {}
+
+    @NgModule({
+      imports: [StoreModule.forRoot({}), FeatureAModule],
+    })
+    class RootModule {}
+
+    beforeEach(() => {
+      TestBed.configureTestingModule({
+        imports: [RootModule],
+      });
+
+      store = TestBed.inject(Store);
+    });
+
+    it('should set up a feature state', () => {
+      store.pipe(take(1)).subscribe((state: State) => {
+        expect(state).toEqual({
+          a: 5,
+        } as State);
+      });
+    });
+  });
 });
--- a/modules/store/spec/modules.spec.ts
+++ b/modules/store/spec/modules.spec.ts
@@ -217,7 +217,7 @@ describe(`Store Modules`, () => {
     });
   });
 
-  describe(`: Slice-Like paramter`, () => {
+  describe(`: With slice object`, () => {
     @NgModule({
       imports: [
         StoreModule.forFeature({ name: 'a', reducer: featureAReducer }),
--- a/modules/store/src/store_module.ts
+++ b/modules/store/src/store_module.ts
@@ -114,6 +114,14 @@ export interface RootStoreConfig<T, V extends Action = Action>
   runtimeChecks?: Partial<RuntimeChecks>;
 }
 
+/**
+ * A pair of reducer and name for the feature.
+ */
+export interface FeatureSlice<T, V extends Action = Action> {
+  name: string;
+  reducer: ActionReducer<T, V> | InjectionToken<ActionReducer<T, V>>;
+}
+
 @NgModule({})
 export class StoreModule {
   static forRoot<T, V extends Action = Action>(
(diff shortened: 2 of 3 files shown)


modules/store/src/store_module.ts
```diff
@@ -114,6 +114,14 @@ export interface RootStoreConfig<T, V extends Action = Action>
   runtimeChecks?: Partial<RuntimeChecks>;
 }
 
+/**
+ * A pair of reducer and name for the feature.
+ */
+export interface FeatureSlice<T, V extends Action = Action> {
```
lacolaco, line 120: Shouldn't it use the word "slice" here?

modules/store/spec/modules.spec.ts
```diff
@@ -216,4 +216,34 @@ describe(`Store Modules`, () => {
       });
     });
   });
+
+  describe(`: Slice-Like paramter`, () => {
```
lacolaco, line 220: Should it be tested with the real RTK slice as an integration spec?

modules/store/src/store_module.ts
```diff
@@ -114,6 +114,14 @@ export interface RootStoreConfig<T, V extends Action = Action>
   runtimeChecks?: Partial<RuntimeChecks>;
 }
 
+/**
+ * A pair of reducer and name for the feature.
```
lacolaco, line 118: To Be Fixed

modules/store/src/store_module.ts
```diff
@@ -114,6 +114,14 @@ export interface RootStoreConfig<T, V extends Action = Action>
...
+ */
+export interface FeatureSlice<T, V extends Action = Action> {
+  name: string;
+  reducer: ActionReducer<T, V> | InjectionToken<ActionReducer<T, V>>;
```
lacolaco, line 122: If this type is named "Slice", Is it necessary to accept an InjectionToken as `reducer`?

modules/store/spec/modules.spec.ts
timdeschryver: I don't think that that is needed.

brandonroberts: I agree

brandonroberts: I agree

modules/store/src/store_module.ts
timdeschryver: No, since the user will provide both options as a "slice"

modules/store/src/store_module.ts
timdeschryver: What do you mean with the comments `Shouldn't it use the word "slice" here?` and this one?
I think it's fine, perhaps the following would read a bit easier.

```suggestion
 * An object with the name and the reducer for the feature.
```

modules/store/spec/modules.spec.ts
timdeschryver: That being said, perhaps we could add a recipe to use NgRx with RTK in the future, as a separate PR.

modules/store/src/store_module.ts
lacolaco: My concern is that possibly "slice" is a word that is only meaningful in the RTK context. If not, I can feel the `FeatureSlice` type is a good name. But then, maybe I should leave a comment to address the question: "What is the Slice?".

timdeschryver (approved):

modules/store/src/store_module.ts
timdeschryver: It doesn't bother me to be honest.
I think some sections in our docs (or in articles) are also using the term slice.
If it isn't clear, we can always add it in the future.

brandonroberts (approved):

brandonroberts:
Thanks @lacolaco!

alex-okrushko:
@lacolaco @brandonroberts 

we'd need to rollback this PR, it's breaking g3 - and any codebase that is still using ViewEngine (the cause itself is my guess).

This is what I have:

```
error TS100: Error during template compile of 'StoreModule'
  Function calls are not supported in decorators but 'StoreModule' was called.
```

alex-okrushko:
@lacolaco Would you be able to take a look at it?
The potential solutions are fairly well described here: https://medium.com/angular-in-depth/solving-aot-error-in-ngrx-function-calls-are-not-supported-in-decorators-5c337381457a

Issue #1781, opened 2019-04-22
Lazy Loading Feature Module with Guards causing Infinite Navigation Loop


BuairtRi:
I have a problem where, if the initial request I make to my application is to a route that requires you to be authenticated, and you're *not* authenticated - the guard shoots you to the login page, which is inside the Authentication feature module - fairly run-of-the-mill logic.

The problem is, the router store is running code which - before the feature module routes are loaded/recognized - attempts to navigate me back to the route. Basically, at some point the Route State is being updated, but the route url is an empty string because the routes haven't been recognized yet.

The routes are set up like this:

```typescript
export const appRoutes : Routes = [
    {
        path         : 'authentication',
        component    : ExternalLayoutComponent,
        loadChildren : './features/authentication/authentication.module#AuthenticationModule'
    },
    {
        path         : 'users',
        canActivate  : [ RoleGuard ],
        loadChildren : './features/users/users.module#UsersModule',
        data         : {
            roles : [ 'Administrator' ]
        }
    },
    {
        path       : '**',
        redirectTo : 'authentication'
    }
];
```

```typescript
export const userRoutes : Routes = [
    {
        path       : '',
        redirectTo : 'list'
    },
    {
        path      : 'list',
        component : UsersListComponent
    },
    {
        path      : 'profile/:id',
        component : ProfileComponent
    }
];
```

```typescript
export const authenticationRoutes : Routes = [
    {
        path      : '',
        component : LoginComponent,
        pathMatch : 'full'
    },
    {
        path      : 'login',
        component : LoginComponent
    }
];
```

So if I request _'/users/list'_, everything works okay until the guard determines the route can't be activated. 

The guard redirects the user to the login page via a call like: `router.navigate(['/authentication/login'])`

The router goes out, loads the module and gets to the "_RouteConfigLoadEnded_" router event. At this point, the Store calls update-reducers to load the store state for that feature module. However, the URL in the route is an empty string at this point, rather than '/authentication/login`. The update-reducers action causes the router-state reducer to execute (I think), which causes the store to update the routerstate URL value to "".

The state update triggers this code in the StoreRouterConnectingModule to execute:

```typescript
private navigateIfNeeded(
    routerStoreState: RouterReducerState,
    storeState: any
  ): void {
    if (!routerStoreState || !routerStoreState.state) {
      return;
    }
    if (this.trigger === RouterTrigger.ROUTER) {
      return;
    }
    if (this.lastEvent instanceof NavigationStart) {
      return;
[...]

PR #2834, opened 2020-12-20
fix(router-store): ignore trailing slash when comparing routes

### LINKED ISSUE #2829: bug(router-store): emitting guard results from state changes can cause infinite loops in certain circumstances

^v8.0.0 https://stackblitz.com/edit/angular-8-getting-started-1koz3g
^v10.0.0 https://stackblitz.com/edit/angular-ivy-bf2jo8

## Expected behavior:

Not an infinite loop, instead a simple navigation cancel resulting in a white-screen.

@ngrx/platform: v8.6.0
@angular/core: v8.0.0

## Other information:
This exists in develop as well, but I simply made a repro on the version of Angular in which I found the bug.

The bug is a result of two things:

1. A guard on the root that utilizes state for its observable emission.
2. A race-condition [here.](https://github.com/ngrx/platform/blob/master/modules/router-store/src/router_store_module.ts#L322). 

The router-store code, which is quite old (pre-ngrx 4), assumes that state reduction follows IMMEDIATELY after an action is dispatched. Yet, this isn't always the case, given that actions are stacked onto an rxjs queue. This means that when [this check](https://github.com/ngrx/platform/blob/master/modules/router-store/src/router_store_module.ts#L214 ) occurs the trigger value is `RouterTrigger.NONE` instead of `RouterTrigger.ROUTER` which results in an infinite loop as it triggers another navigation to the root re-checking the guard.

[] No

alex-okrushko:
## PR Checklist

## PR Type

Ignore trailing slash (`/`) when comparing URLs in the `navigateIfNeed` method of the `router-store`.

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

proceeds with re-navigation, which can lead to infinite loop if there's a guard on `/` or `` path (root).

Closes #2829 

## What is the new behavior?

Doesn't navigate if the changes in Store lead to the URL being the same as the current one (previously only the exact match would've prevented it).

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/router-store/spec/integration.spec.ts
+++ b/modules/router-store/spec/integration.spec.ts
@@ -18,6 +18,7 @@ import {
   ROUTER_NAVIGATED,
   ROUTER_NAVIGATION,
   ROUTER_REQUEST,
+  routerNavigationAction,
   RouterAction,
   routerReducer,
   RouterReducerState,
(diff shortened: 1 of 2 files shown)


damienwebdev:
Commit message has a typo, but otherwise the code LGTM.

brandonroberts (approved):

brandonroberts:
Also related to https://github.com/ngrx/platform/issues/1781

PR #2894, opened 2021-01-25
refactor(store): improve on-function typing

alex-okrushko:
## PR Checklist

## PR Type

Enforces the `State` and `action` type of the callback function.

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```


modules/store/src/reducer_creator.ts
```diff
@@ -1,17 +1,28 @@
...
+> {
+  reducer: OnReducer<State, Creators>;
+  types: ExtractActionTypes<Creators>;
 }
 
 // Specialized Reducer that is aware of the Action type it needs to handle
-export interface OnReducer<S, C extends ActionCreator[]> {
-  (state: S, action: ActionType<C[number]>): S;
+export interface OnReducer<S, C extends readonly ActionCreator[]> {
+  (state: S, action: ActionType<C[number]>): { [P in keyof S]: S[P] };
```
timdeschryver, line 25: So this was what I was missing, nice! 🤩

timdeschryver, line 25: Since we're renamed all `S` types to `State`, should we also do it here?

```suggestion
export interface OnReducer<State, Creators extends readonly ActionCreator[]> {
  (state: State, action: ActionType<Creators[number]>): { [P in keyof State]: State[P] };
```

timdeschryver (approved):
👍

modules/store/src/reducer_creator.ts
alex-okrushko: Yeah. We needed to pull out `Reducer extends OnReducer<State, Creators>` from the signature of `on` function and pass the `OnReducer<State, Creators>` directly to the reducer, but then this was in the way :) It needed to be adjusted to be a bit less strict.

brandonroberts (approved):
🧙‍♂️

markostanimirovic:
Very nice! 🎉

dvabuzyarov:
``` typescript
export interface OnReducer<S, C extends readonly ActionCreator[]> {
  (state: S, action: ActionType<C[number]>): { [P in keyof S]: S[P] };
```

due the signature change

``` typescript
{ [P in keyof S]: S[P] }
```

it is not possible to use reducers on primitive types like number or boolean.


Consider this case:

```typescript
export const setId = createAction('Set Id', props<{Id: number}>());

export interface State {
  id: number;
}

const reducer = createReducer(
  0,
  on(setId, (state, {Id}) => Id) //<-- this line will cause an issue
);

StoreModule.forFeature("feature name", {
    Id: reducer
  })
```

PR #2899, opened 2021-01-29
fix(data): make entity param partial when is not optimistic

### LINKED ISSUE #2870: NgRx Data .add() function can have an optional id parameter if isOptimistic: false but typing does not support that
Any NgRx data entity that does not have it's primary key as optional, ex:

This will not allow you to call `dataService.add({name: 'box'}, {isOptimistic: false})` 
```
interface Product {
  id: string
  name: string
}
```

This does though:
```
interface Product {
  id?: string
  name: string
}
```

## Expected behaviour:
Might be a symptom of something else (providing explicit DTOs on a per implementation basic for each entity) but I don't think every property should be filled in for the `add` method since some defaults (like the id, or other things) could be set on the server. I don't think what I am asking is currently possible with TS since we do not know at compile time what the `id` property is since it can change based on the entity metadata.

This could be solved by providing another generic to the data service so that we could type the `add` method with an overload such as:

```
DataService<T, K> // T is type, K is primary key
add(entity: Partial<T>, options: Omit<EntityActionOptions, 'isOptimistic'> & {isOptimistic: false}): Observable<T> // not sure if this typing is right, let me know if it needs more explanation
add(entity: Required<Pick<T, K>> & Partial<T>, options?: EntityActionOptions): Observable<T> // modified current
```

Angular v11
NgRx v10

[y] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

yharaskrik:
- Make the entity: T parameter for the NgRx Data .add command partial when isOptimistic: false since in some cases the PK will be created on the server

Closes #2870

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently NgRx Data allows the primary key to not be passed into the .add() function for an entity but the typings do not allow for that as they expect `T`

Closes #2870

## What is the new behavior?

Change `T` to be `Partial<T>` so that functionality is represented in the types. The full object may not be available as create time if isOptimistic is false.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/data/src/entity-services/entity-collection-service-base.ts
+++ b/modules/data/src/entity-services/entity-collection-service-base.ts
@@ -141,6 +141,10 @@ export class EntityCollectionServiceBase<
    * @returns Observable of the entity
    * after server reports successful save or the save error.
    */
+  add(
+    entity: Partial<T>,
+    options: Omit<EntityActionOptions, 'isOptimistic'> & { isOptimistic: true }
+  ): Observable<T>;
   add(entity: T, options?: EntityActionOptions): Observable<T> {
     return this.dispatcher.add(entity, options);
   }


yharaskrik:
Workflow failed and I am not entirely sure why as I do not know much about github actions, can anyone point me in the direction of what I need to do to let the workflow run?

brandonroberts (approved):

yharaskrik:
Gotcha! Thank you!

timdeschryver (approved):

timdeschryver:
Thanks @yharaskrik

yharaskrik:
You are welcome!

yharaskrik:
There is a bug in this PR I just found while writing the demo app for ngrx data.

```
this.storyDataService.add(story, { isOptimistic: false });
```

This will error with:

```
TS2322: Type 'false' is not assignable to type 'true'.  entity-collection-service-base.ts(146, 60): The expected type comes from property 'isOptimistic' which is declared here on type 'Pick<EntityActionOptions, "correlationId" | "mergeStrategy" | "tag" | "error" | "skip"> & { isOptimistic: true; }'
```

The function overload for the `Partial<T>` is above the original implementation so I think TS is seeing it first (since options is supplied) and is comparing `isOptimistic: false` to `isOptimistic: true`.

Not sure how exactly to type this correctly, I wonder why TS doesn't look at the second function overload. I will read more into function overloading to see what can be done about this.

yharaskrik:
Solved in: https://github.com/ngrx/platform/pull/2906

Issue #1826, opened 2019-05-06
RFC: Add new Higher Order Effect - createApiEffect


wesleygrimes:
## Scenario

Most apps that use NgRx for API requests involve effects that handle a `request`, `success` and `failure` action. 

- The `request` effect listens for a `request` ofType and then makes an API call that returns an `Observable`. 
- If that `Observable` successfully emits a `success` action is returned with the `Observable` value as the payload.
- If the `Observable` throws an error a `failure` action is returned with the error object as the payload.

### Example using `createEffect`

```typescript
loadRequest$ = createEffect(() =>
    this.actions$.pipe(
      ofType(load),
      startWith(load(5)),
      concatMap(action =>
        this.dataService.getItems(action.payload.count).pipe(
          map(items => loadSuccess({ items })),
          catchError(error => of(loadFailure({ error })))
        )
      )
    )
  );
```
### Example using new `createApiEffect`

With this being a common pattern it might be nice to create a new type of effect that handles this scenario in a predictable way. Since `createEffect` is just a function, we can use composition to wrap `createEffect` with a new function named `createApiEffect`. 

Let's revisit the above example using a newly imagined higher-order effect creator named `createApiEffect`. Our new effect would look like the following:

```typescript
loadRequest$ = createApiEffect({
  actionStream: this.actions$,
  ofType: load,
  apiStream: ({ count }) => this.dataService.getItems(count),
  onSuccess: items => loadSuccess({ items }),
  onFailure: error => loadFailure({ error })
});
```

This provides a much cleaner visual to developers, avoids nesting, and helps prevent developers from placing operators in the wrong areas, e.g. `catchError` in the wrong place.

To accomplish this we will make use of the new `mapToAction` operator in #1822. By default, `createApiEffect` would use `concatMap` but we can pass through `operator` as well. 

```typescript
loadRequest$ = createApiEffect({
  actionStream: this.actions$,
  ofType: load,
  apiStream: ({ count }) => this.dataService.getItems(count),
  onSuccess: items => loadSuccess({ items }),
  onFailure: error => loadFailure({ error }),
  operator: mergeMap
});
```

## Potential Implementation of `createApiEffect`

A first pass of `createApiEffect` that makes use of `mapToAction` would look something like this:

```typescript
import { Actions, createEffect, ofType } from '@ngrx/effects';
import { ActionCreator } from '@ngrx/store';
import { Creator, TypedAction } from '@ngrx/store/src/models';
[...]

PR #2900, opened 2021-01-30
RFC: feat(store) createActionWithModifiers

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Since the introduction of `createAction`, actions have been limited to contain only a type and a payload provided at instantiation.

Although all actions should represent distinct events, they frequently can be categorized into similar _kinds_. Therefore, it may be desired to attach constant properties or have common required, dynamic properties.

Actions should be open to modification at time of declaration in order to allow this.

## What is the new behavior?

`createActionWithModifiers` is introduced, which takes a type and props, but also a variadic number of modifiers.

Modifiers take an existing action creator and produce a new action creator (with internal modifications).

There are definitely still helper functions and other tricks that should be introduced to make the API simpler and the Typescript typing better. Also overloads need to exist for actions with no payload.

```
[ ] Yes
[x] No
```

## Other information

Looking for validation as to whether this idea is one worth pursuing!

### DIFF
--- a/modules/store/spec/action_with_modifiers.spec.ts
+++ b/modules/store/spec/action_with_modifiers.spec.ts
@@ -1,0 +1,52 @@
+import {
+  createActionModifier,
+  createActionWithModifiers,
+} from '../src/action_with_modifiers_creator';
+import { props } from '../src/action_creator';
+
+describe('createActionWithModifiers', () => {
+  it('should inject properties onto action', () => {
+    // given
+    const notificationMessageKey = 'notification-message';
+    const requestIdKey = 'request-id';
+
+    const hasNotification = createActionModifier(
+      (config: { message: string }) => ({
+        [notificationMessageKey]: config.message,
+      })
+    );
+
+    const isEndOfRequest = createActionModifier(
+      (config: { requestState: { id: number } }) => ({
+        [requestIdKey]: config.requestState.id,
+      })
+    );
+
+    const createRequestState = () => ({ id: Math.random() });
+
+    const aFormRequestState = createRequestState();
+
+    const aFormSaveSuccess = createActionWithModifiers(
+      'A Form Save Success',
+      props<{ form: { name: string } }>(),
+      (payload) =>
+        hasNotification({
+          message: `${payload.form.name} has been saved as the new name`,
+        }),
+      () => isEndOfRequest({ requestState: aFormRequestState })
+    );
+
+    // when
+    const result = aFormSaveSuccess({ form: { name: 'David' } });
+
+    // then
+    expect(result).toEqual({
+      type: 'A Form Save Success',
+      form: {
+        name: 'David',
+      },
+      [notificationMessageKey]: 'David has been saved as the new name',
+      [requestIdKey]: aFormRequestState.id,
+    });
+  });
+});
--- a/modules/store/src/action_with_modifiers_creator.ts
+++ b/modules/store/src/action_with_modifiers_creator.ts
@@ -1,0 +1,54 @@
+import {
+  ActionCreator,
+  ActionCreatorProps,
+  NotAllowedCheck,
+  TypedAction,
+} from './models';
+import { createAction } from './action_creator';
+
+export function createActionWithModifiers<
+  T extends string,
+  P extends object,
+  MetaProps extends any[]
+>(
+  type: T,
+  props: ActionCreatorProps<P> & NotAllowedCheck<P>,
+  ...modifiers: [
+    ...{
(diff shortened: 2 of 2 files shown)


modules/store/spec/action_with_modifiers.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const aFormSaveSuccess = createActionWithModifiers(
+      'A Form Save Success',
+      props<{ form: { name: string } }>(),
+      (payload) =>
+        hasNotification({
+          message: `${payload.form.name} has been saved as the new name`,
+        }),
+      () => isEndOfRequest({ requestState: aFormRequestState })
+    );
```
david-shortman, line 48: I was thinking the API could look more like this, but when I tried writing the helper functions I wasn't able to get the type inference to work correctly:

```typescript
const aFormSaveSuccess = createActionWithModifiers(
  'A Form Save Success',
  props<{ form: { name: string } }>(),
  is(({ form })=> aNotification({ message: `${form.name} has been saved as the new name` })),
  is(() => anEndOfRequest({ requestState: aFormRequestState }))
);
```

modules/store/src/action_with_modifiers_creator.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  ActionCreator,
+  ActionCreatorProps,
+  NotAllowedCheck,
+  TypedAction,
+} from './models';
+import { createAction } from './action_creator';
+
+export function createActionWithModifiers<
```
david-shortman, line 9: If this idea is truly worth pursuing, it might be better to make this an overload to `createAction` instead.

modules/store/src/action_with_modifiers_creator.ts
```diff
@@ -0,0 +1,41 @@
...
+  ...modifiers: [
+    ...{
+      [i in keyof MetaProps]: (
+        p: P
+      ) => (
+        actionCreator: ActionCreator<T, (p: P) => TypedAction<T> & P>
+      ) => ActionCreator<T, (p: P) => TypedAction<T> & P & MetaProps[i]>;
+    }
+  ]
+) {
```
david-shortman, line 25: Typescript infers this function as producing an action creator which returns an action that has properties of type `P` and the _intersection_ of all the properties of the payloads of the modifiers.

It would be desired to have the _union_ instead, but I tried using `MetaProps[number]` in a return signature only to continue to get the intersection.

david-shortman: Essentially, if modifier A attaches "propertyOne" and modifier B attaches "propertyOne" and "propertyTwo", then Typescript will think the instantiated action can only have "propertyOne" available, not "propertyTwo."

modules/store/spec/action_with_modifiers.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    >(
+      action: ActionCreator<T, (p: P) => TypedAction<T> & P>
+    ) => {
+      const creator = (p: P) => ({
+        ...action(p),
+        [notificationMessageKey]: config.message,
+      });
+      creator.type = action.type;
+      return creator;
+    };
```
david-shortman, line 20: Ideally there should be a function which creates modifiers for you such that you could do:

```typescript
const hasNotification = createActionModifier(
  (config: { message: string }) => ({ [notificationMessageKey]: config.message })
);
```

david-shortman:
With the ability to apply modifiers to actions, you can have single effects that handle categories of actions.
This is an inverse solution to repetitively crating effects as suggested in https://github.com/ngrx/platform/issues/1826

brandonroberts:
@david-shortman thanks for working on this. This effort would have been better started as an RFC issue instead of a PR. I don't fully understand how this gives you something you can't do with action creators using a callback function instead the `props` method.

With a function:

 ```ts
  export const loginSuccess = createAction(
    '[Auth/API] Login Success',
    (response: Response) => response.user
  );
 ```

As a userland implementation I think it's fine but as an addition to the library itself, I don't see us adding an additional `createAction` API just to support this use case. Subtyping actions in order to reuse them would also violate one principle of Good Action Hygiene.

david-shortman:
> @david-shortman thanks for working on this. This effort would have been better started as an RFC issue instead of a PR.

I've now created https://github.com/ngrx/platform/issues/2902 with a better writeup that hopefully explains what I'm seeking better.

Issue #2715, opened 2020-09-14
Switch to Variadic Tuple Types for reducers and selectors


alex-okrushko:
## Describe any alternatives/workarounds you're currently using

Angular release 10.1 added TS 4.0 support (https://github.com/angular/angular/blob/master/CHANGELOG.md#1010-2020-09-02) so we'd want to get that as well. This will enable us to use Variadic Tuple Types.

## If accepted, I would be willing to submit a PR for this feature

[x] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

david-shortman:
I'm interested in helping add the types. Would want to know how I should go about this since the repo doesn't currently use 4.0+ yet.

markostanimirovic:
Issue with switching `createSelector` to variadic tuples: https://github.com/microsoft/TypeScript/issues/41709

david-shortman:
Here's at first shot at defining them. 

```typescript
export function createSelector<State, S extends any[], Result>(
  selectors: [...{ [i in keyof S]: Selector<State, S[i]> }],
  projector: (...args: [...S]) => Result
): MemoizedSelector<State, Result>;
export function createSelector<S extends any[], State, Result>(
  ...args: [
    ...{ [i in keyof S]: Selector<State, S[i]> },
    (...args: [...S]) => Result
  ]
): MemoizedSelector<State, Result>;
export function createSelector<State, Props, S extends any[], Result>(
  selectors: [...{ [i in keyof S]: SelectorWithProps<State, Props, S[i]> }],
  projector: (...args: [...S]) => Result
): MemoizedSelectorWithProps<State, Props, Result>;
export function createSelector<State, Props, S extends any[], Result>(
  ...args: [
    ...{ [i in keyof S]: SelectorWithProps<State, Props, S[i]> },
    (...args: [...S, Props]) => Result
  ]
): MemoizedSelectorWithProps<State, Props, Result>;
```

But I'm hitting a problem where Typescript is not able to infer `State` properly, resulting in this error:

(in `router-store/src/router_selectors.ts)
```
Argument of type '(state: V) => RouterReducerState<any>' is not assignable to parameter of type 'Selector<unknown, RouterReducerState<any>>'. 
  Types of parameters 'state' and 'state' are incompatible.
    Type 'unknown' is not assignable to type 'V'.
      'V' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.
```

[...]

PR #2905, opened 2021-01-31
refactor(store) used variadic tuple types for createSelector

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: Technically this allows for typed selectors with more than 8 input selectors, so it's kinda a "feature" upgrade as well.
```

Types were manually created for set numbers of arguments.

Partially completes #2715 

## What is the new behavior?

`createSelector` stays strongly typed for an infinite number of arguments.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -118,393 +118,42 @@ export function defaultMemoize(
   return { memoized, reset, setResult, clearResult };
 }
 
-export function createSelector<State, S1, Result>(
-  s1: Selector<State, S1>,
-  projector: (s1: S1) => Result
-): MemoizedSelector<State, Result>;
-export function createSelector<State, Props, S1, Result>(
-  s1: SelectorWithProps<State, Props, S1>,
-  projector: (s1: S1, props: Props) => Result
-): MemoizedSelectorWithProps<State, Props, Result>;
-export function createSelector<State, S1, Result>(
-  selectors: [Selector<State, S1>],
-  projector: (s1: S1) => Result
-): MemoizedSelector<State, Result>;
-export function createSelector<State, Props, S1, Result>(
-  selectors: [SelectorWithProps<State, Props, S1>],
-  projector: (s1: S1, props: Props) => Result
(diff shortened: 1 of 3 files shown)


david-shortman:
oh my god these types are driving me crazy

david-shortman:
opening as a draft to request help

david-shortman:
Some notes:

- The TS 4.2 beta is required for the tuple types to work properly. This is due to the [update for leading and middle rest elements in tuple types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-2-beta/#leading-middle-rest-elements-in-tuple-types).
- This is currently a breaking change, as the type is much stricter than before. For instance, the `selector.spec.ts` file in this repo was using `jasmine.Spy` functions for many selectors, which cannot now be interpreted properly without a type assertion
  - Am looking for feedback on this to know what the typical route is for breaking changes. Here, if we introduce a loose overload that accepts `any` type for the parameters, then the usefulness of the other overloads is greatly reduced.

modules/store/src/selector.ts
```diff
@@ -118,393 +118,42 @@ export function defaultMemoize(
...
+
+export function createSelector<S extends Selector<any, any>[], Result>(
+  ...args: [
+    ...S,
+    (...args: [...{ [i in keyof S]: ValueOfSelector<S[i]> }]) => Result
+  ]
+): MemoizedSelector<StateOfSelectors<S>, Result>;
 export function createSelector<
-  State,
+  S extends SelectorWithProps<any, any, any>[],
```
david-shortman, line 134: The old definition of selectors with props specified that all input selectors had those props, but that doesn't quite make sense to me. Couldn't they be any arbitrary selectors, and the `Props` are merely inferred from the projector?

david-shortman:
Waiting on Angular to support TS 4.2 before continuing

brandonroberts:
Let's revisit this after TS 4.2 lands in Angular

PR #2939, opened 2021-02-20
docs(effects): update documentation to help avoid asynchronous testing mistakes

### LINKED ISSUE #2449: Docs: Use async method when testing effects without TestBed
# Background
I decided to test my effects without using the Angular TestBed because this should be faster then using the entire TestBed when I only want to test the effect in isolation (https://ngrx.io/guide/effects/testing#setup-without-testbed). I am using [Jest](https://jestjs.io) as test runner.

# Repro-Steps
Consider this simplified, generic approach I used.
```ts
  it('should get data on success', () => {
    // Arrange.
    const sampleData = { ... }
    const serviceMock = {
      getMyData: jest.fn().mockReturnValue(of(sampleData)),
    };

    const actions$ = new Actions(
      of(loadData({ id: 42 })),
    );

    const effects = new MyEffects(actions$, serviceMock);

    // Act.
    effects.loadData$.subscribe(action => {
      // Assert.
      expect(action).toEqual(
        loadDataSuccess({ data: sampleData }),
      );
    });
  }));
```

# What's the problem?
Unfortunately the test is green, however the coverage showed me that it is not executed properly. This is due to two problems:
- The jest.fn() call does not execute anything and is wrong. This is an error from my side.
- However, in this case we want to **see that an error occurs and the test should fail**.

# How to fix it? What do change?
If the test itself was run inside of `async`, then we would know that actually the service's `getMyData` has never been called because of the wrong implementation of the mock.

The [docs ](https://ngrx.io/guide/effects/testing#setup-without-testbed) may propose to wrap the async test into the `async` method call like this:

```ts
  it('should get data on success', async(() => { ...
```

# Side Note
The test itself can than be fixed by don't mocking out the call by jest, but letting it go through.

Instead of using this:
```ts
getMyData: jest.fn().mockReturnValue(of(sampleData)),
```

We must use this:
```ts
getMyData: jest.fn(() => of(sampleData)),
```

[ ] No

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The examples may lead users to write tests that silently fail, as noted in https://github.com/ngrx/platform/issues/2449.

Maybe closes #2449 

## What is the new behavior?

Updated examples

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngrx.io/content/guide/effects/testing.md
+++ b/projects/ngrx.io/content/guide/effects/testing.md
@@ -139,7 +139,7 @@ testScheduler.run(({ cold, hot, expectObservable }) => {
 });
 </code-example>
 
-By using the `TestScheduler` we can also test effects dependant on a scheduler.
+By using the `TestScheduler` we can also test effects dependent on a scheduler.
 Instead of creating an effect as a method to override properties in test cases, as shown in [`Effects with parameters`](#effects-with-parameters), we can rewrite the test case by using the `TestScheduler`.
 
 <code-example header="my.effects.spec.ts">
@@ -183,8 +183,10 @@ effects.getAll$.subscribe(action => {
     type: '[Customers API] Get Customers Success',
     customers: [...],
   });
+  done();
 });
 </code-example>
+
 
 ### With `ReplaySubject`
 
@@ -206,6 +208,7 @@ effects.getAll$.subscribe(action => {
     type: '[Customers API] Get Customers Success',
     customers: [...],
   });
+  done();
 });
 </code-example>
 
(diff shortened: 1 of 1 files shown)


projects/ngrx.io/content/guide/effects/testing.md
```diff
@@ -186,6 +188,12 @@ effects.getAll$.subscribe(action => {
 });
 </code-example>
 
+<div class="alert is-important">
```
timdeschryver, line 191: What about adding this note beneath the title `With `ReplaySubject`` ?
We could also clarify/update the test by using the `done` callback.

david-shortman: Is `done` specific to Jest though?

timdeschryver: AFAIK most of the test runners have them

timdeschryver (approved):

brandonroberts (approved):

PR #2940, opened 2021-02-20
feat(router-store) warn about DefaultRouterStateSerializer for runtime checks enabled

### LINKED ISSUE #2686: Warning when using DefaultRouterStateSerializer and runtimeChecks
I'd want to ask if it would be possible to give a warning when using DefaultRouterStateSerializer with `strictStateImmutability` (enabled by default), `strictActionImmutability` (enabled by default) and `strictStateSerializability`. Currently it throws an error really difficult to debug "_Cannot add property NG_ELEMENT_ID, object is not extensible_". I've spent hours until I found the solution on https://github.com/nrwl/nx/issues/3549#issuecomment-678531928.

Not applicable.

## Other information:
I'd want to ask why the `DefaultRouterState` isn't the default one as the name implies? Shouldn't it better to rename it to something like `FullRouterState` or anything else?
Also, how is `RouterState` property different than `Serializer`? I guess I can't use `MinimalRouteStateSerializer` with `RouterStateFull` and vice-versa, right?

As I'm a newbie in NGRX, I confess that these two questions really puzzled me and maybe it's hard to other people learning it as well.

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Without reading the docsite thoroughly, users might miss that the `DefaultRouterStateSerializer` will not conform to `strictStateImmutability` and `strictStateSerializability`.

Closes #2686 

## What is the new behavior?

Always displays warning when using `DefaultRouterStateSerializer`.

```
[ ] Yes
[x] No
```

## Other information

If possible, it would be better to display this warning _only_ when the conflicting runtime checks are enabled. But I'm not sure it's feasible to check the config of an arbitrary module from the initialization of another.


modules/router-store/src/router_store_module.ts
```diff
@@ -155,6 +156,16 @@ export class StoreRouterConnectingModule {
   >(
     config: StoreRouterConfig<T> = {}
   ): ModuleWithProviders<StoreRouterConnectingModule> {
+    if (!isNgrxMockEnvironment() && isDevMode()) {
```
timdeschryver, line 159: At the moment this warning will be logged everytime, right?
I think we only want it when the `DefaultRouterStateSerializer` is used, and preferably together when there are some enabled `runtimeChecks`.

modules/router-store/src/router_store_module.ts
```diff
@@ -35,6 +42,7 @@ import {
   SerializedRouterStateSnapshot,
 } from './serializers/default_serializer';
 import { MinimalRouterStateSerializer } from './serializers/minimal_serializer';
+import { _ACTIVE_RUNTIME_CHECKS } from '../../store/src/tokens';
```
david-shortman, line 45: `_ACTIVE_RUNTIME_CHECKS` was meant to be private to `store`, so this is actually illegal.

looking for your thoughts @timdeschryver if the right approach to having this warning includes exposing the injection token for the active runtime checks

david-shortman: I went ahead and exposed it, YOLO

modules/router-store/src/router_store_module.ts
david-shortman: I'm now injecting the `ACTIVE_RUNTIME_CHECKS` from ngrx/store to show the warning only when the conflicting `runtimeChecks` are enabled

modules/router-store/src/router_store_module.ts
```diff
@@ -35,6 +42,7 @@ import {
   SerializedRouterStateSnapshot,
 } from './serializers/default_serializer';
 import { MinimalRouterStateSerializer } from './serializers/minimal_serializer';
+import { ACTIVE_RUNTIME_CHECKS } from '@ngrx/store';
```
timdeschryver, line 45: Can this be added to the existing `store` imports please?

modules/router-store/src/router_store_module.ts
timdeschryver: Somehow I missed this.
I think this is fine, at least I don't see how we can expose it otherwise.

modules/router-store/src/router_store_module.ts
david-shortman: 😳 whoops, yes

timdeschryver (approved):

brandonroberts (approved):

PR #2941, opened 2021-02-20
feat(store-devtools) add autoPause option

### LINKED ISSUE #2722: Add support for `autoPause` option
We've recently introduced NgRx into one of our applications. As our state object grows very large in our integration environment, we needed to deactivate the NgRx devtools. This was necessary as the overhead of sending the state to devtools brought the application to a crawl. The performance hit is caused by serializing the state object to JSON. This is done by the Redux Devtools Extension itself.

The Redux Devtools Extension has a feature to automatically pause recording the state when the devtools are closed. This reduces the overhead to zero as long as the devtools are closed.
[This feature can be activated by setting `autoPause: true` in the options object.](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#autopause)

It is currently not possible to set `autoPause: true` using the `StoreDevtoolsModule`.

N/A. Disabling the devtools is currently the only option.

## Other information:
It is not possible to reduce the size of our state object.

[ x ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Users cannot configure this property.

Closes #2722

## What is the new behavior?

Adds option to automatically pause capturing actions and state changes when the redux devtools extension window is not open.

```
[ ] Yes
[x] No
```

## Other information

The default is set to `false` as that is the current behavior.

### DIFF
--- a/modules/store-devtools/schematics/ng-add/index.ts
+++ b/modules/store-devtools/schematics/ng-add/index.ts
@@ -51,7 +51,9 @@ function addImportToNgModule(options: StoreDevtoolsOptions): Rule {
     const [instrumentNgModuleImport] = addImportToModule(
       source,
       modulePath,
-      `StoreDevtoolsModule.instrument({ maxAge: ${options.maxAge}, logOnly: environment.production })`,
+      `StoreDevtoolsModule.instrument({ maxAge: ${options.maxAge}, autoPause: ${
+        options.autoPause ?? false
+      }, logOnly: environment.production })`,
       modulePath
     );
 
--- a/modules/store-devtools/schematics/ng-add/index.ts
+++ b/modules/store-devtools/schematics/ng-add/index.ts
@@ -51,9 +51,7 @@ function addImportToNgModule(options: StoreDevtoolsOptions): Rule {
     const [instrumentNgModuleImport] = addImportToModule(
       source,
       modulePath,
-      `StoreDevtoolsModule.instrument({ maxAge: ${options.maxAge}, autoPause: ${
-        options.autoPause ?? false
-      }, logOnly: environment.production })`,
+      `StoreDevtoolsModule.instrument({ maxAge: ${options.maxAge}, logOnly: environment.production })`,
       modulePath
     );
 
--- a/modules/store-devtools/schematics/ng-add/index.spec.ts
+++ b/modules/store-devtools/schematics/ng-add/index.spec.ts
@@ -62,7 +62,7 @@ describe('Store-Devtools ng-add Schematic', () => {
       /import { StoreDevtoolsModule } from '@ngrx\/store-devtools';/
     );
     expect(content).toMatch(
-      /StoreDevtoolsModule.instrument\({ maxAge: 25, logOnly: environment.production }\)/
+      /StoreDevtoolsModule.instrument\({ maxAge: 25, autoPause: false, logOnly: environment.production }\)/
     );
   });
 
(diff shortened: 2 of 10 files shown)


timdeschryver (changes requested):
I don't think this will work because we still need to provide the option to the devtools.
We can do this by defining the `autoPause` property [here](https://github.com/ngrx/platform/blob/master/modules/store-devtools/src/extension.ts#L53) and assign it's value [here](https://github.com/ngrx/platform/blob/master/modules/store-devtools/src/extension.ts#L240).

modules/store-devtools/schematics/ng-add/index.ts
```diff
@@ -51,7 +51,9 @@ function addImportToNgModule(options: StoreDevtoolsOptions): Rule {
     const [instrumentNgModuleImport] = addImportToModule(
       source,
       modulePath,
-      `StoreDevtoolsModule.instrument({ maxAge: ${options.maxAge}, logOnly: environment.production })`,
+      `StoreDevtoolsModule.instrument({ maxAge: ${options.maxAge}, autoPause: ${
```
timdeschryver, line 54: I don't know if we want to add this in the schematics by default.
Personally, I think lesser options is better, and I feel that the  `maxAge` and `logOnly` properties are the most important.
But this got me thinking, perhaps we could follow-up this PR by providing all of the devtools options, and only add them to the config when they deviate from the default values. I don't know if that's worth it tho, what do you think?

timdeschryver (commented):
Hey, thanks for the changes!
In the future, could you push a separate commit with your changes please? This makes it easier to review, otherwise we have to take a look at the whole PR again 😅. 

I also added an additional question that I forgot to ask previously.

david-shortman:
> In the future, could you push a separate commit with your changes please? This makes it easier to review, otherwise we have to take a look at the whole PR again 😅.

I'll do that! Question from that- I noticed that, in order for PRs to be merged to master, they're supposed to be a single commit. Is that something I should just wait to do after getting approvals?

modules/store-devtools/schematics/ng-add/index.ts
david-shortman: I think allowing pass-through of the options could be helpful, but I think it's also worth evaluating how each feature works and if there's anything special we should be doing to support them better.

For instance, I've realized on this PR that, even though auto-pausing is enabled on the devtools side of things, the ngrx-devtools module is still using CPU time by sending data to the not-listening extension. I'm going to look at if it's possible to retrieve the extension status and not bother sending the data when auto-pause is enabled.

timdeschryver:
@david-shortman we can squash the commits into one commit when we approve and merge the PR.

timdeschryver:
So my assumption is that new actions won't be visible in the devtools, but they will still be sent to the devtools extenion. Is that correct?

If it's a big deal we could also prevent that the action is sent to the devtools. Although I'm not sure of the impact it will have, and if it's really needed to do add this.

brandonroberts:
Do we have an update on this PR? If we're blocked on the redux extension, I think we should close this until that's resolved and revisit.

david-shortman:
> Do we have an update on this PR? If we're blocked on the redux extension, I think we should close this until that's resolved and revisit.

Without empirical evidence, my assumption is that this PR is useful even though it would be nice to completely disable sending data to the extension when it is not open.

Allowing passthrough of these options at least prevents the redux devtools from having a performance impact on its side.

timdeschryver (approved):
I think this is already a step in the right decision, for those who encounter problems while using the DevTools. If this issue persists, we can take further steps although I'm not a big fan if this would mean that we have to add some checks whether we need to keep the DevTools state in sync.

brandonroberts (approved):
LGTM

brandonroberts:
Thanks @david-shortman for pushing through as this one took a while.

PR #2976, opened 2021-03-16
fix(effects): add support for Proxy objects in Effects

### LINKED ISSUE #2975: Error when injecting Proxy object in effect class

I've created a minimal project which shows the problem. All the relevant code is in:
https://github.com/MortenGregersen/ngrx-proxy/blob/main/src/app/app.module.ts

The problem is that when a [`Proxy` object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) with a custom getter is injected into the class that contains the `Effect`s, it will cause an error because it is resolved as an `Effect` and `.pipe()` is invoked on it.

## Expected behavior:

When using `createEffect()` to create `Effect`s the error is in the following snippet. In line 83 we ask the Proxy object if it has a `CREATE_EFFECT_METADATA_KEY` property where it (in my reproduction) returns a `Promise` which is resolved to `true`. There should be an addtional condition where we check if the property's `CREATE_EFFECT_METADATA_KEY` property contains a `dispatch` property.
https://github.com/ngrx/platform/blob/8930e22eb8121afb88cff4c6a3a4fc98a97df429/modules/effects/src/effect_creator.ts#L74-L96

@ngrx/effects: 11.0.1
Angular: 11.2.5
Node: 12.14.0

## Other information:

The reason we run into this issue, is because we use custom Ionic Capacitor plugins which in turn is a `Proxy` object. We want to inject them in `Effect`s to be able to mock them in tests.

[ ] No

MortenGregersen:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Right now `mergeEffects` fails with an error when a `Proxy` object is injected in an `Effect`s class.

Closes #2975 

## What is the new behavior?

With this extra condition `Proxy` will be filtered out.

```
[ ] Yes
[x] No
```

## Other information

I don't fully understand how the commit message should be formed and if there should be any documentation changes. I hope you can help with that.

### DIFF
--- a/modules/effects/spec/effect_creator.spec.ts
+++ b/modules/effects/spec/effect_creator.spec.ts
@@ -85,6 +85,14 @@ describe('createEffect()', () => {
       const fakeCreateEffect: any = () => {};
       class Fixture {
         a = fakeCreateEffect(() => of({ type: 'A' }));
+        b = new Proxy(
+          {},
+          {
+            get(_, prop) {
+              return () => Promise.resolve('bob');
+            },
+          }
+        );
       }
 
       const mock = new Fixture();
--- a/modules/effects/src/effect_creator.ts
+++ b/modules/effects/src/effect_creator.ts
@@ -77,11 +77,16 @@ export function getCreateEffectMetadata<
   const propertyNames = Object.getOwnPropertyNames(instance) as Array<keyof T>;
 
   const metadata: EffectMetadata<T>[] = propertyNames
-    .filter(
-      (propertyName) =>
+    .filter((propertyName) => {
+      if (
         instance[propertyName] &&
         instance[propertyName].hasOwnProperty(CREATE_EFFECT_METADATA_KEY)
-    )
+      ) {
+        const property = instance[propertyName] as any;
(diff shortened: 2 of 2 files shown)


timdeschryver (commented):
LGTM, I just would like to add a comment on why the extra check is needed, because this is an edge case.

A bit out of scope but do you have an example of a capacitor plugin that is used as an Effect?

MortenGregersen:
> LGTM, I just would like to add a comment on why the extra check is needed, because this is an edge case.

I've added a comment 👍🏻 

> A bit out of scope but do you have an example of a capacitor plugin that is used as an Effect?

I don't have a public example using a Capacitor plugin. But the reproduction code for the related issue has an example of the use case we want to fix with this pull request. The Proxy object behaves the same way as Capacitor's plugins.
https://github.com/MortenGregersen/ngrx-proxy/blob/main/src/app/app.module.ts

I have created a related issue on Capacitor's repository: https://github.com/ionic-team/capacitor/issues/4356

brandonroberts (approved):

timdeschryver (approved):

PR #2977, opened 2021-03-17
feat(store) mock store.select

### LINKED ISSUE #2962: Feature Request: Allow MockStore.overrideSelector to accept an Observable as its second argument
Hi! ngrx is great!

Coming over from the older unsupported ```angular-redux/store``` framework, one thing I miss is their ability to provide an Observable as a mock response for ```store.select``` calls.

It would be so useful if I could use ```overrideSelector``` with an Observable. This would make it easy to use with Jasmine marbles, for example, when testing Effects or Components/Service methods that listen to the store:

```
      // set up marble diagram for the test using Jasmine marbles
      const promptFlag$ = hot('ff-t-f-t', {f: false, t: true});
      const dataToBeSaved$ = hot('---a---b', {a: {}, b: {}});

      // specify that calls to ```store.select``` should return the marble diagram Observables when the specified selector is sent
      store.overrideSelector(selectShowPromptFlag, promptFlag$);
      store.overrideSelector(selectShowPromptFlag, dataToBeSaved$);
```

If necessary, this could be a new method:
```
      store.overrideSelectorWithObservable(selectShowPromptFlag, promptFlag$);    
```

Here is the workaround I'm currently working on. I think this does the same thing but looks a lot uglier:
```
      store.select = createSpy('select').and.callFake((selector) => {
        if (selector === selectShowPromptFlag) {
          return promptFlag$;
        } else if (selector === selectDataToBeSaved) {
          return dataToBeSaved$;
        } else {
          fail('Unexpected selector');
          return;
        }
      });
```

[ ] No

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Only selectors are natively mockable in tests.

Closes #2962

## What is the new behavior?
`select` can be overridden with observables.

```
[ ] Yes
[x] No
```

## Other information
Looking for feedback on the way select is overridden for selectors with props.

### DIFF
--- a/modules/store/testing/spec/mock_store.spec.ts
+++ b/modules/store/testing/spec/mock_store.spec.ts
@@ -23,8 +23,9 @@ import {
 } from '@ngrx/store';
 import { INCREMENT } from '../../spec/fixtures/counter';
 import { Component, Injector } from '@angular/core';
-import { Observable } from 'rxjs';
+import { Observable, of } from 'rxjs';
 import { By } from '@angular/platform-browser';
+import { cold } from 'jasmine-marbles';
 
 interface TestAppSchema {
   counter1: number;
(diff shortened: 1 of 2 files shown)


brandonroberts:
I don't think we should introduce this change, so that the preference of how to mock the `select` method is up to the developer and testing tool of choice.

PR #2990, opened 2021-04-08
build: don't include un-needed files in the distributed packages

alan-agius4:
At the moment, redundant files are being included in the packages libraries ex: https://unpkg.com/browse/@ngrx/store@11.1.0/esm2015/testing/

This change updates the `tsconfig.json` of the all modules to enable `skipTemplateCodegen` with the recommended setting for libraries.

With this option enabled, ngfactories and ngsummaries which are not needed for libraries are not generated and hence the libraries size footprint in the node_modules directory is also smaller.

This is also the default value for the option when generating a library using Angular CLI https://github.com/angular/angular-cli/blob/e607d4482e761c3e556ccc6f4b870af350a5c86a/packages/angular_devkit/build_angular/test/hello-world-lib/projects/lib/tsconfig.lib.json#L15

More info about this option can be found here: https://angular.io/guide/angular-compiler-options#skiptemplatecodegen

### DIFF
--- a/modules/component-store/tsconfig-build.json
+++ b/modules/component-store/tsconfig-build.json
@@ -25,6 +25,7 @@
     "annotateForClosureCompiler": true,
     // Work around for issue: https://github.com/angular/angular/issues/22210
     "strictMetadataEmit": false,
+    "skipTemplateCodegen": true,
     "flatModuleOutFile": "index.js",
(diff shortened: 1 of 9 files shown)


brandonroberts (approved):

brandonroberts:
Thanks @alan-agius4!

PR #2993, opened 2021-04-11
feat(store): deprecate selectors with props

### LINKED ISSUE #2980: RFC: Deprecate Selectors With Props
# RFC: Deprecate VS Improve Selector With Props
​
I created this to jot down some thoughts about the current state of selectors with props with the intention to come up with a game plan. The result should either be to deprecate selectors with props (and later removed), or to tackle the current problems.
​
## Developer Experience
​
The biggest problem in terms of developer experience is that the types inside the selector aren't inferred correctly
(https://github.com/ngrx/platform/issues/2205). With the release of Angular 12, a new project will be strict by default, so I assume more developers will stumble against this issue.
​
A fix would be to improve the types.
​
## Performance
​
One of the big benefits of using selectors is that a selector is memoized, but with selectors with props it's very easy to "break" this. Most of the problems/questions I see is because selector with props aren't used correctly, resulting in selectors that are executed multiple times. This isn't something a user might expect, and arises some confusion. Even when they're used correctly, shared "child selectors" will be invoked for each time that child selector is used in a selector with props (https://github.com/ngrx/platform/issues/2862).
​
We could improve the docs and try to explain the concept around selector with props better.
Currently, I don't see how we can reduce the number of calls.
​
## Maintenance
​
The NgRx has [a lot of overloads](https://github.com/ngrx/platform/blob/master/modules/store/src/selector.ts#L121-L507) to correctly type selectors and selectors with props. We could refactor this with [Variadic Tuple Types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types), with the benefit that we can support to compose N amount of selectors. I'm not certain, but I think removing selectors with props, would help to refactor the selector's types.
​
## Advocating for factory selectors
​
As mentioned in the earlier sections, I do think that we can improve and fix the problems with the current state of selectors with props.
​
But I do also think that even if we do this, that selector with props will still be "miss-used" simply because it's a concept that has to learned. I've seen numerous questions of users using selector with props, but are still asking on how to pass those props to child selectors (even though the props are passed down to all selectors).
​
So why not keep things simple with factory selectors to pass props to a selector?
By doing this the intention is clear, and it's something that everyone can understand because these are just functions. Simply for this reason, I encourage using factory selectors.
​
Factory selectors also fix the performance hit, although I assume that the difference is negligible.

The only downside is that selectors aren't automatically passed down to all selectors (I think that the majority don't know this about selectors with props, and thus are also not using it). Instead, we have to pass down the props manually.

## TLDR

Selector with props is a new concept, are hard to understand, and are easy to make "mistakes" with.
Instead, we can create the same behavior with factory selectors, without all the overhead.
A bonus would be that I assume that we can get rid of a lot of types (and a bit of code) in the NgRx repo and replace them with variadic tuple types.

To emphasize, we can probably resolve most of the technical problems, but I fear that selectors with props are harder "to get right".

## Demo
​
```ts
interface ReducerState {
  count: number;
}
​
let logs: string[] = [];
​
const selector = createFeatureSelector<ReducerState>('s');
​
const selectCount = createSelector(selector, (s) => {
  logs.push(`[selectCount] ${s.count}`);
  return s.count;
});
​
const selectCountMultipliedProps = () =>
  // 👎 the result type of the selector is lost (`count`)
  // Parameter 'count' implicitly has an 'any' type
  createSelector(
    selectCount,
    (count: number, props: { multiplier: number }) => {
      logs.push(`[selectCountMultipliedProps] ${count} * ${props.multiplier}`);
      return count * props.multiplier;
    }
  );
​
const selectCountMultiplied = (props: { multiplier: number }) =>
  // 👍 `count` knows that it's a number
  createSelector(selectCount, (count) => {
    logs.push(`[selectCountMultiplied] ${count} * ${props.multiplier}`);
    return count * props.multiplier;
  });
​
beforeEach(() => {
  logs.length = 0;
});
​
it('selector with props', () => {
  const state = { s: { count: 5 } };
​
  const selector2 = selectCountMultipliedProps();
  const selector3 = selectCountMultipliedProps();
​
  selector2(state, { multiplier: 2 });
  selector2(state, { multiplier: 2 });
  selector3(state, { multiplier: 3 });
  selector2(state, { multiplier: 2 });
​
  console.log('selector with props', logs);
​
  // selector with props [
  //   '[selectCount] 5',
  //   '[selectCountMultipliedProps] 5 * 2',
  //   '[selectCount] 5',
  //   '[selectCountMultipliedProps] 5 * 2',
  //   '[selectCount] 5',
  //   '[selectCountMultipliedProps] 5 * 3',
  //   '[selectCount] 5',
  //   '[selectCountMultipliedProps] 5 * 2'
  // ]
});
​
it('selector with same props', () => {
  const state = { s: { count: 5 } };
​
  const selector2 = selectCountMultipliedProps();
  const props2 = { multiplier: 2 };
  const selector3 = selectCountMultipliedProps();
  const props3 = { multiplier: 3 };
​
  selector2(state, props2);
  selector2(state, props2);
  selector3(state, props3);
  selector2(state, props2);
​
  console.log('selector with same props', logs);
​
  // selector with same props [
  //   '[selectCount] 5',
  //   '[selectCountMultipliedProps] 5 * 2',
  //   '[selectCount] 5',
  //   '[selectCountMultipliedProps] 5 * 3'
  // ]
});
​
it('factory selector', () => {
  const state = { s: { count: 5 } };
​
  selectCountMultiplied({ multiplier: 2 })(state);
  selectCountMultiplied({ multiplier: 2 })(state);
  selectCountMultiplied({ multiplier: 3 })(state);
  selectCountMultiplied({ multiplier: 2 })(state);
​
  console.log('factory selector', logs);
​
  // factory selector [
  //   '[selectCount] 5',
  //   '[selectCountMultiplied] 5 * 2',
  //   '[selectCountMultiplied] 5 * 2',
  //   '[selectCountMultiplied] 5 * 3',
  //   '[selectCountMultiplied] 5 * 2'
  // ]
});
​
it('factory selector with same reference', () => {
  const state = { s: { count: 5 } };
​
  const select2 = selectCountMultiplied({ multiplier: 2 });
  const select3 = selectCountMultiplied({ multiplier: 3 });
  select2(state);
  select2(state);
  select3(state);
  select2(state);
​
  console.log('factory selector with same reference', logs);
​
  // factory selector with same reference [
  //   '[selectCount] 5',
  //   '[selectCountMultiplied] 5 * 2',
  //   '[selectCountMultiplied] 5 * 3',
  // ]
});

### PR DESCRIPTION
**Pull request by timdeschryver**
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2980

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/models.ts
+++ b/modules/store/src/models.ts
@@ -48,6 +48,9 @@ export interface StoreFeature<T, V extends Action = Action> {
 
 export type Selector<T, V> = (state: T) => V;
 
+/**
+ * @deprecated Selectors with props are deprecated, for more info see {@link https://github.com/ngrx/platform/issues/2980 Github Issue}
+ */
 export type SelectorWithProps<State, Props, Result> = (
   state: State,
   props: Props
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -28,6 +28,9 @@ export interface MemoizedSelector<
   clearResult: () => void;
 }
 
+/**
+ * @deprecated Selectors with props are deprecated, for more info see {@link https://github.com/ngrx/platform/issues/2980 Github Issue}
+ */
 export interface MemoizedSelectorWithProps<
   State,
   Props,
@@ -122,6 +125,9 @@ export function createSelector<State, S1, Result>(
   s1: Selector<State, S1>,
   projector: (s1: S1) => Result
 ): MemoizedSelector<State, Result>;
+/**
+ * @deprecated Selectors with props are deprecated, for more info see {@link https://github.com/ngrx/platform/issues/2980 Github Issue}
+ */
 export function createSelector<State, Props, S1, Result>(
   s1: SelectorWithProps<State, Props, S1>,
   projector: (s1: S1, props: Props) => Result
@@ -130,6 +136,9 @@ export function createSelector<State, S1, Result>(
   selectors: [Selector<State, S1>],
   projector: (s1: S1) => Result
 ): MemoizedSelector<State, Result>;
+/**
+ * @deprecated Selectors with props are deprecated, for more info see {@link https://github.com/ngrx/platform/issues/2980 Github Issue}
+ */
 export function createSelector<State, Props, S1, Result>(
   selectors: [SelectorWithProps<State, Props, S1>],
   projector: (s1: S1, props: Props) => Result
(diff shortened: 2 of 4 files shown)


alex-okrushko (approved):
Would it be possible to add the update schematic for this to ease the migration? (not necessarily for now, but when we remove it)

MikeRyanDev (approved):
Big fan of this change

timdeschryver:
@alex-okrushko we don't have a schematic to fix these selectors.
It's something we'll have to look into, because I'm not sure how we can provide a schematic that covers all cases. For example, props are passed to all selectors and I'm not sure if we can refactor all selectors without breaking something.

PR #3023, opened 2021-05-18
feat(store): use variadic tuple types for createSelector

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Types were manually created for set numbers of arguments.

Completes #2715 (I don't see any reducer code that still needs variadic typing?)

## What is the new behavior?

`createSelector` stays strongly typed for an infinite number of arguments.

```
[x] Yes
[ ] No
```

Some small typing adjustments.


modules/store/src/selector.ts
```diff
@@ -390,44 +263,85 @@ export function createSelector<
...
+
+export function createSelector<State, S extends unknown[], Result>(
+  selectors: [...{ [i in keyof S]: Selector<State, S[i]> }],
+  projector: (...s: S) => Result
+): MemoizedSelector<State, Result>;
+
+/**
+ * @deprecated Selectors with props are deprecated, for more info see {@link https://github.com/ngrx/platform/issues/2980 Github Issue}
+ */
+export function createSelector<State, Props, S1, Result>(
```
david-shortman, line 279: left the existing definitions for `createSelector` when creating selectors with props since it's deprecated

modules/store/spec/selector.spec.ts
```diff
@@ -128,7 +128,7 @@ describe('Selectors', () => {
       const selectorFn = jasmine
         .createSpy(
           'selectorFn',
-          createSelector((state) => state, projectorFn)
+          createSelector((state: unknown) => state, projectorFn)
```
david-shortman, line 131: the typing is slightly stricter now such that new errors like implicit any may occur. therefore, this is a breaking change.

modules/store/src/selector.ts
```diff
@@ -122,34 +122,19 @@ export function defaultMemoize(
   return { memoized, reset, setResult, clearResult };
 }
 
-export function createSelector<State, S1, Result>(
-  s1: Selector<State, S1>,
-  projector: (s1: S1) => Result
+export function createSelector<State, S extends unknown[], Result>(
+  ...args: [...Selector<State, unknown>[], unknown] &
+    [...{ [i in keyof S]: Selector<State, S[i]> }, (...s: S) => Result]
 ): MemoizedSelector<State, Result>;
```
david-shortman, line 128: one might think that this would be sufficient:

```typescript
export function createSelector<State, S extends unknown[], Result>(
  ...args: [...{ [i in keyof S]: Selector<State, S[i]> }, (...s: S) => Result]
): MemoizedSelector<State, Result>;
```

however, TS refuses to match on this override with that as it always infers `State` as `unknown`.

the best I came up with to resolve that was to try to infer `State` by including a type definition specifically for it.

david-shortman:
The example-app tests are failing on circleci and yet

Screen Shot 2021-05-19 at 22 18 06
Screen Shot 2021-05-19 at 22 18 52

timdeschryver:
Does that error disappears if you remove the deprecated selector with props?
The error message in the CI, indicates that that type of selector is used.

david-shortman:
@timdeschryver Am facing an issue with a test using a selector in a way that seems unsupported by the old types:

In integration.spec.ts:

```typescript
const getTodosState = createFeatureSelector<TodoAppSchema, Todo[]>(
  'todos'
);
const getTodos = createSelector(getTodosState, (todos) => todos);
  getTodos, // MemoizedSelector
  (state: TodoAppSchema, id: number) => id, // SelectorWithProps
  (todos, id) => todos.find((todo) => todo.id === id) // projector 
);
```

With the previous types, it seemed that to create a new `MemoizedSelectorWithProps`, the selector must be composed exclusively of other `SelectorWithProps`s.

Yet `getTodosById` is composed of both types of selectors. Stranger yet, I would expect that the projector would be required to have _three_ arguments- the first two being the result of the first two selectors, and the third being the props.

When I change the definition like so, then I no longer have a type error, and the arguments are inferred as expected:

```typescript
          getTodos,
          (state: TodoAppSchema, id: number) => id,
          (todos, id, idProp) => todos.find((todo) => todo.id === id)
        );
```

Yet weirdly, `getTodos` is interpreted as extending `SelectorWithProps` in order for this to work, which is strange.

Similarly, further in the same spec there's this selector:

```typescript
  getTodosState, // MemoizedSelector
  (todos: Todo[], { id }: { id: number }) =>
    todos.find((todo) => todo.id === id)
);
```

Since `getTodosState` does not extend `SelectorWithProps`, my types are unhappy. But looking at the old type definitions, I can't conclude how this ever _was_ supported.

david-shortman:
> Does that error disappears if you remove the deprecated selector with props?
> The error message in the CI, indicates that that type of selector is used.

> Does that error disappears if you remove the deprecated selector with props?

I'm not sure I understand. If I remove the function overload definition, then I have a bigger error because that invocation of `createSelector` will not match the overloads for regular `Selector`s.

Are you suggesting completely removing `SelectorWithProps` code?

modules/entity/src/state_selectors.ts
```diff
@@ -14,11 +14,10 @@ export function createSelectorsFactory<T>() {
     const selectAll = createSelector(
       selectIds,
       selectEntities,
-      (ids: T[], entities: Dictionary<T>): any =>
-        ids.map((id: any) => (entities as any)[id])
+      (ids, entities): any => ids.map((id: any) => (entities as any)[id])
```
david-shortman, line 17: bewilderingly, even though the explicit typing of the projector parameters aligned perfectly with the inferred types, Typescript's brain breaks and thinks this matches the overload for returning `SelectorWithProps` when explicitly typed.

modules/store/spec/selector.spec.ts
```diff
@@ -128,7 +128,7 @@ describe('Selectors', () => {
       const selectorFn = jasmine
         .createSpy(
           'selectorFn',
-          createSelector((state) => state, projectorFn)
+          createSelector((state: any) => state, projectorFn)
```
david-shortman, line 131: i do not know why specifically here TS starting complaining about implicit any, because as far as i can tell this parameter was already implicitly any before this change

david-shortman:
Since there can be minor type adjustments that are breaking, it may be best to pair this with a pr that completely removes `SelectorWithProps` to finish off dealing with selectors for v13: https://github.com/ngrx/platform/issues/3035.

david-shortman:
So it all works now, but in some circumstances (as evidenced by the few changes to other files) the overloads are not properly recognized if the projector functions have explicitly typed parameters. Not sure why and if anyone has advanced TS knowledge to avoid this as to not make this a breaking change, that'd be litty.

brandonroberts:
Is this dependent on selectors with props being removed? If so, I say we close it until then.

david-shortman:
> Is this dependent on selectors with props being removed? If so, I say we close it until then.

This is not dependent on selectors with props being removed.

timdeschryver:
Since the types in some tests are changed, I think this will lead to some breaking changes?
If that's the case, this PR should also have to wait for v13.

david-shortman:
> Since the types in some tests are changed, I think this will lead to some breaking changes?
> If that's the case, this PR should also have to wait for v13.

Right now it's a breaking change, but I'm pretty sure there's something that could be done either with the types or with a Typescript issue that would cause the overloads to not be improperly interpreted... if someone smarter than me could find a solution 😅.

david-shortman:
@timdeschryver This is a breaking change because you cannot provide a specific set of generic type arguments to `createSelector` anymore.

So `createSelector<State, S1, S2, S3, Result>(...)` is invalid now, and instead, if someone were to use explicit generic type arguments, they would have to do this: `createSelector<State, [S1, S2, S3], Result>.

projects/data-example-app/src/state/story.selectors.ts
```diff
@@ -1,7 +1,7 @@
 import { createSelector } from '@ngrx/store';
 import { Stories, Story } from './story';
 
-export const selectStories = createSelector<Story[], Story[], Story[][]>(
+export const selectStories = createSelector<Story[], [Story[]], Story[][]>(
```
david-shortman, line 4: Explicitly provided generic arguments must take the format `<State, [S1, S2, Etc], Result>`

modules/entity/src/state_selectors.ts
david-shortman: you really should be able to explicitly declare the parameter types to a projector function, so I want this to be fixed, I just can't figure out what Typescript is mad about.

david-shortman:
> @MaximSagan we're not there yet. 
> We will try to follow the Angular releases, so we can have v13 released shortly hereafter.

@timdeschryver Is this planned for v13? I noticed it wasn't included for the first beta.

david-shortman:
> > @MaximSagan we're not there yet.
> > We will try to follow the Angular releases, so we can have v13 released shortly hereafter.

@brandonroberts Wondering if there is a blocker for this being included in v13?

brandonroberts:
@david-shortman no blocker. I wasn't sure if this was ready to be merged yet

david-shortman:
> @david-shortman no blocker. I wasn't sure if this was ready to be merged yet

It should be. I'm not sure why it doesn't have any reviews yet, but it is ready.

timdeschryver:
Sorry for my late response, I can take a look at it later this week and provide some feedback then.

timdeschryver (approved):
👏👏 @david-shortman 

This looks good to me.
We could also provide a schematic that provides a fix for the breaking change, e.g.

```ts
BEFORE:

createSelector<Story[], Story[], Story[][]>

AFTER:

//        needs to be a tuple 👇
createSelector<Story[], [ Story[] ] , Story[][]>
```

david-shortman:
> 👏👏 @david-shortman 
> 
> 
> 
> This looks good to me.
> 
> We could also provide a schematic that provides a fix for the breaking change, e.g.
> 
> 
> 
> BEFORE:
> 
> ```ts
> 
> createSelector<Story[], Story[], Story[][]>
> 
> ```
> 
> 
> 
> AFTER:
> 
> ```ts
> 
> //        needs to be a tuple 👇
> 
> createSelector<Story[], [ Story[] ] , Story[][]>
> 
> ```

I haven't written a schematic before, but it could be fun to try.

Should I go for that in a separate PR?

brandonroberts:
@david-shortman @timdeschryver is the tuple only required when you strictly define all the types with `createSelector`?

david-shortman:
> @david-shortman @timdeschryver is the tuple only required when you strictly define all the types with `createSelector`?

The generic arguments are inferred, so that you can use `const inferred = createSelector(() => 1, () => 2, () => 3)`.

When manually specifying the generic arguments, you have to specify the selector's list of selector return values, like `const strictlySpecified = createSelector<State, [1, 2], 3>(...);

modules/store/src/selector.ts
```diff
@@ -122,34 +122,19 @@ export function defaultMemoize(
   return { memoized, reset, setResult, clearResult };
 }
 
-export function createSelector<State, S1, Result>(
-  s1: Selector<State, S1>,
-  projector: (s1: S1) => Result
+export function createSelector<State, S extends unknown[], Result>(
```
markostanimirovic, line 125: It would be great to add a full name for the `S` generic because it's not clear at first what is its purpose. The name could be: `Slices`

modules/store/src/selector.ts
```diff
@@ -122,34 +122,19 @@ export function defaultMemoize(
...
+  ...args: [...Selector<State, unknown>[], unknown] &
```
markostanimirovic, line 126: nice hack 👀

modules/store/src/selector.ts
```diff
@@ -390,44 +264,86 @@ export function createSelector<
     s5: S5,
     s6: S6,
     s7: S7,
+    s8: S8,
     props: Props
   ) => Result
 ): MemoizedSelectorWithProps<State, Props, Result>;
-export function createSelector<State, S1, S2, S3, S4, S5, S6, S7, Result>(
+
+export function createSelector<State, S extends unknown[], Result>(
```
markostanimirovic, line 272: the same here: `S` => `Slices`

markostanimirovic (commented):
Great work David! 🥇 

Btw, it would be great to add type tests for selectors :)

modules/store/src/selector.ts
david-shortman: 🙇‍♂️

david-shortman:
> Great work David! 🥇
> 
> Btw, it would be great to add type tests for selectors :)

Hm, ts-snippet can only validate the return type of the function, not whether the function inferred the parameters correctly.

In #3223, the return type is validated in type checks since the `Slices` are used in `MemoziedSelector`.

david-shortman:
@timdeschryver I was looking into creating a schematic and noticed in https://github.com/ngrx/platform/pull/3074 that migrations might no longer be included in this repo, but somewhere else? Where would that be?

markostanimirovic (approved):
LGTM 👍

markostanimirovic:
Changed PR type from `refactor` to `feature`, for two reasons:

- Breaking change with types
- The `createSelector` function can now accept any number of input selectors.

brandonroberts:
🥳

PR #3025, opened 2021-05-19
feat(schematics): add the ability to create actions with the prefix

### LINKED ISSUE #3012: Allow to specify action name with custom prefix in actions schematics.
Currently, when generating action it's always prefixed with the "**load**", so that when I provide action name as "**DeleteUsers**", I will end-up with following action within generated file "**loadDeleteUsers**". It will be nice to have a possibility to generate action with the custom name without "**load**" prefix. In my case it would be "**deleteUsers**", "**deleteUsersSuccess**" etc. What's more I think it would be great to have an option to add actions to the existing file.

[ ] No

mateuszbasinski:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The actions are created only with the prefix Load.

Closes #3012 

## What is the new behavior?

The actions are created with a specified prefix.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/schematics/src/action/schema.json
+++ b/modules/schematics/src/action/schema.json
@@ -13,6 +13,12 @@
       },
       "x-prompt": "What should be the name of the action?"
     },
+    "prefix": {
+      "description": "The prefix of the action.",
+      "type": "string",
+      "default": "load",
+      "x-prompt": "What should be the prefix of the action?"
+    },
     "path": {
       "type": "string",
       "format": "path",
--- a/modules/schematics/src/action/schema.ts
+++ b/modules/schematics/src/action/schema.ts
@@ -5,6 +5,11 @@ export interface Schema {
   name: string;
 
   /**
+   * The prefix for the actions.
+   */
+  prefix: string;
+
+  /**
    * The path to create the component.
    */
   path?: string;
--- a/modules/schematics/src/action/creator-files/__name@dasherize@if-flat__/__name@dasherize__.actions.spec.ts.template
+++ b/modules/schematics/src/action/creator-files/__name@dasherize@if-flat__/__name@dasherize__.actions.spec.ts.template
@@ -1,7 +1,7 @@
 import * as from<%= classify(name) %> from './<%= dasherize(name) %>.actions';
 
-describe('load<%= classify(name) %>s', () => {
+describe('<%= prefix %><%= classify(name) %>s', () => {
   it('should return an action', () => {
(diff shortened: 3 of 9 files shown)


modules/schematics/src/action/schema.json
```diff
@@ -13,6 +13,15 @@
       },
       "x-prompt": "What should be the name of the action?"
     },
+    "prefix": {
```
timdeschryver, line 16: Does this work? The `name` property already uses the unnamed argument, I think the default of this should be `load`, or the index should be 1 (but I prefer the default option)

modules/schematics/src/action/schema.ts
```diff
@@ -4,6 +4,11 @@ export interface Schema {
    */
   name: string;
 
+  /**
+   * The name of the component.
```
timdeschryver, line 8:
```suggestion
   * The prefix for the actions.
```

modules/schematics/src/action/schema.json
mateuszbasinski: Yes, this works, but the default option is a better idea

mateuszbasinski:
@timdeschryver I came up with the idea to generate actions with many prefixes by using one command. For example - for the action schematics we can pass an array with prefixes ([load, create, update, delete]) and then generate actions with all prefixes from the array. What do you think about this idea?

timdeschryver (approved):

timdeschryver:
@mateuszbasinski I think that's a good idea.
I just don't know if the Angular CLI would parse a comma separated input correctly.
Have you tried that?

The only downside is that this leads to command-driven actions, while we encourage event-driven actions.
But the currently implementation also "violates" this practice 😅

mateuszbasinski:
@timdeschryver I tried to deal with this using split. The output of splitted array looks good.
image

mateuszbasinski:
@timdeschryver I tried to deal with this using split. The output of splitted array looks good.
image

timdeschryver:
Then I think we can extend the schematic to create multiple defaults, as long as the default remains "load".

brandonroberts:
I think this PR is good, but I don't think we should add comma-separated prefixes. As Tim mentioned, we're already violating event-driven actions, so I want to change those to be more correct, such as:

```
[Users Page] (Init/Load Users/Some other Event)
[Users Page/API] Users Loaded Success
[Users Page/API] Users Loaded Failure
```

I also think the goal of the schematics is to generate the minimum of what you need, and not try to auto-generate everything.

brandonroberts (approved):

PR #3040, opened 2021-06-10
fix(data): immutably delete an entity

### LINKED ISSUE #2553: ngrx-data delete error
Goods day.
I've been  stuck here for a few days.And i really need some help.
I'm trying to use ngrx/data,it's working fine when add,update,get. but it gets error when delete.(not 100% happen)

fontend:https://github.com/chobijaeyu/todoListAngular backend:https://github.com/chobijaeyu/todoListGin

service:
```
export class TodoService extends EntityCollectionServiceBase<Todo> {

  constructor(serviceElementsFactory: EntityCollectionServiceElementsFactory) {
    super("Todo", serviceElementsFactory);
  }

}
```

component.ts:
```
 deleteTodo(e: Event, todo: Todo) {
    e.stopPropagation()
    this.todoservice.delete(todo) 
// I tried this.todoservice.delete(todo.id)  same error
  }
```

error:
```
data.js:8106 TypeError: Cannot assign to read only property 'changeType' of object '[object Object]'
    at data.js:5990
    at Array.reduce (<anonymous>)
    at EntityChangeTrackerBase.trackDeleteMany (data.js:5968)
    at EntityChangeTrackerBase.trackDeleteOne (data.js:6023)
    at EntityCollectionReducerMethods.saveDeleteOne (data.js:6711)
    at entityCollectionReducer (data.js:7553)
    at EntityCacheReducerFactory.applyCollectionReducer (data.js:8025)
    at EntityCacheReducerFactory.entityCacheReducer (data.js:7749)
    at store.js:418
    at combination (store.js:306)
```
スクリーンショット 2020-06-03 14 14 05

or
```
data.js:8106 TypeError: Cannot add property skip, object is not extensible
    at EntityCollectionReducerMethods.saveDeleteOne (data.js:6707)
    at entityCollectionReducer (data.js:7553)
    at EntityCacheReducerFactory.applyCollectionReducer (data.js:8025)
    at EntityCacheReducerFactory.entityCacheReducer (data.js:7749)
    at store.js:435
    at combination (store.js:323)
    at store.js:1282
    at store.js:403
    at computeNextEntry (store-devtools.js:999)
    at recomputeStates (store-devtools.js:1052)
```

## Expected behavior:
what i expecte is delete item remote and local store.
actually it's did delete item remote,delete request return 200 even got this TypeError.

```
Angular CLI: 9.1.7
Node: 14.3.0
OS: darwin x64

Angular: 9.1.9
... animations, common, compiler, compiler-cli, core, forms
... language-service, localize, platform-browser
... platform-browser-dynamic, router, service-worker
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.901.7
@angular-devkit/build-angular     0.901.7
@angular-devkit/build-optimizer   0.901.7
@angular-devkit/build-webpack     0.901.7
@angular-devkit/core              9.1.7
@angular-devkit/schematics        9.1.7
@angular/cdk                      9.2.4
@angular/cli                      9.1.7
@angular/flex-layout              9.0.0-beta.29
@angular/material                 9.2.4
@ngtools/webpack                  9.1.7
@schematics/angular               9.1.7
@schematics/update                0.901.7
rxjs                              6.5.5
typescript                        3.8.3
webpack                           4.42.0

    "@ngrx/data": "^9.2.0",
    "@ngrx/effects": "^9.2.0",
    "@ngrx/entity": "^9.2.0",
    "@ngrx/router-store": "^9.2.0",
    "@ngrx/schematics": "^9.2.0",
    "@ngrx/store": "^9.2.0",
    "@ngrx/store-devtools": "^9.2.0",
```

donated $20,thanks for your brilliant works.

donohoea:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2553

## What is the new behavior?

No state mutation on delete and ngrx/data actions are exempt from strictActionImmutability runtime checks.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/runtime_checks.spec.ts
+++ b/modules/store/spec/runtime_checks.spec.ts
@@ -138,15 +138,6 @@ describe('Runtime checks:', () => {
         flush();
       }).not.toThrow();
     }));
-
-    it('should not throw for NgRx actions', fakeAsync(() => {
-      const store = setupStore({ strictStateSerializability: true });
-
-      expect(() => {
-        store.dispatch(makeNgrxAction(invalidAction()));
-        flush();
-      }).not.toThrow();
-    }));
   });
 
   describe('Action Serialization:', () => {
@@ -169,6 +160,15 @@ describe('Runtime checks:', () => {
 
       expect(() => {
         store.dispatch(invalidAction());
+        flush();
+      }).not.toThrow();
+    }));
+
+    it('should not throw for NgRx actions', fakeAsync(() => {
+      const store = setupStore({ strictActionSerializability: true });
+
+      expect(() => {
+        store.dispatch(makeNgrxAction(invalidAction()));
         flush();
       }).not.toThrow();
     }));
@@ -338,14 +338,26 @@ function reducerWithBugs(state: any = {}, action: any) {
         invalidSerializationState: true,
         invalid: new Date(),
       };
-
+      
     case ErrorTypes.UnserializableAction: {
+      return {
+        invalidSerializationAction: true,
+      };
+    }
+    
+    case '@ngrx ' + ErrorTypes.UnserializableAction: {
       return {
         invalidSerializationAction: true,
       };
     }
 
     case ErrorTypes.MutateAction: {
+      action.foo = 'foo';
+      return {
+        invalidMutationAction: true,
+      };
+    }
+    case '@ngrx ' + ErrorTypes.MutateAction: {
       action.foo = 'foo';
       return {
         invalidMutationAction: true,
--- a/modules/data/spec/reducers/entity-change-tracker-base.spec.ts
+++ b/modules/data/spec/reducers/entity-change-tracker-base.spec.ts
@@ -580,6 +580,22 @@ describe('EntityChangeTrackerBase', () => {
       const collection = tracker.trackDeleteMany(['1234', 456], origCollection);
       expect(collection).toBe(origCollection);
     });
+
+    it('should not mutate changeState when called on a tracked "updated" entity', () => {
+      const existingEntity = getFirstExistingEntity();
+      const updatedEntity = toUpdate({
+        ...existingEntity,
+        name: 'test update',
+      });
+      const collection = tracker.trackUpdateOne(updatedEntity, origCollection);
+      const change = collection.changeState[existingEntity!.id];
+      expect(change).toBeDefined();
+      expectChangeType(change, ChangeType.Updated);
+      Object.freeze(change);
+      expect(() => {
+        tracker.trackDeleteMany([existingEntity!.id], collection);
+      }).not.toThrowError();
+    });
   });
 
   describe('#trackUpdateOne', () => {
--- a/modules/data/src/reducers/entity-change-tracker-base.ts
+++ b/modules/data/src/reducers/entity-change-tracker-base.ts
@@ -454,7 +454,7 @@ export class EntityChangeTrackerBase<T> implements EntityChangeTracker<T> {
           } else if (trackedChange.changeType === ChangeType.Updated) {
             // Special case: switch change type from Updated to Deleted.
             cloneChgStateOnce();
-            trackedChange.changeType = ChangeType.Deleted;
+            chgState[id] = { ...chgState[id], changeType: ChangeType.Deleted };
           }
         } else {
           // Start tracking this entity


modules/store/src/runtime_checks.ts
```diff
@@ -69,7 +69,7 @@ export function createImmutabilityCheckMetaReducer({
 }
 
 function ignoreNgrxAction(action: Action) {
-  return action.type.startsWith('@ngrx');
+  return action.type.includes('@ngrx');
```
brandonroberts, line 72: Why do we need this change? I don't think it should be included in this PR

brandonroberts (changes requested):

modules/store/src/runtime_checks.ts
donohoea: One of the errors in #2553 was a `TypeError: Cannot add property skip, object is not extensible` thrown by the strictActionImmutability runtime check. Since action mutation is baked into ngrx/data with use of the skip and error fields in EntityActionOptions it makes sense to exempt ngrx/data actions from the runtime check. As data action types take the form: '[tag || entityname] @ngrx/data.....' they aren't captured by startsWith('@ngrx').

timdeschryver: AFAIK the `delete` action is the only one that's causing this problem (I could be wrong).
Can't we make it immutable instead?

donohoea: Yes this issue is only related to the delete action. But the action is mutated by design. When change tracking is enabled (by default) on entity if an add or update is persisted to the local store but not to the server (ie. no add success action is dispatched) and a subsequent delete action is dispatched the reducer removes the entity from the local store and adds the skip flag to the action so the effects ignore it and do not try to delete a record that never got to the server.

In a similar manner I believe there are cases where other actions would cause an error in a reducer method which would add the error flag to the action so that the action is ignored by the effects. From the definition of the ngrx/data entity-action: 
```javascript
...
 // Mutable actions are BAD.
  // Unfortunately, these mutations are the only way to stop @ngrx/effects
  // from processing these actions.

  /**
   * The action was determined (usually by a reducer) to be in error.
   * Downstream effects should not process but rather treat it as an error.
   */
  error?: Error;

  /**
   * Downstream effects should skip processing this action but should return
   * an innocuous Observable<Action> of success.
   */
  skip?: boolean;
}
```
I don't think we can make data actions immutable without some significant changes.

brandonroberts: I'd rather not change the behavior of the startsWith check, but change the action string for ngrx/data. It could be considered a breaking change is someone is listening based on that action type, but it's unlikely they are using the actual string itself.

timdeschryver: We could keep this in mind while working on v13 and leave it as it for now?

brandonroberts: Yea, i think that would be better

donohoea: Ok I can change the ngrx/data action string instead. There is a test fix and a fix for mutable state change bug also identified in #2553 in this PR as well. These would not introduce breaking changes. Should I spin them out into separate PRs so they can be merged before v13? Maybe they should have been separate from the beginning? New to the PR process.

brandonroberts: You can do them in PR as separate commits and we'll just keep the separate commits instead of squashing them into one

brandonroberts:
@donohoea will you rebase on `master` and condense it down to 2 commits?

brandonroberts:
@donohoea changing the action type format is a breaking change that will have to wait until v13.

donohoea:
@brandonroberts that makes sense. Do you need me to remove that commit or what's the best way to hold on to that for v13?

brandonroberts:
@donohoea yes, just remove the commit and we'll keep it on the list for v13.

brandonroberts (approved):

modules/store/spec/runtime_checks.spec.ts
```diff
@@ -339,6 +339,12 @@ function reducerWithBugs(state: any = {}, action: any) {
         invalid: new Date(),
       };
 
+    case '@ngrx ' + ErrorTypes.UnserializableState:
```
timdeschryver, line 342: Do we need these added cases here?

donohoea: @timdeschryver I found the two `it('should not throw for NgRx actions'...` tests to were passing by default . Because no cases in the reducer match `'@ngrx ' +  ErrorTypes.UnserializableState` or `'@ngrx ' + ErrorTypes.MutateAction` they both fall through to the default which returns the initial state and passes by default. But after fixing that I realized that it should be testing ngrx actions with Unserializable Actions not Unserializable State (as that's the exception granted to ngrx actions). So I changed the test but forgot to change the additional case until you pointed this out thanks!

timdeschryver (approved):

timdeschryver:
Thanks @donohoea

PR #3103, opened 2021-08-01
feat(router-store): add createRouterSelector to select router data for default config

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

## Other information

This is an RFC to make it easier to select the router state when the default router config is used.
Because the router store module is opinionated and uses `router` as the default feature name, I was thinking about adding a bit of default behavior to the current router selectors. This aims to make it easier to get started with using the router store.
 
When the default config is used, I think we can:

- export a router state selector 
- make `getSelectors` smarter and to use the default router state selector to retrieve the router state

I started this PR to promote and use the `DEFAULT_ROUTER_FEATURENAME` (instead of the "magic `router`" string) constant in the example app and the docs. But then I continued to make the selectors more opinionated, because this is how I've been and seen using it in all the projects.

```ts
BEFORE:

export const selectRouter = createFeatureSelector<fromRouter.RouterReducerState>(
  'router'
);

export const { selectRouteData } = fromRouter.getSelectors(selectRouter);

AFTER:

// change 1, select the default router state with createRouterSelector

export const selectRouter = createRouterSelector();

// change 2, the selector argument of getSelectors isn't required

export const { selectRouteData } = fromRouter.getSelectors();
```

### DIFF
--- a/modules/router-store/src/router_selectors.ts
+++ b/modules/router-store/src/router_selectors.ts
@@ -1,9 +1,21 @@
-import { createSelector } from '@ngrx/store';
+import {
+  createFeatureSelector,
+  createSelector,
+  MemoizedSelector,
+} from '@ngrx/store';
 import { RouterStateSelectors } from './models';
 import { RouterReducerState } from './reducer';
+import { DEFAULT_ROUTER_FEATURENAME } from './router_store_module';
+
+function createRouterSelector<State extends {} = {}>(): MemoizedSelector<
+  State,
+  RouterReducerState
+> {
+  return createFeatureSelector(DEFAULT_ROUTER_FEATURENAME);
+}
 
 export function getSelectors<V>(
-  selectState: (state: V) => RouterReducerState<any>
+  selectState: (state: V) => RouterReducerState<any> = createRouterSelector<V>()
 ): RouterStateSelectors<V> {
   const selectRouterState = createSelector(
     selectState,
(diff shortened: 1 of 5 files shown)


projects/ngrx.io/content/guide/router-store/selectors.md
```diff
@@ -13,13 +14,8 @@ Usage:
 
 [Full App Used In This Example](https://stackblitz.com/edit/ngrx-router-store-selectors?file=src/app/car.state.ts)
 
-`router.selectors.ts`
```
timdeschryver, line 16: The changes to the code snippets, migraties the markdown code blocks to `code-example` blocks.

brandonroberts (approved):

modules/router-store/src/router_selectors.ts
```diff
@@ -1,9 +1,21 @@
...
+import {
+  createFeatureSelector,
+  createSelector,
+  MemoizedSelector,
+} from '@ngrx/store';
 import { RouterStateSelectors } from './models';
 import { RouterReducerState } from './reducer';
+import { DEFAULT_ROUTER_FEATURENAME } from './router_store_module';
+
+export function createRouterSelector<State extends {} = {}>(): MemoizedSelector<
```
markostanimirovic, line 10:
```suggestion
export function createRouterSelector<State extends Record<string, any>>(): MemoizedSelector<
```

or

```suggestion
export function createRouterSelector<State extends object>(): MemoizedSelector<
```

because empty object type `{}` is compatible with primitive values.

markostanimirovic (approved):
I like the idea, Tim 👍  Just a small suggestion:

modules/router-store/src/router_selectors.ts
timdeschryver: Thanks, I'm going for the first one because `object` is giving lint warnings.

PR #3109, opened 2021-08-06
WIP: feat: add support for RxJS 7.x

### LINKED ISSUE #3059: Support for rxjs 7
When trying to update to rxjs 7, I run into:

```
Package "@ngrx/store-devtools" has an incompatible peer dependency to "rxjs" (requires "^6.5.3", would install "7.1.0").
```

Is store-devtools compatible as-is with rxjs 7, or is there some work to ensure compatibility?

That would be great.

[ x ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

brandonroberts:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3059 

## What is the new behavior?

Supports RxJS 6/7 as a peer dependency

```
[ ] Yes
[x] No
```


modules/store/src/action_creator.ts
```diff
@@ -138,7 +138,7 @@ export function union<
 
 function defineType<T extends string>(
   type: T,
-  creator: Creator
+  creator: Creator & any
```
brandonroberts, line 141: This needs some type fixes

timdeschryver: I don't know if we can fix this.
We could type assert the return, which is better than the `& any` imho:

```ts
  return Object.defineProperty(creator, 'type', {
    value: type,
    writable: false,
  }) as ActionCreator<T>;
```

modules/effects/src/concat_latest_from.ts
```diff
@@ -29,9 +29,9 @@ export function concatLatestFrom<
       const observablesAsArray = Array.isArray(observables)
         ? observables
         : [observables];
-      return of(value).pipe(
+      return (of(value).pipe(
```
timdeschryver, line 32: By taking a look at the new RxJS types for `withLatestFrom`, I think we can simplify this.
Maybe in another PR though (because it might be breaking if end-users were using the generics).

```ts
export function concatLatestFrom<T, O extends unknown[], V>(
  observablesFactory: (value: V) => [...ObservableInputTuple<O>]
): OperatorFunction<T, [T, ...O]>;
export function concatLatestFrom<T, O, V>(
  observablesFactory: (value: V) => ObservableInput<O>
): OperatorFunction<T, [T, O]>;

/**
 * `concatLatestFrom` combines the source value
 * and the last available value from a lazily evaluated Observable
 * in a new array
 */
export function concatLatestFrom<T, R>(
  observablesFactory: (value: any) => any | any[]
): OperatorFunction<T, R | any[]> {
  return pipe(
    concatMap((value) => {
      const observables = observablesFactory(value);
      const observablesAsArray = Array.isArray(observables)
        ? observables
        : [observables];
      return of(value).pipe(withLatestFrom(...observablesAsArray));
    })
  );
}
```

david-shortman: that's pretty cool!

CrazyBaran:
Up :) We waiting :)

PR #3118, opened 2021-08-17
docs(effects): add concatLatestFrom example to API docs

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

## Other information

I'm currently missing a simple example that shows how we can use the `concatLatestFrom` operator on the API page.
While I was adding it, I also noticed `ofType` was missing an example.


modules/effects/src/concat_latest_from.ts
```diff
@@ -12,6 +12,18 @@ export function concatLatestFrom<T extends Observable<unknown>, V>(
...
+ *
+ * ## Example
+ * Select the active customer from the NgRx Store
+ *
+ * ```ts
+ * import { concatLatestFrom } from '@ngrx/effects';
+ * import * fromCustomers from '../customers';
+ *
+ * this.actions$.pipe(
+ *  concatLatestFrom(() => this.store.select(fromCustomers.selectActiveCustomer))
```
david-shortman, line 24: could be worth adding an example that shows that you can use the value from the stream in the argument list of the function passed into contactLatestFrom, too

like
`concatLatestFrom(action => this.store.select(fromCustomers.selectCustomer(action.id))`

timdeschryver: That's a good idea, thanks!

PR #3223, opened 2021-11-09
feat(store) add overload for projector to inform about incorrect usage

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Starts process on #3097 

## What is the new behavior?

A message on an overload for `projector` when it is used incorrectly:

Screen Shot 2021-11-11 at 15 00 12

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -17,13 +17,21 @@ export type ComparatorFn = (a: any, b: any) => boolean;
 
 export type DefaultProjectorFn<T> = (...args: any[]) => T;
 
+export type ProjectorFn<Args extends unknown[], T> = (...args: Args) => T;
+
+export type StrictMemoizedSelector<
+  Slices extends unknown[],
+  Result,
+  State = unknown
+> = MemoizedSelector<State, Result, (...slices: Slices) => Result>;
+
 export interface MemoizedSelector<
   State,
   Result,
-  ProjectorFn = DefaultProjectorFn<Result>
+  Projector = DefaultProjectorFn<Result>
 > extends Selector<State, Result> {
   release(): void;
-  projector: ProjectorFn;
+  projector: Projector;
   setResult: (result?: Result) => void;
   clearResult: () => void;
 }
@@ -121,6 +129,22 @@ export function defaultMemoize(
 
   return { memoized, reset, setResult, clearResult };
 }
+
+export function createSelector<
+  State,
+  Slices extends unknown[],
+  Result,
+  IsStrict extends boolean
+>(
+  ...args: [...Selector<State, unknown>[], unknown, unknown] &
+    [
+      ...{ [i in keyof Slices]: Selector<State, Slices[i]> },
+      (...s: Slices) => Result,
+      { strict: IsStrict }
+    ]
+): IsStrict extends true
+  ? StrictMemoizedSelector<Slices, Result, State>
+  : MemoizedSelector<State, Result>;
 
 export function createSelector<State, Slices extends unknown[], Result>(
   ...args: [...Selector<State, unknown>[], unknown] &
@@ -271,6 +295,20 @@ export function createSelector<
     props: Props
   ) => Result
 ): MemoizedSelectorWithProps<State, Props, Result>;
+
+export function createSelector<
+  State,
+  Slices extends unknown[],
+  Result,
+  IsStrict extends boolean
+>(
+  selectors: Selector<State, unknown>[] &
+    [...{ [i in keyof Slices]: Selector<State, Slices[i]> }],
+  projector: (...s: Slices) => Result,
+  options: { strict: IsStrict }
+): IsStrict extends true
+  ? StrictMemoizedSelector<Slices, Result, State>
+  : MemoizedSelector<State, Result>;
 
 export function createSelector<State, Slices extends unknown[], Result>(
   selectors: Selector<State, unknown>[] &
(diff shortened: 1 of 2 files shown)


modules/store/src/selector.ts
```diff
@@ -20,10 +20,15 @@ export type DefaultProjectorFn<T> = (...args: any[]) => T;
...
+  ProjectorFn = DefaultProjectorFn<Result>,
+  ProjectorArgs extends unknown[] = unknown[]
 > extends Selector<State, Result> {
   release(): void;
-  projector: ProjectorFn;
+  projector(...args: ProjectorArgs): Result;
+  /** @deprecated Only use valid arguments for the selector's projector */
+  projector(
+    ...args: ProjectorFn extends (...args: infer Args) => unknown ? Args : any[]
+  ): ProjectorFn extends (...args: unknown[]) => infer R ? R : Result;
```
david-shortman, line 31: When invalid arguments are used, then the deprecated signature is used. The deprecated signature is backwards compatible with existing usages of the `ProjectorFn` generic parameter by using its parameters and return value in the type.

david-shortman:
@timdeschryver The docs-app test is failing, seemingly for a reason unrelated to the PR.

timdeschryver:
@david-shortman 👍 you can leave it as is - we'll take a look at it. 
For now, let's try a rerun.

modules/store/src/selector.ts
```diff
@@ -17,13 +17,25 @@ export type ComparatorFn = (a: any, b: any) => boolean;
...
+export type ProjectorFn<Args extends unknown[], T> = (...args: Args) => T;
+
 export interface MemoizedSelector<
   State,
   Result,
   ProjectorFn = DefaultProjectorFn<Result>
 > extends Selector<State, Result> {
   release(): void;
-  projector: ProjectorFn;
+  /** @deprecated A projector must take at least 1 argument */
```
brandonroberts, line 28: We shouldn't deprecate the current signatures just yet. We need to talk more about a plan to eventually remove the less-strict option if that's the long-term goal. Deprecating early is just going to cause lots of warnings and noise.

brandonroberts:
Needs a rebase

modules/store/src/selector.ts
david-shortman: Instead of deprecation, would it be more desirable to make strictness opt-in first? https://github.com/david-shortman/platform/commit/849e569cba8a511bab505f08d2c594ba73c5b36d

brandonroberts: If it doesn't disrupt existing code, then yes

david-shortman: Ok, I'll look to submit with that approach then.
I guess I could also include a schematic to opt-in all existing selectors, too.

brandonroberts: I don't think we need to try to migrate existing code at this point. In general I'd like the experience for testing selectors to be the same as today, just with the added benefit of type checking the projector

david-shortman: I'm now wondering if this is an opportunity to introduce a more clear API for createSelector which comes with the bonus of strict projectors https://github.com/ngrx/platform/discussions/3098#discussioncomment-1626403

brandonroberts: We're not planning to introduce any new APIs for selectors

david-shortman:
@brandonroberts The opt-in for strict projectors can take place in two ways:

```typescript
export const selector = createSelector(
        selectUrl,
        url => url,
        { strict: true } // specify a `strict` flag, which means the developer can allow the slices and return value to be inferred
      );
```

or

```typescript
export const selector = createSelector<[string], string>( // directly supply generic arguments
        selectUrl,
        url => url
      );
```

modules/store/src/selector.ts
```diff
@@ -17,13 +17,21 @@ export type ComparatorFn = (a: any, b: any) => boolean;
 
 export type DefaultProjectorFn<T> = (...args: any[]) => T;
 
+export type ProjectorFn<Args extends unknown[], T> = (...args: Args) => T;
+
+export type StrictMemoizedSelector<
```
david-shortman, line 22: this type is not exported from the index so that developers do not create a dependency on this type (presuming it can one day be removed, and all memoized selectors have strict projectors).

instead, if a dev needs to declare the type of a variable, they can use the form `MemoizedSelector<State, Result, (slices...Slices) => Result>`

modules/store/src/selector.ts
```diff
@@ -17,13 +17,33 @@ export type ComparatorFn = (a: any, b: any) => boolean;
...
 export type DefaultProjectorFn<T> = (...args: any[]) => T;
 
+type StrictMemoizedSelector<
+  State,
+  Slices extends unknown[],
+  Result
+> = MemoizedSelector<
+  State,
+  Result,
+  { strict: true } & ((...slices: Slices) => Result)
```
david-shortman, line 27: `{ strict: true }` is purely type information which is used by `MemoizedSelector` to determine which projector function is used.

david-shortman:
closing and reopening with updates once #3354 is merged

modules/store/src/selector.ts
david-shortman: Apparently this is known as an [opaque type](https://calebmer.com/tangents/opaque-types-in-typescript.html).

PR #3230, opened 2021-11-12
Strict mode fixes

Pingear:
just added some fixes for strict mode

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/example-app/src/app/auth/containers/login-page.component.ts
+++ b/projects/example-app/src/app/auth/containers/login-page.component.ts
@@ -9,7 +9,7 @@ import { LoginPageActions } from '@example-app/auth/actions';
   template: `
     <bc-login-form
       (submitted)="onSubmit($event)"
-      [pending]="pending$ | async"
+      [pending]="(pending$ | async)!"
       [errorMessage]="error$ | async"
     >
     </bc-login-form>
--- a/projects/example-app/src/app/auth/containers/login-page.component.ts
+++ b/projects/example-app/src/app/auth/containers/login-page.component.ts
@@ -9,7 +9,7 @@ import { LoginPageActions } from '@example-app/auth/actions';
   template: `
     <bc-login-form
       (submitted)="onSubmit($event)"
-      [pending]="(pending$ | async)!"
+      [pending]="(pending$ | async) === true"
       [errorMessage]="error$ | async"
     >
     </bc-login-form>
--- a/projects/example-app/src/app/books/components/book-search.component.ts
+++ b/projects/example-app/src/app/books/components/book-search.component.ts
@@ -11,7 +11,7 @@ import { Component, Output, Input, EventEmitter } from '@angular/core';
             matInput
             placeholder="Search for a book"
             [value]="query"
-            (keyup)="search.emit($any($event.target).value)"
+            (keyup)="onSearch($event)"
           />
         </mat-form-field>
         <mat-spinner
--- a/projects/example-app/src/app/books/components/book-search.component.ts
+++ b/projects/example-app/src/app/books/components/book-search.component.ts
@@ -11,7 +11,7 @@ import { Component, Output, Input, EventEmitter } from '@angular/core';
             matInput
             placeholder="Search for a book"
             [value]="query"
-            (keyup)="search.emit($event.target.value)"
+            (keyup)="search.emit($any($event.target).value)"
           />
         </mat-form-field>
         <mat-spinner
--- a/projects/example-app/src/app/books/containers/collection-page.component.ts
+++ b/projects/example-app/src/app/books/containers/collection-page.component.ts
@@ -15,7 +15,7 @@ import * as fromBooks from '@example-app/books/reducers';
       <mat-card-title>My Collection</mat-card-title>
     </mat-card>
 
-    <bc-book-preview-list [books]="books$ | async"></bc-book-preview-list>
+    <bc-book-preview-list [books]="(books$ | async)!"></bc-book-preview-list>
   `,
   /**
    * Container components are permitted to have just enough styles
--- a/projects/example-app/src/app/books/containers/collection-page.component.ts
+++ b/projects/example-app/src/app/books/containers/collection-page.component.ts
@@ -15,7 +15,7 @@ import * as fromBooks from '@example-app/books/reducers';
       <mat-card-title>My Collection</mat-card-title>
     </mat-card>
 
-    <bc-book-preview-list [books]="(books$ | async)!"></bc-book-preview-list>
+    <bc-book-preview-list [books]="(books$ | async) ?? []"></bc-book-preview-list>
   `,
   /**
    * Container components are permitted to have just enough styles
--- a/projects/example-app/src/app/books/containers/find-book-page.component.ts
+++ b/projects/example-app/src/app/books/containers/find-book-page.component.ts
@@ -13,9 +13,9 @@ import * as fromBooks from '@example-app/books/reducers';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-book-search
-      [query]="searchQuery$ | async"
-      [searching]="loading$ | async"
-      [error]="error$ | async"
+      [query]="(searchQuery$ | async)!"
+      [searching]="(loading$ | async)!"
+      [error]="(error$ | async)!"
       (search)="search($event)"
     >
     </bc-book-search>
--- a/projects/example-app/src/app/books/containers/find-book-page.component.ts
+++ b/projects/example-app/src/app/books/containers/find-book-page.component.ts
@@ -13,7 +13,7 @@ import * as fromBooks from '@example-app/books/reducers';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-book-search
-      [query]="(searchQuery$ | async)!"
+      [query]="(searchQuery$ | async) ?? ''"
       [searching]="(loading$ | async)!"
       [error]="(error$ | async)!"
       (search)="search($event)"
--- a/projects/example-app/src/app/books/containers/find-book-page.component.ts
+++ b/projects/example-app/src/app/books/containers/find-book-page.component.ts
@@ -14,7 +14,7 @@ import * as fromBooks from '@example-app/books/reducers';
   template: `
     <bc-book-search
       [query]="(searchQuery$ | async) ?? ''"
-      [searching]="(loading$ | async)!"
+      [searching]="(loading$ | async) === true"
       [error]="(error$ | async)!"
       (search)="search($event)"
     >
--- a/projects/example-app/src/app/books/containers/selected-book-page.component.ts
+++ b/projects/example-app/src/app/books/containers/selected-book-page.component.ts
@@ -12,7 +12,7 @@ import * as fromBooks from '@example-app/books/reducers';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-book-detail
-      [book]="book$ | async"
+      [book]="(book$ | async)!"
       [inCollection]="isSelectedBookInCollection$ | async"
       (add)="addToCollection($event)"
       (remove)="removeFromCollection($event)"


projects/example-app/src/app/auth/containers/login-page.component.ts
```diff
@@ -9,7 +9,7 @@ import { LoginPageActions } from '@example-app/auth/actions';
   template: `
     <bc-login-form
       (submitted)="onSubmit($event)"
-      [pending]="pending$ | async"
+      [pending]="(pending$ | async)!"
```
david-shortman, line 12: is this the same as
```suggestion
      [pending]="(pending$ | async) === true"
```

projects/example-app/src/app/books/containers/find-book-page.component.ts
```diff
@@ -13,9 +13,9 @@ import * as fromBooks from '@example-app/books/reducers';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-book-search
-      [query]="searchQuery$ | async"
-      [searching]="loading$ | async"
-      [error]="error$ | async"
+      [query]="(searchQuery$ | async)!"
```
david-shortman, line 16: is it good practice to assert streams as truthy? I don't see why this is needed as `searchQuery$` is defined as `Observable<string>`, not `Observable<string | undefined>`.

if this behavior is because async can emit `undefined` first, I feel like there's another way to solve as I totally develop apps in strict mode with `async` and haven't run into needing the bang...

Pingear: otherwise IDE swears -
Type 'string | null 'is not assignable to type' string '.
   Type 'null' is not assignable to type 'string'

david-shortman: what about
```suggestion
      [query]="(searchQuery$ | async) ?? ''"
```

in general, I think providing a value when the stream is falsy is better than using `!` to assert that it is not falsy because it avoids the issue that `async` really does output `null` as an initial value, which can cause problems in an input declares it only accepts `string`, for example.

david-shortman:
Is is possible to set the config for this example app to strict mode to ensure it continues to operate correctly?

projects/example-app/src/app/books/containers/collection-page.component.ts
```diff
@@ -15,7 +15,7 @@ import * as fromBooks from '@example-app/books/reducers';
       <mat-card-title>My Collection</mat-card-title>
     </mat-card>
 
-    <bc-book-preview-list [books]="books$ | async"></bc-book-preview-list>
+    <bc-book-preview-list [books]="(books$ | async)!"></bc-book-preview-list>
```
david-shortman, line 18:
```suggestion
    <bc-book-preview-list [books]="(books$ | async) ?? []"></bc-book-preview-list>
```

projects/example-app/src/app/books/containers/find-book-page.component.ts
```diff
@@ -13,9 +13,9 @@ import * as fromBooks from '@example-app/books/reducers';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-book-search
-      [query]="searchQuery$ | async"
-      [searching]="loading$ | async"
-      [error]="error$ | async"
+      [query]="(searchQuery$ | async) ?? ''"
+      [searching]="(loading$ | async)!"
+      [error]="(error$ | async)!"
```
david-shortman, line 18: maybe the `error` should have its type changed to support a falsy value since there may not be an error sometimes

projects/example-app/src/app/books/containers/find-book-page.component.ts
```diff
@@ -13,9 +13,9 @@ import * as fromBooks from '@example-app/books/reducers';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-book-search
-      [query]="searchQuery$ | async"
-      [searching]="loading$ | async"
-      [error]="error$ | async"
+      [query]="(searchQuery$ | async) ?? ''"
+      [searching]="(loading$ | async)!"
```
david-shortman, line 17:
```suggestion
      [query]="(searchQuery$ | async) ?? ''"
      [searching]="(loading$ | async) === true"
```

projects/example-app/src/app/books/containers/selected-book-page.component.ts
```diff
@@ -12,7 +12,7 @@ import * as fromBooks from '@example-app/books/reducers';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-book-detail
-      [book]="book$ | async"
+      [book]="(book$ | async)!"
```
david-shortman, line 15: may be the `book` should have its type changed to support a falsy value since the book may not be initialized

projects/example-app/src/app/books/components/book-search.component.ts
```diff
@@ -11,7 +11,7 @@ import { Component, Output, Input, EventEmitter } from '@angular/core';
             matInput
             placeholder="Search for a book"
             [value]="query"
-            (keyup)="search.emit($event.target.value)"
+            (keyup)="search.emit($any($event.target).value)"
```
david-shortman, line 14: it would be cool to not to have to cast to `any` here... but maybe such is life with JS events. but I coulda sworn there was sufficient typing for a `KeyboardEvent`.

david-shortman (commented):
some type suggestions

projects/example-app/src/app/auth/containers/login-page.component.ts
```diff
@@ -9,7 +9,7 @@ import { LoginPageActions } from '@example-app/auth/actions';
   template: `
     <bc-login-form
       (submitted)="onSubmit($event)"
-      [pending]="pending$ | async"
+      [pending]="(pending$ | async) === true"
```
markostanimirovic, line 12:
```suggestion
      [pending]="(pending$ | async)!"
```

projects/example-app/src/app/books/components/book-search.component.ts
markostanimirovic, line 14:
```suggestion
            (keyup)="onSearch($event)"
```

Add `onSearch` method:

```ts
  onSearch(event: KeyboardEvent): void {
    this.search.emit((event.target as HTMLInputElement).value);
  }
```

projects/example-app/src/app/books/containers/collection-page.component.ts
```diff
@@ -15,7 +15,7 @@ import * as fromBooks from '@example-app/books/reducers';
       <mat-card-title>My Collection</mat-card-title>
     </mat-card>
 
-    <bc-book-preview-list [books]="books$ | async"></bc-book-preview-list>
+    <bc-book-preview-list [books]="(books$ | async) ?? []"></bc-book-preview-list>
```
markostanimirovic, line 18:
```suggestion
    <bc-book-preview-list [books]="(books$ | async)!"></bc-book-preview-list>
```

projects/example-app/src/app/books/containers/find-book-page.component.ts
```diff
@@ -13,9 +13,9 @@ import * as fromBooks from '@example-app/books/reducers';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-book-search
-      [query]="searchQuery$ | async"
-      [searching]="loading$ | async"
-      [error]="error$ | async"
+      [query]="(searchQuery$ | async) ?? ''"
+      [searching]="(loading$ | async) === true"
```
markostanimirovic, line 17:
```suggestion
      [query]="(searchQuery$ | async)!"
      [searching]="(loading$ | async)!"
```

projects/example-app/src/app/books/containers/selected-book-page.component.ts
```diff
@@ -12,7 +12,7 @@ import * as fromBooks from '@example-app/books/reducers';
...
       [inCollection]="isSelectedBookInCollection$ | async"
```
markostanimirovic, line 16:
```suggestion
      [inCollection]="(isSelectedBookInCollection$ | async)!"
```

markostanimirovic: Also, replace line 15 in `app.component.ts` with:

```
<bc-sidenav [open]="(showSidenav$ | async)!" (closeMenu)="closeSidenav()">
```

markostanimirovic (changes requested):
To enable Angular strict mode for the `example-app`, add the following in `example-app/tsconfig.app.json`:

```json
  // line 8
  "angularCompilerOptions": {
    "strictTemplates":  true,
    "strictInjectionParameters": true,
    "strictInputAccessModifiers": true
  },
```

projects/example-app/src/app/auth/containers/login-page.component.ts
david-shortman: Do these have the same behavior?

markostanimirovic: Yes, but to be consistent, let's use the same syntax for all state observables that are unwrapped via `async` pipe.

Suggestion from Angular documentation: https://angular.io/guide/template-typecheck#strict-null-checks

david-shortman: Nice, that's a convenience I didn't know about.

david-shortman: According to that doc, they do not have the same behavior:

> In the template, include the non-null assertion operator ! at the end of a nullable expression, such as <user-detail [user]="user!"></user-detail>.
In this example, the compiler disregards type incompatibilities in nullability, just as in TypeScript code.

It's nbd, but since this is the example app we're choosing to do assertions which may be incorrect instead of truly "strict" handling of `async`'s emissions of `null`.

david-shortman: The doc further advises adjusting Inputs to components to say whether they can handle falsy values: 

> As a library author, you can take several measures to provide an optimal experience for your users. First, enabling strictNullChecks and including null in an input's type, as appropriate, communicates to your consumers whether they can provide a nullable value or not.

markostanimirovic: > they do not have the same behavior

If we still talk about `pending$`, both solutions will work the same: `(pending$ | async)!` and `(pending$ | async) === true` and the return type of both expressions will be `boolean`. I suggested the first solution, to be consistent with the others.

You quoted sections that are not related to the `async` pipe :) I meant this part:

> The async pipe currently assumes that the Observable it subscribes to can be asynchronous, which means that it's possible that there is no value available yet. In that case, it still has to return something—which is null. In other words, **the return type of the async pipe includes null, which might result in errors in situations where the Observable is known to emit a non-nullable value synchronously.**

---

> It's nbd, but since this is the example app we're choosing to do assertions which may be incorrect instead of truly "strict" handling of async's emissions of null.

In case of `pending$` or `books$` for example, `async` will never return `null`.

---

IMO, the best will be to create view model selectors for all pages and avoid non-null assertions at all, but that could be an idea for another PR :)

timdeschryver: Observables have `null` added to their return types because they're observables, and there might not be a value yet.
This isn't the case for selectors, because there will always be an initial value, unless a `null` or `undefined` value is explicitly returned by the selector.
In these cases, I think there's always a value, hence the suggestion to add a bang operator to these async pipes.

david-shortman: Thanks for the investigation on part of my knowledge gaps 😅

david-shortman: Knowing this, I wonder if the ngrxLet directive in coordination with the return type of store.select could meet strictness and know that whatever the observable's value is, is guaranteed.

david-shortman: Just thinking about that lib out loud, not saying it needs to be used here.

markostanimirovic: > Thanks for the investigation on part of my knowledge gaps 😅

No worries David, thank you for your efforts to make NgRx codebase better! :)

PR #3232, opened 2021-11-12
feat(store) use store.select with multiple selectors

### LINKED ISSUE #2839: ComponentStore: Allow multiple selector functions to store.select
Currently, `store.select` can either be passed a selector function, or alternately multiple selectors can be combined by passing 2 or more selectors with a project function.  It  would be convenient if each of the "combine selectors" could accept either a selector or a function, in order to concisely select multiple properties from the state that don't already have selectors defined:  

```ts
this.store.select(
  state => state.foo,
  state => state.bar,
  (foo, bar) => ({ foo, bar })
 );
```

Basically, if the select function encounters a function instead of an observable, all it would have to do is wrap the function internally with `this.select(selectorFn)`

Signature for the overload for two selectors would be something like
```ts
select<R, S1, S2>(
    s1: Observable<S1> | (s: T) => S1, 
    s2: Observable<S2> | (s: T) => S2,
    projector: (s1: S1, s2: S2) => R, 
   config?: SelectConfig
): Observable<R>;
```

```ts

this.store.select(
  this.store.select(state => state.foo),
  this.store.select(state => state.bar),
  (foo, bar) => ({ foo, bar })
 );
```

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2839

## What is the new behavior?

Can select from multiple selectors. Only distinct changes are output.

```typescript
// returns `Observable<number, number, number>`
const counterStateWithFunc = store.select(
    (s) => s.counter1,
    (s) => s.counter2,
    (s) => s.counter3
 );
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/src/store.ts
+++ b/modules/store/src/store.ts
@@ -246,13 +246,16 @@ export function select<T, Props, K>(
       mapped$ = source$.pipe(pluck(pathOrMapFn, ...(pathSlices as string[])));
     } else if (typeof propsOrPathOrMapFn === 'function') {
       mapped$ = source$.pipe(
-        map((source: T) => [
-          pathOrMapFn(source),
-          (propsOrPathOrMapFn as (state: T, props?: Props) => any)(source),
-          ...(
-            (pathsOrMapFns ?? []) as ((state: T, props?: Props) => any)[]
-          ).map((fn) => fn(source)),
-        ])
+        map(
+          (source: T) =>
+            new SelectedValues([
+              pathOrMapFn(source),
+              (propsOrPathOrMapFn as (state: T, props?: Props) => any)(source),
+              ...(
+                (pathsOrMapFns ?? []) as ((state: T, props?: Props) => any)[]
+              ).map((fn) => fn(source)),
+            ])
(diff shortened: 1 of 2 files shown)


modules/store/src/store.ts
```diff
@@ -246,6 +268,27 @@ export function select<T, Props, K>(
       );
     }
 
-    return mapped$.pipe(distinctUntilChanged());
+    return mapped$.pipe(
+      distinctUntilChanged((previous, current) => {
+        if (
+          current instanceof SelectedValues &&
+          previous instanceof SelectedValues
```
david-shortman, line 275: I needed to use a distinct class to hold multiple selected values so that they could be distinguished from a single selected array value.

brandonroberts:
I think the discussion for this PR should be had in the issue, and not in a PR. I'm not in favor of adding new APIs for this as I think combining should be done in selectors and not with `Store.select`.

PR #2115, opened 2019-09-10
Auto cleanup selectors

timdeschryver:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [TODO] Documentation has been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Follow up to #2044 

## What is the new behavior?

After each test, we'll clean the mock selectors. 

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

## Other information

First commit, shows the failing test
With the introduction of the auto cleanup in the second commit, the test passes

/cc @jtcrowson

PR #3245, opened 2021-11-19
fix(store): remove afterEach hook in mock store

### LINKED ISSUE #3243: Error when executing test with jest and jest-circus as testRunner

I am not sure than this issue is related to ngrx, angular or jest package. My apologize if I am in the wrong place

1. Clone this repo: github.com:cedricduffournet/test-ngrx-error.git
2. Install dependencies
```
yarn install
```
3. Run jest test

```
yarn jest
```

The following error occur in all tests with provideMockStore added to providers
```
 Cannot configure the test module when the test module has already been instantiated. Make sure you are not using `inject` before `R3TestBed.configureTestingModule`.
 beforeEach(() => {  
                ^
```

If i set `jest-jasmine2` as `testRunner` in jest config, I don't have any error.
I did not have this issue in  version 12 of angular and ngrx

## Expected behavior:
Test should pass without error

Angular 13, ngrx 13.0.0

cedricduffournet:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3243

```
[x] Yes
[] No
```
`MockStore` will not reset all of the mocked selectors after each test. User will need to call the` MockStore.resetSelectors()` method manually in a `afterEach` hook of a `describe` block


cedricduffournet:
don't know why ci/circleci: schematics-core-check failed on CircleCI

modules/store/testing/src/mock_store.ts
```diff
@@ -18,9 +18,13 @@ import { MOCK_SELECTORS } from './tokens';
 if (typeof afterEach === 'function') {
   afterEach(() => {
     try {
-      const mockStore: MockStore | undefined = TestBed.inject(MockStore);
-      if (mockStore) {
-        mockStore.resetSelectors();
+      const TestBed = getTestBed() as any;
```
brandonroberts, line 21: Add a comment here if why this is needed, so we'll have some context later when we can potentially remove it.

brandonroberts (changes requested):

modules/store/testing/src/mock_store.ts
cedricduffournet: Ok, not sur if my english is good enough, but I tried ...

modules/store/testing/src/mock_store.ts
```diff
@@ -18,9 +18,21 @@ import { MOCK_SELECTORS } from './tokens';
...
+       * Since angular v13, destroyAfterEach part of the teardown option object is enabled by default
+       * which means that at this point TestBed has already been cleaned up.
+       * If destroyAfterEach is enabled, injecting MockStore in TestBed will cause an error when testing with jest
+       *
+       * More info about testing module teardown: https://dev.to/this-is-angular/improving-angular-tests-by-enabling-angular-testing-module-teardown-38kh
+       * Issue related to these changes: https://github.com/ngrx/platform/issues/3243
+       */
+      const TestBed = getTestBed() as any;
+      const shouldResetSelectors = !TestBed.shouldTearDownTestingModule();
+      if (shouldResetSelectors) {
```
timdeschryver, line 31: I think that because we don't reset the selectors anymore, that the selectors keep their "old" value.
What's the consequence of doing this? Should we provide another reset strategy?

brandonroberts: We reset the selectors every time you instantiate the mock store anyway, so I'm wondering about when we'd actually need this also.

cedricduffournet: Seems to be related to this: 
issue: https://github.com/ngrx/platform/issues/2044
PR: https://github.com/ngrx/platform/pull/2115

And from documentation:

> Note: MockStore will reset all of the mocked selectors after each test (in the afterEach() hook) by calling the MockStore.resetSelectors() method.

cedricduffournet: executing mock_store.spec.ts pointed me out issue related to selectors not rese  ( not sur why all test pass in CI)

timdeschryver: @brandonroberts we only reset the selectors that are included in a mock store instance.
With the previous implementation, that instance would be the same in the `afterEach` hook, meaning that we were we able to reset the selectors here.
With the current behavior, I don't think that `resetSelectors` is invoked (with teardown enabled), and thus that selectors won't be cleaned up in between tests.
For example, these tests are passing, while I think they should fail (and I think that these are failing on the previous version).

```ts
  it('set selectors', () => {
    store.overrideSelector(selectCollectionBookIds, [2, 3, 4] as any);
    expect(selectCollectionBookIds({})).toEqual([2, 3, 4] as any);
  });

  it('reset selectors', () => {
    // I would expect this to fail, but the selector doesn't reset in between tests
    expect(selectCollectionBookIds({})).toEqual([2, 3, 4] as any);
  });
```

timdeschryver: Can we can remove this hook and tackle it if it becomes a problem?
A possible solution to keep this functionality is storing the mocked selectors in a global variable, this will be the simplest solution I think.

cedricduffournet: Yes, I'm not sure what you mean and how to handle this. Should I close this PR and let you fix this ?

timdeschryver: @cedricduffournet you can leave the PR open, I'm suggesting to remove the `afterEach` hook until it becomes a problem and then we can think of a solution that we're happy with. 
If @brandonroberts approves, you can remove that hook in this PR.

brandonroberts: I'm good with that

modules/store/testing/spec/mock_store.spec.ts
```diff
@@ -541,81 +541,4 @@ describe('Cleans up after each test', () => {
...
-      ],
-    });
-
-    const store = TestBed.inject(Store);
-    store.pipe(select(selectData)).subscribe((v) => {
-      expect(v).toBe(300);
-      done();
-    });
-  });
-});
```
timdeschryver, line 562: What's the reason that this test is removed?
I think it can be re-added and that it should pass.

timdeschryver (commented):
I left one question, the rest of the PR seems good to me.

modules/store/testing/spec/mock_store.spec.ts
cedricduffournet: I think this is the reason why this hook has been introduced (more info : https://github.com/ngrx/platform/issues/2044 and https://github.com/ngrx/platform/pull/2115)
If selector are not reset, this test fail because selectData return 200, which is the value set in previous test ...

timdeschryver: Oh I didn't notice that, and you're right.
If that's the case, then I think we can completely remove the `Cleans up after each test` block.

cedricduffournet: @timdeschryver I removed useless test.
I wonder if this PR does not introduce a breaking change. Some test that passed before will end up in error.

timdeschryver: Correct, this is indeed a breaking change.
But leaving it as is will also be a breaking change.
If I'm not mistaken, the suggested implementation has the same outcome removing this code.

That's why I'm suggesting to just remove it, which will already help projects and think of a good solution if someone bumps against this problem. I'm thinking that in the majority of the cases that this won't be a problem, I'm assuming that if you're mocking selectors in a test, that you're also mocking those selectors in the following test.

timdeschryver: Update: I verified that the same tests are failing with the suggested implementation:

```ts
      const TestBed = getTestBed() as any;
      const shouldResetSelectors = !TestBed.shouldTearDownTestingModule();
      if (shouldResetSelectors) {
        const mockStore: MockStore | undefined = TestBed.inject(MockStore);
        if (mockStore) {
          mockStore.resetSelectors();
        }
      }
```

cedricduffournet: Yes, I had the the problem with the previous code, and removing the hook seems for me as well the best solution for now. 
Thanks for you help

timdeschryver: No, thank you! 🤗

timdeschryver (approved):

cedricduffournet:
@timdeschryver maybe we should remove or update the following text in the doc (https://ngrx.io/guide/store/testing) :

> Note: MockStore will reset all of the mocked selectors after each test (in the afterEach() hook) by calling the MockStore.resetSelectors() method.

timdeschryver:
@cedricduffournet good catch!

cedricduffournet:
@timdeschryver I updated the doc and added a test, maybe not necessary,  to show how reset selectors in afterEach hook. Let me know if you prefer that I remove it

projects/ngrx.io/content/guide/store/testing.md
```diff
@@ -75,7 +75,7 @@ In this example based on the [walkthrough](guide/store/walkthrough), we mock the
 
 <div class="alert is-helpful">
 
-**Note:** `MockStore` will reset all of the mocked selectors after each test (in the `afterEach()` hook) by calling the `MockStore.resetSelectors()` method.
+**Note:** `MockStore` will **not** reset mocked selectors after each test. You can reset selectors by calling the `MockStore.resetSelectors()` method in `afterEach()` hook.
```
timdeschryver, line 78: I think we can remove this alert, and show how you can reset the store manually after each test.
About the re-added tests. I think we're already testing that  `resetSelector` does what it's supposed to do, so these are redundant and can be removed. 
Thoughts?

cedricduffournet: Ok, that sounds good to me. I was not sure about this test.
I'll add a code example in the doc

projects/ngrx.io/content/guide/store/testing.md
```diff
@@ -73,11 +73,9 @@ Usage:
 
 In this example based on the [walkthrough](guide/store/walkthrough), we mock the `selectBooks` selector by using `overrideSelector`, passing in the `selectBooks` selector with a default mocked return value of an array of books. Similarly, we mock the `selectBookCollection` selector and pass the selector together with another array. In the test, we use `setResult()` to update the mock selectors to return new array values, then we use `MockStore.refreshState()` to trigger an emission from the `selectBooks` and `selectBookCollection` selectors.
 
-<div class="alert is-helpful">
-
-**Note:** `MockStore` will reset all of the mocked selectors after each test (in the `afterEach()` hook) by calling the `MockStore.resetSelectors()` method.
+You can reset selectors by calling the `MockStore.resetSelectors()` method in `afterEach()` hook.
```
timdeschryver, line 76:
```suggestion
You can reset selectors by calling the `MockStore.resetSelectors()` method in the `afterEach()` hook.
```

projects/ngrx.io/content/examples/store-walkthrough/src/app/tests/app.component.1.spec.ts
```diff
@@ -96,3 +97,93 @@ describe('AppComponent', () => {
...
+          id: 'realId',
+          volumeInfo: {
+            title: 'Real Title',
+            authors: ['Real Author'],
+          },
+        },
+      ]);
+      done();
+    });
+  });
```
timdeschryver, line 187: I think it would be better to leave this out, to keep the example concise.
```suggestion
```

projects/ngrx.io/content/examples/store-walkthrough/src/app/tests/app.component.1.spec.ts
```diff
@@ -96,3 +97,93 @@ describe('AppComponent', () => {
     //#enddocregion mockSelector
   });
 });
+
+//#docregion resetMockSelector
+describe('AppComponent reset selectors', () => {
+  let store: MockStore<AppState>;
```
timdeschryver, line 103: Just like `Store`, we want to remove the generic.

```suggestion
  let store: MockStore;
```

projects/ngrx.io/content/examples/store-walkthrough/src/app/tests/app.component.1.spec.ts
```diff
@@ -96,3 +97,93 @@ describe('AppComponent', () => {
...
+        provideMockStore({
+          initialState: [
+            {
+              id: 'firstId',
+              volumeInfo: {
+                title: 'Initial Title',
+                authors: ['Initial Author'],
+              },
+            },
+          ],
```
timdeschryver, line 121: When mocking selectors, the state doesn't matter.
So, I would prefer to leave it out.

```suggestion
```

projects/ngrx.io/content/examples/store-walkthrough/src/app/tests/app.component.1.spec.ts
```diff
@@ -96,3 +97,93 @@ describe('AppComponent', () => {
...
+                    authors: ['Mocked Author'],
+                  },
+                },
+              ],
+            },
+          ],
+        }),
+      ],
+    });
+
```
timdeschryver, line 139: We need to grab an instance of `store` here.

```suggestion

const store = TestBed.inject(MockStore);

```

projects/ngrx.io/content/examples/store-walkthrough/src/app/tests/app.component.1.spec.ts
```diff
@@ -96,3 +97,93 @@ describe('AppComponent', () => {
     //#enddocregion mockSelector
   });
 });
+
+//#docregion resetMockSelector
+describe('AppComponent reset selectors', () => {
+  let store: MockStore<AppState>;
+
+  afterEach(() => {
+    store.resetSelectors();
```
timdeschryver, line 106: Just for safety.

```suggestion
    store?.resetSelectors();
```

timdeschryver (commented):
Thanks again @cedricduffournet .
I left a few comments to make the example more compact.

timdeschryver (approved):
Thanks for your patience and thoroughness @cedricduffournet .

brandonroberts (approved):

PR #3255, opened 2021-11-28
fix(component): fixes recursive rendering

### LINKED ISSUE #3246: *ngrxLet causes ASSERTION ERROR: Reached the max number of directives [Expected=> 1 != 1 <=Actual]
The issue seems to be when the observable emits while angular is still constructing the template.

Here's an example to reproduce

https://github.com/e-davidson/angular-issue-reproduction-2021

## Expected behavior:

The component should render with no errors.

NgRx 13, Angular 13, Node 14.17, OS wsl, browser chrome

## Other information:

I also submitted an issue to the angular team
https://github.com/angular/angular/issues/44231

If they do fix the issue this would still be a problem since createEmbeddedView will have been called an extra time when it shouldn't due to the recurssion.

I believe it can be easily fixed by not relying on the embeddedView to be null and instead have a field if it's the first time or not.

See here
https://github.com/ngrx/platform/blob/master/modules/component/src/let/let.directive.ts#L121

Sure, I'd love to help if the team feels this is correct.

[x ] Yes (Assistance is provided if you need help submitting a pull request)
[ ] No

e-davidson:
Handles edge case where rendering causes the observable to emit which causes
`ASSERTION ERROR: Reached the max number of directives [Expected=> 1 != 1 <=Actual]`

Change the logic to set `embeddedViewCreated` before calling `createEmbeddedView`, so if `ensureEmbeddedViewCreated` is called recursively it won't call `createEmbeddedView` a 2nd time. Also, we're calling `ensureEmbeddedViewCreated` after we setup the view context because if it gets called recursively the previous call stack will override the later value.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3246 

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component/src/let/let.directive.ts
+++ b/modules/component/src/let/let.directive.ts
@@ -94,7 +94,7 @@ export class LetDirective<U> implements OnDestroy {
   // eslint-disable-next-line @typescript-eslint/naming-convention
   static ngTemplateGuard_ngrxLet: 'binding';
 
-  private embeddedviewCreated = false;
+  private embeddedViewCreated = false;
   private readonly viewContext: LetViewContext<U | undefined | null> = {
     $implicit: undefined,
     ngrxLet: undefined,
--- a/modules/component/src/let/let.directive.ts
+++ b/modules/component/src/let/let.directive.ts
@@ -94,7 +94,7 @@ export class LetDirective<U> implements OnDestroy {
   // eslint-disable-next-line @typescript-eslint/naming-convention
   static ngTemplateGuard_ngrxLet: 'binding';
 
-  private embeddedView: any;
+  private embeddedviewCreated = false;
   private readonly viewContext: LetViewContext<U | undefined | null> = {
     $implicit: undefined,
     ngrxLet: undefined,
@@ -107,7 +107,7 @@ export class LetDirective<U> implements OnDestroy {
   private readonly resetContextObserver: NextObserver<void> = {
     next: () => {
       // if not initialized no need to set undefined
-      if (this.embeddedView) {
+      if (this.embeddedviewCreated) {
         this.viewContext.$implicit = undefined;
         this.viewContext.ngrxLet = undefined;
         this.viewContext.$error = false;
@@ -117,26 +117,17 @@ export class LetDirective<U> implements OnDestroy {
   };
   private readonly updateViewContextObserver: Observer<U | null | undefined> = {
     next: (value: U | null | undefined) => {
-      // to have init lazy
-      if (!this.embeddedView) {
-        this.createEmbeddedView();
-      }
       this.viewContext.$implicit = value;
       this.viewContext.ngrxLet = value;
+      this.ensureEmbeddedViewCreated();
     },
     error: (error: Error) => {
-      // to have init lazy
-      if (!this.embeddedView) {
-        this.createEmbeddedView();
-      }
       this.viewContext.$error = true;
+      this.ensureEmbeddedViewCreated();
     },
     complete: () => {
-      // to have init lazy
-      if (!this.embeddedView) {
-        this.createEmbeddedView();
-      }
       this.viewContext.$complete = true;
+      this.ensureEmbeddedViewCreated();
     },
   };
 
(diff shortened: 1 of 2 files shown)


modules/component/src/let/let.directive.ts
```diff
@@ -94,7 +94,7 @@ export class LetDirective<U> implements OnDestroy {
   // eslint-disable-next-line @typescript-eslint/naming-convention
   static ngTemplateGuard_ngrxLet: 'binding';
 
-  private embeddedView: any;
+  private embeddedviewCreated = false;
```
markostanimirovic, line 97:
```suggestion
  private embeddedViewCreated = false;
```

or

```suggestion
  private isEmbeddedViewCreated = false;
```

modules/component/src/let/let.directive.ts
```diff
@@ -168,11 +159,14 @@ export class LetDirective<U> implements OnDestroy {
     this.subscription = this.cdAware.subscribe({});
   }
 
-  createEmbeddedView() {
-    this.embeddedView = this.viewContainerRef.createEmbeddedView(
-      this.templateRef,
-      this.viewContext
-    );
+  ensureEmbeddedViewCreated() {
```
markostanimirovic, line 162:
```suggestion
  private ensureEmbeddedViewCreated(): void {
```

markostanimirovic (changes requested):

modules/component/src/let/let.directive.ts
```diff
@@ -168,11 +159,14 @@ export class LetDirective<U> implements OnDestroy {
...
+    if (!this.embeddedviewCreated) {
+      this.embeddedviewCreated = true;
+      this.viewContainerRef.createEmbeddedView(
+        this.templateRef,
+        this.viewContext
+      );
+    }
```
markostanimirovic, line 169: Removing public `createEmbeddedView` method is a breaking change (if someone uses `LetDirective` outside of the template 👀). Let's keep it for now and remove `ensureEmbeddedViewCreated`:

```ts
  createEmbeddedView() {
    this.isEmbeddedViewCreated = true;
    this.viewContainerRef.createEmbeddedView(
      this.templateRef,
      this.viewContext
    );
  }
```

modules/component/src/let/let.directive.ts
```diff
@@ -117,26 +117,17 @@ export class LetDirective<U> implements OnDestroy {
   };
   private readonly updateViewContextObserver: Observer<U | null | undefined> = {
     next: (value: U | null | undefined) => {
-      // to have init lazy
-      if (!this.embeddedView) {
-        this.createEmbeddedView();
-      }
       this.viewContext.$implicit = value;
       this.viewContext.ngrxLet = value;
+      this.ensureEmbeddedViewCreated();
```
markostanimirovic, line 122: Is there any reason why the `createEmbeddedView` is moved to the end of the callback?

markostanimirovic (changes requested):

modules/component/src/let/let.directive.ts
e-davidson: Yes, if the recursion happens you want to make sure you're setting the latest value last.

I'm trying to avoid the following problem.

`next(0)` called 
Call `this.createEmbeddedView()` 
Set `isEmbeddedviewCreated = true`
Call `viewContainerRef.createEmbeddedView()`
`next(1)` called
Call `this.createEmbeddedView()` 
return to `next(1)`
Set `this.viewContext.ngrxLet = value` (value = 1)
return to `viewContainerRef.createEmbeddedView`
return to `this.createEmbeddedView` 
return to `next(0)`
Set `this.viewContext.ngrxLet = value` (value = 0) OOPS!

modules/component/src/let/let.directive.ts
e-davidson: Thanks, done.

modules/component/src/let/let.directive.ts
e-davidson: Good catch, I put the logic back to the way it was before except for changing order of setting the view context before rendering. Otherwise, the test I added would fail.

On a side point, outside the scope of this PR.
Would it make sense to deprecate and remove the function from the public API in the future?
If you call the function from your component the end result would be the content will get rendered twice. It's very hard to believe that would be desirable behavior.
The one use case that I can see where calling it yourself may be helpful is if you want your template to render even though the observable hadn't yet emitted. To support that scenario we can still stop you from rendering a second time by moving the `if(isEmbeddedviewCreated)`  logic to `createEmbeddedView`. Alternatively we can offer a config option to support this scenario.
We can create a new feature request if that makes sense.

markostanimirovic: Yes, I agree that the `createEmbeddedView` method should not be public. Extending the `LetDirective` is not recommended approach at all. However, someone can do that, and we have to ensure that there are no breaking changes in minor/patch version updates.

markostanimirovic (approved):
LGTM 👍 Thanks @e-davidson!

timdeschryver:
I'm going to trigger the CI by closing and reopening this PR..

timdeschryver (approved):

timdeschryver:
Thanks @e-davidson

PR #3257, opened 2021-12-02
Update my-counter.component.ts

KthProg:
Per the default Angular eslint rules, the non-generic Store should be used instead of the generic Store\<T\>
https://github.com/timdeschryver/eslint-plugin-ngrx/blob/main/docs/rules/no-typed-global-store.md

## PR Checklist

- [ x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit
- [ x] Tests for the changes have been added (for bug fixes / features)
- [ x] Documentation has been added / updated (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[ ] Feature
[ x] Code style update (formatting, local variables)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The generic Store\<T\> is used instead of Store

## What is the new behavior?
Store is used instead of the generic Store\<T\>

```
[ ] Yes
[ x] No
```

### DIFF
--- a/projects/ngrx.io/content/examples/store/src/app/my-counter/my-counter.component.ts
+++ b/projects/ngrx.io/content/examples/store/src/app/my-counter/my-counter.component.ts
@@ -10,7 +10,7 @@ import { increment, decrement, reset } from '../counter.actions';
 export class MyCounterComponent {
   count$: Observable<number>;
 
-  constructor(private store: Store<{ count: number }>) {
+  constructor(private store: Store) {
     this.count$ = store.select('count');
   }
 


projects/ngrx.io/content/examples/store/src/app/my-counter/my-counter.component.ts
```diff
@@ -10,7 +10,7 @@ import { increment, decrement, reset } from '../counter.actions';
 export class MyCounterComponent {
   count$: Observable<number>;
 
-  constructor(private store: Store<{ count: number }>) {
+  constructor(private store: Store) {
```
timdeschryver, line 13: For this case, it's actually "required".
Because the example doesn't use a selector, but simply plucks the `count` state, the generic is required to have the extra type safety. E.g. if you would do `store.select('coun')`, TypeScript would catch this and throw an error.

PR #3297, opened 2022-01-21
Adding a feature selector to the getSelectors method in generated files from entity schematics

### LINKED ISSUE #3270: [@ngrx/schematics] Add Feature Selector and pass it to getSelectors when generating entity

When using ngrx schematics to generate an entity, you can provide a module to register the state as a feature in that module, and it generates a feature key to be added to an action reducer map, but one thing missing that developers must always do when generating this creates a feature selector and pass it to the "getSelectors" method at the bottom of the reducer file to ensure it returns the correct state (feature state) rather than the root app state.

## Expected behavior:

To create a feature selector and pass it to the "getSelectors" method.

[ ] No

PixelatedFigure:
I believe this addition will help as most entities are slices from the root state and this will help with typing and correctly retrieving the state slice. discussed in #3270

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, generating an entity using "ng generate entity" will generate a reducer that contains the entity selectors obtained by adapter.getSelectors that are not properly typed considering the entity is usually a slice from the root state

Closes #3270 

## What is the new behavior?

Added feature selector and passed to the adapter method

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/schematics/src/entity/creator-files/__name@dasherize@if-flat__/__name@dasherize@group-reducers__.reducer.ts.template
+++ b/modules/schematics/src/entity/creator-files/__name@dasherize@if-flat__/__name@dasherize@group-reducers__.reducer.ts.template
@@ -1,4 +1,4 @@
-import { Action, createReducer, on } from '@ngrx/store';
+import { Action, createFeatureSelector, createReducer, on } from '@ngrx/store';
 import { EntityState, EntityAdapter, createEntityAdapter } from '@ngrx/entity';
 import { <%= classify(name) %> } from '<%= featurePath(group, flat, "models", dasherize(name)) %><%= dasherize(name) %>.model';
 import * as <%= classify(name) %>Actions from '<%= featurePath(group, flat, "actions", dasherize(name)) %><%= dasherize(name) %>.actions';
@@ -49,9 +49,11 @@ export const reducer = createReducer(
   ),
 );
 
+export const <%= pluralize(name) %>Feature = createFeatureSelector<State>(<%= pluralize(name) %>FeatureKey);
+
 export const {
   selectIds,
   selectEntities,
   selectAll,
   selectTotal,
-} = adapter.getSelectors();
+} = adapter.getSelectors(<%= pluralize(name) %>Feature);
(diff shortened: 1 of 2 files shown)


modules/schematics/src/entity/creator-files/__name@dasherize@if-flat__/__name@dasherize@group-reducers__.reducer.ts.template
```diff
@@ -49,9 +49,11 @@ export const reducer = createReducer(
   ),
 );
 
+export const <%= pluralize(name) %>Feature = createFeatureSelector<State>(<%= pluralize(name) %>FeatureKey);
```
timdeschryver, line 52: I'm not a big fan of using `createFeatureSelector` (and selectors in general) in a reducer file.
The only exception is `createFeature`.
Leaving this open so someone else can give their opinion.

PixelatedFigure: It does seem to make the file have multiple responsibilities, but you cant have a reducer without selectors, I think of them as getters and setters for the state so, in my opinion, I always put them in the same file

brandonroberts: I'm in agreement with Tim on this one. Feature selectors should be connected elsewhere as it makes the reducer file more portable on its own.

PixelatedFigure: The only purpose of the feature selector is to be passed to the adapter's method so the typing is corrected, otherwise the selectors wont properly be typed.
Feel free to make a decision regarding this pull request, thought this would be a positive addition to this library.

markostanimirovic: @PixelatedFigure AFAIK, entity selectors with or without passed feature selector will be properly typed.

I think we can close this PR and move the discussion to #3270.

PR #3328, opened 2022-02-23
perf(component): reset state / trigger CD only if necessary

### LINKED ISSUE #3327: Push pipe calls detectChanges before setting value
https://stackblitz.com/edit/angular-ivy-ubkqrk?file=src%2Fapp%2Fapp.component.ts

## Expected behavior:
First detect changes should be called after value is set

[ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: Performance improvements
```

1. When the first observable that emits the value synchronously arrives, the change detection will be triggered two times. This causes an error when `changeDetectorRef.detectChanges` is called in zoneless environment (reported in #3327).
2. When another observable that emits the value synchronously arrives, the same will happen.

Closes #3327 

## What is the new behavior?

For both scenarios 👆 the change detection will be triggered once. The behavior for observables that emit asynchronously will remain the same..

```
[ ] Yes
[x] No
```


modules/component/src/core/utils/emits-synchronously.ts
```diff
@@ -0,0 +1,18 @@
+import { Observable } from 'rxjs';
+
+/**
+ * @description
+ * Checks whether the Observable emits next, error, or complete event synchronously.
+ */
+export function emitsSynchronously(observable$: Observable<unknown>): boolean {
+  let emitsSync = false;
+  observable$
+    .subscribe({
```
SebastianPodgajny, line 10: I think this may lead to unexpected behaviour like side effects that will run twice or may cause performance issue if there is sync cpu heavy computation in stream
IMO value that is received here should be reused as renderedValue

SebastianPodgajny: Something like this comes to my mind
```typescript
return new Observable<U | undefined | null>(observer => {
        let emitSync = false;
        const subscription = observable$.subscribe({
          next: value => {
            emitSync = true;
            observer.next(value);
          },
          error: err => {
            emitSync = true;
            observer.error(err);
          },
          complete: () => {
            emitSync = true;
            observer.complete();
          },
        });

        // If a new Observable arrives, reset the value to render_creator
        // only if the new Observable doesn't emit the value synchronously
        if (!emitSync) {
          cfg.resetContextObserver.next();
          cfg.render();
        }

        return subscription;
      }).pipe(
        distinctUntilChanged(),
        tap(cfg.updateViewContextObserver),
        tap(() => cfg.render()),
        ...
 ```
 WDYT?

Comment (markostanimirovic) [reply modules/component/src/core/utils/emits-synchronously.ts:10]: Thanks @SebastianPodgajny! The issue with `PushPipe` will be solved with this proposal. However, both solutions won't work properly for `LetDirective`.

Let's try to find a better solution :)

File: modules/component/src/core/render-creator.ts [hunk 2]
```diff
@@ -1,13 +1,11 @@
...
 
-import { hasZone } from '../utils/has-zone';
-
 export interface RenderConfig {
   ngZone: NgZone;
   cdRef: ChangeDetectorRef;
 }
 
-export function createRender<T>(config: RenderConfig): () => void {
+export function createRender(config: RenderConfig): () => void {
```
brandonroberts, line 8: This was existing but seems like this function should have a different name like `createRenderDetection` or something similar that's more accurate of what its doing.

brandonroberts (commented):
LGTM on code, but if this changes the current behavior of two detection cycles to one, might be an unintentional breaking change for current users.

markostanimirovic:
> LGTM on code, but if this changes the current behavior of two detection cycles to one, might be an unintentional breaking change for current users.

Sure, let's postpone this change for v14.

I also have in mind a few more improvements, such as better typing for `LetDirective`, which would also be a breaking change.

timdeschryver (approved):
Nice work!

modules/component/src/core/render-scheduler.ts
```diff
@@ -0,0 +1,32 @@
...
+): RenderScheduler {
+  function schedule(): void {
+    if (hasZone(config.ngZone)) {
+      config.cdRef.markForCheck();
+    } else {
+      config.cdRef.detectChanges();
+    }
+  }
+
+  return { schedule };
```
timdeschryver, line 23: Why do we wrap `schedule` in an object here?

markostanimirovic: Good point! Creating `RenderScheduler` as object will make it flexible for further improvements. E.g. if we decide to add `local` and `global` strategy, `RenderScheduler` will contain `flush` method. Something like this:

```ts
export interface RenderScheduler {
  schedule(): void;
  flush(): void;
}

export function createRenderScheduler(config: RenderSchedulerConfig): RenderScheduler {
  return config.strategy === RenderStrategy.Local
    ? createLocalRenderScheduler(config)
    : createGlobalRenderScheduler(config);
}

function createLocalRenderScheduler(config: RenderSchedulerConfig): RenderScheduler {
  const scheduledLocalRenders = new ScheduledLocalRenders();
  const context = getChangeDetectorRefContext(config.cdRef);

  function schedule(): void {
    flush();

    const scheduledRender = animationFrameScheduler.schedule(() => {
      config.cdRef.detectChanges();
    });
    scheduledLocalRenders.set(context, scheduledRender);
  }

  function flush(): void {
    const scheduledRender = scheduledLocalRenders.get(context);
    
    if (scheduledRender) {
      scheduledRender.unsubscribe();
      scheduledLocalRenders.delete(context);
    }
  }

  return { schedule, flush };
}
```

`PushPipe`/`LetDirective`:

```ts
export class PushPipe implements OnDestroy {
  ngOnDestroy(): void {
    this.subscription.unsubscribe();
    this.renderScheduler.flush();
  }
}
```

timdeschryver (approved):

modules/component/src/let/let.directive.ts
```diff
@@ -148,35 +163,37 @@ export class LetDirective<U> implements OnDestroy {
...
+  ) {}
 
-  createEmbeddedView() {
-    this.isEmbeddedViewCreated = true;
-    this.viewContainerRef.createEmbeddedView(
-      this.templateRef,
-      this.viewContext
+  ngOnInit(): void {
+    this.subscription.add(
+      this.renderEventManager.handlePotentialObservableChanges().subscribe()
```
markostanimirovic, line 180: After playing around with the suspense template, I realized that the potential observable subscription has to be moved to the `ngOnInit` method in order to get initial input values on directive initialization.

brandonroberts (approved):

PR #3356, opened 2022-03-25
fix(component-store): fix memoization not working when passing selectors directly to select

P4:
This makes selectors created with `createSelector` usable with `ComponentStore.select`
which looks like it should work, but would previously result in the selector getting recomputed each time.

When select method is called with a single argument, the projector function was being passed directly to `map`,
which invokes its callback with an extra `index` argument. This is basically the same problem as the JS gotcha 
where `[1,2,3].map(parseInt)` returns `[1, NaN, NaN]`.

Because the value of `index` is different each time, the memoized selector would get invalidated even if the actual state remained exactly the same.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Given
```
const memoizedSelector = createSelector(...)
```
passing it to a component store directly
```
componentStore.select(memoizedSelector)
```
results in the selector being re-computed on each store update, even if the state does not actually change.

The memoization works fine again if the selector is wrapped inside an extra function:
```
componentStore.select(state => memoizedSelector(state))
```

## What is the new behavior?
Both usages of `select()` shown above should now work the same way.

```
[ ] Yes
[x] No
```


modules/component-store/src/component-store.ts
```diff
@@ -244,7 +244,7 @@ export class ComponentStore<T extends object> implements OnDestroy {
     if (observables.length === 0) {
       observable$ = this.stateSubject$.pipe(
         config.debounce ? debounceSync() : (source$) => source$,
-        map(projector)
+        map((arg) => projector(arg))
```
timdeschryver, line 247: Thanks for creating this PR.
Let's give `arg` a name.

```suggestion
        map((state) => projector(state))
```

P4: Done, I got a bit too influenced by `projectorArgs` below.

brandonroberts (approved):

timdeschryver (approved):

timdeschryver:
Thanks @P4 ! 🏆

PR #3368, opened 2022-04-07
feat(component-store): add OnStoreInit and OnStateInit lifecycle hooks

brandonroberts:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3335 

## What is the new behavior?

Adds two new lifecycle hooks to ComponentStore along with their respective interfaces

- OnStoreInit - is called after the store is created
  - This method throws an error when trying to access the `.get()` method because the state hasn't been set, even if done eagerly.
- OnStateInit - is called after the state is initialized

Adds a new exported function `provideComponentStore(ComponentStore)` that is used to provide the ComponentStore, instantiate it, and run the lifecycle hooks if defined

```ts
import {
  ComponentStore,
  OnStateInit,
  OnStoreInit,
} from '@ngrx/component-store';

@Injectable()
export class LoginPageStore
  extends ComponentStore<{ test: boolean }>
  implements OnStoreInit, OnStateInit
{
  constructor() {
    super({ test: false });
  }

  ngrxOnStoreInit() {
    console.log('store init');
  }

  ngrxOnStateInit() {
    console.log('state init', this.get());
  }
}
```

```
[ ] Yes
[x] No
```

## Other information

- Example App changes are only used for reference and will be reverted


modules/component-store/src/component-store.ts
```diff
@@ -32,6 +32,14 @@ export interface SelectConfig {
   debounce?: boolean;
 }
 
+export interface OnStoreInit {
+  readonly ngrxOnStoreInit: () => void;
+}
+
+export interface OnStateInit {
+  readonly ngrxOnStateInit: () => void;
```
timdeschryver, line 40: Just a question about this method, the rest looks fine to me.
Should this also contain the current `state`?
A use case that I can think of would be to invoke an effect with something that exists in the state.

brandonroberts: I thought about this also, and I'm not strongly opposed to it. We would have to change the interface to use a generic then though, right? Or can we infer it some other way

timdeschryver: Oh right, I didn't think of that.  
I don't think we can infer its type, which also takes the usefulness of this enhancement away.
If we can't infer it, perhaps we should just leave it as is. We can always add it later, when there's a strong need for it.

modules/component-store/src/component-store.ts
```diff
@@ -62,12 +70,24 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
+    // check/call store init hook
+    this.checkInitStoreHook();
+
     // State can be initialized either through constructor or setState.
     if (defaultState) {
       this.initState(defaultState);
     }
   }
 
+  private checkInitStoreHook() {
```
LayZeeDK, line 82: suggestion: consider a different name:
- `triggerInitStoreEvent`
- `callInitStoreHook`

projects/example-app/src/app/auth/containers/login-page.component.spec.ts
```diff
@@ -18,6 +19,7 @@ describe('Login Page', () => {
       imports: [NoopAnimationsModule, MaterialModule, ReactiveFormsModule],
       declarations: [LoginPageComponent, LoginFormComponent],
       providers: [
+        LoginPageStore,
```
LayZeeDK, line 22: issue: cover `OnStateInit` and `OnStoreInit` hooks with tests

LayZeeDK (changes requested):

modules/component-store/src/component-store.ts
brandonroberts: Good suggestions

projects/example-app/src/app/auth/containers/login-page.component.spec.ts
brandonroberts: Agreed. This was for testing example code

modules/component-store/src/component-store.ts
```diff
@@ -149,8 +169,18 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
       this.stateSubject$.next(s);
+
+      if (!isInitialized) {
+        const onStateInit: Function | undefined = (
+          this as unknown as ComponentStore<T> & OnStateInit
+        )['ngrxOnStateInit'];
+        if (typeof onStateInit === 'function') {
+          onStateInit.call(this, s);
+        }
+      }
```
markostanimirovic, line 183: To avoid explicit casting here, I suggest this:

```suggestion
      if (!isInitialized && isOnStateInitDefined(this)) {
        this.ngrxOnStateInit();
      }
```

```ts
function isOnStateInitDefined(cs: unknown): cs is OnStateInit {
  return typeof (cs as OnStateInit).ngrxOnStateInit === 'function';
}
```

The same can be applied for `OnStoreInit` hook

markostanimirovic, line 183: To avoid explicit casting here, I suggest this:

```suggestion
      if (!isInitialized && isOnStateInitDefined(this)) {
        this.ngrxOnStateInit();
      }
```

```ts
function isOnStateInitDefined(cs: unknown): cs is OnStateInit {
  return typeof (cs as OnStateInit).ngrxOnStateInit === 'function';
}
```

The same can be applied for `OnStoreInit` hook

aokrushk: +1

markostanimirovic (approved):

brandonroberts:
Code is refactored, tests are updated and green. I do have some concerns about the feature in general though. Because we are chaining off the constructor of the ComponentStore, it forces us or the developer to consider workarounds to use it effectively. 

Consider the example below:

```ts
import {
  ComponentStore,
  OnStateInit,
  OnStoreInit,
} from '@ngrx/component-store';
import { delay, of, tap } from 'rxjs';

@Injectable()
export class LoginPageStore
  extends ComponentStore<{ test: boolean }>
  implements OnStoreInit, OnStateInit
{
  props = [];
  constructor() {
    super();
  }

  logEffect = this.effect(
    tap<void>(() => {
      console.log('effect');
    })
  );

  ngrxOnStoreInit() {
    this.logEffect(); // doesn't work because constructor isn't done
    console.log(this.props); // undefined
  }

  ngrxOnStateInit() {
    console.log('state init', this.get()); // works
    this.logEffect(); // blows up
  }
}
```

The examples below work

```ts
import {
  ComponentStore,
  OnStateInit,
  OnStoreInit,
} from '@ngrx/component-store';
import { delay, of, tap } from 'rxjs';

@Injectable()
export class LoginPageStore
  extends ComponentStore<{ test: boolean }>
  implements OnStoreInit, OnStateInit
{
  constructor() {
    super();
  }

  logEffect = this.effect(
    tap<void>(() => {
      console.log('effect');
    })
  );

  ngrxOnStoreInit() {
    of(null)
      .pipe(delay(0))
      .subscribe(() => {
        console.log('store init');
        this.logEffect(); // works
      });
  }

  ngrxOnStateInit() {
    setTimeout(() => {
      console.log('state init', this.get()); // works
      this.logEffect();
    });
  }
}
```

From my understanding, setTimeout pushes the task into the next macrotask queue, which allows it to run successfully. It would put the responsibility on the developer to run the code inside the hooks in a way that doesn't break. I'm open to other suggestions also internally we could make to make it more predictable without introducing race conditions.

@aokrushk @markostanimirovic @timdeschryver @nartc

markostanimirovic (commented):
Hoisting 😳 The behavior could be very confusing for users.

This will work for component stores that are explicitly provided in components, modules, or route config:

```ts
const WITH_HOOKS = new InjectionToken<ComponentStore<any>[]>(
  '@ngrx/component-store: ComponentStores with Hooks'
);

export function provideWithHooks(
  componentStoreClass: Type<ComponentStore<any>>
) {
  return [
    { provide: WITH_HOOKS, multi: true, useClass: componentStoreClass },
    {
      provide: componentStoreClass,
      useFactory: () => {
        const componentStore = inject(WITH_HOOKS).pop();

        if (isOnStoreInitDefined(componentStore)) {
          componentStore.ngrxOnStoreInit();
        }

        if (isOnStateInitDefined(componentStore)) {
          componentStore.state$
            .pipe(take(1), takeUntil(componentStore.destroy$))
            .subscribe(() => componentStore.ngrxOnStateInit());
        }

        return componentStore;
      },
    },
  ];
}
```

Example:

```ts
@Injectable()
class LifecycleStore
  extends ComponentStore<LifeCycle>
  implements OnStoreInit, OnStateInit
{
  constructor() {
    super({ init: true });

    console.log('constructor');
    this.logEffect('constructor effect');
  }

  logEffect = this.effect(
    tap<string>((message) => {
      console.log(message);
    })
  );

  ngrxOnStoreInit(): void {
    console.log('onStoreInit');
    this.logEffect('store init effect');
  };

  ngrxOnStateInit(): void {
    console.log('onStateInit');
    this.logEffect('state init effect');
  }
}

@Component({
  providers: [provideWithHooks(LifecycleStore)]
})
export class AppComponent {
  constructor(private readonly lifecycleStore: LifecycleStore) {}
}

// console:
// constructor
// constructor effect
// onStoreInit
// store init effect
// onStateInit
// state init effect
```

brandonroberts:
Nice, I thought of adding a type of `provideComponentStore` function also, but didn't think about using inject. I think it works for the primary use case, and if you want to use the `providedIn` you understand that restriction.

brandonroberts:
I refactored the logic for the init checks to run after the constructor is complete, so the additional token isn't required and it runs consistently whether it's in `providers` or uses `providedIn`.

modules/component-store/spec/component-store.spec.ts
```diff
@@ -1447,4 +1451,83 @@ describe('Component Store', () => {
...
+      }
+    }
+
+    let componentStore: LifecycleStore;
+
+    beforeEach(() => {
+      logs = [];
+    });
+
+    it('should call the OnInitStore lifecycle hook if defined', async () => {
```
timdeschryver, line 1492: These tests shouldn't be `async`

modules/component-store/src/component-store.ts
```diff
@@ -308,6 +331,21 @@ export class ComponentStore<T extends object> implements OnDestroy {
       });
     }) as unknown as ReturnType;
   }
+
+  callInitStateHook() {
+    this.stateSubject$
+      .pipe(
+        take(1),
+        map((val) => {
```
timdeschryver, line 339: Because we're not transforming the data, could we use a `tap` here and also in the other hook?

```suggestion
        tap((val) => {
```

modules/component-store/src/component-store.ts
```diff
@@ -61,6 +71,19 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
 
+  // check/call store init hook
+  private readonly initStoreHook = this.effect(() =>
+    of(null).pipe(($) => {
+      if (isOnStoreInitDefined(this)) {
+        this.ngrxOnStoreInit();
+      }
+      return $;
+    })
+  )();
```
timdeschryver, line 82:
```suggestion
    of(null).pipe(
      tap(() => {
        if (isOnStoreInitDefined(this)) {
          this.ngrxOnStoreInit();
        }
      })
    )
```

modules/component-store/src/component-store.ts
```diff
@@ -32,6 +34,14 @@ export interface SelectConfig {
   debounce?: boolean;
 }
 
+export interface OnStoreInit {
```
timdeschryver, line 37: Should we add JS Doc comments here so it looks nice on the API documentation?

modules/component-store/src/component-store.ts
brandonroberts: Yep. I only thought to inline it to avoid bringing in another operator but using tap is pretty standard

modules/component-store/src/component-store.ts
brandonroberts: Yep. Will add that also

modules/component-store/src/component-store.ts
```diff
@@ -61,6 +71,19 @@ export class ComponentStore<T extends object> implements OnDestroy {
   // Needs to be after destroy$ is declared because it's used in select.
   readonly state$: Observable<T> = this.select((s) => s);
 
+  // check/call store init hook
+  private readonly initStoreHook = this.effect(() =>
```
timdeschryver, line 75: It seems like a comment didn't come through.
Why is `effect` used here, in comparison to the other hook where a manual `subscribe` and `take(1)` is used?

brandonroberts: I wanted to use the effect to handle the `subscribe`, but that did create some inconsistency. I changed them both to manual subscribes inside the methods.

timdeschryver (approved):

modules/component-store/src/component-store.ts
timdeschryver: Thanks Brandon, also a good catch to make those methods private 👍

modules/component-store/src/lifecycle_hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export interface OnStateInit {
+  readonly ngrxOnStateInit: () => void;
+}
+
+export function isOnStoreInitDefined(cs: unknown): cs is OnStoreInit {
+  return typeof (cs as OnStoreInit).ngrxOnStoreInit === 'function';
+}
+
+export function isOnStateInitDefined(cs: unknown): cs is OnStateInit {
```
markostanimirovic, line 17:
```suggestion
function isOnStoreInitDefined(cs: unknown): cs is OnStoreInit {
  return typeof (cs as OnStoreInit).ngrxOnStoreInit === 'function';
}

function isOnStateInitDefined(cs: unknown): cs is OnStateInit {
```

markostanimirovic (commented):
LGTM 🎉 The only thing I would change is not to expose `isStoreInitDefined` and `isStateInitDefined` guards to the public API.

modules/component-store/src/lifecycle_hooks.ts
brandonroberts: Good catch. Fixed

markostanimirovic (approved):

modules/component-store/src/lifecycle_hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+function isOnStateInitDefined(cs: unknown): cs is OnStateInit {
+  return typeof (cs as OnStateInit).ngrxOnStateInit === 'function';
+}
+
+export function provideComponentStore(
+  componentStoreClass: Type<ComponentStore<any>>
+): Provider[] {
+  const CS_WITH_HOOKS = new InjectionToken<ComponentStore<any>>(
```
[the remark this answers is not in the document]
nartc: I think `unknown` will fail `extends object` constraint on `ComponentStore` generics. If we want to absolutely avoid `any`, I think `{}` would do it?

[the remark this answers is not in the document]
brandonroberts: 👍

timdeschryver (approved):
LGTM (again 😅) 
Should we document somewhere why it's done like this instead of "just" adding the hooks to the Store class? (or will this be a part of the docs?).

PR #3379, opened 2022-04-14
feat(component): use global render strategy in zone-less mode

### LINKED ISSUE #3342: @ngrx/component: Use "global" rendering strategy in zoneless mode
`@ngrx/component` uses `changeDetectorRef.detectChanges` in zoneless mode. This method will synchronously run change detection which can cause performance issues.

For example, the following component will be rerendered 3 times when "Emit new value" button is clicked:

```ts
@Component({
  selector: 'ngrx-number',
  template: `
    <p>{{ num$ | ngrxPush }}</p>
    <p>{{ num$ | ngrxPush }}</p>
    <p>{{ num$ | ngrxPush }}</p>

    <button (click)="emitNewValue()">Emit new value</button>
  `,
})
export class NumberComponent {
  readonly num$ = new BehaviorSubject(0);

  emitNewValue(): void {
    this.num$.next(Math.random());
  }
}
```

because the `changeDetectorRef.detectChanges` will be called 3 times under the hood.

As a solution, we can use the `θmarkDirty` function to trigger the change detection in zoneless mode. Also, the rendering strategy in zoneless mode will be "global" with `θmarkDirty`. This will make the rendering strategy consistent in both zoneless and zonefull mode.

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3342 

## What is the new behavior?

`@ngrx/component` uses a global rendering strategy in zone-less mode.

```
[x] Yes
[ ] No

BREAKING CHANGES:

The native local rendering strategy is replaced by global in zone-less mode for better performance.

BEFORE:

The change detection is triggered via `changeDetectorRef.detectChanges` in zone-less mode.

AFTER:

The change detection is triggered via `ɵmarkDirty` in zone-less mode.
```

### DIFF
--- a/modules/component/src/core/render-scheduler.ts
+++ b/modules/component/src/core/render-scheduler.ts
@@ -1,4 +1,8 @@
-import { ChangeDetectorRef, NgZone } from '@angular/core';
+import {
+  ChangeDetectorRef,
+  NgZone,
+  ɵmarkDirty as markDirty,
+} from '@angular/core';
 
 export interface RenderScheduler {
   schedule(): void;
@@ -16,7 +20,8 @@ export function createRenderScheduler(
     if (hasZone(config.ngZone)) {
       config.cdRef.markForCheck();
     } else {
-      config.cdRef.detectChanges();
+      const context = getCdRefContext(config.cdRef);
+      markDirty(context);
     }
   }
 
@@ -29,4 +34,8 @@ export function createRenderScheduler(
  */
 function hasZone(z: NgZone): boolean {
   return z instanceof NgZone;
+}
+
+function getCdRefContext(cdRef: ChangeDetectorRef): object {
+  return (cdRef as unknown as { context: object }).context;
 }
--- a/modules/component/spec/core/render-scheduler.spec.ts
+++ b/modules/component/spec/core/render-scheduler.spec.ts
@@ -1,4 +1,5 @@
-import { NgZone } from '@angular/core';
+import * as angular from '@angular/core';
+import { noop } from 'rxjs';
 import { createRenderScheduler } from '../../src/core/render-scheduler';
 import {
   manualInstanceNgZone,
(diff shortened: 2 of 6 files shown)


modules/component/src/core/render-scheduler.ts
```diff
@@ -30,3 +35,7 @@ export function createRenderScheduler(
 function hasZone(z: NgZone): boolean {
   return z instanceof NgZone;
 }
+
+function getCdRefContext(cdRef: ChangeDetectorRef): object {
+  return (cdRef as unknown as { context: object }).context;
```
timdeschryver, line 40: Does `cdRef` always have a `context` here?

markostanimirovic: When the `ChangeDetectorRef` instance comes from a pipe or directive as in our case, the context will be always defined (it will be a host component).

timdeschryver (approved):

brandonroberts (approved):

markostanimirovic:
@LayZeeDK

The previous way of triggering CD in zone-less mode via `cdr.detectChanges` without coalescing was inefficient and it had to be changed. Because we use the "global" rendering strategy in zone-full mode, we agreed to use the same strategy in zone-less mode.

I chose `ɵmarkDirty` because it already implements what we need in the zone-less mode for the "global" rendering strategy - it marks the passed component as dirty, schedules a `tick`, and coalesces multiple `ɵmarkDirty` calls into a single CD run.

I'm aware that `ɵmarkDirty` can be changed/removed at any moment. However, it has been part of the "ɵpublic" API for more than 2 years. If Angular team decides to remove it, we'll implement a custom solution for scheduling/coalescing `tick` calls in zone-less mode.

LayZeeDK:
@markostanimirovic 
https://github.com/angular/angular/pull/46709

PR #3380, opened 2022-04-14
feat(component): add error as value to LetDirective's context

### LINKED ISSUE #3343: @ngrx/component: Add the `$error` as a value to the `LetDirective`'s context
The type of `$error` property from `LetDirective`'s context is `boolean`. It would be better to add the error as a value, so it can be used in the template:

```html
<div *ngrxLet="books$ as books; let error = $error">
  <p *ngIf="error; else bookList">{{ error.message }}</p>

  <ng-template #bookList>
    <div *ngFor="let book of books">...</div>
  </ng-template>
</div>
```

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `$error` property from `LetDirective`'s view context will be `true` or `false`.

Closes #3343 

## What is the new behavior?

The `$error` property from `LetDirective`'s view context will be thrown error or `undefined`.

```
[x] Yes
[ ] No

BREAKING CHANGES:

The `$error` property from `LetDirective`'s view context is a thrown error or `undefined` instead of `true`/`false`.

BEFORE:

<p *ngrxLet="obs$; $error as e">{{ e }}</p>

- `e` will be `true` when `obs$` emits error event.
- `e` will be `false` when `obs$` emits next/complete event.

AFTER:

<p *ngrxLet="obs$; $error as e">{{ e }}</p>

- `e` will be thrown error when `obs$` emits error event.
- `e` will be `undefined` when `obs$` emits next/complete event.
```

### DIFF
--- a/modules/component/src/let/let.directive.ts
+++ b/modules/component/src/let/let.directive.ts
@@ -26,7 +26,7 @@ export interface LetViewContext<T> {
   /**
    * `*ngrxLet="obs$; let e = $error"` or `*ngrxLet="obs$; $error as e"`
    */
-  $error: boolean;
+  $error: any;
   /**
    * `*ngrxLet="obs$; let c = $complete"` or `*ngrxLet="obs$; $complete as c"`
    */
@@ -66,7 +66,7 @@ export interface LetViewContext<T> {
  *   <app-number [number]="n" *ngIf="!e && !c">
  *   </app-number>
  *
- *   <p *ngIf="e">There is an error.</p>
+ *   <p *ngIf="e">There is an error: {{ e }}</p>
  *   <p *ngIf="c">Observable is completed.</p>
  * </ng-container>
  * ```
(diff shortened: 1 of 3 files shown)


modules/component/src/let/let.directive.ts
```diff
@@ -5,140 +5,170 @@ import {
...
+   * using `ngrxLet` to enable `as` syntax: `*ngrxLet="obs$ as o"`
+   */
   ngrxLet: T;
-  // set context var complete to true (var$; let e = $error)
-  $error: boolean;
-  // set context var complete to true (var$; let c = $complete)
+  /**
+   * `*ngrxLet="obs$; let e = $error"` or `*ngrxLet="obs$; $error as e"`
+   */
+  $error: any;
```
timdeschryver, line 29: This change is great, having access to the error makes more sense.
I do have one question, why use `any` over `unknown` here?

markostanimirovic: I had the same dilemma. `any` seems a better option to me because casting is not possible in the template.

With `$error: unknown` and `strictTemplates: true`, the only way to access the properties of the error object in the template and avoid compiler errors would be to use `$any`:

```html
<p *ngrxLet="obs$; $error as e">
  {{ $any(e).message }}
</p>
```

timdeschryver (approved):

brandonroberts (approved):

PR #3381, opened 2022-04-15
feat(store): add createActionGroup function

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3337 

## What is the new behavior?

```ts
const authApiActions = createActionGroup({
  source: 'Auth API',
  events: {
    'Login Success': props<{ userId: number; token: string }>(),
    'Login Failure': props<{ error: string }>(),
    'Logout Success': emptyProps(),
    'Logout Failure': (error: Error) => ({ error }),
  },
});

authApiActions.loginSuccess({ userId: 10, token: 'ngrx' });
authApiActions.loginFailure({ error: 'Login Failure!' });
authApiActions.logoutSuccess();
authApiActions.logoutFailure(new Error('Logout Failure!'));
```

```
[ ] Yes
[x] No
```


modules/store/spec/action_group_creator.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { createActionGroup, emptyProps, props } from '@ngrx/store';
+
+describe('createActionGroup', () => {
+  const authApiActions = createActionGroup({
+    source: 'Auth API',
+    events: {
+      'Login Success': props<{ userId: number; token: string }>(),
+      'Login Failure': props<{ error: string }>(),
+      'Logout Success': emptyProps(),
```
brandonroberts, line 9: Will a regular empty function work here also?

```ts
'Logout Success': () => {}
```

markostanimirovic: Not for now. We can make `'Logout Success': () => {}` work, but I'd rather expose just a single way to create action without payload (using the `emptyProps` function).

markostanimirovic: Not for now. We can make `'Logout Success': () => {}` work, but I'd rather expose just a single way to create action without payload (using the `emptyProps` function).

brandonroberts: Got it. emptyProps is a reasonable choice

brandonroberts (approved):

modules/store/spec/action_group_creator.spec.ts
timdeschryver: I had the same comment as Brandon. My mind thinks of `null` when I don't want to provide a payload, but that might seem weird here. Instead of adding `emptyProps`, could/would we want to use just `props()`, or do we want to be explicit here?
Reasoning: `emptyProps` is an additional method that people need to know of and also have to import.

markostanimirovic: `'Event Name': props()` for action without payload could be confusing IMO, because `props()` are used for actions with payload. We could make `'Event Name': props<**void**>()` work, but that would require changes in the `props` function signature.

timdeschryver: Not a big fan of `props<void>`. 
And also after letting it sink in, `emptyProps` seems to be the best option 👍

timdeschryver (approved):

modules/store/src/action_group_creator_models.ts
```diff
@@ -0,0 +1,142 @@
+import {
+  ActionCreator,
+  ActionCreatorProps,
+  Creator,
+  FunctionWithParametersType,
+  NotAllowedCheck,
+} from '@ngrx/store';
+import { TypedAction } from './models';
+
+// Generating API docs for the `createActionGroup` function is solved by moving its types
```
markostanimirovic, line 10: There was the same bug with generating API docs for the `createActionGroup` function as there was for the `createFeature` - template literal types. I moved its types to another file to make API generation work.

Xapuu:
I've been doing a version of this for years on my own in each NGRX project that I've worked on, so I'm beyond happy to find out that this is part of the core lib now 🎉 
Just wanted to say thanks for adding this feature.

PR #3410, opened 2022-05-10
Add tapResponse observer signature

MarkKoester:
tapResponse is inconsistent with rxjs since it doesn't accept an observer as argument, making it not match with the tap operator. Passing more than one argument to subscribe/tap is deprecated and will be removed in rxjs 8

https://rxjs.dev/deprecations/subscribe-arguments

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component-store/src/tap-response.ts
+++ b/modules/component-store/src/tap-response.ts
@@ -37,3 +37,13 @@ export function tapResponse<T, E = unknown>(
       catchError(() => EMPTY)
     );
 }
+
+export function tapResponse<T>(
+  observer: Partial<Observer<T>>
+): (source: Observable<T>) => Observable<T> {
+  return (source) =>
+    source.pipe(
+      tap(observer),
+      catchError(() => EMPTY)
+    );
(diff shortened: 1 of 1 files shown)


alex-okrushko:
Thanks @MarkKoester 
It wasn't the intention for `tapResponse` to implement the `Observer` interface. In fact, while `tapResponse` takes multiple arguments, it is calling `tap` with a single argument object (adhering to the non-deprecating API): https://github.com/ngrx/platform/blob/f5b055bae28ad81a875dcaa24beafc04be794f09/modules/component-store/src/tap-response.ts#L32-L36

The goal for the `tapResponse` is to:
- avoid error-prone situation and catch error itself
- **enforce developer to handle the error case**

That's why the second (error-handing) argument is not optional.

Thanks for PR, however I'm going to close it.

PR #3441, opened 2022-05-28
feat(schematics): use schematicCollections instead of defaultCollection

### LINKED ISSUE #3383: use `schematicCollections` instead of `defaultCollection`
### Information

In Angular 14, the `defaultCollection` is deprecated in favor of `schematicCollections`, which supports an array of schematics (instead of a single one).
We should update the NgRx schematics to add the ngrx schematics to the collection, instead of overwriting the `defaultCollection` option. 

See https://github.com/angular/angular-cli/pull/22860 for more info.

The code that does this is located at https://github.com/ngrx/platform/blob/master/modules/schematics/src/ng-add/index.ts.

If you want to pick this issue up, please add a comment below :)  

- [X] Yes
- [ ] No

suke:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3383

## What is the new behavior?
Throw an exception if schematicCollections is not defined because the user may not have upgraded Angular to v14.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/schematics/src/ng-add/index.ts
+++ b/modules/schematics/src/ng-add/index.ts
@@ -1,40 +1,37 @@
 import {
   chain,
-  noop,
   Rule,
   SchematicContext,
   Tree,
 } from '@angular-devkit/schematics';
 import { getWorkspace, getWorkspacePath } from '../../schematics-core';
 
-import { Schema as SchematicOptions } from './schema';
-
-function updateWorkspaceCli(host: Tree, value: any) {
+function updateSchematicCollections(host: Tree) {
   const workspace = getWorkspace(host);
   const path = getWorkspacePath(host);
 
-  workspace['cli'] = {
-    ...workspace['cli'],
-    ...value,
-  };
+  if (!(workspace['cli'] && workspace['cli']['schematicCollections'])) {
+    throw new Error(
+      'schematicCollections is not defined in the global cli options'
+    );
+  }
 
+  workspace['cli']['schematicCollections'] = [
+    ...workspace['cli']['schematicCollections'],
+    '@ngrx/schematics',
+  ];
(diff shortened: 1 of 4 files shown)


modules/schematics/src/ng-add/index.ts
```diff
@@ -1,40 +1,37 @@
...
+function updateSchematicCollections(host: Tree) {
   const workspace = getWorkspace(host);
   const path = getWorkspacePath(host);
 
-  workspace['cli'] = {
-    ...workspace['cli'],
-    ...value,
-  };
+  if (!(workspace['cli'] && workspace['cli']['schematicCollections'])) {
+    throw new Error(
```
brandonroberts, line 14: Why are we throwing an error here? If the schematicCollections property doesn't exist, we should create it and add @ngrx/schematics to it.

brandonroberts (changes requested):

modules/schematics/src/ng-add/index.ts
timdeschryver: That would be on me 😅
https://github.com/ngrx/platform/issues/3383#issuecomment-1140199497

timdeschryver: Since it's 2v1, I'm fine with creating the property.
I don't know what Angular's schematic will do with it though (in the case that NgRx is updated before Angular)

brandonroberts: Ahh ok. I was thinking of cases like Nx where it may not be present, but I think they account for that anyway. Throwing the error should be ok for now

brandonroberts (approved):

PR #3487, opened 2022-07-13
perf(component): do not schedule render for synchronous events

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: Performance improvements
```

The render scheduler unnecessarily schedules render/marks a component as dirty when synchronous events are emitted.

For instance, in the following example:

```ts
@Component({
  template: `
    <p>{{ obs$ | ngrxPush }}</p>
  `
})
export class TestComponent {
  obs$ = of(1, 2, 3);
}
```

The PushPipe will mark `TestComponent` as dirty (and schedule tick in zone-less mode) 4 times (3 next, and 1 complete event) which is not needed because all these events will be emitted synchronously, so the last value will be visible in the template without scheduling render/marking it as dirty.

## What is the new behavior?

The render scheduler schedules render/marks a component as dirty only when asynchronous events are emitted.

- [Zone-Full Playground](https://stackblitz.com/edit/angular-zy74xp?file=src/app/app.component.html)
- [Zone-Less Playground](https://stackblitz.com/edit/angular-zy74xp-g52skq?file=src/app/app.component.html)

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component/src/core/render-event/models.ts
+++ b/modules/component/src/core/render-event/models.ts
@@ -6,13 +6,13 @@ interface BaseRenderEvent {
   /**
    * true if the synchronous event is emitted
    */
-  sync: boolean;
+  synchronous: boolean;
 }
 
 export interface SuspenseRenderEvent extends BaseRenderEvent {
   type: 'suspense';
   reset: true;
-  sync: true;
+  synchronous: true;
 }
 
 export interface NextRenderEvent<T> extends BaseRenderEvent {
--- a/modules/component/src/core/render-event/models.ts
+++ b/modules/component/src/core/render-event/models.ts
@@ -1,10 +1,18 @@
 interface BaseRenderEvent {
+  /**
+   * true if the event is emitted by a new source
+   */
   reset: boolean;
+  /**
+   * true if the synchronous event is emitted
+   */
+  sync: boolean;
 }
 
 export interface SuspenseRenderEvent extends BaseRenderEvent {
   type: 'suspense';
   reset: true;
+  sync: true;
 }
 
 export interface NextRenderEvent<T> extends BaseRenderEvent {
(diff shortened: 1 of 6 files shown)


modules/component/src/let/let.directive.ts
```diff
@@ -233,9 +235,5 @@ export class LetDirective<PO> implements OnInit, OnDestroy {
       this.isSuspenseViewCreated = true;
       this.viewContainerRef.createEmbeddedView(this.suspenseTemplateRef);
     }
-
-    if (this.isMainViewCreated || this.isSuspenseViewCreated) {
-      this.renderScheduler.schedule();
-    }
```
markostanimirovic, line 239: There is no need to schedule render for suspense events at all, because suspense events are synchronous.

modules/component/src/core/render-event/models.ts
```diff
@@ -1,10 +1,18 @@
 interface BaseRenderEvent {
+  /**
+   * true if the event is emitted by a new source
+   */
   reset: boolean;
+  /**
+   * true if the synchronous event is emitted
+   */
+  sync: boolean;
```
timdeschryver, line 9: Initially I was a bit confused about the term `sync`.
I was thinking that it meant that data (or something else) need to be synchronized.
It could be a malformation on my end, but I would rename this to `synchronous` to avoid confusion.  

Again, it it's probably just me, and if you disagree we can leave it as is.

markostanimirovic: Oh 😅 you're right, "sync" can be an abbreviation for multiple things. I'll rename `sync` to `synchronous` then, to avoid potential confusion.

brandonroberts (approved):

PR #3490, opened 2022-07-15
fix(component-store): make synchronous updater errors catchable

### LINKED ISSUE #3266: ComponentStore updater errors are not catchable

https://stackblitz.com/edit/angular-ivy-tgtbh9?devtoolsheight=33&file=src/app/hello.component.ts

## Expected behavior:

I expect the try/catch block to catch the error that is thrown in the ComponentStore updater.

This should all be captured in Stackblitz.

## Other information:

A few notes:

- I suspect this might have something to do with the recent RxJS SuperGrossMode deprecation, which throws unhandled errors in a setTimeout.
- When I originally designed the ComponentStore, I was using a BehaviorSubject and could thus call updater functions directly, ie: `this.state$.next(updaterFn(this.state$.value, updaterArg));`. There was a lot of async logic that was applied to updaters that makes it a little less obvious how to fix this issue. I'm wondering if moving the updaterFn call to a map (or maybe even a concatMap w/ nested try/catch) before the .subscribe() in the updater might fix this.
- I am seemingly able to make things work by manually sending an error through on stateSubject$ (see commented line in my Stackblitz demo). I'm not sure this is really intentional though. It seems to be hitting the `initializationError` re-throw block, which doesn't necessarily seem appropriate (since this isn't an initialization error). 
- I also see that updater errors (or at least the initialization errors mentioned above) are pushed onto the stateSubject$ itself, which isn't a decision I'm familiar with. It doesn't seem like this captures updater errors.

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3266 

## What is the new behavior?

All synchronous errors are now catchable. Also, error handling is changed - an error will not be emitted to the `stateSubject$` anymore, it will be thrown instead.

```
[ ] Yes
[x] No
```

As we discussed at the last team meeting, this should not be considered a breaking change.

### DIFF
--- a/modules/component-store/spec/component-store.spec.ts
+++ b/modules/component-store/spec/component-store.spec.ts
@@ -79,19 +79,8 @@ describe('Component Store', () => {
     it(
       'throws an Error when setState with a function/callback is called' +
         ' before initialization',
-      marbles((m) => {
+      () => {
         const componentStore = new ComponentStore();
-
-        m.expect(componentStore.state$).toBeObservable(
-          m.hot(
-            '#',
-            {},
-            new Error(
-              'ComponentStore has not been initialized yet. ' +
-                'Please make sure it is initialized before updating/getting.'
-            )
-          )
-        );
 
         expect(() => {
           componentStore.setState(() => ({ setState: 'new state' }));
@@ -101,7 +90,7 @@ describe('Component Store', () => {
               'Please make sure it is initialized before updating/getting.'
           )
         );
-      })
+      }
     );
 
     it('throws an Error when patchState with an object is called before initialization', () => {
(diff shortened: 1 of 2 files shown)


timdeschryver (approved):

modules/component-store/spec/component-store.spec.ts
```diff
@@ -638,6 +595,69 @@ describe('Component Store', () => {
...
+
+        expect(() => {
+          updater(asyncObs$);
+          m.flush();
+        }).toThrow(error);
+      })
+    );
+
+    it(
+      'when asynchronous observable throws an error with patchState',
```
alex-okrushko, line 647: Does this test `isSyncUpdate` to be `false`?

Shouldn't the validation be that it **does not throw** then? And once flushed shouldn't `throwError(() => error)` happen?

modules/component-store/src/component-store.ts
```diff
@@ -128,23 +132,26 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          map(([value, currentState]) => updaterFn(currentState, value!)),
+          tap((newState) => this.stateSubject$.next(newState)),
+          catchError((error: unknown) => {
+            if (isSyncUpdate) {
+              syncError = error;
+              return EMPTY;
+            }
+
+            return throwError(() => error);
```
alex-okrushko, line 144: Nice! So this will error the async updater but not the general `state$` 👍

modules/component-store/src/component-store.ts
```diff
@@ -128,23 +132,26 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
-            this.stateSubject$.error(error);
-          },
-        });
+        .subscribe();
 
-      if (initializationError) {
-        // prettier-ignore
-        throw /** @type {!Error} */ (initializationError);
+      if (syncError) {
+        throw syncError;
```
alex-okrushko, line 151: ` /** @type {!Error} */` was needed for Google's Closure compiler, and `\\prettier-ignore` was needed so it doesn't break the lines.

Don't know if `unknown` changes that - we'd need to validate with our friends at Google.

modules/component-store/src/component-store.ts
```diff
@@ -110,7 +113,8 @@ export class ComponentStore<T extends object> implements OnDestroy {
     return ((
       observableOrValue?: OriginType | Observable<OriginType>
     ): Subscription => {
-      let initializationError: Error | undefined;
+      let isSyncUpdate = true;
```
alex-okrushko, line 116: I know why we have it, but could you please add comments for any future readers of this code what's the purpose for this variable.

modules/component-store/src/component-store.ts
timdeschryver: Is there a way that we can test this ourselves?
Or should we add a comment that this was done for the Closure compiler?
The reason why I ask this is that this would have been overlooked if you didn't review it.

modules/component-store/spec/component-store.spec.ts
markostanimirovic: Yes, you're right. I'll update the test to make sure the error is not thrown synchronously, but asynchronously.

modules/component-store/src/component-store.ts
markostanimirovic: Let's check with them. If it's still needed, I'll add a comment.

markostanimirovic: Googlers confirmed that the type comment is not needed anymore. The PR is ready for re-review 👀 cc @alex-okrushko

alex-okrushko (approved):
Fantastic work @markostanimirovic 👍

PR #3492, opened 2022-07-16
fix(component-store): move isInitialized check to queueScheduler context on state update

### LINKED ISSUE #2991: Clarify/improve existing behavior in component-store initState method
## Description:
Displaying of `${this.constructor.name} has not been initialized yet. Please make sure it is initialized before updating/getting.` looks incorrect for multiple component-stores and possible for store + component-store integration in some cases.

Usually call of some store action in subscriber callback is bad practice, but in some cases it could be chain of calls where subscription from one store triggers initialization of another component-store.

<pre><code>
[ * ] Bug report  
[ ] Feature request
</code></pre>

Initialized component-store actions throw the same error: `${this.constructor.name} has not been initialized yet. Please make sure it is initialized before updating/getting.`

## Expected behavior:
Actions of initialized store should work as expected.
Error should not be displayed.

native issue reproduction
```typescript
class Store extends ComponentStore<{ shouldInitDerivedStore: boolean }> {
    constructor() {
        super({ shouldInitDerivedStore: false });
    }

    initDerivedStore = this.updater((state) => ({
        shouldInitDerivedStore: !state.shouldInitDerivedStore,
    }));
}

class DerivedStore extends ComponentStore<{ anyField: boolean }> {
    constructor() {
        super({ anyField: false });
    }

    anyAction = this.updater(() => ({ anyField: true }));
}

const store = new Store();
const initDerivedStoreCallback = () => {
    const derivedStore = new DerivedStore();
    derivedStore.anyAction(); // `DerivedStore has not been initialized yet. Please make sure it is initialized before updating/getting.`
};

store.state$
    .pipe(filter(({ shouldInitDerivedStore }) => shouldInitDerivedStore))
    .subscribe(initDerivedStoreCallback);

store.initDerivedStore();
```

## Other information:
For practice usage in angular world:
Angular supports performance improvements like `changeDetectionStrategy: OnPush`. It means user need to handle view synchronisation manually in case if changes comes outside of angular word (any action which does not trigger change detection). 
For these cases angular provides two options to update view: 
- `marckForCheck` marks all components tree modified to the root entry to validate them on the next change detection call
- `detectChanges` runs change detection directly for the affected component. 
Usage of both methods depends on the used libraries and performance criteria.

If we have two components with different store instances we could not make them dependent if parent component trigger `detectChanges` inside the callback. In other words children component with its own store instance cannot be created from the subscription result of the parent component like next:
```typescript
// in parent component
store.select(shouldDisplayChildSelector).subscribe((shouldDisplayChild) => {
  this.shouldDisplayChild = shouldDisplayChild;
  this.ref.detectChanges(); // this call means initialisation of child component performs before `queueScheduler` of rxjs will call all its actions
})
```

```html
<!--parent component-->
<child-component-with-own-store *ngIf="shouldDisplayChild"><child-component-with-own-store />
```

## Reason:
Looks like it is regression after this PR https://github.com/ngrx/platform/pull/2606 

## Possible solutions if expected behaviour will be confirmed
1. Simple way: refactor this method: https://github.com/ngrx/platform/blob/master/modules/component-store/src/component-store.ts#L149
image

to this way
```typescript
   /**
     * Reason: staying isInitialized flag inside callback looks like code issue
     *
     * Description: staying the flag in the callback means we need to perform side effects (actions queue) before to init the store
     * even if user has already initialized it
     *
     * Fix details: once user init state in constructor or #setState the store will be initialized
     */
    private initState(state: T): void {
        this.isInitialized = true; // the fix

        scheduled([state], queueScheduler).subscribe((s) => {
            this.stateSubject$.next(s);
        });
    }
```

2. Complex way: stop use **singleton** queueScheduler in each component-stores instance  and replace to multiple instances:
```typescript
class ComponentStore {
  constructor() {
    this.queueScheduler = new queueScheduler()
  }
}
```

## workarounds for angular:
1. if the reason is `detectChanges` call and it's correct for the application to postpone to call changeDetection. Simple replace its call to `markForCheck`
2. if the reason in subscriber. You can try to use https://rxjs.dev/api/operators/observeOn (looks like a hack)
```
this.observable$
  .pipe(observeOn(queueScheduler)) // this postpone to call subscribe till all actions in queue will be called (#initState/#setState are placed in this scheduler)
  .subscriber(callback)
```
3. as a hack solution - add any delay

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #2991

## What is the new behavior?

There is no initialization error when ComponentStore initialization and state update are scheduled via `queueScheduler`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component-store/spec/component-store.spec.ts
+++ b/modules/component-store/spec/component-store.spec.ts
@@ -203,7 +203,7 @@ describe('Component Store', () => {
     );
 
     it(
-      'does not throws an Error when updater is called with async Observable' +
+      'does not throw an Error when updater is called with async Observable' +
         ' before initialization, that emits the value after initialization',
       marbles((m) => {
         const componentStore = new ComponentStore();
@@ -235,6 +235,19 @@ describe('Component Store', () => {
           m.hot('(iu)', { i: INIT_STATE, u: UPDATED_STATE })
         );
       })
+    );
+
+    it(
+      'does not throw an Error when ComponentStore initialization and' +
+        ' state update are scheduled via queueScheduler',
+      () => {
+        expect(() => {
+          queueScheduler.schedule(() => {
+            const componentStore = new ComponentStore({ foo: false });
+            componentStore.patchState({ foo: true });
+          });
+        }).not.toThrow();
+      }
     );
   });
 
--- a/modules/component-store/src/component-store.ts
+++ b/modules/component-store/src/component-store.ts
@@ -13,7 +13,6 @@ import {
   EMPTY,
 } from 'rxjs';
 import {
-  concatMap,
   takeUntil,
   withLatestFrom,
   map,
@@ -22,6 +21,7 @@ import {
   take,
   tap,
   catchError,
+  observeOn,
 } from 'rxjs/operators';
 import { debounceSync } from './debounce-sync';
 import {
@@ -61,9 +61,6 @@ export class ComponentStore<T extends object> implements OnDestroy {
 
   private readonly stateSubject$ = new ReplaySubject<T>(1);
   private isInitialized = false;
-  private notInitializedErrorMessage =
-    `${this.constructor.name} has not been initialized yet. ` +
-    `Please make sure it is initialized before updating/getting.`;
   // Needs to be after destroy$ is declared because it's used in select.
   readonly state$: Observable<T> = this.select((s) => s);
   private ɵhasProvider = false;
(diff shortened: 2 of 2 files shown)


modules/component-store/spec/component-store.spec.ts
```diff
@@ -235,6 +235,19 @@ describe('Component Store', () => {
...
+    it(
+      'does not throw an Error when ComponentStore initialization and' +
+        ' state update are scheduled via queueScheduler',
+      () => {
+        expect(() => {
+          queueScheduler.schedule(() => {
+            const componentStore = new ComponentStore({ foo: false });
+            componentStore.patchState({ foo: true });
+          });
+        }).not.toThrow();
```
markostanimirovic, line 248: this test would fail before this PR

modules/component-store/src/component-store.ts
```diff
@@ -125,15 +122,11 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
-            this.isInitialized
-              ? // Push the value into queueScheduler
-                scheduled([value], queueScheduler).pipe(
-                  withLatestFrom(this.stateSubject$)
-                )
-              : // If state was not initialized, we'll throw an error.
-                throwError(() => new Error(this.notInitializedErrorMessage))
-          ),
+          // Push the value into queueScheduler
+          observeOn(queueScheduler),
```
alex-okrushko, line 126: I was very confident that `observeOn` was deprecated in favour of `scheduled`. Apparently I was dreaming?

alex-okrushko (approved):

timdeschryver (approved):

modules/component-store/src/component-store.ts
markostanimirovic: 😁 AFAIK, the signature of the `on` function that accepts scheduler is deprecated in favor of `scheduled`.

PR #3524, opened 2022-08-12
feat(effects): add provideEffects function

### LINKED ISSUE #3522: Effects: Introduce Standalone API for NgRx Effects
### Information

`provideEffects` function should have the same behavior as `EffectsModule.forRoot` and `EffectsModule.forFeature` methods, but it should provide a "module-free" developer experience. This function should have the same signature as `EffectsModule` methods - it should accept an array of effect source classes.

Related discussion: https://github.com/ngrx/platform/discussions/3399

To provide effects with standalone Angular features, we can use `importProvidersFrom` function:

```ts
// main.ts
bootstrapApplication(AppComponent, {
  providers: [importProvidersFrom(EffectsModule.forFeature([RouterEffects]))]
});

// feature.routes.ts:
export const featureRoutes: Route[] = [
  {
    path: '',
    component: FeatureComponent,
    providers: [importProvidersFrom(EffectsModule.forFeature([FeatureApiEffects]))],
  },
];
```

The `provideEffects` function should replace both `EffectsModule` methods, so it can be used in the following way:

```ts
// main.ts
bootstrapApplication(AppComponent, {
  providers: [provideEffects([RouterEffects])]
});

// feature.routes.ts:
export const featureRoutes: Route[] = [
  {
    path: '',
    component: FeatureComponent,
    providers: [provideEffects([FeatureApiEffects])],
  },
];
```

Unlike `EffectsModule.forRoot`, it won't be required to call `provideEffects` at the root level, if there are no root effects.

- [X] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3522 

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/src/provide_effects.ts
+++ b/modules/effects/src/provide_effects.ts
@@ -1,0 +1,79 @@
+import {
+  ENVIRONMENT_INITIALIZER,
+  inject,
+  InjectFlags,
+  Type,
+} from '@angular/core';
+import {
+  EnvironmentProviders,
+  FEATURE_STATE_PROVIDER,
+  ROOT_STORE_PROVIDER,
+  Store,
+} from '@ngrx/store';
+import { EffectsRunner } from './effects_runner';
+import { EffectSources } from './effect_sources';
+import { rootEffectsInit as effectsInit } from './effects_actions';
+
+/**
+ * Runs provided effects. Can be called at the root or feature level.
+ * Unlike the `EffectsModule.forRoot` method, this function does not
+ * need to be called at the root level if there are no root effects.
+ *
+ * @usageNotes
+ *
+ * ### Providing root effects
+ *
+ * ```ts
+ * bootstrapApplication(AppComponent, {
+ *   providers: [provideEffects([RouterEffects])],
+ * });
+ * ```
+ *
+ * ### Providing feature effects
+ *
+ * ```ts
+ * const booksRoutes: Route[] = [
+ *   {
+ *     path: '',
+ *     providers: [provideEffects([BooksApiEffects])],
+ *     children: [
+ *       { path: '', component: BookListComponent },
+ *       { path: ':id', component: BookDetailsComponent },
+ *     ],
+ *   },
+ * ];
+ * ```
+ */
+export function provideEffects(effects: Type<unknown>[]): EnvironmentProviders {
+  return {
+    ɵproviders: [
+      effects,
+      {
+        provide: ENVIRONMENT_INITIALIZER,
+        multi: true,
+        useValue: () => {
+          inject(ROOT_STORE_PROVIDER);
+          inject(FEATURE_STATE_PROVIDER, InjectFlags.Optional);
+
+          const effectsRunner = inject(EffectsRunner);
+          const effectSources = inject(EffectSources);
+          const shouldInitEffects = !effectsRunner.isStarted;
+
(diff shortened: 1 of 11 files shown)


modules/effects/src/provide_effects.ts
```diff
@@ -0,0 +1,64 @@
...
+export function provideEffects(effects: Type<unknown>[]): Provider[] {
+  return [
+    effects,
+    {
+      provide: ENVIRONMENT_INITIALIZER,
+      multi: true,
+      useValue: () => {
+        const effectsRunner = inject(EffectsRunner);
+        const effectSources = inject(EffectSources);
+        const shouldInitEffects = !effectsRunner.isStarted;
```
brandonroberts, line 46: We should use a provider token here that checks to see if there is a parent token above this injector. If not then start the effects and we don't have to use property on the class

modules/effects/src/provide_effects.ts
```diff
@@ -0,0 +1,64 @@
...
+ *       { path: ':id', component: BookDetailsComponent },
+ *     ],
+ *   },
+ * ];
+ * ```
+ */
+export function provideEffects(effects: Type<unknown>[]): Provider[] {
+  return [
+    effects,
+    {
```
brandonroberts, line 40: We also need to add a dependency on the Store/StoreFeature to be injected so we know its loaded. We do this with the StoreRootModule and StoreFeatureModule being injected into the EffectsRootModule and EffectsFeatureModule currently.

modules/effects/src/provide_effects.ts
markostanimirovic: I'm not sure it would work when `provideEffects` is called multiple times within the context of the same injector.

brandonroberts (approved):
🎉

modules/effects/src/provide_effects.ts
```diff
@@ -0,0 +1,79 @@
...
+ * const booksRoutes: Route[] = [
+ *   {
+ *     path: '',
+ *     providers: [provideEffects([BooksApiEffects])],
+ *     children: [
+ *       { path: '', component: BookListComponent },
+ *       { path: ':id', component: BookDetailsComponent },
+ *     ],
+ *   },
+ * ];
```
timdeschryver, line 44: I think "unlike" could be confusing, what about leaving it out?
On the same note, we could also reword the root and feature effects to make it clear that there is no distinction, and thus that you can use it in the various levels.

Does that make sense? 

```suggestion
 * Runs the provided effects. 
 * Can be called at the root and feature levels.
 *
 * @usageNotes
 *
 * ### Providing effects at the root level
 *
 * ```ts
 * bootstrapApplication(AppComponent, {
 *   providers: [provideEffects([RouterEffects])],
 * });
 * ```
 *
 * ### Providing effects at the feature level
 *
 * ```ts
 * const booksRoutes: Route[] = [
 *   {
 *     path: '',
 *     providers: [provideEffects([BooksApiEffects])],
 *     children: [
 *       { path: '', component: BookListComponent },
 *       { path: ':id', component: BookDetailsComponent },
 *     ],
 *   },
 * ];
```

timdeschryver (approved):

timdeschryver:
🚀🚀

PR #3539, opened 2022-08-16
feat(store): add provideStore and provideState functions for standalone APIs

### LINKED ISSUE #3526: Store: Introduce Standalone API for NgRx Store
### Information

Introduce standalone APIs for registering the Store and pieces of state for the `@ngrx/store` package. This should allow developers to use Store in a "module-free" application.

Based on the discussion in #3399 we are looking at adding two new functions

`provideStore(reducerMap, rootConfig)` - Registers the root-level providers for the global Store, and initial setup
`provideState(reducerMap, featureConfig)` - Registers additional features in the global Store

Also under consideration of whether we should separate the runtime checks from the `provideStore()` function so they can be tree-shaken in production builds

```ts
providers: [
  provideStore(),
  provideRuntimeChecks(config)
]
```

Related discussion: https://github.com/ngrx/platform/discussions/3399

Using `importProvidersFrom` function:

```ts
// main.ts
bootstrapApplication(AppComponent, {
  providers: [importProvidersFrom(StoreModule.forRoot({}))]
});

// feature.routes.ts:
export const featureRoutes: Route[] = [
  {
    path: '',
    component: FeatureComponent,
    providers: [importProvidersFrom(StoreModule.forFeature('name', reducer))],
  },
];
```

- [X] Yes
- [ ] No

brandonroberts:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI-related changes
[ ] Other... Please describe:
```

Closes #3526

```
[ ] Yes
[x] No
```


modules/store/src/provide_store.ts
```diff
@@ -0,0 +1,333 @@
...
+ * ### Providing the Global Store
+ *
+ * ```ts
+ * bootstrapApplication(AppComponent, {
+ *   providers: [provideStore({})],
+ * });
+ * ```
+ */
+export function provideStore<T, V extends Action = Action>(
+  reducers: ActionReducerMap<T, V> | InjectionToken<ActionReducerMap<T, V>>,
```
timdeschryver, line 318: Could we make reducers optional?
Reason: I like to just provide the store, and use features/state to build up the state tree. Otherwise, I'm worried that state end up in `provideStore` that doesn't belong there. And `provideStore()` looks better than `provideStore({})` imho.

```suggestion
  reducers?: ActionReducerMap<T, V> | InjectionToken<ActionReducerMap<T, V>>,
```

modules/store/src/store_config.ts
```diff
@@ -0,0 +1,108 @@
...
+  metaReducers: MetaReducer[],
+  userProvidedMetaReducers: MetaReducer[]
+): MetaReducer[] {
+  return metaReducers.concat(userProvidedMetaReducers);
+}
+
+export function _provideForRootGuard(store: Store<any>): any {
+  if (store) {
+    throw new TypeError(
+      `StoreModule.forRoot() called twice. Feature modules should use StoreModule.forFeature() instead.`
```
timdeschryver, line 104: Should this message be changed into something more generic?

projects/standalone-app-e2e/src/integration/app.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { getGreeting } from '../support/app.po';
+
+describe('standalone-app', () => {
+  beforeEach(() => cy.visit('/'));
+
+  it('should display welcome message', () => {
+    // Custom command example, see `../support/commands.ts` file
+    cy.login('my-email@something.com', 'myPassword');
```
timdeschryver, line 8: Great idea to add a new app to test these new features out.

If it's ok for you, I can update the e2e test in a follow-up PR.
Because now, this `login` command confused me because I didn't see a login prompt in the app. 
I would like to clean the cypress setup up, and also make sure that the feature is loaded correctly after navigating to the page.

modules/store/src/provide_store.ts
brandonroberts: Good catch. We've been trying to get to this point by default for a while where nothing is registered in the root

modules/store/src/store_config.ts
brandonroberts: Yep, will fix

projects/standalone-app-e2e/src/integration/app.spec.ts
brandonroberts: Yep, that's fine with me

modules/store/src/provide_store.ts
```diff
@@ -0,0 +1,333 @@
...
+            reducers,
+            initialState: _initialStateFactory(feature.initialState),
+          };
+        });
+
+        reducerManager.addFeatures(feats);
+      };
+    },
+  },
+];
```
markostanimirovic, line 101: We could move feature registration logic to a separate service. This service can be used in `provideEffects` to ensure that the store feature is registered before the effects.

```suggestion
@Injectable()
export class RegisterStateService {
  constructor() {
    // tokens can be injected via constructor 👇
    const features = inject<StoreFeature<any, any>[]>(_STORE_FEATURES);
    const featureReducers =
      inject<ActionReducerMap<any>[]>(FEATURE_REDUCERS);
    const reducerManager = inject(ReducerManager);
    inject(_ACTION_TYPE_UNIQUENESS_CHECK, InjectFlags.Optional);

     // feature registration logic can be moved to a separate method 👇
    const feats = features.map((feature, index) => {
      const featureReducerCollection = featureReducers.shift();
      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
      const reducers = featureReducerCollection! /*TODO(#823)*/[index];

      return {
        ...feature,
        reducers,
        initialState: _initialStateFactory(feature.initialState),
      };
    });

    reducerManager.addFeatures(feats);
  }
}

/**
 * Environment Initializer used in the feature
 * providers to register state features
 */
const ENVIRONMENT_STATE_PROVIDER: Provider[] = [
  RegisterStateService,
  {
    provide: ENVIRONMENT_INITIALIZER,
    multi: true,
    useFactory() {
      return () => inject(RegisterStateService);
    },
  },
];
```

Then, we can inject `RegisterStateService` in `provideEffects` to ensure that the store feature is registered before the effects:

```ts
export function provideEffects(effects: Type<unknown>[]): Provider[] {
  return [
    effects,
    {
      provide: ENVIRONMENT_INITIALIZER,
      multi: true,
      useValue: () => {
        inject(RegisterStateService, InjectFlags.Optional);

        // register effects
      },
    },
  ];
}
```

brandonroberts: Agreed. What about using another InjectionToken with a factory instead of a service?

markostanimirovic: Sounds good 👍

modules/store/src/provide_store.ts
```diff
@@ -0,0 +1,343 @@
...
+    reducerManager.addFeatures(feats);
+  },
+});
+
+/**
+ * Environment Initializer used in the feature
+ * providers to register state features
+ */
+const ENVIRONMENT_STATE_PROVIDER: Provider[] = [
+  { provide: FEATURE_STATE_PROVIDER, deps: [], useValue: true },
```
markostanimirovic, line 102: The `FEATURE_STATE_PROVIDER` factory will not be executed because of `useValue`. Also, when the factory is used for `InjectionToken`, but `providedIn` is not specified, it fallbacks to `root`.

We can refactor this part as follows:

```ts
export const FEATURE_STATE_PROVIDER = new InjectionToken<void>('@ngrx/store Feature State Provider');

function featureStateProviderFactory(): void {
 // logic is moved from `token.factory` to the standalone function
}

const ENVIRONMENT_STATE_PROVIDER: Provider = [
  { provide: FEATURE_STATE_PROVIDER, useFactory: featureStateProviderFactory },
  {
    provide: ENVIRONMENT_INITIALIZER,
    multi: true,
    useFactory() {
      return () => inject(FEATURE_STATE_PROVIDER);
    },
  },
];
```

modules/store/src/provide_store.ts
```diff
@@ -0,0 +1,343 @@
...
+        inject(ActionsSubject);
+        inject(ReducerObservable);
+        inject(ScannedActionsSubject);
+        inject(Store);
+        inject(_ROOT_STORE_GUARD, InjectFlags.Optional);
+        inject(_ACTION_TYPE_UNIQUENESS_CHECK, InjectFlags.Optional);
+      };
+    },
+  },
+];
```
markostanimirovic, line 310: Should we also inject the root provider to the `FEATURE_STATE_PROVIDER` factory to ensure that `provideStore` logic is executed before `provideState`?

```suggestion
export const ROOT_STORE_PROVIDER = new InjectionToken<void>('@ngrx/store Root Store Provider');

function rootStoreProviderFactory(): void {
  inject(ActionsSubject);
  inject(ReducerObservable);
  inject(ScannedActionsSubject);
  inject(Store);
  inject(_ROOT_STORE_GUARD, InjectFlags.Optional);
  inject(_ACTION_TYPE_UNIQUENESS_CHECK, InjectFlags.Optional);
}

const ENVIRONMENT_STORE_PROVIDER: Provider[] = [
  { provide: ROOT_STORE_PROVIDER, useFactory: rootStoreProviderFactory },
  {
    provide: ENVIRONMENT_INITIALIZER,
    multi: true,
    useFactory() {
      return () => inject(ROOT_STORE_PROVIDER);
    },
  },
];
```

Then we can inject the `ROOT_STORE_PROVIDER` token in `featurStateProviderFactory` to ensure that the root logic is executed:

```ts
function featureStateProviderFactory(): void {
  inject(ROOT_STORE_PROVIDER);

  // register store feature
}
```

We'll also need the `ROOT_STORE_PROVIDER` token for `provideEffects` function.

brandonroberts: Yep, added

markostanimirovic (approved):
Great work Brandon! 🏆

timdeschryver (approved):
🚀

PR #3578, opened 2022-09-16
refactor(store): optimize `freeze` helper function

swseverance:
Change implementation of `freeze` helper function so that it does not call `Object.freeze(...)` on an object that is already frozen

## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `freeze` helper function can end up calling `Object.freeze(...)` on the same object even if the object is already frozen.

## What is the new behavior?

An object is frozen only once.

```
[ ] Yes
[X] No
```

### DIFF
--- a/modules/store/src/meta-reducers/immutability_reducer.ts
+++ b/modules/store/src/meta-reducers/immutability_reducer.ts
@@ -14,7 +14,11 @@ export function immutabilityCheckMetaReducer(
   };
 }
 
-function freeze(target: any) {
+export function freeze(target: any) {
+  if (Object.isFrozen(target)) {
+    return target;
+  }
+
   Object.freeze(target);
 
   const targetIsFunction = isFunction(target);
--- a/modules/store/src/meta-reducers/index.ts
+++ b/modules/store/src/meta-reducers/index.ts
@@ -1,3 +1,3 @@
-export { immutabilityCheckMetaReducer } from './immutability_reducer';
+export { immutabilityCheckMetaReducer, freeze } from './immutability_reducer';
 export { serializationCheckMetaReducer } from './serialization_reducer';
 export { inNgZoneAssertMetaReducer } from './inNgZoneAssert_reducer';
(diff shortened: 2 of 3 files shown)


modules/store/src/meta-reducers/immutability_reducer.ts
```diff
@@ -14,33 +14,32 @@ export function immutabilityCheckMetaReducer(
   };
 }
 
-function freeze(target: any) {
-  Object.freeze(target);
+export function freeze(target: any) {
+  if (!Object.isFrozen(target)) {
```
timdeschryver, line 18: Can we invert this check to have an indention less?

```suggestion
  if (Object.isFrozen(target)) {
    return target;
  }
```

timdeschryver (commented):
Thanks for creating this PR.
In which case would the target already be frozen?

modules/store/src/meta-reducers/immutability_reducer.ts
swseverance: Updated!

swseverance:
> Thanks for creating this PR. In which case would the target already be frozen?

It could happen if the same object is passed as part of the payload for multiple actions. Not saying this is best practice but it can happen

modules/store/src/meta-reducers/immutability_reducer.ts
```diff
@@ -14,7 +14,11 @@ export function immutabilityCheckMetaReducer(
   };
 }
 
-function freeze(target: any) {
+export function freeze(target: any) {
```
markostanimirovic, line 17:
```suggestion
function freeze<T extends object>(target: T): T {
```

---

Line 38:

```diff
-const propValue = target[prop];
+const propValue = target[prop as keyof T];
```

modules/store/src/meta-reducers/index.ts
```diff
@@ -1,3 +1,3 @@
-export { immutabilityCheckMetaReducer } from './immutability_reducer';
+export { immutabilityCheckMetaReducer, freeze } from './immutability_reducer';
```
markostanimirovic, line 1: why do we need to export the `freeze` function?

markostanimirovic (changes requested):

PR #3580, opened 2022-09-16
fix(component-store): use asapScheduler to schedule lifecycle hooks check

### LINKED ISSUE #3573: Error: 1 periodic timer(s) still in the queue when testing a store with fakeAsync

component-store

```
it("doesn't work", fakeAsync(() => {
  new ComponentStore<{}>();
  expect(1).toEqual(1);
}));
```
Running this unit test produces
> Error: 1 periodic timer(s) still in the queue.

The test above should pass.

NgRx: 14.0.2
Angular: 14.2

### Other information

I believe it's a regression introduced with the lifecycle hooks. In particular, the run-time check in this PR: https://github.com/ngrx/platform/pull/3432/files#diff-dd156bfdaf294eb2eb0c9c87e1403ddd40190b88afa6ed8b443fb7cb1cd008f6R322

This check uses rxjs `asyncScheduler.schedule()`, which internally uses `setInterval` - a periodic task that offends `fakeAsync` :)
https://github.com/ReactiveX/rxjs/blob/6.x/src/internal/scheduler/AsyncAction.ts#L50

- [X] Yes
- [ ] No

amakhrov:
Switch from asyncScheduler to asapScheduler to prevent the "1 periodic timer(s) still in the queue" error.

## PR Checklist

  there are existing tests for this that still pass

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3573

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component-store/src/component-store.ts
+++ b/modules/component-store/src/component-store.ts
@@ -9,7 +9,7 @@ import {
   Subject,
   queueScheduler,
   scheduled,
-  asyncScheduler,
+  asapScheduler,
   EMPTY,
 } from 'rxjs';
 import {
@@ -322,7 +322,7 @@ export class ComponentStore<T extends object> implements OnDestroy {
    * but not used with provideComponentStore()
    */
   private checkProviderForHooks() {
-    asyncScheduler.schedule(() => {
+    asapScheduler.schedule(() => {
       if (
         isDevMode() &&
         (isOnStoreInitDefined(this) || isOnStateInitDefined(this)) &&
--- a/modules/component-store/spec/component-store.spec.ts
+++ b/modules/component-store/spec/component-store.spec.ts
@@ -37,7 +37,7 @@ import {
   Injector,
   Provider,
 } from '@angular/core';
-import { fakeAsync, tick } from '@angular/core/testing';
+import { fakeAsync, flushMicrotasks } from '@angular/core/testing';
 
 describe('Component Store', () => {
   describe('initialization', () => {
@@ -1651,7 +1651,7 @@ describe('Component Store', () => {
 
       const store = state.injector.get(LifecycleStore);
 
-      tick(0);
+      flushMicrotasks();
       expect(store.ngrxOnStoreInit).toBeDefined();
       expect(store['ɵhasProvider']).toBeTruthy();
       expect(console.warn).not.toHaveBeenCalled();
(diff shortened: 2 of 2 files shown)


modules/component-store/src/component-store.ts
```diff
@@ -322,7 +322,7 @@ export class ComponentStore<T extends object> implements OnDestroy {
    * but not used with provideComponentStore()
    */
   private checkProviderForHooks() {
-    asyncScheduler.schedule(() => {
+    asapScheduler.schedule(() => {
       if (
         isDevMode() &&
```
amakhrov, line 327: Additionally, I'm wondering whether the `isDevMode()` check should be perform at the very beginning, guarding the full `asapScheduler.schedule` call, instead of just scheduled callback body

brandonroberts: I tried using `isDevMode()` outside the callback body, and it doesn't work as intended.

markostanimirovic:
I renamed the PR title to avoid potential confusion because we can use ComponentStore within the `fakeAsync`, but we need to add at least one `tick` that will flush scheduled macrotasks:

```ts
it('works before changes from this PR', fakeAsync(() => {
  new ComponentStore();
  tick(0);
}); // no error
```

modules/component-store/src/component-store.ts
```diff
@@ -322,7 +322,7 @@ export class ComponentStore<T extends object> implements OnDestroy {
    * but not used with provideComponentStore()
    */
   private checkProviderForHooks() {
-    asyncScheduler.schedule(() => {
+    asapScheduler.schedule(() => {
```
markostanimirovic, line 325: Thanks for the PR, @amakhrov! Can you add a unit test that verifies that there is no error?

markostanimirovic (changes requested):

modules/component-store/src/component-store.ts
amakhrov: what if I just change `tick(0)` to `flushMicrotasks()` in the existin tests (there are 2 of them for the hooks feature).

markostanimirovic: sounds good 👍

markostanimirovic (approved):

modules/component-store/src/component-store.ts
aokrushk: @timdeschryver It's hard to test for that. I'd say the `flushMicrotasks();` change in unit tests is the one that "reproduces" the issue.

aokrushk (approved):

timdeschryver (approved):

Issue #3572, opened 2022-09-12
Migration: typed projector methods


timdeschryver:
### Information

This issue is dependent on #3571.
When #3571 is resolved, we want to add a migration so that we don't break any existing codebases.

@david-shortman proposed a migration plan at https://github.com/ngrx/platform/discussions/3097#discussioncomment-3540950, which we can implement. 

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [ ] No

david-shortman:
This can be closed since we're only doing a breaking change

PR #3581, opened 2022-09-17
feat(store): strict projectors

### LINKED ISSUE #3571: Selectors: typed projector methods
### Information

This issue comes from a discussion https://github.com/ngrx/platform/discussions/3097

Currently when using the `selector.projector` method, we can pass it any argument(s) and it won't complain. With this change we want to make the projectors type-safe so that mistakes can be caught earlier. 

/

- [ ] Yes
- [ ] No

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3571 

## What is the new behavior?

The `projector` is strict by default, but can be bypassed with an `any` generic parameter.

```
[x] Yes
[ ] No
```

A migration will be included to include `<any>` for every existing projector.

### DIFF
--- a/modules/store/src/selector.ts
+++ b/modules/store/src/selector.ts
@@ -17,13 +17,32 @@ export type ComparatorFn = (a: any, b: any) => boolean;
 
 export type DefaultProjectorFn<T> = (...args: any[]) => T;
 
+type ProjectorStrictness = 'strict';
+
+type SelectorProjectorArgs<
+  StrictnessConfig extends ProjectorStrictness,
+  ProjectorArgs
+> = StrictnessConfig extends 'strict'
+  ? ProjectorArgs extends unknown[]
+    ? ProjectorArgs
+    : any[]
+  : any[];
+
+type SelectorProjectorFn<ProjectorFn> = ProjectorFn extends (
+  ...args: infer ProjectorArgs
+) => infer ProjectorResult
+  ? <Strictness extends ProjectorStrictness = 'strict'>(
+      ...args: SelectorProjectorArgs<Strictness, ProjectorArgs>
+    ) => ProjectorResult
+  : ProjectorFn;
+
 export interface MemoizedSelector<
   State,
   Result,
   ProjectorFn = DefaultProjectorFn<Result>
 > extends Selector<State, Result> {
   release(): void;
-  projector: ProjectorFn;
+  projector: SelectorProjectorFn<ProjectorFn>;
   setResult: (result?: Result) => void;
   clearResult: () => void;
 }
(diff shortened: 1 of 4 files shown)


david-shortman:
Not sure how to approach the migration (#3572).

Since the `projector` is found on arbitrary objects, then without compiling the code and inspecting type information, you could only naively update any call of a function named `projector` referenced as a property from an object.

modules/store/src/selector.ts
```diff
@@ -17,13 +17,32 @@ export type ComparatorFn = (a: any, b: any) => boolean;
...
+    : any[]
+  : any[];
+
+type SelectorProjectorFn<ProjectorFn> = ProjectorFn extends (
+  ...args: infer ProjectorArgs
+) => infer ProjectorResult
+  ? <Strictness extends ProjectorStrictness = 'strict'>(
+      ...args: SelectorProjectorArgs<Strictness, ProjectorArgs>
+    ) => ProjectorResult
+  : ProjectorFn;
```
timdeschryver, line 37: If I understand this correctly (and feel free to correct me), these types are added to be able to "opt-out" of the strictness by providing `any` as the generic (`projector<any>(...)`)?

Would we still need extra types if we would not support this? 
For example as it was implemented in #2319

> Since the projector is found on arbitrary objects, then without compiling the code and inspecting type information, you could only naively update any call of a function named projector referenced as a property from an object.

You're right. Initially I was thinking we could do this only for spec|test files, but that could possibly update incorrect expressions.
On that note, do we need to allow `any`? I'm of the opinion that we shouldn't, because we want to encourage best practices. If a user still wants to be able to pass incorrect types to `projector` we shouldn't make it that easy, they can always do `(selector.projector as any)(arg1, arg2)` as a workaround.

david-shortman: The hope was that strictly typed projectors could be introduced without being a breaking change. But, perhaps this is a case where a major version will introduce a type change that will require dependents to manually react. Hopefully it is valuable enough to force a change and not time consuming enough for dependents to resolve.

david-shortman: Oh, to answer the other question, yes, the added type complexity is to support passing in the `any` generic. That complexity can be removed if we think people could just cast the projector itself as `any`.

timdeschryver: If providing a migration isn't possible then I'm in favor of removing the complexity, and make this a breaking change.

david-shortman:
Planning to update to remove the conditional generics this week.

brandonroberts:
What's the experience for someone upgrading from non-strict projectors? Are all their existing selectors going to produce type errors?

david-shortman:
> What's the experience for someone upgrading from non-strict projectors? Are all their existing selectors going to produce type errors?

Anywhere a projector is called, it will be required to be called with the correct arguments. Projectors called with incorrect arguments will produce errors.

brandonroberts:
Ok, we definitely need to provide instructions on how to work around this for existing users in the migration guide, as I'm sure there will be some breakages for existing apps.

david-shortman:
> Ok, we definitely need to provide instructions on how to work around this for existing users in the migration guide, as I'm sure there will be some breakages for existing apps.

I'll look to add to this PR.

modules/store/src/selector.ts
```diff
@@ -17,13 +17,19 @@ export type ComparatorFn = (a: any, b: any) => boolean;
 
 export type DefaultProjectorFn<T> = (...args: any[]) => T;
 
+type SelectorProjectorFn<ProjectorFn> = ProjectorFn extends (
+  ...args: infer ProjectorArgs
+) => infer ProjectorResult
+  ? (...args: ProjectorArgs) => ProjectorResult
+  : ProjectorFn;
+
```
markostanimirovic, line 25: Do we really need the extra type here? It seems to work without `SelectorProjectorFn`.

```suggestion

```

modules/store/src/selector.ts
```diff
@@ -17,13 +17,19 @@ export type ComparatorFn = (a: any, b: any) => boolean;
...
 export interface MemoizedSelector<
   State,
   Result,
   ProjectorFn = DefaultProjectorFn<Result>
 > extends Selector<State, Result> {
   release(): void;
-  projector: ProjectorFn;
+  projector: SelectorProjectorFn<ProjectorFn>;
```
markostanimirovic, line 32:
```suggestion
  projector: ProjectorFn;
```

markostanimirovic (changes requested):

modules/store/src/selector.ts
david-shortman: Yeah, this is unneeded now. This was an artifact of attempting to deal with types that may not have been functions passed into the generic, but should have been cleaned up.

timdeschryver (approved):
LGTM 🎉
We should also provide the same changes to `MemoizedSelectorWithProps` in a follow-up PR.

brandonroberts (approved):
📦

david-shortman:
> LGTM 🎉
> 
> We should also provide the same changes to `MemoizedSelectorWithProps` in a follow-up PR.

I assumed, since they're deprecated, they shouldn't receive updates.

PR #3638, opened 2022-11-02
feat(effects): forRoot and forFeature accept spreaded array

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

## What is the new behavior?

Update the signature of `EffectsModule.forRoot` and `EffectsModule.forFeature` to also accept spreaded arrays.
The second commit updates the `provideEffects` method to have the same signature as `EffectsModule`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/spec/types/effects_module.spec.ts
+++ b/modules/effects/spec/types/effects_module.spec.ts
@@ -1,0 +1,66 @@
+import { expecter } from 'ts-snippet';
+import { compilerOptions } from './utils';
+
+describe('EffectsModule()', () => {
+  const expectSnippet = expecter(
+    (code) => `
+      import { EffectsModule } from '@ngrx/effects';
+
+      class EffectFixture {}
+
+      ${code}`,
+    compilerOptions()
+  );
+
+  describe('forRoot', () => {
+    it('should compile without params', () => {
+      expectSnippet(`
+        EffectsModule.forRoot();
+      `).toSucceed();
+    });
+
+    it('should compile with a single param', () => {
+      expectSnippet(`
+        EffectsModule.forRoot(EffectFixture);
+      `).toSucceed();
+    });
+
+    it('should compile with an array param', () => {
+      expectSnippet(`
+        EffectsModule.forRoot([EffectFixture]);
+      `).toSucceed();
+    });
+
+    it('should compile with a spreaded array param', () => {
+      expectSnippet(`
+        EffectsModule.forRoot(EffectFixture, EffectFixture);
+      `).toSucceed();
+    });
+  });
+
+  describe('forFeature', () => {
+    it('should compile without params', () => {
+      expectSnippet(`
+        EffectsModule.forFeature();
+      `).toSucceed();
+    });
+
+    it('should compile with a single param', () => {
+      expectSnippet(`
+        EffectsModule.forFeature(EffectFixture);
+      `).toSucceed();
+    });
+
+    it('should compile with an array param', () => {
+      expectSnippet(`
+        EffectsModule.forFeature([EffectFixture]);
+      `).toSucceed();
+    });
+
+    it('should compile with a spreaded array param', () => {
+      expectSnippet(`
+        EffectsModule.forFeature(EffectFixture, EffectFixture);
+      `).toSucceed();
+    });
+  });
+});
--- a/modules/effects/spec/types/provide_effects.spec.ts
+++ b/modules/effects/spec/types/provide_effects.spec.ts
@@ -1,0 +1,38 @@
+import { expecter } from 'ts-snippet';
+import { compilerOptions } from './utils';
+
+describe('provideEffects()', () => {
+  const expectSnippet = expecter(
+    (code) => `
+      import { provideEffects } from '@ngrx/effects';
+
+      class EffectFixture {}
+
+      ${code}`,
+    compilerOptions()
+  );
+
+  it('should compile without params', () => {
+    expectSnippet(`
+        provideEffects();
+      `).toSucceed();
+  });
+
+  it('should compile with a single param', () => {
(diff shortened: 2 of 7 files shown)


modules/effects/src/provide_effects.ts
```diff
@@ -3,6 +3,7 @@ import {
   EnvironmentProviders,
   inject,
   makeEnvironmentProviders,
+  ModuleWithProviders,
```
brandonroberts, line 6: Remove unused import

modules/effects/src/provide_effects.ts
```diff
@@ -13,6 +14,7 @@ import {
 import { EffectsRunner } from './effects_runner';
 import { EffectSources } from './effect_sources';
 import { rootEffectsInit as effectsInit } from './effects_actions';
+import { EffectsFeatureModule } from './effects_feature_module';
```
brandonroberts, line 17: Remove unused import

modules/effects/spec/types/effects_module.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      expectSnippet(`
+        EffectsModule.forRoot(EffectFixture, EffectFixture);
+      `).toSucceed();
+    });
+  });
+
+  describe('forFeature', () => {
+    it('should compile without params', () => {
+      expectSnippet(`
+        EffectsModule.forFeature();
```
brandonroberts, line 44: Question: Why would we allow this?

modules/effects/spec/types/provide_effects.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      class EffectFixture {}
+
+      ${code}`,
+    compilerOptions()
+  );
+
+  it('should compile without params', () => {
+    expectSnippet(`
+        provideEffects();
```
brandonroberts, line 17: Same question as above

brandonroberts (changes requested):

modules/effects/spec/types/effects_module.spec.ts
timdeschryver: This was previously also supported because we provided an empty array as default argument

https://github.com/ngrx/platform/blob/43198a2fa8d11faa3e4cc080b9468bac385b002b/modules/effects/src/effects_module.ts#L44

modules/effects/spec/types/provide_effects.spec.ts
timdeschryver: To keep the API the same.
I also don't know if it would be possible to throw here because of the spreaded array syntax (it would be an empty array in this case)

modules/effects/spec/types/effects_module.spec.ts
brandonroberts: I see. Thanks

brandonroberts (approved):

modules/effects/src/provide_effects.ts
```diff
@@ -44,10 +44,11 @@ import { rootEffectsInit as effectsInit } from './effects_actions';
  * ```
  */
 export function provideEffects(
-  ...effects: Type<unknown>[]
+  ...effects: Type<unknown>[] | Type<unknown>[][]
 ): EnvironmentProviders {
```
markostanimirovic, line 48: I'd add two signatures here to avoid the ability to pass multiple arrays, for example:

```ts
provideEffects([Effects1, Effects2], [], [Effects3]); // this would work with current impl
```

```suggestion
export function provideEffects(effects: Type<unknown>[]): EnvironmentProviders;
export function provideEffects(...effects: Type<unknown>[]): EnvironmentProviders;
export function provideEffects(
  ...effects: Type<unknown>[] | Type<unknown>[][]
): EnvironmentProviders {
```

timdeschryver: I started out like that, but then we lose the TS Docs - or it should be duplicated for all signatures. Or is there a way around that?

markostanimirovic: I think if we put the TSDoc comment above the implementation, API docs will be generated properly:

```ts
export function provideEffects(effects: Type<unknown>[]): EnvironmentProviders;
export function provideEffects(...effects: Type<unknown>[]): EnvironmentProviders;
/**
 * ...
 */
export function provideEffects(
  ...effects: Type<unknown>[] | Type<unknown>[][]
): EnvironmentProviders {
  // ...
}
```

timdeschryver: Not that I can see:

image

While adding it to the correct overload results in:

image

markostanimirovic: We can add descriptions for signatures too, but the `@usageNotes` section needs to be above the implementation, otherwise it won't be displayed in the [API docs](http://localhost:4200/api/effects/provideEffects). I've just checked this locally.

```ts
/**
 * description for signature 1
 */
export function provideEffects(effects: Type<unknown>[]): EnvironmentProviders;
/**
 * description for signature 2
 */
export function provideEffects(...effects: Type<unknown>[]): EnvironmentProviders;
/**
 * description
 *
 * @usageNotes
 *
 * ### Providing effects at the root level
 *
 * ...
 */
export function provideEffects(
  ...effects: Type<unknown>[] | Type<unknown>[][]
): EnvironmentProviders {
  // ...
}
```

timdeschryver: Oh cool, I didn't know that.
Let me verify it later today, thanks @markostanimirovic !

markostanimirovic (approved):

markostanimirovic:
🚀

PR #3646, opened 2022-11-03
feat(component-store): add a testing library

### LINKED ISSUE #2767: Is it possible to add component-store test example to documentation?
There is store testing in the documentation:  
https://ngrx.io/guide/store/testing
Is it possible to have something similar for component store?

BobobUnicorn:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There is no simple unified way that NgRx provides to mock a `ComponentStore`; this adds a mocking library with a simple API to facilitate writing tests for components that use the `ComponentStore`.

Closes #2767.

```
[ ] Yes
[x] No
```


angular.json
```diff
@@ -1043,5 +1043,8 @@
       "schematics": {},
       "tags": []
     }
+  },
+  "cli": {
+    "analytics": false
```
richieforeman, line 1048: Revert this, they probably don't want that.

modules/component-store/testing/src/testing.ts
```diff
@@ -0,0 +1,211 @@
+import { Provider, Type } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import 'jasmine';
```
richieforeman, line 3: Given the hard dep on jasmine here, you'd need to be using Jasmine to use this (which makes sense, given the contribution here).

Does it make sense to put this in

`~/modules/component-store/testing/jasmine/xxx/yyy`

Imaging that the namespace is..

`@ngrx/component-store/testing/jasmine`

tsconfig.json
```diff
@@ -22,6 +22,7 @@
       "@ngrx/component-store/schematics-core": [
         "./modules/component-store/schematics-core"
       ],
+      "@ngrx/component-store/testing": ["./modules/component-store/testing"],
```
richieforeman, line 25: On my earlier note, should we update this here as well?

modules/component-store/testing/src/testing.ts
```diff
@@ -0,0 +1,211 @@
...
+ * Angular providers and an injection token for a mocked component store.
+ *
+ * For usage, see {@link provideMockComponentStores}.
+ */
+export interface MockComponentStoreFactory<T> {
+  providers: Provider[];
+  token: Type<MockComponentStore<T>>;
+}
+
+interface Constructor<ClassType> {
```
richieforeman, line 21: Nit/opt:  This interface allows you to infer that a constructor for a given type was passed.
- Should this be "`ConstructorOf<T>`" ?

I do not feel strongly, but I did feel like perhaps this might have sounded like a built-in, and it might be good to disambiguate.  /shruggie

richieforeman (changes requested):
Before the core team gets to it, I figured I'd give a review here, based on maybe some initial thoughts.

modules/component-store/testing/src/testing.ts
BobobUnicorn: I wonder if it makes sense to make this generic; the consumer would pass in a parameter that can create a "spy", and a second one that takes a spy and configures it to return a value.

LayZeeDK: Or use a testing framework-agnostic test double library such as [Sinon.JS](https://sinonjs.org/).

markostanimirovic:
I will close this PR as it has not been updated for a long time.

PR #3664, opened 2022-11-15
feat(data): Add HttpOptions to EntityActionOptions (#3663)

### LINKED ISSUE #3663: RFC: Add HttpOptions to EntityActionOptions to allow finer control over HttpClient requests

ngrx/data

### Information

Currently it is difficult to influence http requests sent in response to dispatched `EntityAction`s.  Simple requirements such as adding query parameters or headers require multiple ngrx/data classes to be overriden and provided. Often those implementing ngrx/data have an existing api they need to integrate with, and endpoints that require more than just an id/entity/list of entities to function correctly. 

Proposing to add `HttpOptions` to `EntityActionOptions` that support finer control over the http requests executed by Data Services.

Change summary:

- New `EntityActionOptions` property: `HttpOptions`:

```
/** Options of an EntityAction */
export interface EntityActionOptions {

  /** Correlate related EntityActions, particularly related saves. Must be serializable. */
  readonly correlationId?: any;

  /** other properties hidden for brevity */

  /** Options that will be passed to the dataService http request. Allows setting of Query Parameters and Headers */
  readonly httpOptions?: HttpOptions;
}

/**
 * Options that adhere to the constructor arguments for HttpParams and
 * HttpHeaders.
 */
export interface HttpOptions {
   httpParams?: HttpParamsOptions,
   httpHeaders?: string | {[p: string]: string | string[]};
}
```
- EntityEffects to be updated to pass `HttpOptions` from an `EntityAction` to `DataService` requests
- `EntityCollectionDataService` api to be updated to include optional `HttpOptions` in each supported request
- `DefaultDataService` to be updated to pass `HttpOptions` along to the `HttpClient` requests. Both httpParams and httpHeaders are the correct type for the constructors of their respective angular classes: `HttpParams` and `HttpHeaders` as at Ng14.
- `getWithQuery` should continue to accept `QueryParams` in arg 1 so as not to break clients. `HttpOptions` includes passing Http Params and is available in arg 2. 
  If specified, http params in `HttpOptions` will _replace_ those in arg 1 `QueryParams`. `QueryParams` has been annotated as deprecated, however removing the first arg would be a breaking change so looking for guidance from the core team on whether deprecation is required here.  It might be possible to retire `getWithQuery` and use `getAll` with HttpOptions instead. Food for thought.
- `QueryParams` type has been updated to reflect the (now exported) `HttpParamsOptions` `fromObject` type def (adds `boolean`, `number`, and their respective `ReadOnlyArray` types).

Related Issues
 - #1944
 
This issue proposed passing the entire EntityAction to data services. This [comment](https://github.com/ngrx/platform/pull/1944#issuecomment-508360056) by @wardbell outlines the solution. 
  I've taken a different approach for a couple of reasons. 
  
  1. This solution provides a clear api to query params in the Data Service requests from data passed into the action.  Passing EntityAction would make data available, but it would still require the user to register a custom data services in order to make use of the extra data to form query params or manipulate the url.
  2. I'm erring on the side of not 'oversharing' - provide everything in case its useful. Instead I provide data that directly integrates into the http request so provides a clear way to add query parameters and an encoder, and/or headers. I've currently  I also forsee needs arising for using the `HttpContext` feature, and possibly (for the sake of requirements that can't be anticipated) a `userDefinedOptions: any` property to `HttpOptions` so user's can pass whatever they want.
 - #3093 
  this discussion deals with adding url parameters which this feature does not address, however it also mentions support for pagination which could be supported by the ability to add query parameters.

The workaround to this would be having to provide custom implementations of various ngrx data classes and factories. This ability also offers an alternative to hijacking the EntityAction `tag` property.

- [X] Yes
- [ ] No

cam-m:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Angular HttpParams and HttpHeaders cannot be passed from EntityCollectionDataService to underlying Data Service requests.

Closes #3663

## What is the new behavior?
Http Params and HttpHeaders can be passed from EntityCollectionDataService to underlying Data Service requests.

```
[] Yes
[x] No
```

Method`getWithQuery(queryParams)` can be refactored in a future change to remove the queryParams arg if desired. It is made redundant by the new httpOptions arg.


cam-m:
This is my first ever pull request so please let me know if I've stuffed anything up :)

brandonroberts:
As this would be a breaking change, it would have to wait until the next major release. Also, does this PR https://github.com/ngrx/platform/pull/3717 achieve the same result?

cam-m:
> As this would be a breaking change, it would have to wait until the next major release. Also, does this PR #3717 achieve the same result?

No, and I was able to work around #3717 using the clearCache method. 

This PR provides more general capability - by adding httpoptions to entity actions users can pass query params to any http requests flowing from entity actions - I raised this so I could pass params to a delete request.

I considered doing it as a non-breaking change, but I think the api would be less coherent that way. If you think users would be taking advantage of the current `options` parameter then perhaps I should re-do as non breaking...?

cam-m:
Also if I'm not mistaken, this PR would  solve #3717 by allowing the user to pass query params to the load action.

brandonroberts:
If there is an option to do a non-breaking change in the short term that would be preferable. We can deprecate the current usage and  do a breaking change in the next major if necessary.

modules/data/src/dataservices/interfaces.ts
```diff
@@ -24,9 +31,28 @@ export interface RequestData {
...
+    | ReadonlyArray<string | number | boolean>;
+}
+
+/**
+ * Options that adhere to the constructor arguments for HttpParams and
+ * HttpHeaders.
+ */
+export interface HttpOptions {
+  httpParams?: HttpParamsOptions;
+  httpHeaders?: string | { [p: string]: string | string[] };
```
timdeschryver, line 56: can we use `HttpHeaders` here?

modules/data/src/dataservices/interfaces.ts
```diff
@@ -24,9 +31,28 @@ export interface RequestData {
...
+  [key: string]: any;
```
timdeschryver, line 57: Why do we need to use the index property here?

modules/data/src/dataservices/default-data.service.ts
```diff
@@ -68,67 +70,112 @@ export class DefaultDataService<T> implements EntityCollectionDataService<T> {
...
+        headers: httpOptions?.httpHeaders
+          ? new HttpHeaders(httpOptions?.httpHeaders)
+          : undefined,
+        params: httpOptions?.httpParams
+          ? new HttpParams(httpOptions?.httpParams)
+          : undefined,
+      };
+    }
+
+    // If any options have been specified, pass them to http client. Note
```
timdeschryver, line 159: An idea for until we make the breaking change: I think it would be useful we log (`console.warn`) a message when an option is overridden.

modules/data/src/dataservices/interfaces.ts
cam-m: Wanted to leave the type open so user's could pass custom properties through. Had angular's newish context concept in mind, but happy to remove it if preferred.

modules/data/src/dataservices/interfaces.ts
cam-m: reason was that I think `HttpHeaders` is not serializable - to be in the payload of an action it needs to be, correct?.  The HttpHeaders object is then created 'just in time' for the http request call.

However I now see that `HttpParamsOptions` isn't serializable either because of the `encoder?: HttpParameterCodec;` property.

I should probably really use a custom type for that too. Do you agree?

modules/data/src/dataservices/default-data.service.ts
cam-m: Agree, I'll add this now

modules/data/src/dataservices/interfaces.ts
cam-m: Or could add doco that users shouldn't use the `encoder` property...

modules/data/src/dataservices/default-data.service.ts
cam-m: @timdeschryver do these look ok?
1. All usage of getWithQuery:
`Warning: The queryParams parameter of getWithQuery will be removed in next major release.`

2. Where queryOptions and HttpOptions are both used:
Warning: HttpOptions.httpParams will be merged with `queryParams` when both are are provided. In the event of a conflict HttpOptions.httpParams will override queryParams`.

modules/data/src/dataservices/interfaces.ts
timdeschryver: I think introducing a serializable interface would be the best here.
We can convert it to queryparams/headers before sending the request.

modules/data/src/dataservices/default-data.service.ts
timdeschryver: 2 get's my vote 😀
It can also include `The queryParams parameter of getWithQuery will be removed in next major release`

cam-m: Okeydoke, adding now, thanks Tim

modules/data/src/dataservices/interfaces.ts
cam-m: [HttpContext](https://angular.io/api/common/http/HttpContext)

I'd thought users overriding the default data service may want to be able to pass arbitrary data via the action so they can set HttpContext properties.

Happy to remove it though, I don't need it myself, just seemed potentially useful.

modules/data/src/dataservices/interfaces.ts
cam-m: agree, doing now

modules/data/src/dataservices/interfaces.ts
timdeschryver: If you don't need it, I would say to leave it out.
Otherwise we broaden our surface without anyone using it.
If it's needed in the future we can always add it.

modules/data/src/dataservices/interfaces.ts
cam-m: done

modules/data/src/dataservices/interfaces.ts
cam-m: done

modules/data/src/dataservices/default-data.service.ts
cam-m: done

modules/data/src/dataservices/default-data.service.ts
```diff
@@ -111,7 +111,9 @@ export class DefaultDataService<T> implements EntityCollectionDataService<T> {
         ? { fromString: queryParams }
         : { fromObject: queryParams };
     const params = new HttpParams(qParams);
-
+    console.warn(
+      'Warning: options.httpParams will be merged with queryParams when both are are provided to getWithQuery(). In the event of a conflict HttpOptions.httpParams will override queryParams`. The queryParams parameter of getWithQuery() will be removed in next major release.'
+    );
```
timdeschryver, line 116: Sorry, I went a bit too fast yesterday.
In the other places we also prefix the message with the source.

```suggestion
    console.warn(
      '@ngrx/data: options.httpParams will be merged with queryParams when both are are provided to getWithQuery(). In the event of a conflict HttpOptions.httpParams will override queryParams`. The queryParams parameter of getWithQuery() will be removed in next major release.'
    );
```

Similar to https://github.com/ngrx/platform/blob/e8cf02b4748ea6a3010c9b7db8818e8b666bd73b/modules/entity/src/utils.ts#L7


We also wrap this log in a condition, so that it only logs the message in dev environment, and not in a production environment. If possible, could we also wrap this in a condition so that it will only log this warning when the params will be merged?

cam-m: Cool, done!

cam-m:
> As this would be a breaking change, it would have to wait until the next major release. Also, does this PR https://github.com/ngrx/platform/pull/3717 achieve the same result?

Hey @brandonroberts, just FYI this now a non-breaking change.

brandonroberts (approved):

brandonroberts:
Thanks @cam-m!

e-oz:
It is a breaking change.
In 15.2.1 call `execute('POST', urlWithParams, requestData, {observe: 'response'})` returns HttpResponse object (because of `observe: 'response'`), in 15.3 the same code returns just response body.

timdeschryver:
Do you want to create a Pull Request for this @e-oz ?

e-oz:
Pull Request to change what exactly? I don't mind :)

cam-m:
You're right @e-oz , I missed the case of user's adding http request options _other_ than headers and params. I'm happy to address this in a PR.



 I'm happy to submit a PR to address this.

brandonroberts:
@cam-m that would be great thanks

cam-m:
FYI I've not forgotten about this, have had trouble getting the unit tests to run on my win machine - switched to a mac and should have a PR in today.

PR #3695, opened 2022-12-04
feat(store): make toActionName case more customizable

### LINKED ISSUE #3693: Allows createActionGroup to preserve the case

store

### Information

I would like to be able to preserve the case of the event keys I put :grimacing: 

image

By the way: thank you very much for the improved APIs, I am totally fan! 

image

Unfortunately it will considerably increase any refactoring effort, I really like to have the name of my action being the name of its variable + (removing the spaces for the name of actions allows to have more characters visible in the Redux DevTools which helps a lot in case of long names).

- [X] Yes
- [ ] No

hadrien-toma:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

Closes #3693

## What is the new behavior?

image

- [ ] Yes
- [x] No

### DIFF
--- a/modules/store/spec/action_group_creator.spec.ts
+++ b/modules/store/spec/action_group_creator.spec.ts
@@ -18,11 +18,11 @@ describe('createActionGroup', () => {
   });
 
   it('should create action name by camel casing the event name', () => {
-    expect(booksApiActions.loadBooksSuccess).toBeDefined();
+    expect(booksApiActions.loadBOOKSSuCCess).toBeDefined();
   });
 
   it('should create action type using the "[Source] Event" pattern', () => {
-    expect(booksApiActions.loadBooksSuccess().type).toBe(
+    expect(booksApiActions.loadBOOKSSuCCess().type).toBe(
       '[Books API]  Load BOOKS  suCCess  '
     );
   });
--- a/modules/store/spec/types/action_group_creator.spec.ts
+++ b/modules/store/spec/types/action_group_creator.spec.ts
@@ -80,9 +80,9 @@ describe('createActionGroup', () => {
             },
           });
 
-          let loadBooksSuccess: typeof booksApiActions.loadBooksSuccess;
+          let loadBOOKSSuCCess: typeof booksApiActions.loadBOOKSSuCCess;
         `).toInfer(
-          'loadBooksSuccess',
+          'loadBOOKSSuCCess',
           `ActionCreator<
             "[Books API]  Load BOOKS  suCCess  ",
             () => TypedAction<"[Books API]  Load BOOKS  suCCess  ">
@@ -196,11 +196,11 @@ describe('createActionGroup', () => {
           const booksApiActions = createActionGroup({
             source: 'Books API',
             events: {
-              '  Load BOOks  success ': emptyProps(),
-              'load Books Success': props<{ books: string[] }>(),
+              '  Load BOOks  Success ': emptyProps(),
+              'load BOOks Success': props<{ books: string[] }>(),
             }
           });
-        `).toFail(/loadBooksSuccess action is already defined/);
+        `).toFail(/loadBOOksSuccess action is already defined/);
       });
     });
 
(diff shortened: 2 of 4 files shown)


markostanimirovic:
Hi @hadrien-toma 👋 

Thanks for opening this PR! However, it would introduce a **breaking change**, so it won't be possible to include this change in a minor v15 release.

I'm going to close this PR for now. We'll consider request #3693 for the next major release.

hadrien-toma:
Alright, thank you for your answer. May you help me to understand where the breaking change is? I thought it was retrocompatible beauce I thought it was only enlarging the type 😬.

By the way: the v15 is simply amazing, thank you very much 👌🚀! Can't wait to use actionGroup now 😋

hadrien-toma:
Also @markostanimirovic I propose to give help by adding a unit test highlighting this breaking change when I have understood it if you think it could be interesting 😊

markostanimirovic:
Thank you! 😊

> May you help me to understand where the breaking change is?

If we introduce this change, the following code that is currently correct will not compile:

```ts
const UsersApiActions = createActionGroup({
  source: 'Users API',
  events: {
    'USERS Loaded Success': emptyProps(),
  },
});

// compilation of the line below succeeds with the current NgRx version
UsersApiActions.usersLoadedSuccess();
// however, if we introduce the change from this PR, compilation will fail
// because the action name will be `uSERSLoadedSuccess`
```

These tests validate the current behavior:
- https://github.com/ngrx/platform/pull/3695/files#diff-ea3185170f36adfef9dccbf98bd0b7ab350c483f7807c9691ad50023abf6bb5fL21
- https://github.com/ngrx/platform/pull/3695/files#diff-3384d5be8438d0ef687ffdd0f10b89acaf32209033f91f72a1f1d2331363d1efL83

hadrien-toma:
Ah! Thank you very much @markostanimirovic ! I think I understand better now! So... I dare to ask :sweat_smile: : would it be okay if I suggest a PR with another function named toActionNameCaseSensitive and add the rest of the code to independently support this new functionality by only "adding" codes (and their tests)? It would be so nice if I could group my actions even in v15 :scream: :sweat_smile:. Also I am currently rewritting my Nx generators so it would be very nice if I can have the maximum of new feature in them before the technical debt is spread everywhere :grimacing:

timdeschryver:
We won't extend our current API with `toActionNameCaseSensitive`. Doing this adds another surface that needs to be maintained, and could be removed in v16 (if we change the current behavior) resulting in a breaking change.
If it's really blocking I would suggest you to create `toActionNameCaseSensitive` in your codebase, which could be replaced with our API in the future.

I do think it make sense to keep the casing.
e.g.

```ts
'USERS Loaded Success' => USERSLoadedSuccess
'USERS loaded success' => USERSLoadedSuccess
'Users loaded success' => UsersLoadedSuccess
UsersLoadedSuccess => UsersLoadedSuccess
```

hadrien-toma:
Thank you for your answer @timdeschryver, this is noted and I appreciate you still consider to conserve the case as you presented in the future
 I will create the wrapper on my side for the moment, thanks for the tip. I am also available to help on any PR/changes you think is relevant regarding this behavior 😊.

I take the opportunity to detail a little bit more my usage : on my case I don't like to have to maintain the type of an action + its variable name. All my actions' names are equal to their type which reduces the number of name/case to find, ensure and maintain (it is especially useful when refactoring). Because of that I never have spaces in my actions' names so the casing transformation you suggested would be completely fine on my side 🙂.

TJMaria:
I agree with @hadrien-toma. As a developer i find it more important to easily be able to:
- find usages of an action (dispatch, reducer, effect) --> event/action key is the same everywhere
- easily refactor an actionName

A developer/team should be able to determine the casing used.
Only the actionType is 'negatively' affected, but honestly even there i prefer the type name without spaces.

I created a PR with a different approach (before i read these comments): https://github.com/ngrx/platform/pull/3736

As stated in the comment of my PR:

> Prefered would be to make it configurable, but the current ActionGroup type contains the ActionName type that influence the casing.

TJMaria:
For now i'm using a static class with utility functions as an alternative to createActionGroup
see https://github.com/ngrx/platform/pull/3736#issuecomment-1369785798

PR #3722, opened 2022-12-21
feat(entity): disallow adding ids and entities to additionalState

DaSchTour:
## PR Checklist

- [] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

Typescript does allow adding an object with ids and entities to the getInitialState method

Closes #

## What is the new behavior?

The new signature disallows adding ids or entities props to the getInitialState

Not sure, probably because if used wrong the compiler will fail

```
[ ] Yes
[ ] No
```

### DIFF
--- a/modules/entity/src/entity_state.ts
+++ b/modules/entity/src/entity_state.ts
@@ -9,7 +9,7 @@ export function getInitialEntityState<V>(): EntityState<V> {
 
 export function createInitialStateFactory<V>() {
   function getInitialState(): EntityState<V>;
-  function getInitialState<S extends object>(
+  function getInitialState<S extends object & { ids: never; entities: never }>(
     additionalState: S
   ): EntityState<V> & S;
   function getInitialState(additionalState: any = {}): any {


modules/entity/src/entity_state.ts
```diff
@@ -9,7 +9,7 @@ export function getInitialEntityState<V>(): EntityState<V> {
 
 export function createInitialStateFactory<V>() {
   function getInitialState(): EntityState<V>;
-  function getInitialState<S extends object>(
+  function getInitialState<S extends object & { ids: never; entities: never }>(
```
timdeschryver, line 12: Why should we want to disallow this?

DaSchTour: Because devs might have wrong exceptions in setting this. Although even if ids or entities are set and overwrite `getInitialEntityState()` the entity state works fine. The behavior is very strange. And I have seen some devs adding
```
{
ids: [],
entities: []
}
```

as additional entity state.
It works. But I'm not sure what is the use of having this?
I would expect that `getInitialState` should only take additionalState and not the whole state for an entity state.

DaSchTour: Maybe simply because you can create an invalid entity state: https://stackblitz.com/edit/angular-ivy-yspgnt?file=src/app/state.ts

markostanimirovic: Hi @DaSchTour,

Indeed, the current implementation gives the ability to pass an invalid initial entity state. However, it gives the flexibility to change the default initial entity state if needed. The `@ngrx/entity` package does not always need to be used with asynchronous resources and an empty initial entity state.

Feel free to open an issue, so we can consider it for the next major version. This PR cannot be merged in v15, because it would introduce a breaking change.

brandonroberts:
Closing for a new opened discussion

PR #3736, opened 2023-01-03
feat(store): introduced actionGroupCreator variant that keeps casing

TJMaria:
## PR Checklist

- [x]  The commit message follows our guidelines: https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #
Closes https://github.com/ngrx/platform/issues/3693

## What is the new behavior?
Introduces a new variant of ActionGroup (ActionGroupKeepCasing) where the casing is unchanged.
This allows eventNames to be defined as property keys instead of strings, which is more intuitive.

This also allows usage of actions to be found easier (in: dispatch, effect, reducer)

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/action_group_creator.spec.ts
+++ b/modules/store/spec/action_group_creator.spec.ts
@@ -1,4 +1,5 @@
 import { createActionGroup, emptyProps, props } from '@ngrx/store';
+import { createActionGroupKeepCasing } from '../src/action_group_creator';
 
 describe('createActionGroup', () => {
   const authApiActions = createActionGroup({
@@ -17,6 +18,16 @@ describe('createActionGroup', () => {
     },
   });
 
+  const authorApiActions = createActionGroupKeepCasing({
+    source: 'Author API keep casing',
+    events: {
+      loadAuthors: emptyProps(),
+      loadAuthorsSUCceS: emptyProps(),
+      loadAuthorsFailed: emptyProps(),
+      'reload authors': emptyProps(),
+    },
+  });
+
   it('should create action name by camel casing the event name', () => {
     expect(booksApiActions.loadBooksSuccess).toBeDefined();
   });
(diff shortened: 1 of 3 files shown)


TJMaria:
Prefered would be to make it configurable, but the current ActionGroup type contains the ActionName type that influence the caseing.

It should't be hard to do

brandonroberts:
Thanks @TJMaria. Adding a new API for just this use case is not that simple and not something we will do. There may be other options to explore though

TJMaria:
@brandonroberts I agree. If I have the time i will explore adding an optional configuration and changing the ActionGroup and ActionName type.

It should be possible to expand the API without breaking

TJMaria:
For now i'm using this as an alternative for createActionGroup:
``` 
const actionType = (type: string) => `[DATA] ${ type }`;

function action<T extends string, P extends object>(type: T, propz?: ActionCreatorProps<P> & NotAllowedCheck<P>) {
    return createAction(actionType(type), propz) as ActionCreator<T, (props: P & NotAllowedCheck<P>) => P & TypedAction<T>> ;
}

function actionNoProps<T extends string>(type: T): ActionCreator<T, () => TypedAction<T>>{
    return createAction(actionType(type)) as ActionCreator<T, () => TypedAction<T>>;
}

export class DataActions {
    static getData = actionNoProps('Get data');
    static setData = action('Set data', props<{ data: Data[] }>());
}
```

markostanimirovic:
Thanks for the PR @TJMaria! However, we decided to change the `createActionGroup` function instead of introducing an additional creator function. More info: https://github.com/ngrx/platform/issues/3693#issuecomment-1494510641

PR #3745, opened 2023-01-06
feat(router-store): rename getSelectors to getRouterSelectors

### LINKED ISSUE #3738: @ngrx/router-store: Rename `getSelectors` to `getRouterSelectors`

router-store

### Information

All public APIs from the `@ngrx/router-store` package have precise names (`routerReducer`, `RouterReducerState`, `createRouterSelector`, etc.) except the `getSelectors` function. Therefore we should rename `getSelectors` to `getRouterSelectors`.

This won't be a breaking change - the alias `getSelectors` will be still available, but it will be deprecated.

_No response_

- [X] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3738 

## What is the new behavior?

The `getSelectors` function renamed to `getRouterSelectors`. The `getSelectors` alias is still available but is deprecated.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngrx.io/content/guide/migration/v15.md
+++ b/projects/ngrx.io/content/guide/migration/v15.md
@@ -221,3 +221,26 @@ export class TestComponent {
 }
 ```
 
+## Deprecations
+
+### @ngrx/router-store
+
+#### Renamed `getSelectors` Function
+
+The `getSelectors` function is deprecated in favor of `getRouterSelectors`.
+
+BEFORE:
+
+```ts
+import { getSelectors } from '@ngrx/router-store';
+
+const routerSelectors = getSelectors();
+```
+
+AFTER:
+
+```ts
+import { getRouterSelectors } from '@ngrx/router-store';
+
+const routerSelectors = getRouterSelectors();
+```
--- a/projects/ngrx.io/content/guide/migration/v15.md
+++ b/projects/ngrx.io/content/guide/migration/v15.md
@@ -225,7 +225,7 @@ export class TestComponent {
 
 ### @ngrx/router-store
 
-#### Renamed `getSelectors` Function
+#### Renamed `getSelectors` Function (Introduced in v15.2)
 
 The `getSelectors` function is deprecated in favor of `getRouterSelectors`.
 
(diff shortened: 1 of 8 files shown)


projects/ngrx.io/content/guide/migration/v15.md
```diff
@@ -221,3 +221,26 @@ export class TestComponent {
 }
 ```
 
+## Deprecations
```
timdeschryver, line 224: I don't think we need to add this here.
We usually keep this for migrations between v14 -> v15.
This could also be confusing when peoples are on v15.0.0, in which this method is not available (I don't know why they shouldn't upgrade to the latest v15, but it happens 🤷).

I like that it's already documented though, maybe we should mention the required NgRx version, or already add a v16 file?

brandonroberts: I agree with Tim. Maybe we need a separate guide for deprecations between each major

markostanimirovic: What do you think about adding a note in the title, like this:

```diff
## Deprecations

### @ngrx/router-store

-#### Renamed `getSelectors` Function
+#### Renamed `getSelectors` Function (Introduced in v15.2)

The `getSelectors` function is deprecated in favor of `getRouterSelectors`.
```

timdeschryver: Angular does the following in [upgrade.angular.com](https://update.angular.io/?v=15.0-16.0)
I like the additional note
> Warning: Plans for releases after the current major release are not finalized and may change. These recommendations are based on scheduled deprecations.

They also version their website with the `rc` and `next` tag

image

I don't think we need `rc`, but we could add a `next` migration page with the above note where all the deprecations can be added.

When we release that version, we can transform the deprecations to breaking change.


I think doing this would make our intentions clear.

If that's not good or would cause confusion, I also like Marko's proposal :)

brandonroberts: Currently, our `next.ngrx.io` page points at what's latest on `master`, so we may have some extra work there. I think Marko's note on the deprecations page is a good start

brandonroberts (approved):

timdeschryver (approved):

PR #3791, opened 2023-02-28
feat(schematics): Use createActionGroup in schematics

### LINKED ISSUE #3543: Add ability to generate actions with createActionGroup using schematics
### Information

We are currently be able to generate actions using schematics. 

https://ngrx.io/guide/schematics/action

The generated file contains all actions  like this: 

```ts
export const loadBooks = createAction(
  '[Books] Load Books'
);

export const loadBooksSuccess = createAction(
  '[Books] Load Books Success',
  props<{ books: Book[] }>()
);

export const loadBooksFailure = createAction(
  '[Books] Load Books Failure',
  props<{ error: any }>()
);
```

It will be helpful if the developer have the ability to group them using the new `createActionGroup` function to reach following result directly during file generation. 

```ts
export const BooksActions = createActionGroup({
  source: 'Books',
  events: {
    'Load Books': emptyProps(),
    'Load Books Success': props<{ books: Book[] }>(),
    'Load Books Failure': props<{ error: any }>(),
  }
})

```

## Proposal

`ng generate action ActionName --group` is already use to group actions within files an alternative can be 

- `ng generate action ActionName --compact`  or
- `ng generate action ActionName --action-group` or
- `ng generate action ActionName --minimal` or
...
any other proposal? Does this make sense? 

_No response_

- [X] Yes
- [ ] No

faragos:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

It generates the actions with `createAction`

Closes #3543 

## What is the new behavior?

It generates the actions with `createActionGroup`

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/schematics/src/entity/files/__name@dasherize@if-flat__/__name@dasherize@group-actions__.actions.ts.template
+++ b/modules/schematics/src/entity/files/__name@dasherize@if-flat__/__name@dasherize@group-actions__.actions.ts.template
@@ -1,53 +1,20 @@
-import { createAction, props } from '@ngrx/store';
+import { createActionGroup, emptyProps, props } from '@ngrx/store';
 import { Update } from '@ngrx/entity';
 
 import { <%= classify(name) %> } from '<%= featurePath(group, flat, "models", dasherize(name)) %><%= dasherize(name) %>.model';
 
-export const load<%= classify(name) %>s = createAction(
-  '[<%= classify(name) %>/API] Load <%= classify(name) %>s', 
-  props<{ <%= camelize(name) %>s: <%= classify(name) %>[] }>()
-);
-
-export const add<%= classify(name) %> = createAction(
-  '[<%= classify(name) %>/API] Add <%= classify(name) %>',
-  props<{ <%= camelize(name) %>: <%= classify(name) %> }>()
-);
-
-export const upsert<%= classify(name) %> = createAction(
-  '[<%= classify(name) %>/API] Upsert <%= classify(name) %>',
-  props<{ <%= camelize(name) %>: <%= classify(name) %> }>()
-);
-
-export const add<%= classify(name) %>s = createAction(
-  '[<%= classify(name) %>/API] Add <%= classify(name) %>s',
-  props<{ <%= camelize(name) %>s: <%= classify(name) %>[] }>()
-);
-
-export const upsert<%= classify(name) %>s = createAction(
-  '[<%= classify(name) %>/API] Upsert <%= classify(name) %>s',
-  props<{ <%= camelize(name) %>s: <%= classify(name) %>[] }>()
-);
-
-export const update<%= classify(name) %> = createAction(
-  '[<%= classify(name) %>/API] Update <%= classify(name) %>',
-  props<{ <%= camelize(name) %>: Update<<%= classify(name) %>> }>()
-);
-
-export const update<%= classify(name) %>s = createAction(
-  '[<%= classify(name) %>/API] Update <%= classify(name) %>s',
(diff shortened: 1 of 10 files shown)


markostanimirovic (approved):
🤩

modules/schematics/src/entity/files/__name@dasherize@if-flat__/__name@dasherize@group-actions__.actions.ts.template
```diff
@@ -1,53 +1,20 @@
...
-
-export const delete<%= classify(name) %>s = createAction(
-  '[<%= classify(name) %>/API] Delete <%= classify(name) %>s',
-  props<{ ids: string[] }>()
-);
-
-export const clear<%= classify(name) %>s = createAction(
-  '[<%= classify(name) %>/API] Clear <%= classify(name) %>s'
-);
+export const <%= classify(name) %>Actions = createActionGroup({
```
timdeschryver, line 6: Should action groups be an option to the schematic?
I do think some teams prefer single actions because you can use some refactoring tools with them, which you can't with groups.

timdeschryver (commented):
Thanks for the PR, it's looking good.
I just have one concern about it (see comment) for which I'd like to hear some thoughts.

modules/schematics/src/entity/files/__name@dasherize@if-flat__/__name@dasherize@group-actions__.actions.ts.template
faragos: I see your point. But in my opinion it reducers duplication and implements a namespace for the action. This outweights the keeping of the status-quo. 
I was also pretty surprised that something like `createActionGroup` exist and it isn't the default. Because a lot of people don't know about this feature and the main critisism against ngrx is the boilerplate which this feature reduces.
We could either do a `--legacy`- or a `--verbose`-flag or create a new schematics, but maintaining 2 variants is time-consuming. Maybve you have a way to ask different usergroups? For me so far everybody I showed this liked the new method more than the old one.

timdeschryver: Sorry for my delayed answer.

AFAIK we don't have a way to ask different usergroups.

I do think that having this as a default is good, the only question that remains open is if we should provide a flag which keeps the current implementation.
After giving this some more thought I think we can keep this PR as is, we could revisit this in the feature if it's really needed.

Is this ok for you @brandonroberts and @markostanimirovic ?

timdeschryver (approved):

modules/schematics/src/entity/files/__name@dasherize@if-flat__/__name@dasherize@group-actions__.actions.ts.template
brandonroberts: I think it's reasonable to move forward without it being behind a flag. We can mark it as breaking for v16 so people are aware of it

markostanimirovic: Sounds good to me 👍

timdeschryver:
@faragos could you resolve these conflicts please?

faragos:
> 



> @faragos could you resolve these conflicts please?

Done. Hopefully everything correct 🙂

timdeschryver (approved):
LGTM

brandonroberts (approved):

brandonroberts:
Thanks @faragos!

PR #3819, opened 2023-03-31
docs: use select method instead of state pipe

fabioemoutinho:
accessing the store directly may confuse users because they might get weird behaviors when using observables like this in the state (like infinite loops), I think we should probably discourage examples that access the store directly without the select method to avoid confusion.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

confusing example (accessing state directly) that may lead to wrong implementation if people don't read full docs

Closes #

## What is the new behavior?

use select method to get the state

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngrx.io/content/guide/component-store/initialization.md
+++ b/projects/ngrx.io/content/guide/component-store/initialization.md
@@ -43,9 +43,7 @@ would be to initialize the state lazily by calling [`setState`](guide/component-
   providers: [ComponentStore],
 })
 export class MoviesPageComponent {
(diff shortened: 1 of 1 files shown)


m-thompson-code:
https://github.com/m-thompson-code/rxjs-state-example/commit/b2c3928329302a5a493d7833299c6bdace91eb64

Ran into this confusion because of that code snippet

brandonroberts (approved):
Good idea

PR #3832, opened 2023-04-05
feat(store): preserve the event name case with createActionGroup

### LINKED ISSUE #3693: Allows createActionGroup to preserve the case

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3693

All letters of the event name will be lowercase, except for the initial letters of words starting from the second word, which will be uppercase.

## What is the new behavior?

The initial letter of the first word of the event name will be lowercase, and the initial letters of the other words will be uppercase. The case of other letters in the event name will remain the same.

```
[x] Yes
[ ] No
```

BREAKING CHANGES:

The event name case is preserved when converting to the action name by using the `createActionGroup` function.

BEFORE:

All letters of the event name will be lowercase, except for the initial letters of words starting from the second word, which will be uppercase.

```ts
const authApiActions = createActionGroup({
  source: 'Auth API',
  events: {
    'LogIn Success': emptyProps(),
    'login failure': emptyProps(),
    'Logout Success': emptyProps(),
    logoutFailure: emptyProps(),
  },
});

// generated actions:
const {
  loginSuccess,
  loginFailure,
  logoutSuccess,
  logoutfailure,
} = authApiActions;
```

AFTER:

The initial letter of the first word of the event name will be lowercase, and the initial letters of the other words will be uppercase. The case of other letters in the event name will remain the same.

```ts
const {
  logInSuccess,
  loginFailure,
  logoutSuccess,
  logoutFailure,
} = authApiActions;
```

### DIFF
--- a/modules/store/src/action_group_creator_models.ts
+++ b/modules/store/src/action_group_creator_models.ts
@@ -95,7 +95,7 @@ type EventCreator<
   : never;
 
 export type ActionName<EventName extends string> = Uncapitalize<
-  Join<TitleCase<Lowercase<Trim<EventName>>>>
+  Join<TitleCase<EventName>>
 >;
 
 export interface ActionGroupConfig<
--- a/modules/store/spec/action_group_creator.spec.ts
+++ b/modules/store/spec/action_group_creator.spec.ts
@@ -14,16 +14,21 @@ describe('createActionGroup', () => {
     source: 'Books API',
     events: {
       ' Load BOOKS  suCCess  ': emptyProps(),
+      loadBooksFailure: emptyProps(),
     },
   });
 
   it('should create action name by camel casing the event name', () => {
-    expect(booksApiActions.loadBooksSuccess).toBeDefined();
+    expect(booksApiActions.loadBOOKSSuCCess).toBeDefined();
+    expect(booksApiActions.loadBooksFailure).toBeDefined();
   });
 
   it('should create action type using the "[Source] Event" pattern', () => {
-    expect(booksApiActions.loadBooksSuccess().type).toBe(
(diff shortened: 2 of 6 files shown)


modules/store/src/action_group_creator_models.ts
```diff
@@ -95,7 +95,7 @@ type EventCreator<
   : never;
 
 export type ActionName<EventName extends string> = Uncapitalize<
-  Join<TitleCase<Lowercase<Trim<EventName>>>>
+  Join<TitleCase<EventName>>
```
markostanimirovic, line 98: `Trim` is also removed here because spaces will be removed by `Join` anyway.

markostanimirovic:
> LGTM 👍
> 
> Do you think that we also need to update the docs?
> 
> > An action group uses the event descriptions to create properties within the group that represent the action(s). The property names are auto-generated and are the camelCased version of the event description

This sentence is still valid because we're still camel-casing the event name. 😅 

I plan to improve the "Action Group" page and also add an alternative way of defining event names in a separate PR.

brandonroberts (approved):
🔥

PR #3838, opened 2023-04-11
feat(effects): accept ObservableInput as concatLatestFrom argument

benlesh:
+ removes identity use of `pipe(fn)`.
+ removes use of `of(x).pipe(withLatestFrom(...args)))` in favor of just `forkJoin(args).pipe(map(r => [x, ...r]))` because it's more efficient and does basically the same thing.
+ No longer checks to see if there's an array
+ Changes the typings to use `ObservableInput` rather than `Observable`. The code always supported Promise, Iterable, AsyncIterable, etc, but the TypeScript typings prevented it.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: Fixed the types
```

```
[ ] Yes
[x] No
```

## Other information

This is a refactor with the exception of type changes. Sort of a drive-by on my part because someone pointed me to this operator during an interview. If y'all want to add type changes or have someone take this PR over and run with it, that's fine.


modules/effects/src/concat_latest_from.ts
```diff
@@ -1,5 +1,4 @@
-import { Observable, ObservedValueOf, of, OperatorFunction, pipe } from 'rxjs';
-import { concatMap, withLatestFrom } from 'rxjs/operators';
```
brandonroberts, line 2: We're still using `rxjs/operators` for backwards compatibility with RxJS 6.5.x, so we still need those imported from here also. And the `map` operator needs to be imported for the changes below.

brandonroberts (changes requested):

modules/effects/src/concat_latest_from.ts
benlesh: Ew. But okay.

And sorry about the mistakes. I'm just using the GitHub IDE.

brandonroberts: No problem. Yea, we'd like to move off support for 6.5.x also but the existing apps themselves need to be upgraded. New apps are on RxJS 7.x

modules/effects/src/concat_latest_from.ts
```diff
@@ -38,24 +38,19 @@ export function concatLatestFrom<T extends Observable<unknown>, V>(
...
-      const observablesAsArray = Array.isArray(observables)
-        ? observables
-        : [observables];
-      return of(value).pipe(
-        withLatestFrom(...observablesAsArray)
-      ) as unknown as Observable<R>;
-    })
+  return concatMap(
+    (value) =>
+      forkJoin(observablesFactory(value)).pipe(
```
markostanimirovic, line 52: I'm not sure if `forkJoin` is the right fit here because it waits for all observables to complete.

The `concatLatestFrom` operator is usually used with state observables that never complete:

```ts
concatLatestFrom(() =>
  this.store.select(selectFoo) // this observable never completes
)
```

benlesh:
Looking at this, I just reverted to remove the superfluous `pipe` call. It's mostly fine, but it feels like this could be more efficient.

Sorry for the churn BTW, the GitHub Web API isn't amazing for iterating on development, and I didn't feel like setting things up locally.

Basically, I'd really need to understand the use-cases here, because it seems expensive to have to subscribe to N observables for each dispatched action. What I suspect is that most of the use cases this is serving would be better served by a more specialized set of code, but I don't have context.

In any case, that additional `pipe` isn't needed.

brandonroberts:
There's context related to why we introduced `concatLatestFrom` here https://github.com/ngrx/platform/issues/2222.

On the development side of things, we also have built-in support for using GitHub codespaces in this repo, so you can still edit on the web with a better interface.

https://github.com/ngrx/platform/blob/master/CONTRIBUTING.md

brandonroberts (approved):

markostanimirovic (approved):

PR #3856, opened 2023-04-23
feat(store): add selectSignal method for interop with Angular Signals

brandonroberts:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

## What is the new behavior?

Introduces the `Store.selectSignal()` method for returning a signal of the provided selector function

- It accepts a selector function and returns a Signal of its value
- It throws an error when used outside the injection context 
  - Underlying issue here is the use of the `select` operator that uses the `lift` on the `Observable` to create a new instance of the `Store`. This is done outside the injection context, causing the error.
- Docs incoming also

```
[x] Yes
[ ] No
```

## Other information

Converted almost entire NgRx example-app to use `Store.selectSignal()` for exploring, but will revert most of it.


projects/example-app/src/app/core/containers/app.component.ts
```diff
@@ -12,10 +11,10 @@ import { LayoutActions } from '@example-app/core/actions';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <bc-layout>
-      <bc-sidenav [open]="(showSidenav$ | async)!" (closeMenu)="closeSidenav()">
+      <bc-sidenav [open]="showSidenav()!" (closeMenu)="closeSidenav()">
```
timdeschryver, line 14: We shouldn't need these anymore:

```suggestion
      <bc-sidenav [open]="showSidenav()" (closeMenu)="closeSidenav()">
```

modules/store/src/store.ts
```diff
@@ -93,6 +94,17 @@ export class Store<T = object>
     return (select as any).call(null, pathOrMapFn, ...paths)(this);
   }
 
+  /**
+   * Returns a signal of the provided selector.
+   * This method must be run within in injection context.
+   *
+   * @param selectorFn selector function
+   */
+  selectSignal<K>(selectorFn: (state: T) => K): Signal<K>;
```
timdeschryver, line 103: 🚀

modules/store/src/store.ts
```diff
@@ -93,6 +94,17 @@ export class Store<T = object>
...
+  selectSignal<K = unknown>(selectorFn: (state: T) => K): Signal<unknown> {
```
markostanimirovic, line 104: it seems that an overload signature is not needed here:

```suggestion
  selectSignal<K>(selectorFn: (state: T) => K): Signal<K> {
```

modules/store/src/store.ts
```diff
@@ -93,6 +94,17 @@ export class Store<T = object>
...
+    return toSignal(this.select(selectorFn), { requireSync: true });
+  }
```
markostanimirovic, line 106: What do you think about the following implementation:

```diff
@Injectable()
export class Store<T = object>
  extends Observable<T>
  implements Observer<Action>
{
+  private readonly state: Signal<T>;

  constructor(
    state$: StateObservable,
    private actionsObserver: ActionsSubject,
    private reducerManager: ReducerManager,
+    private injector: Injector
  ) {
    super();

    this.source = state$;
+    this.state = runInInjectionContext(injector, () =>
+      toSignal(state$ as Observable<T>, { requireSync: true })
+    );
  }

+  selectSignal<R>(selector: (state: T) => R): Signal<R> {
+    return computed(() => selector(this.state()));
+  }
}
```

It would not require the `selectSignal` method to be called within the injection context.

brandonroberts: Yea, I tried this option also.  It is better that it doesn't need to be called within the injection context.

It makes it a breaking change because of the added dependency on the `Injector` and breaks some tests where we are creating a `new Store()` manually. We also have to update the `Store.lift` method.

This may be the same pattern we end up using with ComponentStore also

timdeschryver (approved):
LGTM! 🚀

modules/store/src/to_signal.ts
```diff
@@ -0,0 +1,28 @@
...
+export function toSignal<T, U>(source: StateObservable): Signal<T | U> {
+  const state = signal<T | U>(undefined as T);
+
+  source.subscribe({
+    next: (value) => state.set(value),
+    error: (error) => state.set(error),
+  });
+
+  return state.asReadonly();
+}
```
markostanimirovic, line 28: We can remove the second generic here:

```suggestion
export function toSignal<T>(state$: StateObservable): Signal<T> {
  const state = signal<T>(undefined as T);
  state$.subscribe((value) => state.set(value));
  
  return state.asReadonly();
}
```

If error handling is needed for the `StateObservable`, I think it's better to throw an error instead of setting it as a signal value:

```ts
export function toSignal<T>(state$: StateObservable): Signal<T> {
  const state = signal<State<T>>({ kind: StateKind.NoValue });
  
  state$.subscribe({
    next: (value) => state.set({ kind: StateKind.Value, value }),
    error: (error) => state.set({ kind: StateKind.Error, error }),
  });

  return computed(() => {
    const currentState = state();

    switch (currentState.kind) {
      case StateKind.Value:
        return currentState.value;
      case StateKind.Error:
        throw currentState.error;
      case StateKind.NoValue:
        throw new Error('The state observable must emit the initial value synchronously');
    }
  });
}

enum StateKind {
  NoValue,
  Value,
  Error,
}

type NoValueState = { kind: StateKind.NoValue };
type ValueState<T> = { kind: StateKind.Value; value: T };
type ErrorState = { kind: StateKind.Error; error: unknown };

type State<T> = NoValueState | ValueState<T> | ErrorState;
```

projects/example-app/src/app/books/containers/collection-page.component.ts
```diff
@@ -34,11 +32,9 @@ import * as fromBooks from '@example-app/books/reducers';
   ],
 })
 export class CollectionPageComponent implements OnInit {
-  books$: Observable<Book[]>;
+  books = this.store.selectSignal(fromBooks.selectBookCollection);
```
markostanimirovic, line 35: I'm fine with `example-app` changes, but it's maybe better to keep the observable implementation at least until signals become stable.

markostanimirovic (commented):
It looks much better now! 🤘

projects/example-app/src/app/books/containers/collection-page.component.ts
brandonroberts: Agreed. It was mainly for testing in a few areas

modules/store/src/to_signal.ts
brandonroberts: Good call

markostanimirovic (approved):
🚀

PR #3861, opened 2023-04-26
feat(component-store): add selectSignal method for interop with Angular Signals

brandonroberts:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

## What is the new behavior?

`ComponentStore.selectSignal` returns a `Signal` of the provided selector function
`ComponentStore.effect` can receive a `Signal` as an input argument and will push a new value to the observable on each value produced by the Signal

```
[ ] Yes
[x] No
```

This allows existing usage and testing of ComponentStore without requiring an injection context and/or TestBed.

**NOTE** ComponentStore.effect changes reverted, leaving notes for reference.

In order to use `ComponentStore.effect` with a `Signal`, you must use the `provideComponentStore(ComponentStore)` function, as an Angular `effect` is tied to change detection, and tied to an injection context. 

The `ComponentStore.effect` uses the `toObservable` function internally from the `@angular/core/rxjs-interop` package to read a Signal. `toObservable` needs an `Injector` to be provided, or it uses `inject()` internally which throws an error outside an injection context, which would happen when the ComponentStore effect function is called.

Another option would be to pass the `injector` to the `ComponentStore.effect` with the signal which is less than ideal. Open to suggestions here.

## Other information

- Early draft for early review
- Docs incoming
~~- Needs more tests for a `ComponentStore.effect` that receives a `Signal`~~

### DIFF
--- a/modules/component-store/src/component-store.ts
+++ b/modules/component-store/src/component-store.ts
@@ -70,7 +70,7 @@ export class ComponentStore<T extends object> implements OnDestroy {
   private ɵhasProvider = false;
 
   // Signal of state$
-  private readonly state: Signal<T>;
+  readonly state: Signal<T>;
 
   constructor(@Optional() @Inject(INITIAL_STATE_TOKEN) defaultState?: T) {
     // State can be initialized either through constructor or setState.
@@ -79,7 +79,7 @@ export class ComponentStore<T extends object> implements OnDestroy {
     }
 
     this.checkProviderForHooks();
-    this.state = toSignal(this.stateSubject$, this.destroy$);
+    this.state = toSignal(this.stateSubject$.pipe(takeUntil(this.destroy$)));
   }
 
   /** Completes all relevant Observable streams. */
--- a/modules/component-store/src/component-store.ts
+++ b/modules/component-store/src/component-store.ts
@@ -32,8 +32,14 @@ import {
   InjectionToken,
   Inject,
   isDevMode,
+  Signal,
+  computed,
+  isSignal,
+  Injector,
 } from '@angular/core';
+import { toObservable } from '@angular/core/rxjs-interop';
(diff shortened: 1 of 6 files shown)


modules/component-store/src/component-store.ts
```diff
@@ -66,13 +69,17 @@ export class ComponentStore<T extends object> implements OnDestroy {
   readonly state$: Observable<T> = this.select((s) => s);
   private ɵhasProvider = false;
 
+  // Signal of state$
+  private readonly state: Signal<T>;
```
markostanimirovic, line 73: should we make the `state` signal public? 👀

```suggestion
  readonly state: Signal<T>;
```

so consumers can use `computed` instead of `selectSignal` if that's a preference:

```ts
@Injectable()
class CounterStore extends ComponentStore<{ count: number }> {
  readonly count = computed(() => this.state().count);
  readonly doubleCount = computed(() => this.count() * 2);
}
```

brandonroberts: Good catch, yes this should be public

modules/component-store/src/component-store.ts
```diff
@@ -66,13 +69,17 @@ export class ComponentStore<T extends object> implements OnDestroy {
...
+
   constructor(@Optional() @Inject(INITIAL_STATE_TOKEN) defaultState?: T) {
     // State can be initialized either through constructor or setState.
     if (defaultState) {
       this.initState(defaultState);
     }
 
     this.checkProviderForHooks();
+    this.state = toSignal(this.stateSubject$, this.destroy$);
```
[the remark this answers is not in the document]
brandonroberts: Sounds good to me, then we can potentially swap it out

markostanimirovic (approved):
LGTM 🚀

It would be also good to add tests that verify `selectSignal` behavior.

PR #3883, opened 2023-05-08
fix(store): add Signal equal function for immutable object comparison

### LINKED ISSUE #3882: selectSignal() fires on every global state change

store

All selectors created with `selectSignal()` fire on every global state change, even if the related state received no update.

```
store = inject(Store);

constructor() {
   const myState = this.store.selectSignal(mySelector);
   effect(() => console.log(myState());
   setTimeout(() => {
      this.store.dispatch(
         changeOtherStateAction({ info: 'globale state changed' })
      );
   },  5_000);
}
```

**Why does this happen?**

The Angular team decided to emit a new Signal value for each object set, even if the reference stays the same.
This is an issue for working with immuable data (the default for NgRx state management).

The behavior can be fixes manually by configuring an equal function:

```
this.store.selectSignal(
   mySelector,
   { equal: (a, b) => a === b }
);
```

**Necessary change**

This equal compare behavior should be set automatically for `selectSignal()` as the current behavior is not expected by the developer that works with an immutable state pattern.

NgRx: 16 RC.0
Angular 16.0.0

### Other information

May affect `@ngrx/component-store` as well.

Credits to @manfredsteyer who found this issue.

- [X] Yes
- [ ] No

mikezks:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

All selectors created with `store.selectSignal()` fire on every global state change, even if the related state received no update.

Closes #3882 

## What is the new behavior?

If `selector(this.state())` returns the same object reference the computed Signal does not trigger a value update anymore.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store/spec/integration_signals.spec.ts
+++ b/modules/store/spec/integration_signals.spec.ts
@@ -12,6 +12,7 @@ import {
   VisibilityFilters,
   resetId,
 } from './fixtures/todos';
+import { computed } from '@angular/core';
 
 interface Todo {
   id: number;
@@ -128,6 +129,32 @@ describe('NgRx and Signals Integration spec', () => {
       }
 
       expect(error).toBeUndefined();
+    });
+  });
+
+  describe('immutable state integration spec', () => {
+    it('Store.selectSignal should not trigger on unrelated global state changes', () => {
+      let todosTriggerCount = 0;
+
+      const todos = TestBed.runInInjectionContext(() =>
+        store.selectSignal((state) => state.todos)
+      );
+
+      const todosTriggerState = TestBed.runInInjectionContext(() =>
+        computed(() => {
+          todos();
+          return ++todosTriggerCount;
+        })
+      );
+
+      store.dispatch({ type: ADD_TODO, payload: { text: 'first todo' } });
+      expect(todosTriggerState()).toBe(1);
+
+      store.dispatch({
+        type: SET_VISIBILITY_FILTER,
+        payload: VisibilityFilters.SHOW_ACTIVE,
+      });
+      expect(todosTriggerState()).toBe(1);
     });
   });
 });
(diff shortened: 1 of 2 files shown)


modules/store/src/store.ts
```diff
@@ -112,7 +112,9 @@ export class Store<T = object>
     selector: (state: T) => K,
     options?: SelectSignalOptions<K>
   ): Signal<K> {
-    return computed(() => selector(this.state()), { equal: options?.equal });
+    return computed(() => selector(this.state()), {
+      equal: options?.equal || ((previous, current) => previous === current),
+    });
```
markostanimirovic, line 117: LGTM 👍 

Can you add a test to the `integration_signals.spec.ts` to verify this fix?

modules/store/spec/integration_signals.spec.ts
```diff
@@ -130,4 +131,30 @@ describe('NgRx and Signals Integration spec', () => {
...
+      const todos = TestBed.runInInjectionContext(() =>
+        store.selectSignal((state) => state.todos)
+      );
+
+      const todosTriggerState = TestBed.runInInjectionContext(() =>
+        computed(() => {
+          todos();
+          return ++todosTriggerCount;
+        })
+      );
```
markostanimirovic, line 148: we should be able to use `selectSignal` and `computed` out of injection context:

```suggestion
      const todos = store.selectSignal((state) => state.todos);

      const todosTriggerState = computed(() => {
        todos();
        return ++todosTriggerCount;
      });
```

markostanimirovic (approved):

mikezks:
If fix is OK:

- [ ] Review `@ngrx/component-store` `selectSignal()` as well.

brandonroberts (approved):

brandonroberts:
Thanks @mikezks!

PR #3936, opened 2023-06-29
fix(eslint): fix inject function based injection not detecting store (#3834)

### LINKED ISSUE #3834: ESLint plugin does not work correctly when the `inject` function is used

eslint-plugin

For example, we want the linter to catch issues like subscribing to the store directly in components. It works correctly when the store is injected via a constructor, but fails to catch issues when the `inject` function is used

Here the plugin will correctly catch a listing a issue:

```typescript
@Components({
  // metadata
})
export class MyComponent implements OnInit {
  constructor(private store: Store) {}

  ngOnInit() {
    this.store.select(selectSomething).subscribe(() => {
      // do something
    });
  }
}
```

But will fail to do so in the following case:

```typescript
@Components({
  // metadata
})
export class MyComponent implements OnInit {
  private store = inject(Store);

  ngOnInit() {
    this.store.select(selectSomething).subscribe(() => {
      // do something
    });
  }
}
```

It is possible the issue exists in the case of other plugin rules, but I have managed to find this one to fail 100% of the time.

NgRx: 15.4.0
Angular: 15.1.0
Node: 18.10.0
Browsers: N/A
Operating systems: Windows 11

### Other information

_No response_

- [ ] Yes
- [ ] No

Flusinerd:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, the eslint plugin does not detect the store, if it is injected via the new ``inject`` method

Closes #3834

## What is the new behavior?

Updated store detection logic to also check for stores injected via ``inject`` method from the ``@angular/core`` package

```
[ ] Yes
[x] No
```

## Other information
I only updated a single spec file for now, so check if the implementation approach is okay. I'll add more test files later.
Also maybe other rules are affected as well, since for example ``Actions`` could also be injected using ``inject``

### DIFF
--- a/modules/eslint-plugin/spec/rules/no-store-subscription.spec.ts
+++ b/modules/eslint-plugin/spec/rules/no-store-subscription.spec.ts
@@ -48,19 +48,9 @@ class Ok3 {
   constructor(private store: Store) {}
 }`,
   `
-class Ok4 {
-  readonly items$: Observable<readonly Item[]>
-  readonly metrics$: Observable<Metric>
-
-  constructor(store: Store) {
-    this.items$ = store.pipe(select(selectItems))
-    this.metrics$ = store.select(selectMetrics)
-  }
-}`,
-  `
 import { Store } from '@ngrx/store'
 
-class Ok5 {
+class Ok4 {
   readonly items$: Observable<readonly Item[]>
   readonly metrics$: Observable<Metric>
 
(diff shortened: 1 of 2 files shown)


timdeschryver (approved):
Nice! 🤩
The fact that other tests are still passing is already something.
How do you want to go forward with this, because as you mentioned more tests/implementations would be needed.

I propose to merge this change in and create another issue to go over each rule and update the tests and/or the implementation.

Flusinerd:
Hi, whatever works for you. Splitting it would add the benefit of multiple people being able to work on it.

brandonroberts (approved):

PR #3965, opened 2023-07-22
fix(store-devtools): resolve memory leak when devtools are destroyed

arturovt:
This commit updates the implementation of the `StoreDevtools` and adds the `ngOnDestroy` hook, which ensures the teardown of internal subscriptions. Previously, this caused an excessive memory leak because subscription callbacks created closures that captured `this`, preventing `this` from being garbage collected.

### DIFF
--- a/modules/store-devtools/src/devtools.ts
+++ b/modules/store-devtools/src/devtools.ts
@@ -1,4 +1,4 @@
-import { Injectable, Inject, ErrorHandler } from '@angular/core';
+import { Injectable, Inject, ErrorHandler, OnDestroy } from '@angular/core';
 import { toSignal } from '@angular/core/rxjs-interop';
 import {
   Action,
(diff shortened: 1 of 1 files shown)


arturovt:
The failing CIrcle is unrelated since it fails to install the Chrome driver.

brandonroberts:
Thanks @arturovt. I pushed up a fix for CI to run the browser tests correctly.

modules/store-devtools/src/devtools.ts
```diff
@@ -121,13 +133,15 @@ export class StoreDevtools implements Observer<any> {
       value: toSignal(state$, { manualCleanup: true, requireSync: true }),
     });
 
-    this.extensionStartSubscription = extensionStartSubscription;
-    this.stateSubscription = liftedStateSubscription;
```
brandonroberts, line 125: For reference, why not just unsubscribe from these in the `ngOnDestroy` instead of adding a destroy Subject?

arturovt: Well, my VSCode initially marked them as unused, so I removed them first. Then I added `takeUntil` since I saw it's already being used in other places too.

arturovt: I can revert properties back and unsubscribe.

brandonroberts: I think that approach would be better in this case because the subscription references are already there.

arturovt: Updated

brandonroberts (approved):

markostanimirovic (approved):

PR #3971, opened 2023-07-24
feat(component-store): tree-shake hooks check from production bundle

arturovt:
This commit allows for tree-shaking `checkProviderForHooks` from production bundles since its expression is wrapped with `isDevMode`, which is supposed to be called only in development mode. This will reduce the number of change detection cycles in production for existing component stores due to the use of the `asapScheduler`, which schedules a macrotask within the Angular zone. This leads to redundant change detections in production because the callback acts as a "noop" (as `isDevMode` returns false).

### DIFF
--- a/modules/component-store/src/component-store.ts
+++ b/modules/component-store/src/component-store.ts
@@ -40,6 +40,12 @@ import {
 import { isOnStateInitDefined, isOnStoreInitDefined } from './lifecycle_hooks';
 import { toSignal } from '@angular/core/rxjs-interop';
 
+// This is a global variable provided by Angular in development mode
+// (when running `serve`, for example) or by Terser (through its global definitions)
+// when running build with optimizer enabled. This allows us to tree-shake the code
+// from production bundles, ensuring that it is only called during development when needed.
+declare const ngDevMode: boolean;
+
 export interface SelectConfig<T = unknown> {
   debounce?: boolean;
   equal?: ValueEqualityFn<T>;
@@ -97,7 +103,9 @@ export class ComponentStore<T extends object> implements OnDestroy {
       this.initState(defaultState);
     }
 
-    this.checkProviderForHooks();
+    if (typeof ngDevMode === 'undefined' || ngDevMode) {
+      checkProviderForHooks(this);
+    }
   }
 
   /** Completes all relevant Observable streams. */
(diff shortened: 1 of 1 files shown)


modules/component-store/src/component-store.ts
```diff
@@ -40,6 +40,12 @@ import {
 import { isOnStateInitDefined, isOnStoreInitDefined } from './lifecycle_hooks';
 import { toSignal } from '@angular/core/rxjs-interop';
 
+// This is a global variable provided by Angular in development mode
+// (when running `serve`, for example) or by Terser (through its global definitions)
+// when running build with optimizer enabled. This allows us to tree-shake the code
+// from production bundles, ensuring that it is only called during development when needed.
+declare const ngDevMode: boolean;
```
markostanimirovic, line 47: Thanks for this PR @arturovt! We have already considered using `ngDevMode`. However, we're not using it because it's an internal API that can be changed in any patch/minor version.

arturovt: Well, that's true that they may find reasons to remove it. However, let's consider the case where the variable is removed or renamed. The code would still maintain its backward compatibility due to the use of `isDevMode`. Since we're accessing it safely (with the `typeof` operator), the behavior wouldn't change. If the Angular team decides to remove the variable from their code, it would be a simple matter of removing it from our codebase as well.

markostanimirovic: Have you tested this PR?

Tree shaking is not working because of the `|| isDevMode()` condition.

You can test it by executing: `yarn nx build --project=standalone-app`

arturovt: Yeah, I tested changes in another project, just build the `component-store` and linked it to `@ngrx/component-store` in another folder.

markostanimirovic:
https://github.com/angular/angular/issues/51175

modules/component-store/src/component-store.ts
```diff
@@ -97,7 +103,14 @@ export class ComponentStore<T extends object> implements OnDestroy {
       this.initState(defaultState);
     }
 
-    this.checkProviderForHooks();
+    // Note that we have still kept the `isDevMode()` condition for cases where
+    // `ngDevMode` would be undefined. Under normal conditions, `ngDevMode` would
+    // always be defined, and it would be evaluated during the compilation into
+    // `if (false)` during the production build, allowing the below expression to
+    // be tree-shaken.
+    if ((typeof ngDevMode === 'undefined' && isDevMode()) || ngDevMode) {
```
JeanMeche, line 111: I'm not sure this makes sense. since `isDevMode()` is `ngDevMode === 'undefined' || !!ngDevMode;`

arturovt: I haven't seen that `isDevMode` implementation was changed to using `ngDevMode`, that's why I used this conditional expression. @markostanimirovic can confirm that using `isDevMode()` is not being stripped out, see comments above.

modules/component-store/src/component-store.ts
JeanMeche: Following what @arturovt wrote on angular/issues#51175 I'd be inclinded to think that this is the best solution right now. 

* `ngDevMode` is used - everywhere - in the angular code base, it's not going to change anytime soon. It isn't a standard constant as it's set by the CLI and only declared as a global.

brandonroberts: I agree, but the other side of this is that if we use this, and the bundle size changes significantly, and then it gets removed, we have no way of allowing it to be tree-shaken again unless something else is introduced. Is this bit of code that much to need this change?

No one else AFAIK, even library authors outside of the framework is using it.

brandonroberts:
Closing this one as we still don't have an official API for tree-shaking or consensus around using this.

PR #3984, opened 2023-08-02
fix(effects): register functional effects from object without prototype

### LINKED ISSUE #3972: "Cannot read properties of undefined (reading 'name') at isClassInstance" when importing effects from a library

effects

1. Import into your app a library that exports a provider with effects;
2. Run your app.
3. ERROR TypeError: Cannot read properties of undefined (reading 'name')
    at isClassInstance (ngrx-effects.mjs:135:29)

```diff
// node_modules/@ngrx/effects/fesm2022/ngrx-effects.mjs
function isClassInstance(obj) {
-     return (obj.constructor.name !== 'Object' && obj.constructor.name !== 'Function');
+     return (obj.constructor && obj.constructor.name !== 'Object' && obj.constructor.name !== 'Function');
}

function mergeEffects(sourceInstance, globalErrorHandler, effectsErrorHandler) {
-    const sourceName = getSourceForInstance(sourceInstance).constructor.name;
+    const sourceName = getSourceForInstance(sourceInstance)
+      ? getSourceForInstance(sourceInstance).constructor.name
+      : null;
```

NgRx: ^16.1.0
Angular: ^16.1.7
Node: 18.17.0

### Other information

_No response_

- [X] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3972

## What is the new behavior?

It's possible to import functional effects that grouped via named imports from buildable libraries.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/spec/effect_sources.spec.ts
+++ b/modules/effects/spec/effect_sources.spec.ts
@@ -208,8 +208,9 @@ describe('EffectSources', () => {
       const recordB = {
         b: createEffect(() => alwaysOf(b), { functional: true }),
       };
-      // named imports from built packages that contain functional effects
-      // don't have a prototype
+      // a record with functional effects that is defined as
+      // a named import in a built package doesn't have a prototype
+      // for more info see: https://github.com/ngrx/platform/issues/3972
       const recordC = Object.freeze({
         __proto__: null,
         c: createEffect(() => alwaysOf(c), { functional: true }),
--- a/modules/effects/spec/effect_sources.spec.ts
+++ b/modules/effects/spec/effect_sources.spec.ts
@@ -208,9 +208,15 @@ describe('EffectSources', () => {
       const recordB = {
         b: createEffect(() => alwaysOf(b), { functional: true }),
       };
+      // named imports from built packages that contain functional effects
+      // don't have a prototype
       const recordC = Object.freeze({
         __proto__: null,
         c: createEffect(() => alwaysOf(c), { functional: true }),
+      });
+      const recordD = Object.freeze({
+        __proto__: null,
+        d: createEffect(() => alwaysOf(d as any), { functional: true }),
       });
 
       it('should resolve effects from class instances', () => {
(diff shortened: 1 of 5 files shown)


modules/effects/spec/effect_sources.spec.ts
```diff
@@ -207,6 +208,10 @@ describe('EffectSources', () => {
       const recordB = {
         b: createEffect(() => alwaysOf(b), { functional: true }),
       };
+      const recordC = Object.freeze({
```
timdeschryver, line 211: I think it's useful to add a comment here on why this record/test exists.

modules/effects/src/effects_resolver.ts
```diff
@@ -13,7 +13,8 @@ export function mergeEffects(
   globalErrorHandler: ErrorHandler,
   effectsErrorHandler: EffectsErrorHandler
 ): Observable<EffectNotification> {
-  const sourceName = getSourceForInstance(sourceInstance).constructor.name;
+  const source = getSourceForInstance(sourceInstance);
+  const sourceName = source ? source.constructor.name : 'Object';
```
timdeschryver, line 17: Unrelated, but seeing this made me wonder how `sourceName` is used throughout the whole flow.
I ended up that it's only used to throw an error when an effect doesn't dispatch an action (an object with `type` as property).
Do we want to change the message for functional effects?
It currently throws the following message:

```js
ERROR Error: Effect "Object.logDispatchedActions()" dispatched an invalid action: {}
``` 

We can prettify this a little bit to

```js
//           👇 does this add value?
//                                👇 remove `Object.`
ERROR Error: (Functional) Effect "logDispatchedActions()" dispatched an invalid action: {}
```

brandonroberts (approved):

modules/effects/src/effects_resolver.ts
markostanimirovic: Good point!

timdeschryver (approved):
👍

PR #3988, opened 2023-08-05
fix(data): make DataServiceError extend from Error

rob4226:
I was kind of surprised when `DataServiceError instanceof Error` returned `false` but then I saw the note in the source code (4 years ago) about how `dse instanceof DataServiceError` didn't work in unit tests when extending from `Error`.  As long as `super()` is called in the child class constructor when extending from `Error`, then `instanceof` should work fine.  I made the change, then ran the unit tests and they all passed so it seems it's ok.

This is not a big deal but it's usually good form to have `Error` in the inheritance chain of custom errors. I don't believe this is a breaking change.


timdeschryver (approved):
Weird, I wonder why this failed before.
It seems to work now 👍

brandonroberts (approved):

PR #4026, opened 2023-09-01
feat(eslint-plugin): add new avoid-mapping-component-store-selectors rule

### LINKED ISSUE #3940: eslint-plugin(component-store): avoid-mapping-selectors

component-store, eslint-plugin

### Information

I think an NgRx ComponentStore variant of the NgRx Store ESLint rule [avoid-mapping-selectors](https://ngrx.io/guide/eslint-plugin/rules/avoid-mapping-selectors) would be helpful to avoid anti-patterns.

#### Incorrect usage

```typescript
export class UserStore extends ComponentStore<UserState> {
  loggedInUser$ = this.select((state) => state.loggedInUser);

  name$ = this.select(this.loggedInUser$).pipe(map((user) => user.name));
}
```

#### Correct usage

```typescript
export class UserStore extends ComponentStore<UserState> {
  loggedInUser$ = this.select((state) => state.loggedInUser);

  name$ = this.select(this.loggedInUser$, (user) => user.name);
}
```

Currently, I'm maintaining internal developer guides for NgRx ComponentStore usage.

- [ ] Yes
- [X] No

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #3940

```
[ ] Yes
[x] No
```

## Other information

The first commit (61bc231ccb5ee1f8dee90da2cb201e390343f0c8) moves the spec files into directories so it becomes easier to find.

### DIFF
--- a/modules/eslint-plugin/spec/rules/component-store/avoid-mapping-component-store-selectors.spec.ts
+++ b/modules/eslint-plugin/spec/rules/component-store/avoid-mapping-component-store-selectors.spec.ts
@@ -1,0 +1,140 @@
+import type {
+  ESLintUtils,
+  TSESLint,
+} from '@typescript-eslint/experimental-utils';
+import { fromFixture } from 'eslint-etc';
+import * as path from 'path';
+import rule, {
+  messageId,
+} from '../../../src/rules/component-store/avoid-mapping-component-store-selectors';
+import { ruleTester } from '../../utils';
+
+type MessageIds = ESLintUtils.InferMessageIdsTypeFromRule<typeof rule>;
+type Options = ESLintUtils.InferOptionsTypeFromRule<typeof rule>;
+type RunTests = TSESLint.RunTests<MessageIds, Options>;
+
+const validConstructor: () => RunTests['valid'] = () => [
+  `
+import { ComponentStore } from '@ngrx/component-store'
+
+class Ok extends ComponentStore<MoviesState> {
+  movies$ = this.select((state) => state.movies);
+  firstMovie$ = this.select(this.movies$, (movies) => movies[0]);
+
+  constructor() {
+    super({ movies: [] })
+  }
+}`,
+  `
+import { ComponentStore } from '@ngrx/component-store'
+
+class Ok {
+  readonly movies$ = this.store.select((state) => state.movies);
+  firstMovie$ = this.store.select(this.movies$, (movies) => movies[0]);
+
+  constructor(private readonly store: ComponentStore<MoviesState>) {}
+}`,
+  `
+import { ComponentStore } from '@ngrx/component-store'
+
+class Ok {
+  readonly movies$: Observable<unknown>
+  readonly firstMovie$: Observable<unknown>
+
+  constructor(private customStore: ComponentStore<MoviesState>) {
+    const movies = customStore.select((state) => state.movies);
+    this.firstMovie$ = customStore.select(movies, (movies) => movies[0]);
+    
+    this.movies$ = this.customStore.select((state) => state.movies);
+    this.firstMovie$ = this.customStore.select(this.movies$, (movies) => movies[0]);
+  }
+}`,
+  `
+import { ComponentStore } from '@ngrx/component-store'
+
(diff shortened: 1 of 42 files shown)


modules/eslint-plugin/spec/rules/component-store/avoid-mapping-component-store-selectors.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+  firstMovie$ = this.select(this.movies$, (movies) => movies[0]);
+
+  constructor() {
+    super({ movies: [] })
+  }
+}`,
+  `
+import { ComponentStore } from '@ngrx/component-store'
+
+class Ok {
```
va-stefanek, line 31: I saw in other test cases that we used to add incremental number value to each Ok/NotOk cases. I am not sure whether is somehow required.

modules/eslint-plugin/spec/rules/component-store/avoid-mapping-component-store-selectors.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+import rule, {
+  messageId,
+} from '../../../src/rules/component-store/avoid-mapping-component-store-selectors';
+import { ruleTester } from '../../utils';
+
+type MessageIds = ESLintUtils.InferMessageIdsTypeFromRule<typeof rule>;
+type Options = ESLintUtils.InferOptionsTypeFromRule<typeof rule>;
+type RunTests = TSESLint.RunTests<MessageIds, Options>;
+
+const validConstructor: () => RunTests['valid'] = () => [
```
va-stefanek, line 16: I saw that in same rule for store we had an valid tests case which allows to use map operator on select when it was in scope of an effect, it's name is "OkBecauseOfEffect". Does we should respect mapping in scope of component store effects to? If yes maybe it's worth adding test case for that?

modules/eslint-plugin/spec/rules/component-store/avoid-mapping-component-store-selectors.spec.ts
timdeschryver: It's not required, I didn't add them here because other test cases weren't consistent and it can sometimes take some times to increment the index while adding tests 😅

modules/eslint-plugin/spec/rules/component-store/avoid-mapping-component-store-selectors.spec.ts
timdeschryver: Good question. 
I didn't add this here because the use-case is different. 
Within an effect that uses the global store there are some cases where you'd want to read a different slice of the state for a particular effect.

I don't think this is a good practice to do with component store effects. Because the state is already scoped.
For the rare cases you can use `get` to read the state, https://ngrx.io/guide/component-store/read#get-method in a component store effect.

projects/ngrx.io/content/guide/eslint-plugin/rules/avoid-mapping-component-store-selectors.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```ts
+export class UserStore extends ComponentStore<UserState> {
+  loggedInUser$ = this.select((state) => state.loggedInUser);
+  //                                           ⚠ Avoid mapping logic outside the selector level.
+  name$ = this.select(this.loggedInUser$).pipe(map((user) => user.name));
```
markostanimirovic, line 23: passing observable without a projector to the `select` method will throw a compilation error, so maybe this example would be better:

```suggestion
  name$ = this.select((state) => state.loggedInUser).pipe(map((user) => user.name));
```

timdeschryver: Thanks Marko!

brandonroberts (approved):

markostanimirovic (approved):
🚀

Issue #4033, opened 2023-09-04
@ngrx/signals: Test Coverage


rainerhahnekamp:
### Which @ngrx/* package(s) are relevant/related to the feature request?

store

### Information

Given the high amount of logic, investing in many unit tests is important. A high code coverage would be desirable as it also acts as documentation.

Additionally, the sample projects should be covered via E2E tests. In the best case supported via GitHub actions. I would go with Playwright as the E2E tool.

### I would be willing to submit a PR to fix this issue

- [X] Yes
- [ ] No

markostanimirovic:
I'm going to close this issue because `@ngrx/signals` coverage is already high with unit, integration, and type tests.

We'll create the issue to add e2e tests for the NgRx Signals example app when we create it.

PR #4046, opened 2023-09-17
test(signals): verify that patchState does not emit on empty change

rainerhahnekamp:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Part of #4033

## What is the new behavior?

Test checks that `patchState` with an empty object or  updateFn returning the state untouched, does not cause the underlying signal to emit.

```
[ ] Yes
[x] No
```


modules/signals/spec/patch-state.spec.ts
```diff
@@ -74,4 +76,20 @@ describe('patchState', () => {
...
+      let emitCount = 0;
+      const state = signalState(initialState);
+      effect(() => emitCount++);
+
+      tick();
+      expect(emitCount).toBe(1);
+      patchState(state, {});
+      patchState(state, (state) => state);
+
+      expect(emitCount).toBe(1);
```
markostanimirovic, line 92: if we read the `state` value in the `effect` and invoke `tick` after `patchState`:

```suggestion
      effect(() => {
        state();
        emitCount++;
      });

      tick();
      expect(emitCount).toBe(1);
      patchState(state, {});
      patchState(state, (state) => state);
      tick();
      
      expect(emitCount).toBe(1);
```

this test will fail because the state value reference will be changed. However, nested signals will not emit. Here is the test that verifies this behavior: https://github.com/ngrx/platform/blob/main/modules/signals/spec/signal-state.spec.ts#L54

I suggest moving this test to the `signal-state.spec.ts` and verifying that nested signals don't emit when the signal state is patched with an empty object/updater that doesn't change the state.

rainerhahnekamp:
Sorry for the late reply, but I've spent some time understanding the inner workings of the Signals. Some remarks or requests for clarification:

1. Do you intend to use `patchState` for both `signalState` and `signalStore`? If yes, then I would have put all `patchState`-specific tests into `patch-state.spec.ts`. That was my initial thought.
2. `DeepSignal`: You're referring to that feature as nested Signal, and I also explain it like that. Why not rename it to `NestedSignal`? 
3. `tick`: We use `tick` quite often in combination `fakeAsync`. Since this is something completely different, I would rename `tick` to `triggerEffects / trigger / cd` or something else.
4. Is it the intended behaviour that `patchState(state, value => value)` and `patchState({})` change the object reference? I am usually an advocate for DX but a pre-check inside `patchState` would result in better performance.


I can create a separate issue but I thought it would be faster in this PR.

modules/signals/spec/patch-state.spec.ts
rainerhahnekamp: Done, please see my remarks below.

markostanimirovic:
> 1. Do you intend to use `patchState` for both `signalState` and `signalStore`? If yes, then I would have put all `patchState`-specific tests into `patch-state.spec.ts`. That was my initial thought.

Yes, it's used for both.

> 2. `DeepSignal`: You're referring to that feature as nested Signal, and I also explain it like that. Why not rename it to `NestedSignal`?

We can rename it easily since it's an internal API. However, `deep` seems more precise to me, because `nested` can refer to one level of nesting for example. If we choose `nested`, we would need to call it `DeepNestedSignal` to avoid confusion. Therefore, I think `DeepSignal` is a proper name.

> 3. `tick`: We use `tick` quite often in combination `fakeAsync`. Since this is something completely different, I would rename `tick` to `triggerEffects / trigger / cd` or something else.

It's also an internal API, so it can be easily renamed.

> 4. Is it the intended behaviour that `patchState(state, value => value)` and `patchState({})` change the object reference? I am usually an advocate for DX but a pre-check inside `patchState` would result in better performance.

I mean, we can add a check -> if an empty object is passed or the same state reference, then do not spread the current state, but I'm not sure if that would bring an improvement. Let me know if I'm missing something here. All nested slices are 'memoized', so they'll not emit a change.

markostanimirovic:
Thanks Rainer! There is a conflict in the `signal-state.spec.ts` file.

rainerhahnekamp:
Hi Marko,

I've pushed but there is one thing about `patchState` which I find important.

> 2. `DeepState` / `NestedState`: We can rename it easily since it's an internal API. However, deep seems more precise to me, because nested can refer to one level of nesting for example. If we choose nested, we would need to call it DeepNestedSignal to avoid confusion. Therefore, I think DeepSignal is a proper name.

Internal yes, but not hidden. IDEs show that type. Last workshop I've explained the nested signal, and participants asked what a `DeepSignal` is all about. I've left it as `DeepSignal`, though.

Screenshot 2023-10-19 at 13 57 19

> 3. `tick`: It's also an internal API that can be easily renamed.

I have replaced the complete `testEffects` function with the native TestBed.

> 4. `patchState()` I mean, we can add a check -> if an empty object is passed or the same state reference, then do not spread the current state, but I'm not sure if that would bring an improvement. Let me know if I'm missing something here. All nested slices are 'memoized', so they'll not emit a change.

- `patchState(state, {})`: I see here a use case whenever we deal with `Partial<T>`. Its values could directly come from the user or somewhere else. Either way, it could be an empty object, and it should not trigger the signal.

- `patchState(state, value => value)`: Since the stable version of `Signal::update(value => value)` doesn't trigger, I would suggest we make `patchState` fully compatible.

Let me know, what you think.

markostanimirovic:
Cool, thanks!

`Signal.update` does not support partial updates, so I don't think they're fully comparable. `patchState` is closer to `ComponentStore.patchState` which would also update the state observable if we pass an empty object or the same state reference.

Consuming nested signals will be a common case and patching the SignalState/Store's state signal with an empty object or the same state reference will not affect them:

```ts
const state = signalState({ firstName: 'John', lastName: 'Doe'  });

const Store = signalStore(
  withState({ firstName: 'John', lastName: 'Doe' })
);
const store = new Store();

let executionCount = 0;
effect(() => {
  state.firstName();
  state.lastName();
  store.firstName();
  store.lastName();

  executionCount++;
});

patchState(state, {});
patchState(store, {});
patchState(state, (s) => s);
patchState(store, (s) => s);
```

👆 `executionCount` will be 1.

I mean, we can add the check for an empty object and the same state reference, but would it be worth introducing additional implementation complexity? I'm not sure. 🤔

If we realize that it can be useful, we can consider introducing `setState` helper that would work the same as `Signal.update/set` and `ComponentStore.setState`.

markostanimirovic (approved):

rainerhahnekamp:
> I mean, we can add the check for an empty object and the same state reference, but would it be worth introducing additional implementation complexity? I'm not sure. 🤔

It is your call. I would add it without a doubt.

For me, it is important that you are aware of it. We can also raise a separate issue once the beta version is out and we see a need for it. 👍

rainerhahnekamp:
Thanks @markostanimirovic (and @timdeschryver)!

PR #4049, opened 2023-09-18
feat(signals): add signalStore and signalStoreFeature

### LINKED ISSUE #4000: @ngrx/signals: Add initial `signalStore` implementation
### Information

Add initial `signalStore` and `signalStoreFeature` implementation.

- [x] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4000

## What is the new behavior?

`signalStore` and `signalStoreFeature` functions are added to the `@ngrx/signals` package.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/spec/signal-store-feature.spec.ts
+++ b/modules/signals/spec/signal-store-feature.spec.ts
@@ -1,0 +1,94 @@
+import { Signal, signal } from '@angular/core';
+import {
+  selectSignal,
+  signalStore,
+  signalStoreFeature,
+  type,
+  withMethods,
+  withSignals,
+  withState,
+} from '../src';
+import { STATE_SIGNAL } from '../src/signal-state';
+
+describe('signalStoreFeature', () => {
+  function withCustomFeature1() {
+    return signalStoreFeature(
+      withState({ foo: 'foo' }),
+      withSignals(({ foo }) => ({ bar: selectSignal(() => foo() + 1) })),
+      withMethods(({ foo, bar }) => ({
+        baz: () => foo() + bar() + 2,
+      }))
+    );
+  }
+
+  function withCustomFeature2() {
+    return signalStoreFeature(
+      withCustomFeature1(),
+      withMethods(({ foo, baz }) => ({
+        bar: (value: number) => value,
+        m: () => foo() + baz() + 3,
+      }))
+    );
+  }
+
+  function withCustomFeatureWithInput<_>() {
+    return signalStoreFeature(
+      {
+        state: type<{ foo: string }>(),
+        signals: type<{ s: Signal<number> }>(),
+      },
+      withState({ foo1: 1 }),
+      withState({ foo2: 2 })
+    );
+  }
+
+  it('creates a custom feature by combining base features', () => {
+    const Store = signalStore(
+      withCustomFeature1(),
+      withSignals(({ bar }) => ({
+        s: selectSignal(() => bar() + 's'),
+      }))
+    );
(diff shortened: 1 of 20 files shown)


brandonroberts (approved):

modules/signals/src/signal-store-feature.ts
```diff
@@ -0,0 +1,224 @@
+import {
+  EmptyFeatureResult,
+  MergeFeatureResults,
+  SignalStoreFeature,
+  SignalStoreFeatureResult,
+} from './signal-store-models';
+
+export function signalStoreFeature<F1 extends SignalStoreFeatureResult>(
```
timdeschryver, line 8: Can these overload types be refactored with variadic tuple types?

markostanimirovic: No, it's not possible. Every next feature needs the type of the previous one, similar to [RxJS `pipe`](https://github.com/ReactiveX/rxjs/blob/master/src/internal/util/pipe.ts#L4).

modules/signals/spec/signal-store-feature.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    return signalStoreFeature(
+      withCustomFeature1(),
+      withMethods(({ foo, baz }) => ({
+        bar: (value: number) => value,
+        m: () => foo() + baz() + 3,
+      }))
+    );
+  }
+
+  function withCustomFeatureWithInput<_>() {
```
timdeschryver, line 34: Does this generic has a meaning?

```suggestion
  function withCustomFeatureWithInput() {
```

modules/signals/spec/signal-store-feature.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    return signalStoreFeature(
+      {
+        state: type<{ foo: string }>(),
+        signals: type<{ s: Signal<number> }>(),
```
timdeschryver, line 38: I got a question about the usage of `type`.
If I understand this correctly, the `type` is used here to make the store typesafe? In this case, this feature expects that it's used within a store that has a `foo` state property, and a `s` signal property.

timdeschryver (approved):
I like the different files for each feature 🤩
As mentioned earlier, it seems like my WebStorm fails to infer some types, while VSCode can.

image

modules/signals/spec/signal-store-feature.spec.ts
markostanimirovic: I noticed a weird typing bug when custom features with input are used in complex use cases.

[Here is the reproduction.](https://github.com/markostanimirovic/ngrx-signal-store-playground/blob/main/src/tests/test2.ts#L20) If we remove `<_>` from `withY`, the compilation error will be thrown in SignalStores that use `withY` feature although the generic isn't used anywhere. 👀

This is a rare case, but I'll document this behavior and there will be a recommendation in the docs.

modules/signals/spec/signal-store-feature.spec.ts
markostanimirovic: Exactly!

timdeschryver:
Congrats on closing #4000! 🚀

Maciejlys:
When can we expect release of this? I assume this will be a npm package ngrx/signals?

PR #4118, opened 2023-11-09
feat(signals): add `getState` function

### LINKED ISSUE #4116: [Signal Store] Access the whole state object

signals

### Information

Using `@ngrx/signals` 17.0.0-beta.0

It doesn't look like it's currently possible to access the whole state object from a `signalStore` instance as a single signal.

It would be useful to be able to do so, e.g. in cases where the state object is modelled as a state machine using type unions and a discriminated field.

I've noticed the autocompletion does allow you to do something like:

```ts
import { STATE_SIGNAL } from '@ngrx/signals/src/signal-state';
...
vm = this.#store[STATE_SIGNAL];
```

… which compiles fine, but at runtime there is an error:

```shell
✘ [ERROR] Could not resolve "@ngrx/signals/src/signal-state"

    src/app/login/feature/login-page.component.ts:3:29:
      3 │ import { STATE_SIGNAL } from '@ngrx/signals/src/signal-state';
        ╵                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  The path "./src/signal-state" is not exported by package "@ngrx/signals":

    node_modules/@ngrx/signals/package.json:55:13:
      55 │   "exports": {
         ╵              ^

  You can mark the path "@ngrx/signals/src/signal-state" as external to exclude it from the bundle, which will remove this error and leave the unresolved path in the bundle.
```

Note: it would also be very useful to access this single signal in the `withMethods`, `withHooks` etc. A concrete use case is setting up an `effect` in `onInit` to log state changes in the whole state object.

No simple working alternatives found.

- [X] Yes
- [ ] No

jits:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, it's not possible to obtain the state object as a whole, via a single signal.

Closes #4116

## What is the new behavior?

Using the `getState` function will return the whole state object,  via the state signal under the hood.

```ts
import { getState, signalState } from '@ngrx/signals';

const itemsState = signalState({
  loading: false,
  items: [] as Item[]
});

getState(itemsState);  // returns the state object
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/spec/get-state.spec.ts
+++ b/modules/signals/spec/get-state.spec.ts
@@ -1,0 +1,53 @@
+import { TestBed } from '@angular/core/testing';
+import {
+  getState,
+  patchState,
+  signalState,
+  signalStore,
+  withState,
+} from '../src';
+import { effect } from '@angular/core';
+
+describe('getState', () => {
+  const initialState = {
+    user: {
+      firstName: 'John',
+      lastName: 'Smith',
+    },
+    foo: 'bar',
+    numbers: [1, 2, 3],
+    ngrx: 'signals',
+  };
+
+  describe('with signalStore', () => {
+    it('returns the state object', () => {
+      const Store = signalStore(withState(initialState));
+      const store = new Store();
+
+      expect(getState(store)).toEqual(initialState);
+    });
+
+    it('executes in the reactive context', () => {
+      const Store = signalStore(withState(initialState));
+      const store = new Store();
+
+      let executionCount = 0;
+
+      TestBed.runInInjectionContext(() => {
+        effect(() => {
+          getState(store);
+          executionCount++;
+        });
+      });
+
+      TestBed.flushEffects();
+      expect(executionCount).toBe(1);
+
+      patchState(store, { foo: 'baz' });
+
+      TestBed.flushEffects();
+      expect(executionCount).toBe(2);
+      expect(getState(store)).toEqual({ ...initialState, foo: 'baz' });
+    });
+  });
+});
--- a/modules/signals/src/get-state.ts
+++ b/modules/signals/src/get-state.ts
@@ -1,0 +1,7 @@
+import { STATE_SIGNAL, SignalStateMeta } from './signal-state';
+
+export function getState<State extends Record<string, unknown>>(
+  signalState: SignalStateMeta<State>
+): State {
+  return signalState[STATE_SIGNAL]();
+}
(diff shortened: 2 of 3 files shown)


jits:
Note: I've followed the exact same naming conventions as used in the [`patchState` function](https://github.com/jits/ngrx-platform/blob/2b7906fe54d80af5478a85e6302fb14e198a5b6f/modules/signals/src/patch-state.ts#L7).

jits:
@markostanimirovic — a minor concern: when calling `getState(…)` it's not immediately clear that it's accessing the getter on a signal, under the hood, and thus tying into the publishing/consuming mechanism of signals. Maybe this is not a big deal and could be mitigated by making it clear in docs what's happening, but thought I'd mention it in case there's a better way of, err, signalling (😆) to users that it's using a signal getter under the hood.

modules/signals/spec/get-state.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+  ].forEach(({ name, stateFactory }) => {
+    describe(name, () => {
+      it('returns the state object', () => {
+        const state = stateFactory();
+
+        expect(getState(state)).toEqual(initialState);
+      });
+    });
+  });
```
markostanimirovic, line 34: It should be enough to test only the `signalStore` because `signalState` already returns the signal that contains the full state. Also, it would be great if you could add a test that verifies behavior when `getState` is called within the reactive context. Something like this:

```ts
it('executes in the reactive context', () => {
  const Store = signalStore(withState(initialState));
  const store = new Store();
  const executionCount = 0;

  TestBed.runInInjectionContext(() => {
    effect(() => {
      getState(store);
      executionCount++;
    });
  });

  expect(executionCount).toBe(1);

  patchState(store, { foo: 'baz' });
  TestBed.flushEffects();

  expect(executionCount).toBe(2);
});
```

markostanimirovic (changes requested):

modules/signals/spec/get-state.spec.ts
jits: Updated and pushed.

I've also added a check that a subsequent call to `getState` will result in an updated state object.

markostanimirovic (approved):
LGTM 👍

brandonroberts (approved):

brandonroberts:
Thanks @jits!

Issue #3700, opened 2022-12-06
No Provider issue on lazy load module and ViewContainerRef with Angular 14


Julienbideau:
### Which @ngrx/* package(s) are the source of the bug?

router-store

### Minimal reproduction of the bug/regression with instructions

Since we bumped versions from Angular V13 to V14, and ngrx 13.0 to 14.2, we have an error : 
NullInjectorError: NullInjectorError: No provider for Store!

We need to declare our Store in a lazy loaded module : 
``` ts
StoreModule.forRoot({}),
EffectsModule.forRoot([])
```

We also don't use routing in our app, just component initialization like so : 
``` ts
 @ViewChild('transactionContainer', { read: ViewContainerRef, static: true })
  public transactionContainer: ViewContainerRef | undefined;

  constructor(private readonly injector: Injector) {}

  async loadLazyComponent() {

    const lazyStoreModuleNgModuleRef = createNgModule(
      LazyStoreModule,
      this.injector
    );
    this.transactionContainer!.createComponent(LazyStoreComponent, {
      ngModuleRef: lazyStoreModuleNgModuleRef,
    });
  }
```

If I declare it in the app.module.ts, It solves the problem but we can't declare the store there because **we need our stores to be isolated in each component** 

I made a stackblitz : 
https://stackblitz.com/edit/angular-ivy-u9b5yo?file=package.json

If you downgrade to 13.2.0, we have no issues anymore. 

### Minimal reproduction of the bug/regression with instructions

The store to be correctly provided

### Versions of NgRx, Angular, Node, affected browser(s) and operating system(s)

NgRx: 14.3.2
Angular 14.2.12
Node 16

### Other information

Thank you in advance ! :-)

### I would be willing to submit a PR to fix this issue

- [X] Yes
- [X] No

Phusty:
We have the same problem in our project. Are there any news?

rudobri:
Same Problem on our side. Please let us know. Thanks

Julienbideau:
I tryed the 15.1.0 version and we still have the same problem

brandonroberts:
We did not change anything with how injection works between those versions. Will look into it though

vavon92:
We experienced this issue with the 14.3.0 version, we solved it by downgrading to the 14.2.0 version.
Maybe it could be related to the change on how the services are provided.
Hope it helps!

brandonroberts:
[...]

PR #4119, opened 2023-11-09
fix(effects): allow effects providers to be registered outside of the root injector

brandonroberts:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Effects classes are registered using `providedIn: 'root'`. This doesn't allow effects to be used in a remote application when used with Module Federation unless the Store/Effects are registered within the host AppModule.

Closes https://github.com/ngrx/platform/issues/3700

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/effects/src/provide_effects.ts
+++ b/modules/effects/src/provide_effects.ts
@@ -15,6 +15,9 @@ import { EffectSources } from './effect_sources';
 import { rootEffectsInit as effectsInit } from './effects_actions';
 import { FunctionalEffect } from './models';
 import { getClasses, isClass } from './utils';
+import { EFFECTS_ERROR_HANDLER } from './tokens';
+import { defaultEffectsErrorHandler } from './effects_error_handler';
+import { Actions } from './actions';
 
 /**
  * Runs the provided effects.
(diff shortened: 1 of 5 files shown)


modules/effects/src/provide_effects.ts
```diff
@@ -66,6 +69,13 @@ export function provideEffects(
 
   return makeEnvironmentProviders([
     effectsClasses,
+    EffectSources,
```
brandonroberts, line 72: I think this will cause an issue when `provideEffects` is used in a child route, but wanted to open the discussion for the fix.

aokrushk: Yes, that would be a problem.

modules/effects/src/provide_effects.ts
```diff
@@ -66,6 +69,13 @@ export function provideEffects(
...
+    EffectsRunner,
+    Actions,
+    {
+      provide: EFFECTS_ERROR_HANDLER,
+      useFactory: () => defaultEffectsErrorHandler,
+    },
```
markostanimirovic, line 78: Would it be still possible to use `provideEffects` at the child injector level? 🤔 We would need to add a check to ensure that these services are not provided multiple times.

Currently, we don't need to call `provideEffects` at the root level if there are no root effects. Even if we add a check to ensure that they're not provided multiple times, this change would introduce a breaking change, so using `provideEffects` at the root level would be required similar to `EffectsModule.forRoot`.

An option that we can consider is keeping the same behavior for `provideEffects` and introducing this change only for module-based apps.

brandonroberts: Yea, good point. I don't know what the behavior is for standalone APIs with module federation, or if we'll have the same issue there.

brandonroberts:
Going to close this as its not a viable solution to this issue

PR #4124, opened 2023-11-11
fix(signals): remove state checks for better DX

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

After some research, we realized that compilation errors that prevent creating deep signals with keys `name` or `length` (overriding Signal/Function properties) can be hard to debug, especially when migrating existing stores to the SignalStore.

```ts
// compilation error - it's not allowed to override `name` which is a property of `state` Signal
const state = signalState({ name: 'Marko' });
```

## What is the new behavior?

We're able to override Signal/Function properties and create strongly typed deep signals with keys `name` or `length`. We can also pass optional properties or unknown records as state type, which is not recommended, but everything will be still properly typed.

There is more flexibility but also more responsibility on the user side when defining state.

```ts
const state = signalState({ name: 'Marko' });
const name = state.name; // type: Signal<string> ✅ 
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/src/deep-signal.ts
+++ b/modules/signals/src/deep-signal.ts
@@ -1,5 +1,17 @@
-import { computed, Signal, untracked } from '@angular/core';
+import {
+  computed,
+  isSignal,
+  Signal as NgSignal,
+  untracked,
+} from '@angular/core';
 import { IsUnknownRecord } from './ts-helpers';
+
+// An extended Signal type that enables the correct typing
+// of nested signals with the `name' or `length' key.
+interface Signal<T> extends NgSignal<T> {
+  name: unknown;
+  length: unknown;
+}
 
 export type DeepSignal<T> = Signal<T> &
   (T extends Record<string, unknown>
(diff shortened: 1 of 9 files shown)


modules/signals/src/deep-signal.ts
```diff
@@ -1,6 +1,18 @@
...
+  computed,
+  isSignal,
+  Signal as NgSignal,
+  untracked,
+} from '@angular/core';
 import { IsUnknownRecord } from './ts-helpers';
 
+// An extended Signal type that enables the correct typing
+// of nested signals with the `name' or `length' key.
+interface Signal<T> extends NgSignal<T> {
```
timdeschryver, line 11: I think creator our own type with the same name can be confusing.
Could we rename this to `ExtendedSignal` or something else?

timdeschryver (commented):
👍 the types are also simpler now, but it's sad to see `K extends keyof Function ? true : HasFunctionKeys<T[K]>;` go 😅

modules/signals/src/deep-signal.ts
markostanimirovic: If we rename it, IDE will show `ExtendedSignal` instead of `Signal` for SignalState/Store properties, which will be confusing to users. Extended signal type is private and fully compatible with Angular's Signal type.

There is no other way to provide correct typing if a SignalState/Store contains `name` or `length` properties (at least I didn't find it). If we do: `Omit<Signal<T>, 'name' | 'length'>, TypeScript will not recognize that type as a function, so it won't be executable anymore.

timdeschryver: I see, I thought this was just for internal use.

timdeschryver (approved):

brandonroberts (approved):

PR #4152, opened 2023-11-24
fix(signals): allow using signalStore and signalState in TS libs

### LINKED ISSUE #4151: ngrx/signals type error "Exported variable 'CounterStore' has or is using name 'STATE_SIGNAL' from external module "node_modules/@ngrx/signals/src/signal-state" but cannot be named"

signals

Copy the code from the ngrx/signals anouncement:
[Announcing NgRx v17: Introducing NgRx Signals](https://dev.to/ngrx/announcing-ngrx-v17-introducing-ngrx-signals-operators-performance-improvements-workshops-and-more-55e4)

```typescript

export const CounterStore = signalStore(withState({ count: 0 }));
```

image

### Expected behavior

I am working in an Nx mono repo. If I add the store ts file directly in the app (`apps` folder), the error is not present. The error only happens when the file is created in an nx library (`libs` folder).

This works correctly.
image

This does not work.
image

ngrx/signals: v17.0.0
Angular: v17.0.1
Nx: v17.1.2
Node: v18.17.1

### Other information

_No response_

- [ ] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4151 

## What is the new behavior?

`signalStore` and `signalState` can be used/exported in projects that have `tsconfig` with `"declaration": true`.

```
[ ] Yes
[x] No
```

## Other information

This is a workaround for the following TS bug: https://github.com/microsoft/TypeScript/issues/37888

### DIFF
--- a/modules/signals/src/deep-signal.ts
+++ b/modules/signals/src/deep-signal.ts
@@ -8,7 +8,7 @@ import { IsKnownRecord } from './ts-helpers';
 
 // An extended Signal type that enables the correct typing
 // of nested signals with the `name` or `length` key.
-interface Signal<T> extends NgSignal<T> {
+export interface Signal<T> extends NgSignal<T> {
   name: unknown;
   length: unknown;
 }
--- a/modules/signals/src/signal-store-models.ts
+++ b/modules/signals/src/signal-store-models.ts
@@ -1,6 +1,6 @@
 import { Signal } from '@angular/core';
 import { DeepSignal } from './deep-signal';
-import { SignalStateMeta } from './signal-state';
+import { StateSignal } from './state-signal';
 import { IsKnownRecord, Prettify } from './ts-helpers';
 
 export type SignalStoreConfig = { providedIn: 'root' };
@@ -15,12 +15,11 @@ export type SignalStoreSlices<State> = IsKnownRecord<
     }
   : {};
 
-export type SignalStore<FeatureResult extends SignalStoreFeatureResult> =
+export type SignalStoreProps<FeatureResult extends SignalStoreFeatureResult> =
   Prettify<
     SignalStoreSlices<FeatureResult['state']> &
       FeatureResult['signals'] &
-      FeatureResult['methods'] &
-      SignalStateMeta<Prettify<FeatureResult['state']>>
+      FeatureResult['methods']
   >;
 
 export type SignalsDictionary = Record<string, Signal<unknown>>;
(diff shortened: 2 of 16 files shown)


modules/signals/src/signal-store-models.ts
```diff
@@ -15,12 +15,11 @@ export type SignalStoreSlices<State> = IsKnownRecord<
...
   : {};
 
-export type SignalStore<FeatureResult extends SignalStoreFeatureResult> =
+export type SignalStoreProps<FeatureResult extends SignalStoreFeatureResult> =
   Prettify<
     SignalStoreSlices<FeatureResult['state']> &
       FeatureResult['signals'] &
-      FeatureResult['methods'] &
-      SignalStateMeta<Prettify<FeatureResult['state']>>
+      FeatureResult['methods']
```
markostanimirovic, line 22: A workaround for this TS bug is to always put a unique symbol inside of the named type.

**Before:**

The `STATE_SIGNAL` was appended to the dynamic `SignalStore` type by prettifying the intersection.

```ts
const CounterStore = signalStore(withState({ count: 0 }));
// type: Type<{ count: Signal<number>; [STATE_SIGNAL]: WritableSignal<{ count: number }> }>
```

**After:**

The `STATE_SIGNAL` is always exported as a part of the `StateSignal` type.

```ts
const CounterStore = signalStore(withState({ count: 0 }));
// type: Type<{ count: Signal<number>; } & StateSignal<{ count: number }>>
```

modules/signals/src/deep-signal.ts
```diff
@@ -8,7 +8,7 @@ import { IsKnownRecord } from './ts-helpers';
 
 // An extended Signal type that enables the correct typing
 // of nested signals with the `name` or `length` key.
-interface Signal<T> extends NgSignal<T> {
+export interface Signal<T> extends NgSignal<T> {
```
markostanimirovic, line 11: This export fixes the same compilation error in TS libraries with `declaration: true` when `signalState` is used. The problem was caused by another unique symbol - Angular's `SIGNAL`.

alex-okrushko: 👍

modules/signals/src/signal-store-models.ts
alex-okrushko: Makes sense. 

`StateSignal` has to be added to the public APIs `index.ts` as well, right?

markostanimirovic: It's not needed. We can keep it private.

alex-okrushko (approved):

otkrivashkin:
Please, somebody approve it)

brandonroberts (approved):

PR #4196, opened 2023-12-31
fix(signals): destroy hook doesn't run in injection context.

rainerhahnekamp:
## PR Checklist

## PR Type

It fixes a testing issues when a store, which is provided in root, has an onDestroy hook.

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The Signal Store saves the injection context upon instantiation and re-uses it in the `onDestroy` hook.

If the Store is provided in `root`, any test with an `onDestroy` hook fails with the error message: `NG0205: Injector has already been destroyed`.

This commit requires `withHook` to access the DI outside of `onDestroy`.

With the main branch, it is very easily to test. Just run the following test:

```typescript
it('should fail', () => {
  const Store = signalStore(
    { providedIn: 'root' },
    withState({ name: 'Coffee' }),
    withHooks({
      onDestroy() {
        console.log('will throw...');
      },
    })
  );

  TestBed.inject(Store);
});
```

## What is the new behavior?

`onDestroy` runs outside of the injection context.

I would like to add, that this PR makes `withHooks` in its usage fully compatible with `withMethods`, `withComputed` and are similar extensions that provide access to the DI.

```
[x] Yes
[ ] No
```

We add a note the changelog, describe the reasons and that's it. Signal Store is, as far as I know, still in developer preview.

## Other information

I left the old notation of `withHooks`. If that PR would be accepted, I'd suggest to remove that one as well, so that `withHooks` accepts only a `HookSupplier` as argument.

### DIFF
--- a/modules/signals/spec/helpers.ts
+++ b/modules/signals/spec/helpers.ts
@@ -2,7 +2,8 @@ import { Component, inject, Type } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 
 export function createLocalService<Service extends Type<unknown>>(
-  serviceToken: Service
+  serviceToken: Service,
+  provideLocally = true
 ): {
   service: InstanceType<Service>;
   flushEffects: () => void;
@@ -11,7 +12,7 @@ export function createLocalService<Service extends Type<unknown>>(
   @Component({
     standalone: true,
     template: '',
-    providers: [serviceToken],
+    providers: provideLocally ? [serviceToken] : [],
   })
   class TestComponent {
     service = inject(serviceToken);
--- a/modules/signals/spec/signal-store.spec.ts
+++ b/modules/signals/spec/signal-store.spec.ts
@@ -1,5 +1,11 @@
-import { inject, InjectionToken, isSignal, signal } from '@angular/core';
-import { TestBed } from '@angular/core/testing';
+import {
+  inject,
+  Injectable,
+  InjectionToken,
+  isSignal,
+  signal,
+} from '@angular/core';
+import { TestBed, waitForAsync } from '@angular/core/testing';
 import {
   patchState,
   signalStore,
@@ -275,15 +281,16 @@ describe('signalStore', () => {
         factory: () => 'ngrx',
       });
       const Store = signalStore(
-        withHooks({
-          onInit() {
-            inject(TOKEN);
-            messages.push('onInit');
-          },
-          onDestroy() {
-            inject(TOKEN);
-            messages.push('onDestroy');
-          },
+        withHooks(() => {
+          const token = inject(TOKEN);
+          return {
+            onInit() {
+              messages.push('onInit');
+            },
+            onDestroy() {
+              messages.push('onDestroy');
+            },
+          };
         })
       );
       const { destroy } = createLocalService(Store);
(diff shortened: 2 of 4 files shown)


markostanimirovic:
Thanks Rainer! I suggest splitting the bugfix and new feature into separate PRs.

---

Regarding the new signature, it seems useful! But I'd slightly change it in terms of verbosity:

```ts
withHooks((store) => {
  // on init logic

  return () => /* on destroy logic */
})
```

> Inspired by: https://svelte.dev/docs/svelte#onmount

Returning the `onInit` callback seems unnecessary to me because init logic can be performed within the `withHooks` factory (out of `onInit`) anyway:

```ts
withHooks(() => {
  // init logic can be performed here

  return {
    onInit() {
      // and here as well
    }
  };
})
```

rainerhahnekamp:
I'll split it into two parts @markostanimirovic. But please help me out here. What would be the new feature here? As far as I see it, the fix is the new feature.

markostanimirovic:
> I'll split it into two parts @markostanimirovic. But please help me out here. What would be the new feature here? As far as I see it, the fix is the new feature.

We can split it in the following way:

- Bugfix: Executing `onDestroy` out of injection context with tests that verify the fix
- Feature: Proposal for new `withHooks` signature

rainerhahnekamp:
I'm sorry, but I don't know how to provide a fix without breaking change.

Currently, it is like this:

```typescript
inject(DestroyRef).onDestroy(() => {
  runInInjectionContext(injector, hooks.onDestroy!);
});
```

The way I understand the proposed fix:

```typescript
inject(DestroyRef).onDestroy(() => {
  hooks.onDestroy!();
});
```

But then any hook like 

```typescript
withHooks({onDestroy(store) {
  const service = inject(SomeService);
}});
```

will fail. So I would fix one thing and introduce a bug at the same time.

Maybe you can give me a hint. Thanks!

markostanimirovic:
> I'm sorry, but I don't know how to provide a fix without breaking change.
> 
> Currently, it is like this:
> 
> ```ts
> inject(DestroyRef).onDestroy(() => {
>   runInInjectionContext(injector, hooks.onDestroy!);
> });
> ```
> 
> The way I understand the proposed fix:
> 
> ```ts
> inject(DestroyRef).onDestroy(() => {
>   hooks.onDestroy!();
> });
> ```
> 
> But then any hook like
> 
> ```ts
> withHooks({onDestroy(store) {
>   const service = inject(SomeService);
> }});
> ```
> 
> will fail. So I would fix one thing and introduce a bug at the same time.
> 
> Maybe you can give me a hint. Thanks!

Yeah, it will be a breaking change, but if I understood correctly, the current behavior breaks tests when SignalStore is provided at the root level, so I think it's fine to move the `onDestroy` execution out of the injection context and provide the additional `withHooks` signature in a separate PR once we agree about its design. Both - feature and bugfix can be released together.

Btw, `@ngrx/signals` is in developer preview, so we can introduce breaking changes in minor/patch releases as Angular does with signals. Of course, breaking changes will be reduced to a minimum, but this one seems necessary.

rainerhahnekamp:
This one gets split up into two different PRs (fix and feature)

PR #4199, opened 2024-01-03
feat(signals): Protection for Overriding Properties

rainerhahnekamp:
This Proof of Concept intrdocues a feature which prevents that feature override properties (`state`, `signals`, `methods`).

It does that by adding a conditional type `NoOverride` which as a constraint to the parameters of `signalState`.

Example:
```typescript
signalState(
  withState({ id: 1, name: 'hallo', prettyName: 'hi' }),
  withComputed((store) => {
    return {
      prettyName: computed(() => store.name()),
    };
  })
)
```

This would fail to compile because `prettyName` is overriden.

The overriding protection is enabled by default. It is very likely that overriding doesn't happen on purpose.

An opt-out is also possible via the configuration:

```typescript
signalState(
  {allowOverrides: true},
  withState({ id: 1, name: 'hallo', prettyName: 'hi' }),
  withComputed((store) => {
    return {
      prettyName: computed(() => store.name()),
    };
  })
)
```

Screenshot 2024-01-03 at 19 48 31

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Features can override properties of other features.

Closes #

## What is the new behavior?

By default we get a compilation error. Allowing overrides is possible via the configuration.

```
[x] Yes
[ ] No
```

Honestly, I think that developers would welcome that breaking change. Overriding properties can cause bugs which might be hard to detect.

## Other information

This is just a POC. If accepted, it has to be integrated into the original `withState` and probably `signalStoreFeature`.

### DIFF
--- a/modules/signals/src/overriding-feature-poc.ts
+++ b/modules/signals/src/overriding-feature-poc.ts
@@ -1,0 +1,318 @@
+import { computed, Signal, Type } from '@angular/core';
+import {
+  EmptyFeatureResult,
+  MergeFeatureResults,
+  SignalStoreConfig,
+  SignalStoreFeature,
+  SignalStoreFeatureResult,
+  SignalStoreProps,
+} from './signal-store-models';
+import { StateSignal } from './state-signal';
+import { withState } from './with-state';
+import { withComputed } from './with-computed';
+import { withMethods } from '@ngrx/signals';
+import { Prettify } from './ts-helpers';
+
+/**
+ * This Proof of Concept would introduce a feature which prevents
+ * that feature override properties (`state`, `signals`, `methods`).
+ *
+ * It does that by adding a conditional type `NoOverride` which
+ * as a constraint to the parameters of `signalState`.
+ *
+ * Example:
+ * signalState(
+ *   withState({ id: 1, name: 'hallo', prettyName: 'hi' }),
+ *   withComputed((store) => {
+ *     return {
+ *       prettyName: computed(() => store.name()),
+ *     };
+ *   })
+ * )
+ *
+ * This would fail to compile because `prettyName` is overriden.
+ *
+ * The overriding protection is enabled by default. It is very likely
+ * that overriding doesn't happen on purpose.
+ *
+ * An opt-out is also possible via `signalState({allowOverrides: true})`.
+ */
+
+type OverridenProperties<T extends string> = `overriding property: '${T}'`;
+
+export type NestedProperties<Type> = {
+  [Property in keyof Type]: keyof Type[Property];
+} extends Record<string, infer P>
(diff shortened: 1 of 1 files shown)


rainerhahnekamp:
@markostanimirovic, since I don't know which process you prefer, I only created a PR.

If you prefer to open an issue first and do all the discussions there, please let me know.

modules/signals/src/overriding-feature-poc.ts
```diff
@@ -0,0 +1,318 @@
...
+ *     };
+ *   })
+ * )
+ *
+ * This would fail to compile because `prettyName` is overriden.
+ *
+ * The overriding protection is enabled by default. It is very likely
+ * that overriding doesn't happen on purpose.
+ *
+ * An opt-out is also possible via `signalState({allowOverrides: true})`.
```
timdeschryver, line 38: First of all thanks for all these PRs and feedback @rainerhahnekamp.

If we go through with this change, what would be the benefit of having this configurable?
Is there a use case where this would be desired to turn this off, and override properties?

rainerhahnekamp: Hi Tim, thanks a lot as well.

To me, the use case for disabling the override protection would be in enterprise-like environments, where teams share huge extensions and I - as a consumer - want to customize or change parts of them. So a little bit similar to what we can do with class inheritance.

For example, I'd like to override `undo` of a `withUndoRedo` extension:

```
signalStore(
  // ...
  withUndoRedo(),
  withMethods(store => {
    return {
      undo: (times = 1) => {
        for(const i = 0; i < times; i++) {
          store.undo();
        }
      }
    }
  })
);
```

Maybe it doesn't have to be a configuration value but it could be a function we can add selectively:

```
signalStore(
  // ...
  withUndoRedo(),
  withOverride(
    withMethods(store => {
      return {
        undo: (times = 1) => {
          for(const i = 0; i < times; i++) {
            store.undo();
          }
        }
      }
    })
  )
);
```

---

> If we go through with this change...

Before we add new features to https://github.com/angular-architects/ngrx-toolkit, I'd like to post them here first.

If you see it as something which should be part of the core, we can move forward here. If you see, it is a better fit for a community contribution, we would integrate it into the toolkit. Could also serve as an incubator.

In the same manner, I also plan to push the branch for an encapsulation feature.

timdeschryver: Thanks for the elaboration @rainerhahnekamp

markostanimirovic:
closing this PR in favor of https://github.com/ngrx/platform/pull/4424

PR #4200, opened 2024-01-03
fix(signals): run `onDestroy` outside of injection context

rainerhahnekamp:
## PR Checklist

## PR Type

The Signal Store saves the injection context upon instantiation and re-uses it in the `onDestroy` hook.

If the Store is provided in root, any test with an `onDestroy` hook fails with the error message: NG0205: Injector has already been destroyed.

This PR runs `onDestroy` outside of the injection context.

Since this will break any code which depends on the injection context in `onDestroy`, there will be follow-up (breaking change) feature which will provide the injection context before the hook is executed.

 Discussed in: https://github.com/ngrx/platform/pull/4196

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Any test with a store `providedIn: root` fails.

Closes #

## What is the new behavior?

`onDestroy` runs outside of the injection context.

```
[x] Yes
[ ] No
```

There will be a feature PR which re-enables the injection context, but in a different way.

### DIFF
--- a/modules/signals/src/signal-store.ts
+++ b/modules/signals/src/signal-store.ts
@@ -325,16 +325,14 @@ export function signalStore(
         (this as any)[key] = props[key];
       }
 
-      if (hooks.onInit) {
-        hooks.onInit();
+      const { onInit, onDestroy } = hooks;
+
+      if (onInit) {
+        onInit();
       }
 
-      if (hooks.onDestroy) {
-        const injector = inject(Injector);
-
-        inject(DestroyRef).onDestroy(() => {
-          runInInjectionContext(injector, hooks.onDestroy!);
-        });
+      if (onDestroy) {
+        inject(DestroyRef).onDestroy(onDestroy);
       }
     }
   }
--- a/modules/signals/src/signal-store.ts
+++ b/modules/signals/src/signal-store.ts
@@ -1,21 +1,13 @@
-import {
-  DestroyRef,
-  inject,
-  Injectable,
-  Injector,
-  runInInjectionContext,
-  signal,
-  Type,
-} from '@angular/core';
+import { DestroyRef, inject, Injectable, signal, Type } from '@angular/core';
 import { STATE_SIGNAL, StateSignal } from './state-signal';
 import {
   EmptyFeatureResult,
   InnerSignalStore,
   MergeFeatureResults,
-  SignalStoreProps,
   SignalStoreConfig,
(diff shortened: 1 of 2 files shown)


modules/signals/src/signal-store.ts
```diff
@@ -325,16 +325,14 @@ export function signalStore(
...
-        const injector = inject(Injector);
+      if (onInit) {
+        onInit();
+      }
 
-        inject(DestroyRef).onDestroy(() => {
-          runInInjectionContext(injector, hooks.onDestroy!);
-        });
+      if (onDestroy) {
+        inject(DestroyRef).onDestroy(onDestroy);
```
markostanimirovic, line 335: Can we add the test(s) that verifies the fixed bug?

modules/signals/src/signal-store.ts
```diff
@@ -325,16 +325,14 @@ export function signalStore(
...
+      const { onInit, onDestroy } = hooks;
 
-      if (hooks.onDestroy) {
-        const injector = inject(Injector);
+      if (onInit) {
+        onInit();
+      }
 
-        inject(DestroyRef).onDestroy(() => {
-          runInInjectionContext(injector, hooks.onDestroy!);
```
markostanimirovic, line 336: `Injector` and `runInInjectionContext` APIs are unused and can be removed from imports

markostanimirovic (changes requested):

modules/signals/src/signal-store.ts
rainerhahnekamp: Done

modules/signals/src/signal-store.ts
rainerhahnekamp: Done

modules/signals/src/signal-store.ts
markostanimirovic: I was thinking about verifying that the test will not throw an error when the store is provided in root and the `onDestroy` hook is used. Sorry if it wasn't clear!

rainerhahnekamp: No problem, I've replaced the test. I hope it is better now.

markostanimirovic: Thanks Rainer!

markostanimirovic (approved):

brandonroberts (approved):

PR #4208, opened 2024-01-09
feat(signals): add `withHooks` signature with factory input

### LINKED ISSUE #4201: provide injection context to `onDestroy`

signals

### Information

As discussed in https://github.com/ngrx/platform/pull/4196, we have to run `onDestroy` outside of the injection context.

At the moment, there are two options on the table:

**Version 1: "Svelte style":**

```typescript
withHooks((store) => {
  // on init logic

  const service = inject(SomeService);

  return () => {
    service.doSomethingAtTheEnd();
  }
})
```

**Version 2:**

```typescript
withHooks((store) => {
  // access to DI

  const service = inject(SomeService);

  return {
    onInit() {
      service.doSomethingAtTheStart();
    },

    onDestroy() {
      service.doSomethingAtTheStart();
    }
  }
})
```

I opt for version 2 because
1. it follows the same pattern of `withMethods` & `withComputed`, where we have the first part which provides access to the DI and the second contains the implementation.
2. It allows us to add further hooks. For example there is a discussion, asking for a `onChange`: https://github.com/ngrx/platform/discussions/4192
3. Having an explicit `onInit` might be more familiar to an Angular developer because of the same hook in the framework.

---

In the end, it is a question of personal style. From a technical perspective, both options are ok.

_No response_

- [X] Yes
- [ ] No

rainerhahnekamp:
This PR re-introduces the injection context
into hooks after they were removed in e21df19.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, hooks have no way to access the dependency injection.

Closes #4201

## What is the new behavior?

Hook can access DI.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/spec/signal-store.spec.ts
+++ b/modules/signals/spec/signal-store.spec.ts
@@ -280,17 +280,38 @@ describe('signalStore', () => {
       });
       const Store = signalStore(
         withState({ name: 'NgRx Store' }),
-        withHooks(() => {
+        withHooks((store) => {
           const tokenInit = inject(TOKEN_INIT);
           const tokenDestroy = inject(TOKEN_DESTROY);
           return {
-            onInit(store) {
+            onInit() {
               messages.push(`${tokenInit} ${store.name()}`);
             },
-            onDestroy(store) {
+            onDestroy() {
               messages.push(`${tokenDestroy} ${store.name()}`);
             },
           };
+        })
+      );
+      const { destroy } = createLocalService(Store);
+
+      expect(messages).toEqual(['init NgRx Store']);
+
+      destroy();
+      expect(messages).toEqual(['init NgRx Store', 'destroy NgRx Store']);
+    });
+
+    it('executes hooks without injection context', () => {
+      const messages: string[] = [];
+      const Store = signalStore(
+        withState({ name: 'NgRx Store' }),
+        withHooks({
+          onInit(store) {
+            messages.push(`init ${store.name()}`);
+          },
+          onDestroy(store) {
+            messages.push(`destroy ${store.name()}`);
+          },
         })
       );
       const { destroy } = createLocalService(Store);
--- a/modules/signals/spec/signal-store.spec.ts
+++ b/modules/signals/spec/signal-store.spec.ts
@@ -268,32 +268,37 @@ describe('signalStore', () => {
       expect(message).toBe('onDestroy');
     });
 
-    // FIX: injection context will be provided for `onDestroy` in a separate PR
-    // see https://github.com/ngrx/platform/pull/4196#issuecomment-1875228588
     it('executes hooks in injection context', () => {
       const messages: string[] = [];
-      const TOKEN = new InjectionToken('TOKEN', {
+      const TOKEN_INIT = new InjectionToken('TOKEN_INIT', {
         providedIn: 'root',
-        factory: () => 'ngrx',
+        factory: () => 'init',
+      });
+      const TOKEN_DESTROY = new InjectionToken('TOKEN_DESTROY', {
+        providedIn: 'root',
+        factory: () => 'destroy',
       });
       const Store = signalStore(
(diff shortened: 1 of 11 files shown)


modules/signals/src/with-hooks.ts
```diff
@@ -16,13 +16,23 @@ type HooksFactory<Input extends SignalStoreFeatureResult> = (
...
+};
+
+export function withHooks<Input extends SignalStoreFeatureResult>(
+  hooks:
+    | {
+        onInit?: HooksFactory<Input>;
+        onDestroy?: HooksFactory<Input>;
+      }
+    | HooksSupplier<Input>
+): SignalStoreFeature<Input, EmptyFeatureResult> {
```
markostanimirovic, line 31: Instead of:

```ts
withHooks(() => ({
  onInit(store) {},
  onDestroy(store) {},
}))
```

Let's use:

```ts
withHooks((store) => ({
  onInit() {},
  onDestroy() {},
}))
```

I also suggest defining two signatures for the `withHooks` function instead of a union type for better readability, but also for easier potential deprecation of the first signature in the future.

markostanimirovic (changes requested):

modules/signals/src/with-hooks.ts
rainerhahnekamp: Done, since the old might become a candidate for deprecation, I named the new version's type `HookFactory` and the old one is now just a `HookFn`. Let me know, if should revert the renaming.

modules/signals/spec/signal-store.spec.ts
```diff
@@ -268,32 +268,58 @@ describe('signalStore', () => {
       expect(message).toBe('onDestroy');
     });
 
-    // FIX: injection context will be provided for `onDestroy` in a separate PR
-    // see https://github.com/ngrx/platform/pull/4196#issuecomment-1875228588
     it('executes hooks in injection context', () => {
```
markostanimirovic, line 271:
```suggestion
    it('executes hooks factory in injection context', () => {
```

Additionally, we can add a test to ensure that `onInit` is also executed within the injection context for the first (old) signature.

markostanimirovic (changes requested):

modules/signals/src/with-hooks.ts
```diff
@@ -16,13 +16,46 @@ type HooksFactory<Input extends SignalStoreFeatureResult> = (
...
   return (store) => {
-    const createHook = (name: keyof typeof hooks) => {
-      const hook = hooks[name];
+    const _hooks =
+      typeof hooks === 'function'
+        ? hooks({
+            [STATE_SIGNAL]: store[STATE_SIGNAL],
+            ...store.slices,
+            ...store.signals,
+            ...store.methods,
```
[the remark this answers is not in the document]
rainerhahnekamp: Done

rainerhahnekamp:
Pushed latest changes as requested and updated merged from main

modules/signals/src/with-hooks.ts
```diff
@@ -16,11 +16,45 @@ type HooksFactory<Input extends SignalStoreFeatureResult> = (
...
+export function withHooks<Input extends SignalStoreFeatureResult>(
+  hooksOrFactory:
+    | {
+        onInit?: HookFn<Input>;
+        onDestroy?: HookFn<Input>;
+      }
+    | HooksFactory<Input>
+): SignalStoreFeature<Input, EmptyFeatureResult> {
   return (store) => {
+    const storeProps = {
```
markostanimirovic, line 48: to avoid redundant code, `storeProps` can be used on line 68

rainerhahnekamp: Sorry, missed that one. Should be all good now.

markostanimirovic (approved):

PR #4224, opened 2024-01-31
fix(signals): run `rxMethod` outside of reactive context

rainerhahnekamp:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

It looks like the latest refactoring in `rxMethod` (https://github.com/ngrx/platform/commit/fa45d92a3b0669f88a0cc2daeaa00f805aa1b205) causes it now to run inside an `effect`. Practically, every `withMethods`
 which uses `rxMethod` cannot call `patchState` anymore.

I've added two tests. Once you use the new version of `rxState` from the main branch, you will see that these tests will fail.

Closes #

## What is the new behavior?

I played safe and reverted back to the old version. Maybe one can find a better alternative.

```
[] Yes
[ ] No
```


modules/signals/rxjs-interop/src/rx-method.ts
```diff
@@ -30,23 +30,20 @@ export function rxMethod<Input>(
   destroyRef.onDestroy(() => sourceSub.unsubscribe());
 
   const rxMethodFn = (input: RxMethodInput<Input>) => {
-    if (isSignal(input)) {
-      const watcher = effect(() => source$.next(input()), { injector });
```
markostanimirovic, line 34: Thanks for the quick catch Rainer!

No need to revert `rxMethod` implementation. It's enough to change this line to:

```ts
      const watcher = effect(() => {
        const value = input();
        untracked(() => source$.next(value));
      }, { injector });
```

and new tests will pass. 👌

rainerhahnekamp: Yup, done! The test also passes. I removed the second test because I don't think it brings any additional benefit.

markostanimirovic (approved):

alex-okrushko (approved):

PR #4240, opened 2024-02-11
perf(signals): avoid creating unnecessary objects in excludeKeys

### LINKED ISSUE #4238: minor perf improvement

signals

### Information

Some of the internal functions are executing too much object destructuring, like this one:

https://github.com/ngrx/platform/blob/main/modules/signals/src/helpers.ts#L4C1-L9C2

```ts
export function excludeKeys<
  Obj extends Record<string, unknown>,
  Keys extends string[]
>(obj: Obj, keys: Keys): Omit<Obj, Keys[number]> {
  return Object.keys(obj).reduce(
    (acc, key) => (keys.includes(key) ? acc : { ...acc, [key]: obj[key] }),
    {}
  ) as Omit<Obj, Keys[number]>;
}
```
this line: `    (acc, key) => (keys.includes(key) ? acc : { ...acc, [key]: obj[key] }),`
The thing is - depending on the number of keys (n) above code is creating n+1 new objects, where only 1 is needed (the new obj with excluded keys). Since reduce is sync anyway, nobody can have reference to the newly created object before reduce is finished. It's immutable-safe to go on with mutable reducer as long as the initial remains a new object `{}`. New object getting mutated right after creation (and before anyone can actually use it) breaks nothing. But is faster, slightly.

Since excludeKeys is a helper, it might be executed multiple times within an app, especially a big one.

I'd be happy to provide a simple PR and start contributing to the signal store 😅

the alternative is to leave it as it is.

- [X] Yes
- [ ] No

ducin:
Make the `excludeKeys` create only one object as it's enough for the function to work correctly. Currently the `excludeKeys` helper function creates a new object on every iteration of the non-removed keys. It's unnecessary as - before the function returns the result outside, nothing can have the reference to the newly created object. It's safe to mutate the temporarily created result.

no breaking changes, fixes #4238

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4238

```
[ ] Yes
[X] No
```

### DIFF
--- a/modules/signals/src/helpers.ts
+++ b/modules/signals/src/helpers.ts
@@ -2,8 +2,10 @@ export function excludeKeys<
   Obj extends Record<string, unknown>,
   Keys extends string[]
 >(obj: Obj, keys: Keys): Omit<Obj, Keys[number]> {
-  return Object.keys(obj).reduce(
-    (acc, key) => (keys.includes(key) ? acc : { ...acc, [key]: obj[key] }),
-    {}
-  ) as Omit<Obj, Keys[number]>;
+  return Object.keys(obj).reduce<Record<string, unknown>>((acc, key) => {
+    if (!keys.includes(key)) {
+      acc[key] = obj[key];
+    }
+    return acc;
+  }, {}) as Omit<Obj, Keys[number]>;
 }


markostanimirovic (approved):
Thanks Tomasz!

brandonroberts (approved):
🔥

PR #4247, opened 2024-02-14
fix(signals): add `StateSignal` to the public API

ducin:
expose `StateSignal` store type in order to make it easier for 3rd party extensions to wrap signal store APIs, e.g. patchState with immer immutability - generic `StateSignal<State>` is difficult to obtain via `ReturnType<typeof signalStore>` or any other way.

no breaking changes

## PR Checklist

## PR Type

```
```

```
[ ] Yes
[X] No
```

### DIFF
--- a/modules/signals/src/index.ts
+++ b/modules/signals/src/index.ts
@@ -4,6 +4,7 @@ export { signalState } from './signal-state';
 export { signalStore } from './signal-store';
 export { signalStoreFeature, type } from './signal-store-feature';
 export { SignalStoreFeature } from './signal-store-models';
+export { StateSignal } from './state-signal';
 
 export { withComputed } from './with-computed';
 export { withHooks } from './with-hooks';


rainerhahnekamp:
I think that one was already on the table: https://github.com/ngrx/platform/issues/4116#issuecomment-1803971396

ducin:
@rainerhahnekamp not sure if that's the same thing - I don't care about the symbol.

Lacking the type is a blocker when it comes to providing a more specialized `patchState`. The original one is:

```ts
export declare function patchState<State extends object>(stateSignal: StateSignal<State>, ...updaters: Array<Partial<State & {}> | PartialStateUpdater<State & {}>>): void;
```

I just need to know the `StateSignal` type for the wrapping function; it's type-only, no effect on runtime whatsoever.

rainerhahnekamp:
Ah, I see; I remember struggling with that `object` as well and giving up in the end.

markostanimirovic (approved):

markostanimirovic:
Thanks Tomasz! Btw, public API exports are usually released as features. Since `@ngrx/signals` is still in the dev preview, I changed the PR type to `fix`, so it can be released in the patch release too.

brandonroberts (approved):

PR #4271, opened 2024-03-05
refactor(signals): simpler entity named types

flensrocker:
While playing with the signal store (very nice thing!) I found a nicer way to map the types of `withEntities` to the named types.

It makes use of the `Capitalize` utility type and maps all properties in one line instead of summing up all properties with intersections.


timdeschryver (approved):
I also find this cleaner, and it seems like the behavior is the same 👍

markostanimirovic (approved):

markostanimirovic:
thanks @flensrocker!

PR #4276, opened 2024-03-12
docs: providing more explicit detail

goodnic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Consuming Signals is implicitly described in the example. But the only explicit mention is of the `getState` function.

Closes #

## What is the new behavior?

Also mention that Signals can be consumed without any overhead. This should help devs getting into the Signals topic and in turn the `SignalStore` more easily.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngrx.io/content/guide/signals/signal-store/index.md
+++ b/projects/ngrx.io/content/guide/signals/signal-store/index.md
@@ -141,6 +141,8 @@ export class BooksComponent {
 
 </code-example>
 
+The value wrapped inside a Signal object can be easily accessed at any time by calling the Signal as a function with no arguments.
+
 The `@ngrx/signals` package also offers the `getState` function to get the current state value of the SignalStore.
 When used within the reactive context, state changes are automatically tracked.
 


projects/ngrx.io/content/guide/signals/signal-store/index.md
```diff
@@ -141,6 +141,8 @@ export class BooksComponent {
 
 </code-example>
 
+The value wrapped inside a Signal object can be easily accessed at any time by calling the Signal as a function with no arguments.
```
timdeschryver, line 144: Thanks for the PR @goodnic 

I think when this isn't clear that this text isn't that helpful, it probably means the user is new to signals. Instead it would be better to point to the signals docs in that case.

This can also be applicable to other packages of NgRx, but this isn't something that we do.
That's why I'm not a big fan of this PR, and I would rather leave it as is.

Issue #4283, opened 2024-03-26
RFC: SignalStore with WritableSignals and ReadOnlySignalStore


marklagendijk:
### Which @ngrx/* package(s) are relevant/related to the feature request?

signals

### Information

When thinking about how Signals should change the Store concept, I thought: there should be a SignalStore.
So I googled for 'signal store' and found the beautiful implementation of that has been created in NgRx.
There were two main features that I loved:
1. It brings the power of 'composition over inheritance' that Signals bring us, to the Store world. And in a very powerful way, I might add.
2. It cleverly builds on the existing Signal functionality, instead of doing too much itself.

However, I also found two points that were disappointing:
1. A Store service allows modifying the data from the outside. 
2. Updating the data in the store is done via the `patchState` which is called with the Store instance, instead of being done via `WritableSignal`. This point goes against 'building on existing Signal functionality'.

This morning I woke up with an idea for tackling both these points. I do not know if:
1. It is possible to implement this, or whether there is something I don't know about that is blocking this.
2. It will be acceptable to introduce such a big change to the SignalStore. I would expect it would at least be a major version.

The idea:
1. A `SignalStore` will hold `DeepWritableSignal`s. 
2. All the `withX` calls get the `SignalStore` as argument, and can update any signal value via the normal `set` and `update` methods of those signals. For nested values this would look like this: `store.person.address.street.set(value)`.
3. The result of calling `signalStore` is not a `SignalStore` but a `ReadonlySignalStore`. The `ReadOnlySignalStore` is a clone of its `SignalStore`, except that all it's signals are read only ([asReadOnly](https://angular.io/api/core/WritableSignal#asReadonly))

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [X] No

marklagendijk:
This issue relates to #4210 feat(signals): Encapsulation

rainerhahnekamp:
@marklagendijk: I have a prototype for that `DeepWritableSignals` but it is far from stable. My use case was the support for the banana box syntax for Signals introduced in 17.2.

So if you have 

```typescript
const user = signalStore(withState({firstname: 'Hugo', lastname: 'Schmidt'}));
```

It would be great to have an easy option to use it in the template like

```html
<input [(ngModel)]="user.firstname">
<input [(ngModel)]="user.lastname">
```

marklagendijk:
That could indeed be nice. Especially for component level stores.
[...]

PR #4309, opened 2024-04-29
feat: make signalStore entries starting with _ not available outside of the store (interface, first level)

mauriziocescon:
## Summary
As (more or less) mentioned in https://github.com/ngrx/platform/issues/4283 and https://github.com/ngrx/platform/pull/4210, providers returned by the factory `signalStore` are currently lacking some sort on encapsulation. 

As I see it, there are 2 problems: 
1.  there is no way to declare that some slices of the state / computeds / methods are private, meaning that they are not meant to be used by consumers, 
2. `patchState` and `getState` are defined as utilities applicable on both objects returned by `signalStore` and `signalState`. This is good for reusability, but it comes with the undesired side effect of making a signalStore patchable anywhere (like in components consuming it). Considering the store mental model, `patchState` and `getState` should probably be available only in methods / hooks. 

The goal of this PR is targeting point 1, following (hopefully) the suggestion in https://github.com/ngrx/platform/pull/4210#issuecomment-1966810494. Here is a use case where the feature would come in handy (there are many others): 
``` typescript
/**
 * Modified example at https://ngrx.io/guide/signals/signal-store#putting-it-all-together
 * where the store is defined as a global provider.
 *
 * In this case, you probably don't want to have loadByQuery callable from consumers
 * because you don't want to have more than one subscription.
 */
const BooksStore = signalStore(
  { providedIn: 'root' },
  withState({/* ... */ filter: { query: '', order: 'asc' } }),
  // ...
  withMethods((store, booksService = inject(BooksService)) => ({
      // not callable from components injecting BooksStore
      _loadByQuery: rxMethod<string>(/* ... */),
    }),
  ),
  withHooks({
    onInit(store): void {
      store._loadByQuery(store.filter.query);
    },
  }),
);
```
With this PR, prefixing any root slice of the state / computed / method by `_` should make it unavailable (ts interface) in the returned `signalStore` provider. 

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Providers returned by the factory `signalStore` are currently lacking encapsulation

Partially address https://github.com/ngrx/platform/issues/4283 and https://github.com/ngrx/platform/pull/4210

## What is the new behavior?
Prefixing any root slice of the state / computed / method by `_` should make it unavailable (ts interface) in the returned `signalStore` provider. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/spec/types/signal-store.types.spec.ts
+++ b/modules/signals/spec/types/signal-store.types.spec.ts
@@ -198,6 +198,125 @@ describe('signalStore', () => {
     expectSnippet(snippet).toInfer('storeKeys', 'unique symbol');
   });
 
+  it('does not access state slices starting with _ from store = signalStore()', () => {
+    expectSnippet(`
+      const Store = signalStore(
+        withState({ _user: { _userId: '' } }),
+        withComputed(({ _user }) => ({
+          _userId: computed(() => _user._userId()),
+        })),
+        withMethods((store) => ({
+            doNothing(): void {
+              store._user();
+            },
+          }),
+        ),
+        withHooks({
+          onInit(store): void {
+            store.doNothing();
+          },
+        }),
+      );
+
+      const store = new Store();
+      const value = store._user;
+    `).toFail();
+
+    expectSnippet(`
+      const Store = signalStore(
+        withState({ _user: { _userId: '' } }),
+        withComputed(({ _user }) => ({
+          _userId: computed(() => _user._userId()),
+        })),
+        withMethods((store) => ({
+            doNothing(): void {
+              store._user();
+            },
+          }),
+        ),
+        withHooks({
+          onInit(store): void {
+            store.doNothing();
+          },
+        }),
+      );
+
+      const store = new Store();
+      const value = store._userId;
+    `).toFail();
+
+    expectSnippet(`
+      const Store = signalStore(
+        withState({ _user: { _userId: '' } }),
+        withComputed(({ _user }) => ({
+          _userId: computed(() => _user._userId()),
+        })),
+        withMethods((store) => ({
+            _doNothing(): void {
+              store._user();
+            },
+          }),
+        ),
+        withHooks({
+          onInit(store): void {
+            store._doNothing();
+          },
+        }),
+      );
+
+      const store = new Store();
+      const value = store._doNothing;
+    `).toFail();
+
+    expectSnippet(`
+      const Store = signalStore(
+        withState({ user: { _userId: '' } }),
+        withComputed(({ user }) => ({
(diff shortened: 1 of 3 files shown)


rainerhahnekamp:
Hey @mauriziocescon, good stuff!

In your case, would you also have to prefix the properties in the state with `_` to make them unpatchable from the outside?

Have you considered combining #4210 so that the whole state would be patchable?

By combining, I meant that for the state we could use `signalStore({readonly: true})` and methods and computed stick to `_`.

mauriziocescon:
@rainerhahnekamp yep! You can quickly take a look at the tests to see what's supported. 

@markostanimirovic any thought about this PR? In line with what the team has in mind or not really? Thanks!

markostanimirovic:
There are several open questions:

- Should private state slices be excluded from (readonly)`StateSource`?

```ts
const Store = signalStore(withState({ _foo: 'foo', bar: 'bar' }));

const store = inject(Store);
const state = getState(store); // should the type be { _foo: string; bar: string } or { bar: string } ?
```

- Should private state slices be excluded from `WritableStateSource` when `protectedState` is set to `false`?

```ts
const Store = signalStore(
  { protectedState: false },
  withState({ _foo: 'foo', bar: 'bar' })
);

const store = inject(Store);
patchState(store, { _foo: 'should it be allowed to patch foo because the state is unprotected?' });
```

- What's the expected behavior when `_` is used within the custom SignalStore feature? Should private members be available in a store that uses such a feature?

-  With this feature available, it will probably be expected to allow defining private entity properties. How will the `withEntities` signature look in this case? What about entity updaters?

```ts
export const TodosStore = signalStore(
  withEntities({ entity: type<Todo>(), privateMembers: ['entityMap', 'ids'], collection: 'todo' }),
  withMethods((store) => ({
    addTodo(todo: Todo) {
      patchState(store, addEntity(todo, { collection: 'todo', privateMembers: ['entityMap', 'ids'] })); // 😑
    },
  }))
);
```

Additionally, this will add significant complexity to the TypeScript part of the `entities` plugin implementation.

---

There is a flow that we try to follow when it comes to new features: Suggestion (RFC) => Discussion/Feedback => PR

SignalStore is already widely used and we cannot rapidly add new features without ensuring that all use cases are covered in a consistent way. Therefore, I suggest opening the RFC, covering open questions, and collecting feedback first.

mauriziocescon:
Yeah, I understand. Closing it!

mauriziocescon:
Ok, created a dedicated ticket https://github.com/ngrx/platform/issues/4446

PR #4315, opened 2024-05-07
feat(signal): extends merged features results up to 25

### LINKED ISSUE #4314: RFC: Allow more than 10 signal store features

signals

### Information

The typescript definition of `signalStore` function contains 10 overrides, which allows a maximum of 10 signal store `with` methods to be used. That includes all the custom features and the features they are using. In real life scenario I have about 20 reusable custom features that can be used in my signal stores, but when I combine more than 3-4 of them together, the typescript compiler marks errors becuase there is no overload for a type that merges more than 10 feature results.

I am combining sets of custom features together to reduce the total amount of `with` methods used. So one custom feature actually adds 2 features at the same time, with merged state, merged sets of computed signals, and so on. This of course reduces the modularity of the application - which goes against the whole idea of custom features (which is brilliant, BTW)

I think a more realistic limit to the number of merged features results is about 25 :-)

- [ ] Yes
- [ ] No

SonyStone:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4314

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/src/signal-store.ts
+++ b/modules/signals/src/signal-store.ts
@@ -11,7 +11,7 @@ import {
 } from './signal-store-models';
 import { Prettify } from './ts-helpers';
 
-// @ts-ignore
+// @ts-expect-error
 export function signalStore<F1 extends SignalStoreFeatureResult>(
   f1: SignalStoreFeature<EmptyFeatureResult, F1>
 ): Type<SignalStoreProps<F1> & StateSignal<Prettify<F1['state']>>>;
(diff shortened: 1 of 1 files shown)


brandonroberts:
Let's discuss this in the issue first

SonyStone:
You could fix this issue now just by those add types, which has no breaking changes for curent users.

markostanimirovic:
> You could fix this issue now just by those add types, which has no breaking changes for curent users.

Let's see. The question is - is this really an issue that should be fixed?

For example, `Observable.pipe` supports typing for up to 9 operators. If you have more than 9 operators in your chain, you're probably on the wrong way.

Before trying to find a solution or workaround for a problem that occurs in a system, we should first check if the system is properly designed. In other words, having a SignalStore with 20+ features is probably a wrong design choice, because that store does too many things and it could be very hard to maintain.

Anyway, adding additional signatures will be considered, but not 20+.

abhijit-chikane:
any update on this?

rainerhahnekamp:
@abhijit-chikane you find the discussion and issue at #4314

annyDot:
Any updates on how to work around the signalStore feature limit? I don't see how it's bad design to have more than 20+ features. For example, a page with many user actions + stuff like logging, withDevTools etc can easily hit that if we put each action logic into a separate feature.

PR #4183, opened 2023-12-15
feat(store): export TypedAction

NathanLaing:
Export TypedAction from Store

Closes #4136

## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/main/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[X] Other... Please describe:
```
API change - exports new type

## What is the current behavior?
No export

Closes #4136 

## What is the new behavior?
Exports `TypedAction`

## Does this PR introduce a breaking change?

```
[ ] Yes
[X] No
```

PR #4318, opened 2024-05-07
feat(store): merge Action and TypedAction intefaces

### LINKED ISSUE #4136: Can TypedAction be exported from @ngrx/store?

store

### Information

`@ngrx/store` exports `Action` but not `TypedAction` from:
https://github.com/ngrx/platform/blob/b5385dbe243c2f2025682290e194124076be086f/modules/store/src/models.ts#L8

There is no documentation why `TypedAction` should be internal only.

There are several projects that import it by reaching into the source file:
* https://github.com/gabrielguerrero/ngrx-traits/blob/9b1ba42120b4856123b57af9856f4482a1273f30/libs/ngrx-traits/common/src/lib/async-action/async-action.model.ts#L3
* https://github.com/flyflex/ngflex-libs/blob/5b0569a21cd6d71bf7a6f943665df43334a22030/packages/ngrx-firebase/lib/operators/map-to-actions.operator.ts#L1
* https://github.com/VladiRR/svvs/blob/554732f32e7981b3962d51949bbf16f617ebfdd9/libs/shared/utils/store/src/lib/effects/abstract-effects.ts#L2

_No response_

- [ ] Yes
- [ ] No

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4136
Closes #4183

```
[x] Yes
[ ] No
```

```
BREAKING CHANGE:

The Action and TypedAction interfaces are merged into one interface.

BEFORE:

There was a separation between the Action and TypedAction interfaces.

AFTER:

The Action interface accepts a generic type parameter that represents the payload type (defaults to string).
The TypedAction interface is removed.
```

### DIFF
--- a/modules/effects/spec/types/effect_creator.spec.ts
+++ b/modules/effects/spec/types/effect_creator.spec.ts
@@ -22,7 +22,7 @@ describe('createEffect()', () => {
       expectSnippet(`
         const effect = createEffect(() => of({ foo: 'a' }));
       `).toFail(
-        /Type 'Observable<{ foo: string; }>' is not assignable to type 'EffectResult<Action>'./
+        /Type 'Observable<{ foo: string; }>' is not assignable to type 'EffectResult<Action<string>>'./
       );
     });
 
@@ -50,7 +50,7 @@ describe('createEffect()', () => {
       expectSnippet(`
         const effect = createEffect(() => of({ foo: 'a' }), { dispatch: true });
       `).toFail(
-        /Type 'Observable<{ foo: string; }>' is not assignable to type 'EffectResult<Action>'./
+        /Type 'Observable<{ foo: string; }>' is not assignable to type 'EffectResult<Action<string>>'./
       );
     });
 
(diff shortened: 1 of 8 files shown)


markostanimirovic (approved):

brandonroberts:
I like this change. I also think its going to be a big breaking change for people who have reached into our `src/models` directory for that interface that will no longer be there. We've always declared that those are not public API and could be broken without warning, but just something we may want to note in the migration guide specifically.

brandonroberts (approved):

trevorade:
> I like this change. I also think its going to be a big breaking change for people who have reached into our `src/models` directory for that interface that will no longer be there. We've always declared that those are not public API and could be broken without warning, but just something we may want to note in the migration guide specifically.

Could NgRx be structured in such a way such that only the index.ts files are accessible to remove the possibility of reaching into the implementation files?

brandonroberts:
> > I like this change. I also think its going to be a big breaking change for people who have reached into our `src/models` directory for that interface that will no longer be there. We've always declared that those are not public API and could be broken without warning, but just something we may want to note in the migration guide specifically.
> 
> Could NgRx be structured in such a way such that only the index.ts files are accessible to remove the possibility of reaching into the implementation files?

We're already structured that way. We only expose code through the public API, but the types are included as deep imports. This is how the Angular library build process works with ng-packagr and not something custom we've set up.

PR #4404, opened 2024-06-17
fix(signals): allow modifying entity id on update

### LINKED ISSUE #4235: updateEntity method don't update key in entityMap

entity, signals

```typescript
interface Todo {
  id: string;
  name: string;
}

const TodoStore = signalStore(withEntities<Todo>());

// Add model with temp id to store
patchState(TodoStore, addEntity({id: 'tmp_id', name: 'some_name'}))

/**
 * At whis point we have TodoStore.entities() = [{id: 'tmp_id', name: 'some_name'}]
 * and TodoStore.entityMap() = { 'tmp_id':  {id: 'tmp_id', name: 'some_name'}}
 */

// Then i wants to update entity with real data, for example from API
patchState(TodoStore, updateEntity({id: 'tmp_id', changes: {id: 'real_uuid_from_api', name: 'some_real_name'}}))

/**
 * Now we have TodoStore.entities() = [{id: 'real_uuid_from_api', name: 'some_real_name'}]
 * ❗and TodoStore.entityMap() = { 'tmp_id':  {id: 'real_uuid_from_api', name: 'some_real_name'}}❗
 */

// Then i want's to delete this entity, but nothing happens and state of entities don't changed
patchState(TodoStore, removeEntity('real_uuid_from_api'))

/**
 * I think, this because of removeEntity teying to find id in entityMap
 * but entityMap was not changed bu updateEntity method
 */
```

### Expected behavior

If i call `updateEntity` method with changed entity ID, `entityMap` should be updated too

Against we can't remove model by new setted ID

"@ngrx/signals": "^17.1.0",
"@angular/core": "~17.0.8",
Node: v21.4.0
OS: Win 11
Browser: MS Edge

### Other information

_No response_

- [ ] Yes
- [X] No

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

It's not possible to modify entity ID via `update*` functions. Instead, it's necessary to remove the existing one and add a new entity with an updated ID:

```ts
const myEntity = { ...entityMap()['tmp_id'], id: 'real_id' };

patchState(store, removeEntity('tmp_id'), addEntity(myEntity));
```

Closes #4235 

## What is the new behavior?

The `update*` updaters can be used to modify entity ID:

```ts
patchState(store, updateEntity({ id: 'tmp_id', changes: { id: 'real_id' } }));
```

```
[ ] Yes
[x] No
```

## Other information

After this change, it's necessary to provide `selectId` to all `update*` updaters if an entity has a custom ID.

```diff
type Todo = { _id: string; text: string };

+const selectId: SelectEntityId<Todo> = (todo) => todo._id;

const TodosStore = signalStore(
  withEntities<Todo>(),
  withMethods((store) => ({
    updateTodo(id: string, changes: Partial<Todo>): void {
-      patchState(store, updateEntity({ id, changes }));
+      patchState(store, updateEntity({ id, changes }, { selectId }));
    },
  }))
);
```

This is not considered as a breaking change because the `@ngrx/signals` package is in developer preview.

### DIFF
--- a/modules/signals/entities/src/helpers.ts
+++ b/modules/signals/entities/src/helpers.ts
@@ -7,6 +7,7 @@ import {
   SelectEntityId,
 } from './models';
 
+declare const ngDevMode: unknown;
 const defaultSelectId: SelectEntityId<{ id: EntityId }> = (entity) => entity.id;
 
 export function getEntityIdSelector(config?: {
@@ -166,11 +167,13 @@ export function removeEntitiesMutably(
 export function updateEntitiesMutably(
   state: EntityState<any>,
   idsOrPredicate: EntityId[] | EntityPredicate<any>,
-  changes: EntityChanges<any>
+  changes: EntityChanges<any>,
+  selectId: SelectEntityId<any>
 ): DidMutate {
   const ids = Array.isArray(idsOrPredicate)
     ? idsOrPredicate
     : state.ids.filter((id) => idsOrPredicate(state.entityMap[id]));
+  let newIds: Record<EntityId, EntityId> | undefined = undefined;
   let didMutate = DidMutate.None;
 
   for (const id of ids) {
@@ -181,7 +184,31 @@ export function updateEntitiesMutably(
         typeof changes === 'function' ? changes(entity) : changes;
       state.entityMap[id] = { ...entity, ...changesRecord };
       didMutate = DidMutate.Entities;
+
+      const newId = selectId(state.entityMap[id]);
+      if (newId !== id) {
+        state.entityMap[newId] = state.entityMap[id];
+        delete state.entityMap[id];
+
+        newIds = newIds || {};
+        newIds[id] = newId;
+      }
     }
+  }
+
+  if (newIds) {
+    state.ids = state.ids.map((id) => newIds[id] ?? id);
+    didMutate = DidMutate.Both;
+  }
+
+  if (ngDevMode && state.ids.length !== Object.keys(state.entityMap).length) {
+    console.warn(
+      '@ngrx/signals/entities: Entities with IDs:',
+      ids,
+      'are not updated correctly.',
+      'Make sure to apply valid changes when using `updateEntity`,',
+      '`updateEntities`, and `updateAllEntities` updaters.'
+    );
   }
 
   return didMutate;
--- a/modules/signals/entities/spec/updaters/update-all-entities.spec.ts
+++ b/modules/signals/entities/spec/updaters/update-all-entities.spec.ts
@@ -51,8 +51,10 @@ describe('updateAllEntities', () => {
 
     patchState(
       store,
-      updateAllEntities({ text: '' }),
-      updateAllEntities((todo) => ({ completed: !todo.completed }))
+      updateAllEntities({ text: '' }, { selectId: (todo) => todo._id }),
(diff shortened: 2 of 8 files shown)


timdeschryver (commented):
With the current implementation, the ids can get out of sync with the entity keys. This happens when we update an entity with an existing id.
Is this expected? If so, I think adding a warning can be useful.

```ts
  it('updates entity ids', () => {
    const Store = signalStore(withEntities<User>());
    const store = new Store();

    patchState(
      store,
      addEntities([user1, user2]),
      updateEntities({
        ids: [user1.id],
        changes: ({ id }) => ({ id: user2.id, firstName: `Updated Jimmy` }),
      }),
    );

    // 👇 only has user 2
    expect(store.entityMap()).toEqual({
      [user2.id]: { ...user1, id: user2.id, firstName: 'Updated Jimmy' },
    });

    // 👇 user id 2 is added twice
    expect(store.ids()).toEqual([user2.id, user2.id]);
  });
```

markostanimirovic:
@timdeschryver With allowing ID changes there are several ways to break/move entity collection to the inconsistent state. That's the reason why I was against this change initially. Users now have better control, but also responsibility.

Let's check how `@ngrx/entity` handles this case. It probably works the same.

I agree that warning will be useful 👍

markostanimirovic:
I just checked `@ngrx/entity` - the behavior is the same. Let's add a warning then.

modules/signals/entities/src/helpers.ts
```diff
@@ -7,6 +7,7 @@ import {
   SelectEntityId,
 } from './models';
 
+declare const ngDevMode: unknown;
```
markostanimirovic, line 10: 1) There is an option to avoid declaring `ngDevMode` in each file where it's needed by using:

```ts
declare global {
  declare const ngDevMode: unknown;
}
```

in a single place. However, in that case, the `ngDevMode` global variable will be available in applications that use the `@ngrx/signals` package.

2) I also tried to define `ngDevMode` in a separate file and export it as renamed `ngrxDevMode`:

```ts
// file: dev-mode.ts
declare const ngDevMode: unknown;
export const ngrxDevMode = ngDevMode;

// file: helpers.ts
import { ngrxDevMode } from './dev-mode';


if (ngrxDevMode) {
  // ...
}
```

However, tree-shaking will not work in apps if `ngDevMode` is renamed. cc @brandonroberts

brandonroberts: Good catch. We'll go with it as-is for now and adjust if they change it in the framework 👍

modules/signals/entities/src/helpers.ts
```diff
@@ -181,8 +184,32 @@ export function updateEntitiesMutably(
...
+      }
     }
   }
 
+  if (newIds) {
+    state.ids = state.ids.map((id) => newIds[id] || id);
+    didMutate = DidMutate.Both;
+  }
+
+  if (ngDevMode && state.ids.length !== Object.keys(state.entityMap).length) {
```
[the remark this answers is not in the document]
timdeschryver: Thanks @markostanimirovic !

timdeschryver (approved):

brandonroberts (approved):

PR #4418, opened 2024-07-04
fix(eslint-plugin): do not report non-array returns in no-multiple-actions-in-effects

jsaguet:
…tions-in-effects

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Eslint rules incorrectly reports an error for the rule no-multiple-actions-in-effects for the following code:
```typescript
import { Actions, createEffect, ofType } from '@ngrx/effects';
import { createAction, props } from '@ngrx/store';
import { of, switchMap } from 'rxjs';

const foo = createAction('foo', props<{ payload: string }>());
const bar = createAction('bar');
const baz = createAction('baz');

@Injectable()
export class Effects {
  constructor(private actions$: Actions) {}
  effect$ = createEffect(() =>
      this.actions$.pipe(
          ofType(foo),
          switchMap(({ payload }: { payload: string }) => (payload === 'value' ? of(bar()) : of(baz()))),
      ),
  );
}

```

Closes #

## What is the new behavior?
The eslint rule does not report an error on the valid code.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/eslint-plugin/src/rules/effects/no-multiple-actions-in-effects.ts
+++ b/modules/eslint-plugin/src/rules/effects/no-multiple-actions-in-effects.ts
@@ -58,7 +58,7 @@ export default createRule<Options, MessageIds>({
           });
         } else if (
           type.isUnion() &&
-          type.types.some((ut) => !typeChecker.isArrayType(ut))
+          type.types.some((ut) => typeChecker.isArrayType(ut))
         ) {
           context.report({
             node: nodeToReport,
--- a/modules/eslint-plugin/spec/rules/effects/no-multiple-actions-in-effects.spec.ts
+++ b/modules/eslint-plugin/spec/rules/effects/no-multiple-actions-in-effects.spec.ts
@@ -94,6 +94,24 @@ export const saveSearchCriteria$ = createEffect(
   { functional: true }
 );
   `,
+  `
+import { Actions, createEffect, ofType } from '@ngrx/effects';
+import { createAction, props } from '@ngrx/store';
+import { of, switchMap } from 'rxjs';
+const foo = createAction('foo', props<{ payload: string }>());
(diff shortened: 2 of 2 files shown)


modules/eslint-plugin/src/rules/effects/no-multiple-actions-in-effects.ts
```diff
@@ -58,7 +58,7 @@ export default createRule<Options, MessageIds>({
           });
         } else if (
           type.isUnion() &&
-          type.types.some((ut) => !typeChecker.isArrayType(ut))
+          type.types.some((ut) => typeChecker.isArrayType(ut))
```
timdeschryver, line 61: Good catch!
Could you also add an invalid code snippet please?

jsaguet: Isn't the invalid code snippet already handled by cases like this one:
```typescript
import { of } from 'rxjs'
@Injectable()
export class Effects {
  readonly effectNOK3$ = createEffect(() =>
    this.actions$.pipe(concatMapTo(condition ? [foo(), bar()] : of(foo())))
                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  [${messageId}]
  )
}
```

timdeschryver: You're right! Sorry, I was only looking at the changes, not the already existing cases.

timdeschryver (approved):

markostanimirovic (approved):

markostanimirovic:
Thank you @jsaguet!

PR #4423, opened 2024-07-14
Entity: Strengthen typing of getInitialState

### LINKED ISSUE #4422: NGRX Entity, Stronger Typing on getInitialState

entity

### Information

Hi NGRX Team, 

When using NGRX Entity package and setting the initial state, unfortunately it is possible to introduce unknown properties.

I'd like to show briefly how I am solving this and ask if it's possible for you to implement this somehow for all to enjoy? Thanks in advance:

Firstly, in my reducers, consider this simplified initial Entity State, with an irrelevant entity type of "ExampleEntity":

```
export interface State extends EntityState<ExampleEntity> {
  selectedId?: string;
  loaded: boolean;
  error?: HttpErrorResponse;
  searchText: string;
}
```

when calling 

```
myAdapter.getInitialState({ 
  loaded: false,
  searchText: '',
  unkownProperty: "test" <--- This does NOT throw an error as I expect (since it is not a part of ExampleEntity)
})
```

as recommended in the documentation, unfortunately it is possible to introduce unknown properties without any error, such as the "unkownProperty" shown above. 

To resolve this, I have type casted typing of getInitialState:

```
export const initialState: State = myAdapter.getInitialState<Omit<State, keyof EntityState<ExampleEntity>>>({
  loaded: false,
  searchText: '',
  unkownProperty: "test" <--- This will now throw an error
});
```

If you have the time, perhaps you could introduce some more strict typing similar to the above to allow for IDE's to catch these sorts of errors without casting? Thanks very much for your time.

export const initialState: State = myAdapter.getInitialState<Omit<State, keyof EntityState<ExampleEntity>>>({
  loaded: false,
  searchText: '',
  unkownProperty: "test" <--- This will now throw an error
});

Using TypeScript's "Omit" as show above allows for this to work.

- [X] Yes
- [ ] No

GavynHolt:
## PR Checklist

## PR Type

```
[] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When initializing an `EntityAdapter<T>` and subsequently setting the `initialState` using `...adapter.getInitialState({ <additional state here>})`, it is possible, given the current interface, to introduce unknown variables that do not conform to the user's specified state. This could result in bugs related to mistyped variable names, resulting in confusing user error scenarios.

Closes #4422

## What is the new behavior?

This small change modifies the interface to require that the input state `S` extends `EntityState<T>`, requires that the additionalState `state` argument conforms to an Omit version of this `S` type, which removes the 'id' and 'entities' properties by way of `keyof EntityState<T>`, and then finally, since the argument now already extends `EntityState<T>`, we can now simply return `S`.

```
[x] Yes
[ ] No
```

I have answered "Yes", but realistically, if a user of the Entity package has entered in correct code already, there will be no breaking changes. Any breaking changes should be in the form of a user now being aware of incompatible property types in their additional state, or if their State interface does not contain `extends EntityState<T>`.

## Other information

This change has helped my place of employment from unknown state properties quite nicely, but I must say this is my very first community/open source pull request. I would be happy to take any constructive criticism or guidance on anything that I could do better here. Very excited to be contributing and hope this is of help. 

Thank you, -Gavyn Holt

### DIFF
--- a/modules/entity/src/entity_state.ts
+++ b/modules/entity/src/entity_state.ts
@@ -9,9 +9,9 @@ export function getInitialEntityState<V>(): EntityState<V> {
 
 export function createInitialStateFactory<V>() {
   function getInitialState(): EntityState<V>;
-  function getInitialState<S extends object>(
-    additionalState: S
-  ): EntityState<V> & S;
+  function getInitialState<S extends EntityState<V>>(
+    additionalState: Omit<S, keyof EntityState<V>>
+  ): S;
   function getInitialState(additionalState: any = {}): any {
     return Object.assign(getInitialEntityState(), additionalState);
   }
--- a/modules/entity/src/models.ts
+++ b/modules/entity/src/models.ts
@@ -110,7 +110,9 @@ export interface EntityAdapter<T> extends EntityStateAdapter<T> {
   selectId: IdSelector<T>;
   sortComparer: false | Comparer<T>;
   getInitialState(): EntityState<T>;
-  getInitialState<S extends object>(state: S): EntityState<T> & S;
+  getInitialState<S extends EntityState<T>>(
+    state: Omit<S, keyof EntityState<T>>
+  ): S;
   getSelectors(): EntitySelectors<T, EntityState<T>>;
   getSelectors<V>(
     selectState: (state: V) => EntityState<T>


timdeschryver (commented):
Thanks for the PR @GavynHolt .
Imho this is a useful change, but we cannot merge this now because this is a breaking change. I'll leave this open as we can revisit this on our next major version release.

timdeschryver (changes requested):
Wanted to mention the above comment as a request for changes to prevent an accidental merge.

brandonroberts:
I think we should close this PR and revisit closer to the v19 release cycle for breaking changes

timdeschryver:
> I think we should close this PR and revisit closer to the v19 release cycle for breaking changes

Sounds good!

PR #4424, opened 2024-07-14
feat(signals): prevent overriding of SignalStore members

### LINKED ISSUE #4144: [signals] Prevent overriding state fields with a computed signal of the same name

signals

I'm not sure if this is intentional or not: it's possible to define a computed field with the same name as a field already defined in the initial state.

**Reproduction:**

https://stackblitz.com/edit/stackblitz-starters-atpbi6?file=src%2Fmain.ts

For posterity, here's the store definition used in the StackBlitz above:

```ts
const Store = signalStore(
  { providedIn: 'root' },
  withState({ value: 1 }),
  withComputed((store) => ({
    value: computed(() => store.value() * 2),
  }))
);
```

This may lead to unintentional bugs.

### Expected behavior

You should get an error if you try to define a computed field with same name as an existing field in the initial state.

Angular: v17
NgRx: v17.0.0

### Other information

_No response_

- [X] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4144

It's allowed to override SignalStore members:

```ts
const DemoStore = signalStore({
  withState({ foo: 'bar' }),
  withMethods(({ foo }) => ({ // `foo` type: Signal<string>
    foo() {
      console.log(foo());
    }
  }),
  withHooks({
    onInit({ foo }) { // `foo` type: () => void
      foo();
    },
  })
});
```

## What is the new behavior?

- The warning will be displayed in development mode if a user tries to override SignalStore members.
- `StateSignal` is renamed to `StateSource` for clarity - avoiding confusion between `StateSignal` and `SignalState`.

```
[ ] Yes
[x] No
```

## Other information

After this change, signal stores that rely on overriding have to be changed. Also, if `StateSignal` is used, it should be replaced with `StateSource`:

```diff
-import { patchState, StateSignal } from '@ngrx/signals';
+import { patchState, StateSource } from '@ngrx/signals';

-resetCount(state: StateSignal<{ count: number }>): void {
+resetCount(state: StateSource<{ count: number }>): void {
  patchState(state, { count: 0 });
}
```

This is not considered as a breaking change because the `@ngrx/signals` package is in developer preview.


modules/signals/src/signal-store-assertions.ts
```diff
@@ -0,0 +1,29 @@
+import { InnerSignalStore } from './signal-store-models';
+
+declare const ngDevMode: unknown;
+
+export function assertUniqueStoreMembers(
+  store: InnerSignalStore,
+  newMemberKeys: string[]
+): void {
+  if (!ngDevMode) {
```
brandonroberts, line 9: Can we wrap this whole check inside the `ngDevMode` block so it gets tree-shaken during the prod build?

brandonroberts (changes requested):

modules/signals/src/signal-store-assertions.ts
markostanimirovic: Tree-shaking works with existing implementation as well. Tested on `standalone-app`.

brandonroberts (approved):

modules/signals/src/index.ts
```diff
@@ -1,11 +1,14 @@
...
+export { signalState, SignalState } from './signal-state';
 export { signalStore } from './signal-store';
 export { signalStoreFeature, type } from './signal-store-feature';
 export { SignalStoreFeature } from './signal-store-models';
-export { StateSignal } from './state-signal';
+export {
+  getState,
+  PartialStateUpdater,
+  patchState,
+  StateSource,
```
timdeschryver, line 10: Unrelated to the PR, but do we want to have migration in place to update StateSignal to StateSource ?

timdeschryver (approved):

modules/signals/src/index.ts
markostanimirovic: yes, that would be great 👍

PR #4492, opened 2024-08-14
ci: increase test timeout

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[ ] No
```

## Other information

All of our builds are suddenly failing.
After some research it seems due to a timeout during the CI build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,6 +82,7 @@ jobs:
       - run:
           name: Run All Unit Tests
           command: yarn nx run-many -t test
+          no_output_timeout: 20m
 
   test-affected:
     <<: *run_in_browser
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ version: 2.1
 # See https://blog.daemonl.com/2016/02/yaml.html
 # To validate changes, use an online parser, eg.
 # https://yaml-online-parser.appspot.com/
-var_1: &cache_key yarn-cache-{{ checksum "yarn.lock" }}-0.18.0
+var_1: &cache_key yarn-cache-{{ checksum "yarn.lock" }}-0.18.1
 var_2: &run_in_node
   docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -82,6 +82,7 @@ jobs:
       - run:
           name: Run All Unit Tests
           command: yarn nx run-many -t test
+          no_output_timeout: 20m
```
timdeschryver, line 85: The default is 10 minutes, and it seems like we're just surpassing it with 10m 25s...

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ version: 2.1
 # See https://blog.daemonl.com/2016/02/yaml.html
 # To validate changes, use an online parser, eg.
 # https://yaml-online-parser.appspot.com/
-var_1: &cache_key yarn-cache-{{ checksum "yarn.lock" }}-0.18.0
+var_1: &cache_key yarn-cache-{{ checksum "yarn.lock" }}-0.18.1
```
rainerhahnekamp, line 13: Out of interest, why do we have the suffix 0.18.1? Is the checksum not already enough?

.circleci/config.yml
rainerhahnekamp: Yeah, and we already see that the build took more than twelve minutes and was successful. So it works. 👍

.circleci/config.yml
timdeschryver: The checksum should be enough, and was in this case not the culprit.
In the past we've had some experiences where the cache could go corrupt causing the CI build to act funky. That's why my first idea was to create a new cache.

rainerhahnekamp (approved):

markostanimirovic (approved):

PR #4519, opened 2024-09-07
feat(signals): set state in `patchState` to readonly

### LINKED ISSUE #4030: @ngrx/signals: Improve Developer Experience for Immutability

store

### Information

Ensuring that state updates are immutable is crucial for Signal to emit correctly and trigger necessary DOM updates, derived signals, or side effects. Currently, an immutable runtime check is available, but only in development mode due to its potential performance impact.

To enhance the robustness of applications and facilitate debugging, I propose two additional behaviours:

- Option to enable the immutable check in production mode: There are many Angular applications where this “safety net” outweighs performance penalties by far. Especially “classic enterprise applications” containing many forms where the business logic runs in the backend.
- $update with Readonly Types: Compilation checks are highly preferable to runtime checks because they allow potential issues to be discovered before the runtime. The $update method could be enhanced to provide the state as a recursive Readonly<T> type to achieve this. This would extend to arrays as ReadonlyArray<T>, removing mutable methods like push.

__Example__:

```typescript
import { signalStore, withMethods, withState } from '@ngrx/signals';
type User = { id: number; name: string };
type UsersState = { users: User[]; loading: boolean };

declare function withImmutabilityCheck(mode: 'disabled' | 'devMode' | 'always');

const UsersStore = signalStore(
  withImmutabilityCheck('always'), // <— immutable check
  withState<UsersState>({ users: [], loading: false }),
  withMethods(({ $update, users }) => {
    return {
      addUser(user: User): void {
        $update((state) => {
          state.users.push(user); // 💥compilation error because ReadonlyArray<User>
          state.loading = true; //  💥compilation error because Readonly<UsersState>

          return state;
        });
      },
    };
  })
);
```

If accepted, I would be happy to provide an initial PR.

_No response_

- [X] Yes
- [ ] No

rainerhahnekamp:
As discussed in #4030, the update function of `patchState`, provides the state as deep readonly.

Notes on the test: I've decided to go with `@ts-expect-error` instead of ts-snippet. ts-snippet requires the source code as string, whereas @ts-expect-error can be applied to normal code. This makes it easier to write but also to refactor.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4030 

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/src/ts-helpers.ts
+++ b/modules/signals/src/ts-helpers.ts
@@ -27,3 +27,11 @@ export type IsKnownRecord<T> = IsRecord<T> extends true
 export type OmitPrivate<T> = {
   [K in keyof T as K extends `_${string}` ? never : K]: T[K];
 };
+
+export type DeepReadonly<T> = T extends object
+  ? {
+      readonly [P in keyof T]: DeepReadonly<T[P]>;
+    }
+  : T extends []
+  ? ReadonlyArray<DeepReadonly<T>>
+  : T;
--- a/modules/signals/spec/types/patch-state.types.spec.ts
+++ b/modules/signals/spec/types/patch-state.types.spec.ts
@@ -1,5 +1,6 @@
 import { expecter } from 'ts-snippet';
 import { compilerOptions } from './helpers';
+import { patchState, signalState } from '@ngrx/signals';
 
 describe('patchState', () => {
   const expectSnippet = expecter(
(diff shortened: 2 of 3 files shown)


modules/signals/src/ts-helpers.ts
```diff
@@ -27,3 +27,11 @@ export type IsKnownRecord<T> = IsRecord<T> extends true
 export type OmitPrivate<T> = {
   [K in keyof T as K extends `_${string}` ? never : K]: T[K];
 };
+
+export type DeepReadonly<T> = T extends object
```
jits, line 31: The [`type-fest`](https://github.com/sindresorhus/type-fest) library has a more comprehensive version of this: https://github.com/sindresorhus/type-fest/blob/main/source/readonly-deep.d.ts. I can't tell the differences off the top of my head but thought I'd mention it.

(Note: I'm not advocating using the `type-fest` library as a dependency — more providing this in case of inspiration for some good 'ol copy-pasting :))

rainerhahnekamp: Hey @jits, I took a look at it before. It supports a lot of types that are not allowed to be part of the state object. So I took arrays, primitive types and nested objects out of it.

jits: Hi @rainerhahnekamp — ahh, perfect! Thanks for clarifying. (And that makes a lot of sense)

markostanimirovic:
I suggest keeping types as they are and displaying a warning in dev mode if the state is mutated.

rainerhahnekamp:
Withdrawed in favor of #4526

PR #4526, opened 2024-09-18
feat(signals): throw error in dev mode on state mutation

### LINKED ISSUE #4030: @ngrx/signals: Improve Developer Experience for Immutability

rainerhahnekamp:
Alternative version to protected the state from mutable changes via `Object.freeze`.

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Mutable changes in `patchState` don't trigger any kind of warning

Closes #4030

## What is the new behavior?

In development mode (`ngDevMode`), `Object.freeze` is applied to the state, causing a runtime error on a mutable change.

```
[x] Yes
[ ] No
```

```
BREAKING CHANGES:

The `patchState` function applies a deep freeze on the state in dev mode.

BEFORE:

const userState = signalState(initialState);
patchState(userState, (state) => {
  state.user.firstName = 'mutable change'; // mutable change which went through
  return state;
});

AFTER:

const userState = signalState(initialState);
patchState(userState, (state) => {
  state.user.firstName = 'mutable change'; // throws in dev mode
  return state;
});

```

### DIFF
--- a/modules/signals/src/state-source.ts
+++ b/modules/signals/src/state-source.ts
@@ -8,6 +8,7 @@ import {
   WritableSignal,
 } from '@angular/core';
 import { Prettify } from './ts-helpers';
+import { freezeInDevMode } from './deep-freeze';
 
 const STATE_WATCHERS = new WeakMap<Signal<object>, Array<StateWatcher<any>>>();
 
@@ -37,10 +38,11 @@ export function patchState<State extends object>(
 ): void {
   stateSource[STATE_SOURCE].update((currentState) =>
     updaters.reduce(
-      (nextState: State, updater) => ({
-        ...nextState,
-        ...(typeof updater === 'function' ? updater(nextState) : updater),
-      }),
+      (nextState: State, updater) =>
+        freezeInDevMode({
+          ...nextState,
+          ...(typeof updater === 'function' ? updater(nextState) : updater),
+        }),
       currentState
     )
   );
--- a/modules/signals/spec/deep-freeze.spec.ts
+++ b/modules/signals/spec/deep-freeze.spec.ts
@@ -1,0 +1,115 @@
+import { getState, patchState } from '../src/state-source';
+import { signalState } from '../src/signal-state';
+import { signalStore } from '../src/signal-store';
+import { TestBed } from '@angular/core/testing';
+import { withState } from '../src/with-state';
+
+describe('deepFreeze', () => {
+  const initialState = {
+    user: {
+      firstName: 'John',
+      lastName: 'Smith',
+    },
+    foo: 'bar',
+    numbers: [1, 2, 3],
+    ngrx: 'signals',
+  };
+
+  for (const { stateFactory, name } of [
+    {
+      name: 'signalStore',
+      stateFactory: () => {
+        const Store = signalStore(
+          { protectedState: false },
+          withState(initialState)
+        );
+        return TestBed.configureTestingModule({ providers: [Store] }).inject(
+          Store
+        );
+      },
+    },
+    { name: 'signalState', stateFactory: () => signalState(initialState) },
+  ]) {
+    describe(name, () => {
+      it(`throws on a mutable change`, () => {
+        const state = stateFactory();
+        expect(() =>
+          patchState(state, (state) => {
+            state.ngrx = 'mutable change';
+            return state;
+          })
+        ).toThrowError("Cannot assign to read only property 'ngrx' of object");
+      });
+
+      it('throws on a nested mutable change', () => {
+        const state = stateFactory();
+        expect(() =>
+          patchState(state, (state) => {
+            state.user.firstName = 'mutable change';
+            return state;
+          })
+        ).toThrowError(
+          "Cannot assign to read only property 'firstName' of object"
+        );
+      });
+      describe('mutable changes outside of patchState', () => {
+        it('throws on reassigned a property of the exposed state', () => {
+          const state = stateFactory();
+          expect(() => {
+            state.user().firstName = 'mutable change 1';
+          }).toThrowError(
+            "Cannot assign to read only property 'firstName' of object"
+          );
+        });
+
+        it('throws when exposed state via getState is mutated', () => {
+          const state = stateFactory();
+          const s = getState(state);
+
+          expect(() => (s.ngrx = 'mutable change 2')).toThrowError(
+            "Cannot assign to read only property 'ngrx' of object"
+          );
+        });
+
+        it('throws when mutable change happens for', () => {
+          const state = stateFactory();
+          const s = { user: { firstName: 'M', lastName: 'S' } };
+          patchState(state, s);
+
+          expect(() => {
+            s.user.firstName = 'mutable change 3';
+          }).toThrowError(
+            "Cannot assign to read only property 'firstName' of object"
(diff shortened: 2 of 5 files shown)


modules/signals/src/state-source.ts
```diff
@@ -49,6 +53,14 @@ export function patchState<State extends object>(
   notifyWatchers(stateSource);
 }
 
+function freezeInDevMode<State extends object>(value: State): State {
+  if (ngDevMode) {
+    Object.freeze(value);
```
markostanimirovic, line 58: To ensure that nested state properties are also not mutated, we can deeply freeze the state object.

rainerhahnekamp: I've copied and slightly modified the existing freeze function of the global store.

modules/signals/src/state-source.ts
```diff
@@ -49,6 +54,10 @@ export function patchState<State extends object>(
   notifyWatchers(stateSource);
 }
 
+function freezeInDevMode<State extends object>(value: State): State {
+  return ngDevMode ? deepFreeze(value) : value;
```
timdeschryver, line 58: I think we should add this to the v19 list as a breaking change.
I suppose this could break some applications in dev mode.

rainerhahnekamp: Fair enough. It really throws an error and I couldn't find a way how to "map" it to a warning message.

rainerhahnekamp:
@samuelfernandez 

> Any reasons for not returning a Readonly type

Yes, experience. It has shown that if you start to bend the type system too much, you might end up in some unpredictable issues.

For example, if people provide some type, which could be any or maybe a very complicated type that we cannot predict, then we might run into issues where suddenly it stops working.

That's the reasoning behind this decision.

rainerhahnekamp:
@samuelfernandez I think this is an issue for Angular. `patchState` is an NgRx feature, so we can improve that one. computed() and the Signal type comes from the framework itself.

brandonroberts (approved):
We should add a note to the docs on this behavior

modules/signals/src/state-source.ts
```diff
@@ -49,6 +54,10 @@ export function patchState<State extends object>(
...
+}
+
```
markostanimirovic, line 60: `freezeInDevMode` (and `ngDevMode`) can be moved to the `deep-freeze.ts` file as well. Since it's not directly related to the state object, the generic can be named `T`:

```suggestion
function freezeInDevMode<T>(target: T): T {
  return ngDevMode ? deepFreeze(target) : target;
}
```

modules/signals/spec/signal-state.spec.ts
```diff
@@ -194,4 +194,28 @@ describe('signalState', () => {
...
+      expect(() =>
+        patchState(userState, (state) => {
+          state.user.firstName = 'mutable change';
+          return state;
+        })
+      ).toThrowError(
+        "Cannot assign to read only property 'firstName' of object"
+      );
+    });
+  });
```
markostanimirovic, line 220: I suggest moving these tests to `deep-freeze.spec.ts` and checking the behavior for both - `signalState` and `signalStore`. Take a look at these tests for inspiration: https://github.com/ngrx/platform/blob/main/modules/signals/spec/state-source.spec.ts#L49

Also, in the following cases, error should be thrown as well:

```ts
userState.user().firstName =  'mutable change 1'; // error

// ---

getState(userState).ngrx = 'mutable change 2'; // error

// ---

const s = { user: { firstName: 'M', lastName: 'S' } };
patchState(userState, s);
s.user.firstName = 'mutable change 3'; // error
```

markostanimirovic (changes requested):

markostanimirovic:
To mention new behavior, we can update this alert with an additional sentence that an error will be thrown in dev mode on state mutations: https://github.com/ngrx/platform/blob/main/projects/ngrx.io/content/guide/signals/signal-state.md?plain=1#L59

markostanimirovic:
Since this PR introduces a breaking change, we should add this to the PR description in the following format:

```
BREAKING CHANGES:

__description__

BEFORE:

__...__

AFTER:

__...__
```

Check this PR for more info: https://github.com/ngrx/platform/pull/4584

rainerhahnekamp:
@markostanimirovic, all your requests have been incorporated:

- Breaking Changes in PR docs
- consolidation of methods in `deep-freeze.ts``
- own testing file for `deep-freeze.ts`
- type-renaming in `freezeInDevMode()`

markostanimirovic:
@rainerhahnekamp recheck this comment: https://github.com/ngrx/platform/pull/4526#discussion_r1838866481

Currently, the error will be thrown only if the state is mutated within the `patchState` updater. It would be good to do the full state freeze:

> Also, in the following cases, errors should be thrown as well:
>
> userState.user().firstName =  'mutable change 1'; // error
> 
> // ---
>
> getState(userState).ngrx = 'mutable change 2'; // error
>
> // ---
>
> const s = { user: { firstName: 'M', lastName: 'S' } };
> patchState(userState, s);
> s.user.firstName = 'mutable change 3'; // error
>
> Currently, there are no errors in these cases.

markostanimirovic:
@rainerhahnekamp It will also be useful to update documentation: https://github.com/ngrx/platform/pull/4526#issuecomment-2471739661

rainerhahnekamp:
@markostanimirovic, as requested, the state itself is now frozen, not just the changes coming from `patchState`.
Thanks for pointing it out. This feature is became now much better.

I'll update the PR about the documentation as well.

Commits have been squashed into a "conventional commit"-friendly message.

markostanimirovic (approved):
LGTM 👍 Thanks Rainer!

PR #4529, opened 2024-09-22
fix(signals): use `Injector` of `rxMethod` instance caller if available

### LINKED ISSUE #4528: @ngrx/signals: Potential memory leak in `rxMethod`

signals

Any component that passes on a Signal to an `rxMethod` of a service provided in root, will cause a memory leak, because the rxMethod's `effect` is using the `RootInjector`:

https://stackblitz.com/edit/github-kvz3tp?file=src%2Fmain.ts

### Expected behavior

If a component passes a Signal to an `rxMethod` instance that has been created in the root injector, the internal effect should run with the component's injector.

Latest versions

### Other information

_No response_

- [X] Yes
- [ ] No

rainerhahnekamp:
Ensure that an instance of `rxMethod` uses the `Injector` of the caller where the instance was invoked, rather than the injector of the node where `rxMethod` was created.

This change addresses a potential memory leak where a `SignalStore` method, generated by `rxMethod` and provided at the root level, continues to track a signal via its `effect` using the `RootInjector`. In cases where the component calling the method is destroyed, the `Signal` remains tracked, leading to unintended behavior and memory retention.

With this commit, `rxMethod` now prioritizes the injector of the instance caller, falling back to the creator's injector only if no injection context exists. Additionally, a custom injector can be provided by the caller and will take precedence when specified.

For `Observable`, since it has a different completion mechanism, the instance caller is still responsible for handling unsubscription or completion.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4528

```
[ ] Yes
[x] No
```


modules/signals/rxjs-interop/src/rx-method.ts
```diff
@@ -24,22 +25,32 @@ export function rxMethod<Input>(
...
           untracked(() => source$.next(value));
         },
-        { injector }
+        { injector: instanceInjector }
       );
-      const instanceSub = { unsubscribe: () => watcher.destroy() };
+
+      instanceInjector.get(DestroyRef).onDestroy(() => {
+        sourceSub.unsubscribe();
+      });
```
markostanimirovic, line 49: when we unsubscribe from the root/source subscription, rxMethod won't be alive anymore

with this logic, unsubscribing from the source subscription will happen when the first instance injector is destroyed

modules/signals/rxjs-interop/spec/rx-method.spec.ts
```diff
@@ -231,4 +238,122 @@ describe('rxMethod', () => {
...
+            },
+          ]),
+          provideLocationMocks(),
+        ],
+      });
+
+      return TestBed.inject(GlobalService);
+    }
+
+    it('should destroy with component injector when rxMethod is in root and RxMethod in component', async () => {
```
markostanimirovic, line 292: for consistency with other tests from the signals package:

```suggestion
    it('destroys with component injector when rxMethod is in root and RxMethod in component', async () => {
```

the same convention should be applied to other tests

modules/signals/rxjs-interop/src/rx-method.ts
```diff
@@ -11,7 +11,8 @@ import {
 import { isObservable, noop, Observable, Subject, Unsubscribable } from 'rxjs';
 
 type RxMethod<Input> = ((
-  input: Input | Signal<Input> | Observable<Input>
+  input: Input | Signal<Input> | Observable<Input>,
+  injector?: Injector
```
markostanimirovic, line 15: for consistency with the `rxMethod` signature, I suggest:

```suggestion
  config?: { injector?: Injector }
```

markostanimirovic (commented):
Left a few suggestions. It will be useful to add a concise section on this topic to the [RxJS Integration](https://ngrx.io/guide/signals/rxjs-integration#rxjs-integration) page.

rainerhahnekamp:
@markostanimirovic, I've incorporated all your suggestions and fixed the issues with the source subscription. As a result, the current version of `rxMethod` has fewer changes.

If you are OK with the implementation, I'd do a last push and add the addition for the docs.

modules/signals/rxjs-interop/src/rx-method.ts
```diff
@@ -30,14 +31,21 @@ export function rxMethod<Input>(
...
 
-  const rxMethodFn = (input: Input | Signal<Input> | Observable<Input>) => {
+  const rxMethodFn = (
+    input: Input | Signal<Input> | Observable<Input>,
+    config?: { injector?: Injector }
+  ) => {
     if (isSignal(input)) {
+      const callerInjector = getCallerInjectorIfAvailable();
+      const customInjector = config?.injector;
+      const instanceInjector = customInjector ?? callerInjector ?? injector;
```
timdeschryver, line 41: perf: can we skip the execution of `getCallerInjectorIfAvailable` when a `customInjector` is provided via the config?

```suggestion
      const instanceInjector = config?.injector ?? getCallerInjectorIfAvailable() ?? injector;
```

modules/signals/rxjs-interop/src/rx-method.ts
rainerhahnekamp: Done

modules/signals/rxjs-interop/src/rx-method.ts
rainerhahnekamp: Sure

markostanimirovic (approved):
I've added the same behavior for observables - unsubscribe from an instance subscription when an injector is destroyed.

modules/signals/rxjs-interop/spec/rx-method.spec.ts
```diff
@@ -231,4 +238,221 @@ describe('rxMethod', () => {
...
+
+      expect(globalService.globalObservableChangeCounter).toBe(2);
+
+      await harness.navigateByUrl('/without-store');
+      globalService.incrementObservable();
+
+      expect(globalService.globalObservableChangeCounter).toBe(2);
+    });
+
+    it('falls back to source injector when reactive method is called is outside of injection context', async () => {
```
timdeschryver, line 428:
```suggestion
    it('falls back to source injector when reactive method is called outside of the injection context', async () => {
```

timdeschryver (approved):

PR #4539, opened 2024-09-28
feat(signals): add deepComputed function

### LINKED ISSUE #4202: @ngrx/signals: add `deepComputed` function

signals

### Information

I would like to have a ``computed()`` function similar to the one provided by Angular for ``Signals`` but in this case, instead of returning a Signal, it would return a ``DeepSignal``.

This would allow to optimise template rendering for computed states based on objects.

Exemples:

- Current way of using computed variables:
```html
<p>
    {{ computedState().firstName }}
</p>
```
- Proposed way of using computed variables:
```html
<p>
    {{ computedState.firstName() }}
</p>
```

I'm using the ``computed()`` function provided by Angular which returns a Signal.

- [ ] Yes
- [X] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4202

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/src/deep-computed.ts
+++ b/modules/signals/src/deep-computed.ts
@@ -1,0 +1,8 @@
+import { computed } from '@angular/core';
+import { DeepSignal, toDeepSignal } from './deep-signal';
+
+export function deepComputed<T extends object>(
+  computation: () => T
+): DeepSignal<T> {
+  return toDeepSignal(computed(computation));
+}
--- a/modules/signals/spec/deep-computed.spec.ts
+++ b/modules/signals/spec/deep-computed.spec.ts
@@ -1,0 +1,32 @@
+import { isSignal, signal } from '@angular/core';
+import { deepComputed } from '../src';
+
+describe('deepComputed', () => {
+  it('creates a deep computed signal when computation result is an object literal', () => {
+    const source = signal(0);
+    const result = deepComputed(() => ({ count: { value: source() + 1 } }));
+
+    expect(isSignal(result)).toBe(true);
+    expect(isSignal(result.count)).toBe(true);
+    expect(isSignal(result.count.value)).toBe(true);
+
+    expect(result()).toEqual({ count: { value: 1 } });
+    expect(result.count()).toEqual({ value: 1 });
+    expect(result.count.value()).toBe(1);
+
+    source.set(1);
+
+    expect(result()).toEqual({ count: { value: 2 } });
+    expect(result.count()).toEqual({ value: 2 });
+    expect(result.count.value()).toBe(2);
+  });
+
+  it('does not create a deep computed signal when computation result is an array', () => {
+    const source = signal(0);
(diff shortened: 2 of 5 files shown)


timdeschryver (approved):
👍

brandonroberts:
Q: Why wouldn't `signalState` use this?

markostanimirovic:
> Q: Why wouldn't `signalState` use this?

`signalState` provides a state source to the `toDeepSignal` function while `deepComputed` provides a computed signal:

```ts
// signalState
const stateSource = signal(initialState);
toDeepSignal(stateSource.asReadonly());

// deepComputed
toDeepSignal(computed(computation));
```

Instead of `stateSource.asReadonly()` we can do the following:

```ts
// signalState
deepComputed(computed(() => stateSource()));
```

But I don't see a benefit because the writable signal provides the `asReadonly` method.

modules/signals/src/deep-computed.ts
```diff
@@ -0,0 +1,8 @@
+import { computed } from '@angular/core';
+import { DeepSignal, toDeepSignal } from './deep-signal';
+
+export function deepComputed<T extends object>(
```
matheo, line 4: @markostanimirovic what's the reasoning to restrict `T` to be an object if `toDeepSignal` supports "any"?
sometimes we have selectors that returns primitives or objects depending on the context :thinking:

markostanimirovic: Yes, that's a requirement. `deepComputed` uses the initial value to create a proxy for deeply nested signals.

matheo: @markostanimirovic it's basically an alias of `toDeepSignal` which doesn't have that requirement

markostanimirovic: If a signal with a primitive value is provided to the `toDeepSignal` function, it's just returned without further processing. Check the implementation: https://github.com/ngrx/platform/blob/main/modules/signals/src/deep-signal.ts#L26-L29

If a primitive type or union that contains primitive is passed to the `DeepSignal` type, it will be resolved to `Signal`:

```ts
import { DeepSignal } from '@ngrx/signals';

declare const s1: DeepSignal<string>; // type: Signal<string>
declare const s2: DeepSignal<{ foo: string } | string>; // type: Signal<{ foo: string } | string>
```

I could remove the `extends object` restriction from the `deepComputed` input, but then, users may be confused why the return type is resolved to `Signal`:

```ts
const s1 = deepComputed(() => 1 as number | { foo: string }); // type: Signal<number | { foo: string }>
```

In other words, there is no difference in this example whether you use `computed` or `deepComputed`.

With the `extends object` restriction it's clear - use `deepComputed` for objects, otherwise, it has no impact and you can go with `computed`.

PR #4605, opened 2024-11-22
fix: exclude class instance from `DeepSignal`

### LINKED ISSUE #4604: signalState is not creating DeepSignal property signals when the value object is an instance of a class

signals

Using signalState, create a new store where the type of the value in the store is a class, and the value is set to an instance of that class. The documentation states that when the value in the store is an object, signals will be created for each property on the object (DeepSignal). But in the case where the object is a class instance, signals are not being created for each property, they are undefined.

If this isn't supported, that's one thing, but TypeScript still types each property on the value as a signal.

Minimal reproduction: https://stackblitz.com/edit/typescript-dxfawb?file=index.ts

### Expected behavior

Signals are created for each property on the signalStore value when the value is an instance of a class.

NgRx: 18+

### Other information

_No response_

- [ ] Yes
- [ ] No

rainerhahnekamp:
This is a draft. Whoever has an idea, for the fix, please let me know. It tries to solve the following problem.

```typescript
class User {
  id = 0;
  name = 'Konrad';
}

const state = signalState({ user: new User() });
const user = state.user; // should not be DeepSignal<User>
```

`IsRecord` tries to differentiate a class instance from object literal by checking
if a prototype is there or not.

Fore more details, checkout the commit.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4604

```
[ ] Yes
[ ] No
```

### DIFF
--- a/modules/signals/spec/types/signal-state.types.spec.ts
+++ b/modules/signals/spec/types/signal-state.types.spec.ts
@@ -160,6 +160,22 @@ describe('signalState', () => {
     );
   });
 
+  it('does not create deep signals for a class instance', () => {
+    const snippet = `
+      class User {
+        id = 0;
+        name = 'Konrad';
+      }
+
+      const state = signalState({ user: new User() });
+      const user = state.user;
+    `;
+
+    expectSnippet(snippet).toSucceed();
+
+    expectSnippet(snippet).toInfer('user', 'Signal<User>');
+  });
+
   it('does not create deep signals for optional state slices', () => {
     const snippet = `
       type State = {
--- a/modules/signals/src/ts-helpers.ts
+++ b/modules/signals/src/ts-helpers.ts
@@ -1,3 +1,6 @@
+import { Signal } from '@angular/core';
+import { signalState } from './signal-state';
+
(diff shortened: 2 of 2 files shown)


jdegand:
I assume you used `constructor: Function` and that didn't work.  

Maybe, check if the `constructor.name` is 'Object' (true for object literals but not for class instances)?

```
export type IsRecord<T> = T extends object
  ? T extends unknown[]
    ? false
    : T extends Set<unknown>
    ? false
    : T extends Map<unknown, unknown>
    ? false
    : T extends Function
    ? false
    : T extends { prototype: unknown } // Checks for class instances
    ? false
    : T extends { constructor: { name: string } }
    ? T['constructor']['name'] extends 'Object'
      ? true
      : false
    : false
  : false;

```

rainerhahnekamp:
@jdegand: Nope, in the meantime, I've pulled out a TypeScript Playground, which allows us to experiment a little bit.

I am going to share the Link in the original issue, and I'm going to close this PR since it is not ready yet.

PR #4611, opened 2024-12-01
feat(eslint-plugin): add new rule require-super-ondestroy

### LINKED ISSUE #4505: ESLint rule: require-super-ondestroy

eslint-plugin

### Information

The documentation explains that we need to call `super.ngOnDestroy()` in our component stores: https://ngrx.io/guide/component-store/lifecycle#ondestroy
However, it's easy to forget.
Ideally, it should be enforced automatically. An ESLint rule sounds like a relatively low effort way to achieve that

Custom eslint rule

- [x] Yes
- [ ] No

markoblagdan:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No message from ESLint when overriding `ngOnDestroy()` in a class extending `ComponentStore` from `@ngrx/component-store` and not calling the parent's method via `super.ngOnDestroy()`.

Closes #4505 

## What is the new behavior?

Running eslint or opening the relevant class file in an editor supporting ESLint shows an error message when overriding `ngOnDestroy()` in a class extending `ComponentStore` and not calling the parent's method via `super.ngOnDestroy()`.

```
[ ] Yes
[x] No
```


markoblagdan:
Maybe I should've mentioned the following in the PR under _Other information_, wasn't sure:

- I do not handle the case when `ComponentStore` is a custom class not imported from `@ngrx/component-store`, which means this rule will raise a false error if one extends from a custom `ComponentStore` class and overrides `ngOnDestroy`. Seemed like a weird edge case so I did not handle it, but let me know if it would be better to do so anyway

Additional changes unrelated to #4505 (not sure if it was OK to include this):
- added a new util function for the `component-store` rules, `getNgrxComponentStoreNames`, and used it where appropriate
- removed deprecated method call `context.getSourceCode()` and replaced it with the suggested `context.sourceCode` property
- removed a line in ng-add `schema.ts` which disables the `@typescript-eslint/no-empty-interface` rule

modules/eslint-plugin/spec/rules/component-store/require-super-ondestroy.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+import rule, {
+  messageId,
+} from '../../../src/rules/component-store/require-super-ondestroy';
+import { fromFixture, ruleTester } from '../../utils';
+import path = require('path');
+
+type MessageIds = ESLintUtils.InferMessageIdsTypeFromRule<typeof rule>;
+type Options = ESLintUtils.InferOptionsTypeFromRule<typeof rule>;
+
+const valid: () => (string | ValidTestCase<Options>)[] = () => [
```
timdeschryver, line 12: Thanks, @markoblagdan this is looking good!

> I do not handle the case when ComponentStore is a custom class not imported from @ngrx/component-store, which means this rule will raise a false error if one extends from a custom ComponentStore class and overrides ngOnDestroy. Seemed like a weird edge case so I did not handle it, but let me know if it would be better to do so anyway

Do you want to add a test for this case please, that way it's also "documented"

timdeschryver (changes requested):

modules/eslint-plugin/spec/rules/component-store/require-super-ondestroy.spec.ts
markoblagdan: Of course, added valid and invalid cases for that scenario, but I wonder if I should leave a comment at the test case(s), maybe linking to this comment or PR?

timdeschryver: Feel free to add a comment.
I had a different test in my mind though:

```ts
import { BookStore } from '../components/component-store';
class BooksStore extends BookStore implements OnDestroy
{
  cleanUp() {}
  override ngOnDestroy(): void {
    this.cleanUp();
    super.ngOnDestroy();
  }
}
```

markoblagdan: That will not raise a false error, that is, no error will be shown when _not_ calling super.ngOnDestroy in your example. The error will be wrongly raised only if the extended class is literally called `ComponentStore`.

So I'm not sure it makes sense to add your example, unless I'm missing something, since it doesn't (and shouldn't) trigger the rule.

Should I keep the case I've added?

markoblagdan: > Of course, added valid and invalid cases for that scenario, but I wonder if I should leave a comment at the test case(s), maybe linking to this comment or PR?

Was bothered by this so I handled the case with the rule after all, it wasn't so hard. Basically now I'm also checking if `ComponentStore` is imported from ngrx/component-store. Also updated thevalid test case for this.

timdeschryver (approved):
Thanks @markoblagdan !

modules/eslint-plugin/spec/rules/component-store/require-super-ondestroy.spec.ts
timdeschryver: Nice, that's a good solution!

markostanimirovic (approved):

modules/eslint-plugin/src/rules/component-store/require-super-ondestroy.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    let hasNgrxComponentStoreImport = false;
+
+    return {
+      [`ImportDeclaration[source.value='@ngrx/component-store'] ImportSpecifier[imported.name='${componentStoreClassName}']`](
+        _: TSESTree.ImportSpecifier
+      ) {
+        hasNgrxComponentStoreImport = true;
+      },
+      [`ClassDeclaration[superClass.name=${componentStoreClassName}] ${ngOnDestroyMethodSelector}:not(:has(CallExpression[callee.object.type='Super'][callee.property.name='ngOnDestroy'])) > .key`](
```
amakhrov, line 37: this uses `:has` - which is technically supported by the underlying `esquery`, but is not officially supported by `eslint`. It's not mentioned in t[he docs here](https://eslint.org/docs/latest/extend/selectors#what-syntax-can-selectors-have), and also see this issue: https://github.com/eslint/eslint/issues/14076#issuecomment-926983076.

do you think this approach is reasonably future-proff?

markoblagdan: I was not aware of this - doesn't seem like support for it would be dropped, but maybe it's safer to use another approach.

@timdeschryver If I wanted to make a PR with an alternative to using `:has`, what would be the best way - should we create a new issue and I make a bugfix PR referencing it, or some other way?

timdeschryver: I prefer to keep it as-is.
There are multiple rules using `:has`, and I like it because it's easy to read.
If it does become deprecated/removed in the future we can update it.

PR #4614, opened 2024-12-02
fix(signals): create deep signals for custom class instances

### LINKED ISSUE #4604: signalState is not creating DeepSignal property signals when the value object is an instance of a class

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

- Deep signals are generated by mistake for built-in types: `Error`, `Date`, and `RegExp`.
- Deep signals are not generated for custom class instances even though they were typed as deep signals.

Closes #4604

## What is the new behavior?

- Deep signals are not generated for built-in types: `Error`, `Date`, and `RegExp`.
- Deep signals are generated for custom class instances:

```ts
class User {
  constructor(readonly name: string) {}
}

const state = signalState({ user: new User('John') });
console.log(state.user()); // { name: 'John' }
console.log(state.user.name()); // 'John'
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/src/ts-helpers.ts
+++ b/modules/signals/src/ts-helpers.ts
@@ -1,13 +1,14 @@
 export type Prettify<T> = { [K in keyof T]: T[K] } & {};
 
 export type IsRecord<T> = T extends object
-  ? T extends unknown[]
-    ? false
-    : T extends Set<unknown>
-    ? false
-    : T extends Map<unknown, unknown>
-    ? false
-    : T extends Function
+  ? T extends
+      | unknown[]
+      | Set<unknown>
+      | Map<unknown, unknown>
+      | Date
+      | Error
+      | RegExp
+      | Function
     ? false
     : true
   : false;
--- a/modules/signals/spec/deep-signal.spec.ts
+++ b/modules/signals/spec/deep-signal.spec.ts
@@ -1,0 +1,113 @@
+import { toDeepSignal } from '../src/deep-signal';
+import { isSignal, signal } from '@angular/core';
+
+describe('toDeepSignal', () => {
+  it('creates deep signals for plain objects', () => {
+    const sig = signal({ m: { s: 't' } });
+    const deepSig = toDeepSignal(sig);
+
+    expect(sig).not.toBe(deepSig);
+
+    expect(isSignal(deepSig)).toBe(true);
+    expect(deepSig()).toEqual({ m: { s: 't' } });
+
+    expect(isSignal(deepSig.m)).toBe(true);
+    expect(deepSig.m()).toEqual({ s: 't' });
+
+    expect(isSignal(deepSig.m.s)).toBe(true);
+    expect(deepSig.m.s()).toBe('t');
+  });
+
+  it('creates deep signals for custom class instances', () => {
+    class User {
+      constructor(readonly firstName: string) {}
+    }
+
+    class UserState {
+      constructor(readonly user: User) {}
+    }
+
+    const sig = signal(new UserState(new User('John')));
+    const deepSig = toDeepSignal(sig);
+
+    expect(sig).not.toBe(deepSig);
+
+    expect(isSignal(deepSig)).toBe(true);
+    expect(deepSig()).toEqual({ user: { firstName: 'John' } });
+
+    expect(isSignal(deepSig.user)).toBe(true);
+    expect(deepSig.user()).toEqual({ firstName: 'John' });
+
+    expect(isSignal(deepSig.user.firstName)).toBe(true);
+    expect(deepSig.user.firstName()).toBe('John');
+  });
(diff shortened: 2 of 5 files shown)


modules/signals/src/ts-helpers.ts
```diff
@@ -1,13 +1,14 @@
...
-    ? false
-    : T extends Function
+  ? T extends
+      | unknown[]
+      | Set<unknown>
+      | Map<unknown, unknown>
+      | Date
+      | Error
+      | RegExp
+      | Function
```
markostanimirovic, line 11: ESLint rules `with-state-no-arrays-at-root-level` and `signal-state-no-arrays-at-root-level` can be extended to match all these types. cc @timdeschryver

brandonroberts (approved):

rainerhahnekamp:
Hi @markostanimirovic, why do we want to support deep signals for class instances? I’m not sure we should “split it up” as we would with a normal object literal.

To me, a class instance is an atomic unit—it has its own state, encapsulated properties, methods that define its logic, and potentially inheritance. Given these characteristics, I don’t see it as a good fit for a deep signal.


image

markostanimirovic:
> Hi @markostanimirovic, why do we want to support deep signals for class instances? I’m not sure we should “split it up” as we would with a normal object literal.
> 
> To me, a class instance is an atomic unit—it has its own state, encapsulated properties, methods that define its logic, and potentially inheritance. Given these characteristics, I don’t see it as a good fit for a deep signal.

I fully agree with you! Because of that, this won't be shown in official docs or be recommended anywhere. There are two main reasons for this change:

- `DeepSignal` types are already created for custom class instances, but deep signals are not generated. This leads to confusion and bad DX - errors occur in runtime. Since I couldn't find a way to restrict the usage of custom class instances with `signalStore`/`signalState` at the type level, it's better to make it work than have runtime errors in my opinion.
- Compromise - there are still many apps that use classes for models. This would facilitate their migration to `@ngrx/signals`.

If you remember in the beginning, SignalStore didn't support interfaces, and the state type was restricted to `Record<string, unknown>`. To support interfaces, we made the state type less strict which has this issue as a consequence.

modules/signals/src/ts-helpers.ts
```diff
@@ -1,13 +1,14 @@
...
 
 export type IsRecord<T> = T extends object
-  ? T extends unknown[]
-    ? false
-    : T extends Set<unknown>
-    ? false
-    : T extends Map<unknown, unknown>
-    ? false
-    : T extends Function
+  ? T extends
```
timdeschryver, line 4: Do want to extend this check (and within `isRecord`) to include all other types as well, including lesser known/used ones?

With help from Copilot:

```ts
type NonRecordTypes =
  | unknown[]
  | Set<unknown>
  | Map<any, unknown>
  | WeakSet<unknown>
  | WeakMap<any, unknown>
  | Date
  | Error
  | RegExp
  | Function
  | Promise<unknown>
  | ArrayBuffer
  | SharedArrayBuffer
  | DataView
  | Int8Array
  | Uint8Array
  | Uint8ClampedArray
  | Int16Array
  | Uint16Array
  | Int32Array
  | Uint32Array
  | Float32Array
  | Float64Array;

export type IsRecord<T> = T extends object
  ? T extends NonRecordTypes
    ? false
    : true
  : false;
```

Copilot also suggested this can be made simpler, but AFAIK there's no way of making `isRecord` more generic, instead of defining all possible types.

```ts
type IsRecord<T> = T extends object
  ? T extends Error
    ? false
    : keyof T extends string | number
    ? true
    : false
  : false;
```

markostanimirovic: PR is updated with checks for additional built-in object types. Array, set, map, and uint/float arrays are covered with a check for `Iterable`.

timdeschryver (approved):
@rainerhahnekamp is this also OK for you? (because you added a comment)

rainerhahnekamp:
@timdeschryver 

> @rainerhahnekamp is this also OK for you? (because you added a comment)

All good 👍

PR #4648, opened 2025-01-02
fix(signals): remove `signalMethod` instance watcher on destroy

### LINKED ISSUE #4644: signalMethod cleans up watcher after first effect run

signals

The Cleanup callback registration as shown https://github.com/ngrx/platform/blob/34d18af5520bc671d9e7b3b4fda038357517f7a3/modules/signals/src/signal-method.ts#L41in the signalMethod effect gets executed on every effect run. 

This would remove the effect watcher entry from the Watchers list after the first effect run and therefore will remove the capability to destroy all the watchers in one go via the  signalMethodFn.destroy() call.

### Expected behavior

We dont need to register a cleanup callback for the underlying effect used to track the signal when calling signalMethod. This will ensure that a call to signalMethodFn.destroy() will work correctly and therefore will destroy all the effects that are created via the signalMethod call

NgRx - 19.0.0 

### Other information

Reproduction - https://stackblitz.com/edit/github-nagwuftn?file=src%2Fmain.ts.
As you can see even after destroying the signalMethod the logging continues in the console. 

- [ ] Yes
- [ ] No

rainerhahnekamp:
Each `signalMethod` instance uses a watcher assigned to it. The watcher is destroyed when the `signalMethod` itself gets destroyed. However, watchers were removed prematurely due to `onDestroy` behavior in `effect`. • `onDestroy` doesn’t run when an effect is destroyed but after each execution. • This caused watchers to be removed after the first execution unexpectedly.

Problematic Code: `watchers.splice(watchers.indexOf(watcher), 1);`

The removal mistakenly affected other watchers, keeping instances running, even after the signalMethod was destroyed.

Co-authored-by: @davdev82

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

effects are not destroyed.

Closes #4644

```
[ ] Yes
[x] No
```


modules/signals/src/signal-method.ts
```diff
@@ -35,15 +36,21 @@ export function signalMethod<Input>(
...
+        () => {
           const value = input();
           untracked(() => processingFn(value));
-          onCleanup(() => watchers.splice(watchers.indexOf(watcher), 1));
         },
         { injector: instanceInjector }
       );
       watchers.push(watcher);
 
+      instanceInjector.get(DestroyRef).onDestroy(() => {
```
davdev82, line 47: @rainerhahnekamp , I think effect() in angular already does the destroying of the effect using DestroyRef internally using the passed in injector as the source of DestroyRef. So this code might not be needed

rainerhahnekamp: @davdev82 that is for removing the `watcher` reference from the `watchers` array. Otherwise the GC (at least in my understanding) cannot remove `watcher`.

davdev82: @rainerhahnekamp sorry my bad

modules/signals/spec/signal-method.spec.ts
```diff
@@ -72,18 +72,15 @@ describe('signalMethod', () => {
       const adder = createAdder((value) => (a += value));
       adder(summand1);
       adder(summand2);
-
-      summand1.set(2);
-      summand2.set(3);
       TestBed.flushEffects();
-      expect(a).toBe(6);
+      expect(a).toBe(4);
```
timdeschryver, line 76: I was doing some testing and it didn't work the way as I expected when an effect is destroyed manually.
Is this the desired behavior?

```ts
it('stops specific tracking when calling destroy manually on an effect', () => {
      let a = 1;
      const summand1 = signal(1);
      const summand2 = signal(2);
      const adder = createAdder((value) => (a += value));
      adder(summand1);

      const s2 = adder(summand2);
      // 👇 immediately destroy
      s2.destroy();

      TestBed.flushEffects();
      expect(a).toBe(4);

      summand1.set(100);
      summand2.set(3000);

      TestBed.flushEffects();

      // 👇 Fails, received 3104
      expect(a).toBe(1004); 
    });
```

rainerhahnekamp: Ouch, that hurts. Thanks Tim, let me fix that as well.

rainerhahnekamp: Ok, now it starts to become interesting. That is a behavior of `effect`. That test also fails:

```typescript
  it('should destroy an effect', () => {
    let effectCounter = 0;
    const counter = signal(1);
    const effectRef = TestBed.runInInjectionContext(() =>
      effect(() => {
        counter();
        effectCounter++;
      }, { injector: TestBed.inject(EnvironmentInjector) })
    );
    expect(effectCounter).toBe(0);
    effectRef.destroy();
    TestBed.flushEffects();
    expect(effectCounter).toBe(0);
  });
  ```

rainerhahnekamp: @timdeschryver, I've now added a test, that destroys the instance manually after the first `flushEffects()` and that one works as expected.

I have been passed on the behavior of a non working destroy before the first flush but it is something that needs to be fixed by Angular.

Therefore, I'd suggest that this one is seen as resolved.

timdeschryver: Thanks for testing this @rainerhahnekamp.
I agree that this is then the expected Angular behavior.

timdeschryver (approved):

markostanimirovic (approved):

rainerhahnekamp:
For the record. The not-working `destroy` is now officially a bug: https://github.com/angular/angular/issues/59410

PR #4656, opened 2025-01-03
fix(signals): enable `withProps` to handle Symbols

### LINKED ISSUE #4655: `withProps` doesn't allow Symbols as keys

signals

`signalStore` skips any symbols of `withProps`.

In this example, `secret` is undefined

```typescript
const SECRET = Symbol('SECRET');

const Store = signalStore(withProps(() => ({ [SECRET]: 'secret' })));

@Component({
  selector: 'app-root',
  template: `Secret: {{ secret }}`,
  providers: [Store],
})
export class App {
  protected store = inject(Store);
  protected secret = this.store[SECRET];
}
```

https://stackblitz.com/edit/github-wnvurbbx?file=src%2Fmain.ts

### Expected behavior

Symbols should be used as well

NgRx 19, Angular 19

### Other information

_No response_

- [X] Yes
- [ ] No

rainerhahnekamp:
Resolve an issue in `signalStore` that prevented Symbols from being added as property keys when merging
different features.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The value is just undefined.

Closes #4655

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/spec/with-props.spec.ts
+++ b/modules/signals/spec/with-props.spec.ts
@@ -1,7 +1,8 @@
 import { signal } from '@angular/core';
 import { of } from 'rxjs';
-import { withMethods, withProps, withState } from '../src';
+import { signalStore, withMethods, withProps, withState } from '../src';
 import { getInitialInnerStore } from '../src/signal-store';
+import { TestBed } from '@angular/core/testing';
 
 describe('withProps', () => {
   it('adds properties to the store immutably', () => {
@@ -46,5 +47,16 @@ describe('withProps', () => {
       'Trying to override:',
       's1, p2, m1'
     );
+  });
+
+  it('allows symbols as props', () => {
+    const SECRET = Symbol('SECRET');
+
+    const Store = signalStore(withProps(() => ({ [SECRET]: 'secret' })));
+    const store = TestBed.configureTestingModule({ providers: [Store] }).inject(
+      Store
+    );
+
+    expect(store[SECRET]).toBe('secret');
   });
 });
--- a/modules/signals/spec/with-props.spec.ts
+++ b/modules/signals/spec/with-props.spec.ts
@@ -1,8 +1,7 @@
 import { signal } from '@angular/core';
 import { of } from 'rxjs';
-import { signalStore, withMethods, withProps, withState } from '../src';
+import { withMethods, withProps, withState } from '../src';
 import { getInitialInnerStore } from '../src/signal-store';
-import { TestBed } from '@angular/core/testing';
 
 describe('withProps', () => {
   it('adds properties to the store immutably', () => {
@@ -47,25 +46,5 @@ describe('withProps', () => {
       'Trying to override:',
       's1, p2, m1'
     );
-  });
-
-  it('allows symbols as props', () => {
-    const SECRET = Symbol('SECRET');
-
-    const Store = signalStore(withProps(() => ({ [SECRET]: 'secret' })));
-    const store = TestBed.configureTestingModule({ providers: [Store] }).inject(
-      Store
-    );
-
-    expect(store[SECRET]).toBe('secret');
-  });
-
-  it('allows numbers as props', () => {
-    const Store = signalStore(withProps(() => ({ 1: 'Number One' })));
-    const store = TestBed.configureTestingModule({ providers: [Store] }).inject(
-      Store
-    );
-
-    expect(store[1]).toBe('Number One');
   });
 });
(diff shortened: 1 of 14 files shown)


modules/signals/src/signal-store.ts
```diff
@@ -1355,12 +1355,18 @@ export function signalStore(
...
+        ...props,
+        ...methods,
+      };
 
       (this as any)[STATE_SOURCE] = innerStore[STATE_SOURCE];
 
-      for (const key in storeMembers) {
-        (this as any)[key] = storeMembers[key];
+      for (const key of Reflect.ownKeys(storeMembers)) {
+        if (typeof key === 'string' || typeof key === 'symbol') {
```
timdeschryver, line 1367: Can we also include `number`s?
Because these can also be used as property names.

Or maybe this isn't a problem?

rainerhahnekamp: Short answer: No, and I've added a test as well.

In my understanding, numbers are always serialized during the runtime. That's why Reflect.ownKeys returns 'symbol | string`.

Example

```typescript
const o = {1: 'a'}
o[1]; // returns 'a'
o['1']; // returns 'a'
```

The if condition was necessary because of the TypeScript compiler. It doesn't accept a union type.

timdeschryver (approved):

modules/signals/src/signal-store.ts
```diff
@@ -1355,12 +1355,18 @@ export function signalStore(
...
+          (this as any)[key] = storeMembers[key];
+        }
```
markostanimirovic, line 1369: what's the reason for adding this `if`? is there any other type of key other than string or symbol that should be ignored?
```suggestion
        (this as any)[key] = storeMembers[key];
```

markostanimirovic (changes requested):

modules/signals/src/signal-store.ts
rainerhahnekamp: `key` in this case is of type `string | symbol` and TypeScript doesn't allow an access with that type. That's why I had to apply type narrowing.

Litte bit strange, since I've never experienced that before.

markostanimirovic: I just checked and it seems to work on my machine 😅 



Anyway, instead of adding additional LOC that will end up in the final bundle unnecessarily in order to satisfy TS, I suggest using type casting if it doesn't work without `if (typeof key === 'string' ...`

modules/signals/spec/with-props.spec.ts
```diff
@@ -47,4 +48,24 @@ describe('withProps', () => {
...
+  });
+
+  it('allows numbers as props', () => {
+    const Store = signalStore(withProps(() => ({ 1: 'Number One' })));
+    const store = TestBed.configureTestingModule({ providers: [Store] }).inject(
+      Store
+    );
+
+    expect(store[1]).toBe('Number One');
+  });
```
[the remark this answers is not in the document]
rainerhahnekamp: @markostanimirovic. I've integrated the support for symbol in the following areas:
- `withState`
- `withMethods`
- `withComputed`
- Slices (`DeepSignal`)
- `deepFreeze`
- override protection

Think, I've covered them all.

modules/signals/src/signal-store-models.ts
```diff
@@ -11,9 +11,9 @@ export type StateSignals<State> = IsKnownRecord<Prettify<State>> extends true
     }
   : {};
 
-export type SignalsDictionary = Record<string, Signal<unknown>>;
+export type SignalsDictionary = Record<string | symbol, Signal<unknown>>;
```
rainerhahnekamp, line 14: @markostanimirovic @timdeschryver: This could be seen as a breaking change.

If I leave it to string only (`SignalsDictionary` & `MethodsDictionary`), tests and build would also be successful.

What's your take on this?

rainerhahnekamp: @timdeschryver honestly, I have no idea. I just saw that this is a change to public type, and we don't know if there is some code out there that depends, that it is only a string and nothing else...

markostanimirovic: Since `Record<string, ...>` already supports symbols as keys, we can avoid changing `SignalsDictionary` and `MethodsDictionary` types just in case.

rainerhahnekamp: Done

timdeschryver (approved):

PR #4662, opened 2025-01-07
refactor(store-devtools): resolve import cycle to improve tool compatibility

raphinesse:
This change resolves the cyclical import dependency `store-devtools/src/{provide-store-devtools <-> instrument}.ts` by moving `createStateObservable` to `provide-store-devtools.ts`, its only user.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The import cycle breaks TS language tools/environments (e.g. Google's TS build) that don't support cyclical imports. 

## What is the new behavior?

Support TS language tools/environments that don't support cyclical imports.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/store-devtools/src/provide-store-devtools.ts
+++ b/modules/store-devtools/src/provide-store-devtools.ts
@@ -19,7 +19,6 @@ import {
   StoreDevtoolsOptions,
 } from './config';
 import { ReducerManagerDispatcher, StateObservable } from '@ngrx/store';
-import { createStateObservable } from './instrument';
 import { StoreDevtools } from './devtools';
 
 export const IS_EXTENSION_OR_MONITOR_PRESENT = new InjectionToken<boolean>(
@@ -44,6 +43,12 @@ export function createReduxDevtoolsExtension() {
   } else {
     return null;
   }
+}
+
+export function createStateObservable(
+  devtools: StoreDevtools
+): StateObservable {
+  return devtools.state;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


modules/store-devtools/src/provide-store-devtools.ts
```diff
@@ -19,9 +19,14 @@ import {
   StoreDevtoolsOptions,
 } from './config';
 import { ReducerManagerDispatcher, StateObservable } from '@ngrx/store';
-import { createStateObservable } from './instrument';
 import { StoreDevtools } from './devtools';
 
+export function createStateObservable(
```
raphinesse, line 24: Arguably the export here is not required anymore.

```suggestion
function createStateObservable(
```

timdeschryver: Just like the other factory functions this can remain to be exported.

timdeschryver, line 24: To be inline with the rest this `method` should be moved after `IS_EXTENSION_OR_MONITOR_PRESENT`.

timdeschryver (changes requested):
Thanks for the PR, I left some comments

modules/store-devtools/src/provide-store-devtools.ts
raphinesse: I moved it after `createReduxDevtoolsExtension` so that the factory function definitions are in the same order as their usages in the `EnvironmentProvider` below.

markostanimirovic (approved):

timdeschryver (approved):

PR #4698, opened 2025-02-09
fix: ng-add should also work for ESLint 9 flat config files (#4643)

### LINKED ISSUE #4643: ng-add should also work for ESLint 9 flat config files

eslint-plugin

### Information

Currently the NgRx eslint plugin only configures the NgRx plugin within a json file.
We should also add the option to configure our plugin within flat config files.

The schematic should automatically detect which file needs to be updated.
We can take a look at the Angular ESlint plugin for inspiration https://github.com/angular-eslint/angular-eslint/blob/390df5a211f7016be1bcc12c4e139796c1867867/packages/schematics/src/ng-add/index.ts#L41

_No response_

- [ ] Yes
- [ ] No

adxdits:
## PR Checklist

- [x] The commit message follows our guidelines: [CONTRIBUTING.md](https://github.com/ngrx/platform/blob/main/CONTRIBUTING.md#commit)

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

Today: When someone runs the NgRx ESLint schematic (for example via an ng add command), it updates the ESLint configuration by editing a JSON file (like .eslintrc.json) to add the NgRx plugin and its rules.
Closes #4643

## What is the new behavior?

- Detect if `eslint.config.js` exists and update it with NgRx plugin
- Fallback to `.eslintrc.json` when flat config is missing
- Preserve existing configuration

---

**Steps to push the change:**

- [ ] Yes
- [x] No


modules/eslint-plugin/schematics/ng-add/index.ts
```diff
@@ -4,17 +4,79 @@ import type { Schema } from './schema';
 
 export default function addNgRxESLintPlugin(schema: Schema): Rule {
   return (host: Tree, context: SchematicContext) => {
-    const eslintConfigPath = '.eslintrc.json';
+    const jsonConfigPath = '.eslintrc.json';
+    const flatConfigPath = 'eslint.config.js';
```
timdeschryver, line 8: This should also cover commonjs and modulejs files:

- eslint.config.js
- eslint.config.mjs
- eslint.config.cjs

timdeschryver (changes requested):
Hello, thanks for this PR ~ I will review this later this week.
I already noticed that there are no test cases, could you please write some tests to cover these changes?
I'm thinking of:

- a test to detect the correct file is changed
- a test that adds the NgRx plugin to a flat config file
- a test that verifies the plugin isn't added twice

modules/eslint-plugin/schematics/ng-add/index.ts
markostanimirovic: should we also support `.eslintrc.(m|c)js`?

timdeschryver: @markostanimirovic I don't think so, those aren't supported anymore.
https://eslint.org/docs/latest/use/configure/configuration-files#configuration-

adxdits: just as @timdeschryver said i updated the schematics to support only the flat ESLint config files (eslint.config.js, eslint.config.mjs, and eslint.config.cjs) since .eslintrc.(m|c)js files are no longer supported.  :)

adxdits:
> Hello, thanks for this PR ~ I will review this later this week. I already noticed that there are no test cases, could you please write some tests to cover these changes? I'm thinking of:
> 
> * a test to detect the correct file is changed
> * a test that adds the NgRx plugin to a flat config file
> * a test that verifies the plugin isn't added twice

Hello, of course. Thanks for pointing that out!

modules/eslint-plugin/schematics/ng-add/index.ts
```diff
@@ -4,17 +4,79 @@ import type { Schema } from './schema';
...
+    if (host.exists(flatConfigPath)) {
+      let content = host.read(flatConfigPath)?.toString('utf-8');
+      if (!content) {
+        context.logger.warn(`Could not read the ESLint flat config at \`${flatConfigPath}\`.`);
+        return host;
+      }
+
+      // Add the import for the NgRx ESLint plugin if not already present.
+      if (!content.includes("from '@ngrx/eslint-plugin'")) {
+        content = `import ngrxEslintPlugin from '@ngrx/eslint-plugin';\n` + content;
```
timdeschryver, line 21: Is it possible to match this with the current import statements?
For example, when it's not the module syntax we should use 

```ts
const ngrx = require("@ngrx/eslint-plugin");
```

modules/eslint-plugin/schematics/ng-add/index.ts
```diff
@@ -4,17 +4,79 @@ import type { Schema } from './schema';
...
+      const ngrxConfig = `
+{
+  files: ['*.ts'],
+  extends: [\`plugin:@ngrx/${schema.config}\`],
+},`;
+
+      // Check if the flat config already contains an NgRx config.
+      if (!content.includes('plugin:@ngrx')) {
+        // Look for an exported default array.
+        const exportDefaultRegex = /export\s+default\s+\[([\s\S]*?)\];/;
```
timdeschryver, line 35: Same comment here, ideally we also support `module.exports = ...`.
As I expect most eslint configs use tseslint, checking for `tseslint.config` is probably the simplest solution. Updating it is harder than a JSON file, I think it can be easier if we use the TS AST to find the `tseslint.config` CallExpression, and add an argument.

modules/eslint-plugin/schematics/ng-add/index.ts
```diff
@@ -4,17 +4,79 @@ import type { Schema } from './schema';
...
   return (host: Tree, context: SchematicContext) => {
-    const eslintConfigPath = '.eslintrc.json';
+    const jsonConfigPath = '.eslintrc.json';
+    const flatConfigPath = 'eslint.config.js';
     const docs = 'https://ngrx.io/guide/eslint-plugin';
 
-    const eslint = host.read(eslintConfigPath)?.toString('utf-8');
+    // If a flat config exists, update it
+    if (host.exists(flatConfigPath)) {
+      let content = host.read(flatConfigPath)?.toString('utf-8');
```
timdeschryver, line 13: This method is becoming lengthy, could you split it up into at least two methods, one for the flat config, the other for the JSON file?

timdeschryver (changes requested):
C

modules/eslint-plugin/schematics/ng-add/index.ts
timdeschryver: I only see `eslint.config.js` here, could you also want to include the other extensions?

modules/eslint-plugin/schematics/ng-add/index.ts
```diff
@@ -1,70 +1,304 @@
...
+  const ngrxConfig = `
+{
+  files: ['*.ts'],
+  extends: [\`plugin:@ngrx/${schema.config}\`],
+},`;
+
+  // Prefer AST-based update if using tseslint.config(…)
+  if (content.includes('tseslint.config(')) {
+    content = updateTsEslintConfigCall(content, schema, context);
+  } else if (!content.includes('plugin:@ngrx')) {
```
timdeschryver, line 74: To keep things simple I was thinking to only support tseslint.config.
We can always revisit it later when there's a demand.
What do you think about this?

modules/eslint-plugin/schematics/ng-add/index.ts
```diff
@@ -1,70 +1,304 @@
...
+ *
+ * It finds the call expression where the callee is "tseslint.config" and whose first argument is an array.
+ * If that array does not already contain an object configuring the NgRx plugin, it appends one.
+ *
+ * @param content The original file content.
+ * @param schema The provided schema (contains the config name).
+ * @param context The schematic context (used for logging).
+ * @returns The updated file content.
+ */
+function updateTsEslintConfigCall(content: string, schema: Schema, context: SchematicContext): string {
```
timdeschryver, line 126: I will have a look at this, I think it can be easier to update the config.

modules/eslint-plugin/schematics/ng-add/index.ts
adxdits: Good suggestion! i will add the new version today

timdeschryver (changes requested):
It seems as if I can't push to your branch, please update the files with https://gist.github.com/timdeschryver/bb49c998c5a39ebe289b77412aa8fd0f.
The main changes are to use Angular's recorder instead of TypeScript (`ts.factory`) to make the changes, this keeps the implementation consistent with our codebase.
It also updates the tests.
For the next time, please make sure the tests are green while creating a Pull Requests. 

Also, please revert the changes to the package and package-lock files.

timdeschryver:
Since this has become stale, I'm closing this PR in favor of #4747.

Issue #4683, opened 2025-01-24
SignalStore: RFC: Reverting `Object.freeze` Protection in SignalStore


rainerhahnekamp:
### Which @ngrx/* package(s) are relevant/related to the feature request?

signals

### Information

In v19, we introduced a protection mechanism against mutable changes in the SignalStore. This mechanism recursively applied `Object.freeze` to the state object.

Since its introduction, we’ve received numerous issues and reports from the community:

1. **Unintentional Mutations**: Many users were already mutating their state without realizing it.  
2. **Incompatibility**: Some users used types that `Object.freeze` was unable to process, causing runtime errors.

The SignalStore builds on top of Angular’s Signal type, which itself does not enforce immutability through an internal mechanism like `Object.freeze`. This means, when developers switch to the SignalStore, they encounter a different behavior.

In certain edge cases, mutable changes within Signals are necessary to achieve specific outcomes. The current protection mechanism has proven to be more restrictive than beneficial in such scenarios, leading to frustration and workarounds.

Therefore: **We propose reverting the `Object.freeze` protection in the SignalStore.**

If immutability enforcement remains a priority for certain developers, we encourage the community to come up with solutions.

### Issues and Disucssions on `Object.freeze`:

https://github.com/ngrx/platform/issues/4649

https://github.com/ngrx/platform/issues/4645

https://github.com/ngrx/platform/issues/4635

https://github.com/ngrx/platform/discussions/4676

https://github.com/ngrx/platform/discussions/4677

https://github.com/ngrx/platform/discussions/4636

### I would be willing to submit a PR to fix this issue

- [x] Yes
- [ ] No

hafnerpw:
This is needed asap

timdeschryver:
@steven-excelmans this is released a couple hours ago in [19.0.1](https://www.npmjs.com/package/@ngrx/store/v/19.0.1)

PR #4713, opened 2025-03-01
feat(signals): define SignalStore members as readonly

Chiorufarewerin:
## PR Checklist

## PR Type

```
[x] Feature
```

https://github.com/ngrx/platform/discussions/4712

## What is the new behavior?

After these changes, the store and entities will become type-safe:

```ts
this.userStore.entities = signal([{ id: 1000, name: 'Vova' }]); // error
this.userStore.entityMap()[0] = { id: 2000, name: 'Dima' }; // error
this.userStore.entities()[0] = { id: 2000, name: 'Dima' }; // error
this.userStore.entities = signal([]); // error
this.userStore.loadAll = () => undefined; // error
```

```
[x] Yes
[ ] No
```

If someone modifies the store or directly edits entities, they will need to fix these errors (or add just `as any`). However, this will help prevent future errors by enforcing type safety.

Also, I want to mention that it doesn't introduces "breaking changes" in logic, like it was with `Object.freeze` that was reverted https://github.com/ngrx/platform/issues/4683.

## Other information

In my application, almost every type includes readonly to ensure type safety. I always know when this rule is broken—only when using the `as` keyword. It would be great to have similar safety in @ngrx/signals.

This PR provides the minimum type safety I want. Additionally, I suggest changing every instance like this:

```ts
SignalStoreFeature<Input, { state: {}; props: {}; methods: Methods }>
```

To:
```ts
{
    readonly state: {};
    readonly props: {};
    readonly methods: Methods;
  }
```

I can also implement this change if the idea is accepted.

### DIFF
--- a/modules/signals/entities/src/helpers.ts
+++ b/modules/signals/entities/src/helpers.ts
@@ -2,6 +2,7 @@ import {
   DidMutate,
   EntityChanges,
   EntityId,
+  EntityMap,
   EntityPredicate,
   EntityState,
   SelectEntityId,
@@ -9,6 +10,11 @@ import {
 
 declare const ngDevMode: unknown;
 const defaultSelectId: SelectEntityId<{ id: EntityId }> = (entity) => entity.id;
+
+type EntityStateMutable<Entity = any> = {
+  entityMap: Record<EntityId, Entity>;
+  ids: EntityId[];
+};
 
 export function getEntityIdSelector(config?: {
   selectId?: SelectEntityId<any>;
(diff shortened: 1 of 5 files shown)


rainerhahnekamp (approved):
I don't know if there are any side-effects or edge cases where this might become problematic. I don't see any and I do agree this would be a helpful addition.

The PR looks solid. So green light from my side.

timdeschryver:
We're going to postpone this change.
To have a clean PR list, we're going to close this one.

PR #4725, opened 2025-03-20
feat(signals): add unprotected testing helper

### LINKED ISSUE #4540: RFC(`@ngrx/signals`): Add `unprotected` helper for testing

signals

### Information

This RFC suggests introducing the `unprotected` function that would be exported from `@ngrx/signals/testing`. It would allow updating protected SignalStore's state in tests as follows:

```ts
// users.store.ts
const UsersStore = signalStore(withEntities<User>());

// users.store.spec.ts
import { unprotected } from '@ngrx/signals/testing';

const store = TestBed.inject(UsersStore);
patchState(unprotected(store), setAllEntities(usersMock));
```

_No response_

- [X] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4540

## What is the new behavior?

The new `testing` plugin is added with `unprotected` helper:

```ts
import { unprotected } from '@ngrx/signals/testing';

const CounterStore = signalStore(
  { providedIn: 'root' },
  withState({ count: 0 }),
  withComputed(({ count }) => ({
    doubleCount: computed(() => count() * 2),
  })),
);

it('works', () => {
  const store = TestBed.inject(CounterStore);
  patchState(unprotected(store), { count: 1 });

  expect(store.doubleCount()).toBe(2);
});
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/src/state-source.ts
+++ b/modules/signals/src/state-source.ts
@@ -3,6 +3,7 @@ import {
   DestroyRef,
   inject,
   Injector,
+  isSignal,
   Signal,
   untracked,
   WritableSignal,
@@ -28,6 +29,17 @@ export type PartialStateUpdater<State extends object> = (
 export type StateWatcher<State extends object> = (
   state: NoInfer<State>
 ) => void;
+
+export function isWritableStateSource<State extends object>(
+  stateSource: StateSource<State>
+): stateSource is WritableStateSource<State> {
+  return (
+    'set' in stateSource[STATE_SOURCE] &&
+    'update' in stateSource[STATE_SOURCE] &&
+    typeof stateSource[STATE_SOURCE].set === 'function' &&
+    typeof stateSource[STATE_SOURCE].update === 'function'
+  );
+}
 
 export function patchState<State extends object>(
   stateSource: WritableStateSource<State>,
--- a/modules/signals/spec/state-source.spec.ts
+++ b/modules/signals/spec/state-source.spec.ts
@@ -3,17 +3,21 @@ import {
   effect,
   EnvironmentInjector,
   Injectable,
+  signal,
 } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import {
   getState,
+  isWritableStateSource,
   patchState,
   signalState,
   signalStore,
+  StateSource,
   watchState,
   withHooks,
   withMethods,
   withState,
+  WritableStateSource,
 } from '../src';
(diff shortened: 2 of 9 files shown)


modules/signals/src/state-source.ts
```diff
@@ -29,6 +30,17 @@ export type StateWatcher<State extends object> = (
   state: NoInfer<State>
 ) => void;
 
+export function isWritableStateSource<State extends object>(
+  stateSource: StateSource<State>
+): stateSource is WritableStateSource<State> {
+  return (
```
Chiorufarewerin, line 36: Why not [isWritableSignal](https://github.com/angular/angular/blob/main/packages/core/src/render3/reactivity/signal.ts#L107C17-L107C33)?

markostanimirovic: Because `isWritableSignal` is not part of the public API: https://angular.dev/api/core/isWritableSignal

rainerhahnekamp:
I went through the discussion on the issue and just wanted to share a thought. I feel there’s a bit of a risk that unprotected might be used in unintended ways in application code — something we might be able to reduce by adjusting the design so that unprotected is only accessible via a TestBed context.

There are two areas that could lead to accidental usage:
- AI-assisted coding tools may suggest `unprotected` in application code.
- IDEs often auto-import from testing, and it’s easy to overlook where the import is coming from.

One idea might be to use something like `injectedUnprotectedStore`, or even take a similar approach to Angular’s routing and HTTP testing setup by providing a fake service:

```typescript
TestBed.configureTestingModule({ provideNgRxTesting() });
const { unprotected } = TestBed.inject(NgRxTesting);
```

I know this could still be misused, but the design would more clearly indicate that it’s intended for testing.

timdeschryver (approved):
Nice! 💪

timdeschryver:
Sorry @rainerhahnekamp , I missed your comment.

I get your point, on the other hand, I like the simplicity of the `unprotected` in this PR.
Because we cannot guarantee it will be misused, I'm fine with the simplicity.
This is better than manually casting a protected state in a test, or in production code.

I hope teams will review their code to catch these misuses,  and/or create a rule that `testing` imports are forbidden in "normal" `*.ts` files.

EinfachHans:
Just read the docs and wondered why it is not working for me 😃 

Any plans about when the next release including this will happen?

PR #4780, opened 2025-05-13
docs(ngrx/signals): mention re-use of `withComputed/withMethods`

### LINKED ISSUE #4669: docs(`@ngrx/signals`): add a mention of restructuring `withComputed` or `withMethods` for re-use in those features
### Information

The first major stumbling point I had with using features like `withComputed` or `withMethods` was the need to reference other computeds or methods in those features. For example, if I wanted a `withComputed` to expose be able to return a `fullName` that was a store's `firstName` and store's `lastName`, and then also expose a `allCapsFullName` that was the full name computed inside another computed. But do to what I thought was a limitation of how to return an object like `withComputed(({ firstName, lastName }) => ({...})`, I had to make a second `withComputed`

```ts
  withComputed(({ firstName, lastName }) => ({
    fullName: computed(() => {
      return `${firstName()} ${lastName()}`;
    }),
  })),
  withComputed(({ fullName }) => ({
    allCapsFullName: computed(() => {
      return fullName().toUpperCase();
    }),
  })),
```

However, someone pointed out that the first `withComputed` could be restructured to have a return block, with helper `const`s or functions.

```ts
  withComputed(({ firstName, lastName }) => {
    const fullName = computed(() => {
      return `${firstName()} ${lastName()}`;
    })
    return {
      fullName: fullName,
      allCapsFullName: computed(() => {
        return fullName().toUpperCase();
      }),
    }
  }),
```

In retrospect this is fairly obvious, but I see the former approach all the time and was stuck in that mindset myself. I think partially because most examples of things like `withComputed` have those returns without a `return` block, and because there is a lot of little syntax nuances with functions and objects in the signal store to become comfortable with that add up.

I think something like an FAQ page point on this is warranted with how often this little nuance turns into what is perceived as a bigger problem than it is with an easy solution. Or some examples of the second syntax baked into other pages. Willing to formalize/condense this language to make a PR.

### Documentation page

https://ngrx.io/guide/signals/faq

- [X] Yes
- [ ] No

michael-small:
Issue: https://github.com/ngrx/platform/issues/4669

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue: [docs(@ngrx/signals): add a mention of restructuring withComputed or withMethods for re-use in those features](https://github.com/ngrx/platform/issues/4669)

Closes #4669

## What is the new behavior?

Documentation mentions how to re-use a computed or function across or within a `withComputed` or `withMethods`

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ngrx.io/content/guide/signals/signal-store/index.md
+++ b/projects/ngrx.io/content/guide/signals/signal-store/index.md
@@ -285,6 +285,38 @@ export const BooksStore = signalStore(
 
 </code-example>
 
+<div class="alert is-helpful">
+
+It may be necessary for a computed in a `withComputed` feature to need to reference another computed value, 
+or a method in a `withMethods` feature to refer to another method. To do so, either: 
+
+(A) Create another `withComputed` feature  or `withMethods` feature, or 
+
+(B) Break out the common piece with a helper `const`.
+
+```ts
+export const BooksStore = signalStore(
+  withState(initialState),
+  // 👇 Accessing previously defined state signals and properties.
+  withComputed(({ books, filter }) => ({
+    booksCount: computed(() => books().length),
+    sortDirection: computed(() => filter.order() === 'asc' ? 1 : -1),
+  })),
+  // 👇 (A) Also access previously defined computed properties (or functions).
+  withComputed(({ books, sortDirection }) => {
+    // 👇 (B) Define helper functions (or computeds).
+    const sortBooks = (direction: number) =>
+      books().toSorted((a, b) => direction * a.title.localeCompare(b.title));
+
+    return {
+      sortedBooks: computed(() => sortBooks(sortDirection())),
+      reversedBooks: computed(() => sortBooks(-1 * sortDirection())),
+    };
+  }),
+);
+```
+</div>
+
 ### Reactive Store Methods
 
 In more complex scenarios, opting for RxJS to handle asynchronous side effects is advisable.


michael-small:
@timdeschryver @rainerhahnekamp since you two coached me in the issue

I tweaked the example code comments a little.

I added the block towards the end of the "Core Concepts > Defining Store Methods" as this pertains to both `withMethods` and `withComputed` which by then are both explained.

projects/ngrx.io/content/guide/signals/signal-store/index.md
```diff
@@ -285,6 +285,38 @@ export const BooksStore = signalStore(
...
+```ts
+export const BooksStore = signalStore(
+  withState(initialState),
+  // 👇 Accessing previously defined state signals and properties.
+  withComputed(({ books, filter }) => ({
+    booksCount: computed(() => books().length),
+    sortDirection: computed(() => filter.order() === 'asc' ? 1 : -1),
+  })),
+  // 👇 (A) Also access previously defined computed properties (or functions).
+  withComputed(({ books, sortDirection }) => {
```
markostanimirovic, line 306: When I was writing this page, I assumed that readers are familiar with core JavaScript concepts, so they're aware that `() => ({})` is a sugar syntax for `() => { return {} }`.

I'm not a fan of using multiple `withComputed` calls. The native JS way should be preferred and recommended in my opinion:

```ts
withComputed({ books, filter }) => {
  const booksCount = computed(() => books().length);
  const sortDirection = computed(() => /* ... */);
  // ...
  
  return { /* export public computed signals */ };
}),
```

Additionally, it provides the ability not to export private computed signals if necessary.

michael-small: > When I was writing this page, I assumed that readers are familiar with core JavaScript concepts, so they're aware that () => ({}) is a sugar syntax for () => { return {} }.

(sorry if this is overselling but I have thought on this a lot) In retrospect that made sense when I learned this after seeking it out, but I have seen people with this same problem frequently across different platforms. It has even been news to some GDEs who make signal store content. I can't speak for why the various other people didn't recognize this sugar syntax, but for me I just think going from a class based store to a function based one has a lot of little moments like this. And this point for people who wonder about this seems to solve a large hurdle with a small tip.

> I'm not a fan of using multiple withComputed calls. The native JS way should be preferred and recommended in my opinion:

I agree. That said, in the places this has come up, it feels like 2/3 of people are like "great, no need for another `withComputed`" and the other 1/3 are like "that's cool that the `const` method works good but I prefer multiple `withComputed`". So in my opinion, I feel like this `const` helper restructuring is worth a tip somewhere and that people who prefer multiple `withComputed` are just going to keep doing that lol.

michael-small: What about having this PR as still adding this tip with respect to the native `const` approach, but dropping the repeated `withComputed/withMethods` reference?

@timdeschryver you suggested having them both, so I'm curious if my proposed compromise which tentatively drop the repeated `withComputed` sounds legit.

markostanimirovic: If both options look good to Tim and Rainer, then I'm also fine with adding both of them to the documentation.

However, I'd move this from "Core Principles" to the "FAQ" page with a few changes:

- Separate examples for each case (A) and (B).
- Concise examples that are focused on the tip. For example, the `sortBooks` helper is not needed to show how computed signals/methods can be reused.

timdeschryver: Hey, sorry if I mentioned something that wasn't recommended.
I'm following @markostanimirovic's lead here, let's move it to the FAQ page as was originally planned. I do think that it can be useful to be mentioned.

michael-small: I like @rainerhahnekamp's take on that - showing option B but giving a passing mention of option A in text. That seems inline from how I am reading everyone's take on this - emphasizing option B, but having some mention of option A.

Additionally, move this to a FAQ point.

markostanimirovic:
Hi @michael-small,

Still interested in adding this tip to the FAQ page?

michael-small:
Yeah @markostanimirovic, I was curious about what I think can be agreed upon since 3 maintainers all had their own takes. But I didn't want to ping everyone when you guys were working hard on a release.

Thoughts?

> I like rainerhahnekamp's take on that - showing option B but giving a passing mention of option A in text. That seems inline from how I am reading everyone's take on this - emphasizing option B, but having some mention of option A.
> 
> Additionally, move this to an FAQ point.

So to the FAQ for sure, but the former point I was wondering about before fleshing out the next iteration of the tip

michael-small:
@markostanimirovic @timdeschryver @rainerhahnekamp 

I did the following after making some assumptions on what would work well between everyone's takes:

- Moved the tip to the FAQ
- With respects of "Option A" (two consecutive feature blocks)
    - Removed a showcase of "A" in code
    - Mentioned that "A" was possible in text, but took the opportunity to say why it is not encouraged
- Streamlined the example by removing `sortBooks`.
- Adjusted the example to be concisely about a `computed` that is returned but also a `computed` made from that one in the same scope
    - Perhaps it is a bit contrived, but I think it is more concise and easy to follow
    - I didn't want to stray too far from the type of example that was workshopped starting in the issue. However, if this example as it is now seems off then I can rework it to be something else

PR #4795, opened 2025-05-25
feat(signals): allow user-defined signals in `withState` and `signalState` by splitting `STATE_SOURCE`

rainerhahnekamp:
BREAKING CHANGES:

`withState` and `signalState` now support user-defined signals like
`linkedSignal`, `resource.value`, or any other `WritableSignal`.

For example:

```ts
const user = signal({ id: 1, name: 'John Doe' });
const userClone = linkedSignal(user);
const userValue = resource({
  loader: () => Promise.resolve('user'),
  defaultValue: ''
});

const Store = signalStore(
  withState({ user, userClone, userValue: userValue.value })
);
```

The state slices don't change:

```ts
store.user;       // DeepSignal<{ id: number, name: string }>
store.userClone;  // DeepSignal<{ id: number, name: string }>
store.userValue;  // Signal<string>
```

The behavior of `linkedSignal` and `resource` is preserved. Since the
SignalStore no longer creates the signals internally in these cases,
signals passed into `withState` can also be changed externally.

This is a foundational change to enable features like `withLinkedState`
and potential support for `withResource`.

The internal `STATE_SOURCE` is no longer represented as a single
`WritableSignal` holding the entire state object. Instead, each top-level
property becomes its own `WritableSignal`, or remains as-is if the user
already provides a `WritableSignal`.

## Motivation

- Internal creation of signals limited flexibility; users couldn’t bring
their own signals into the store
- Reusing existing signals enables future features like `withLinkedState`
or `withResource`.
- Splitting state into per-key signals improves the performance, because
the root is not the complete state anymore.

## Change to `STATE_SOURCE`

Given:

```ts
type User = {
  firstname: string;
  lastname: string;
};
```

### Before

```ts
STATE_SOURCE: WritableSignal<User>;
```

### Now

```ts
STATE_SOURCE: {
  firstname: WritableSignal<string>;
  lastname: WritableSignal<string>;
};
```

### 1. Different object reference

The returned object from `signalState()` or `getState()` no longer keeps
the same object identity:

```ts
const obj = { ngrx: 'rocks' };
const state = signalState(obj);
```

**Before:**

```ts
state() === obj; // ✅ true
```

**Now:**

```ts
state() === obj; // ❌ false
```

---

### 2. No signal change on empty patch

Empty patches no longer emit updates, since no signal is mutated:

```ts
const state = signalState({ ngrx: 'rocks' });

let count = 0;
effect(() => count++);

TestBed.flushEffects();
expect(count).toBe(1);

patchState(state, {});
```

**Before:**

```ts
expect(count).toBe(2); // triggered
```

**Now:**

```ts
expect(count).toBe(1); // no update
```

---

### 3. No wrapping of top-level `WritableSignal`s

```ts
const Store = signalStore(
  withState({ foo: signal('bar') })
);
const store = new Store();
```

**Before:**

```ts
store.foo; // Signal<Signal<string>>
```

**Now:**

```ts
store.foo; // Signal<string>
```

---

### 4.: `patchState` no longer supports `Record` as root state

Using a `Record`as the root state is no longer supported by `patchState`.

**Before:**

```ts
const Store = signalStore(
  { providedIn: 'root' },
  withState<Record<number, number>>({}),
  withMethods((store) => ({
    addNumber(num: number): void {
      patchState(store, {
        [num]: num,
      });
    },
  }))
);

store.addNumber(1);
store.addNumber(2);

expect(getState(store)).toEqual({ 1: 1, 2: 2 });
```

**After:**

```ts
const Store = signalStore(
  { providedIn: 'root' },
  withState<Record<number, number>>({}),
  withMethods((store) => ({
    addNumber(num: number): void {
      patchState(store, {
        [num]: num,
      });
    },
  }))
);

store.addNumber(1);
store.addNumber(2);

expect(getState(store)).toEqual({}); // ❌ Nothing updated
```

If dynamic keys are needed, consider managing them inside a nested signal instead.

## Further Changes

- `signalStoreFeature` updated due to changes in `WritableStateSource`
- `patchState` now uses `NoInfer` on `updaters` to prevent incorrect type
  inference when chaining
## PR Checklist

- [] Documentation has been added / updated (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[x] Yes
[ ] No
```


markostanimirovic (changes requested):

markostanimirovic:
@rainerhahnekamp It's also necessary to write `BREAKING CHANGES: ...` in a plain text format before the `Other Information` section. It will be copied to the commit body on squash merge. See the example here: https://github.com/ngrx/platform/pull/4584

rainerhahnekamp:
> It's also necessary to write BREAKING CHANGES: ... 

Got it: Plural and then a line break.

It looks like my IDE messed up the subject a little bit. Will check that one as well.

modules/signals/src/state-source.ts
```diff
@@ -30,40 +32,70 @@ export type StateWatcher<State extends object> = (
...
+    return isWritableSignal(signals[key]);
+  });
 }
 
 export function patchState<State extends object>(
   stateSource: WritableStateSource<State>,
   ...updaters: Array<
-    Partial<Prettify<State>> | PartialStateUpdater<Prettify<State>>
+    | Partial<Prettify<NoInfer<State>>>
+    | PartialStateUpdater<Prettify<NoInfer<State>>>
```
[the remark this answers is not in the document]
rainerhahnekamp: Without `NoInfer`, a sequence of updaters would not work anymore.

You can try it out by removing the `NoInfer` and then navigating to the test in `state-source.spec.ts` in line 122.

Not sure if we can drop `Prettify`. I didn't because it was already there before.

modules/signals/src/state-source.ts
```diff
@@ -30,40 +32,70 @@ export type StateWatcher<State extends object> = (
...
+      ...(typeof updater === 'function' ? updater(nextState) : updater),
+    }),
+    currentState
   );
 
+  const signals = stateSource[STATE_SOURCE];
+  const stateKeys = Reflect.ownKeys(stateSource[STATE_SOURCE]);
+  for (const key of Reflect.ownKeys(newState)) {
+    if (!stateKeys.includes(key)) {
+      // TODO: Optional properties which don't exist in the initial state will not be added
```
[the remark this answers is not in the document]
rainerhahnekamp: We have here another breaking change. If we have a state with an optional root property and the initial state skips that one, `patchState` will always skip that one.

We could do a check for `undefined` in `withState` and/or come with an ESLint rule.

I've added some additional tests.

modules/signals/src/with-state.ts
```diff
@@ -1,32 +1,37 @@
...
   SignalStoreFeature,
   SignalStoreFeatureResult,
 } from './signal-store-models';
+import { isWritableSignal, STATE_SOURCE } from './state-source';
 
+type StripWritableSignals<State extends object> = {
+  [Property in keyof State]: State[Property] extends WritableSignal<infer Type>
+    ? Type
+    : State[Property];
+};
```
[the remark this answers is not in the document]
rainerhahnekamp: Done renaming

modules/signals/src/with-state.ts
```diff
@@ -35,21 +40,23 @@ export function withState<State extends object>(
...
+      string | symbol,
+      Signal<unknown>
+    >;
+    const stateSignals = {} as SignalsDictionary;
+    for (const key of stateKeys) {
+      const signalValue = stateAsRecord[key];
+      stateSource[key] = isWritableSignal(signalValue)
+        ? signalValue
+        : signal(signalValue);
+      stateSignals[key] = toDeepSignal(stateSource[key]);
```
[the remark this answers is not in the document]
rainerhahnekamp: Marko, don't we have to assign them twice? Once for `STATE_SOURCE` and once for the state signals? That would be variables `stateSource` and `stateSignals`.

modules/signals/src/signal-store-feature.ts
```diff
@@ -369,13 +369,13 @@ export function signalStoreFeature<
...
-    typeof featureOrInput === 'function'
-      ? [featureOrInput, ...restFeatures]
-      : restFeatures;
+  ...args:
+    | [Partial<SignalStoreFeatureResult>, ...SignalStoreFeature[]]
+    | SignalStoreFeature[]
+): SignalStoreFeature<EmptyFeatureResult, EmptyFeatureResult> {
+  const features = (
+    typeof args[0] === 'function' ? args : args.slice(1)
+  ) as SignalStoreFeature[];
```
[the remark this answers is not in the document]
rainerhahnekamp: With the change for the types, the typings stopped working for `signalStoreFeature` (not for `signalStore`). I had to apply this change to make it work again.

modules/signals/src/signal-state.ts
```diff
@@ -1,18 +1,48 @@
...
 
 export function signalState<State extends object>(
   initialState: State
 ): SignalState<State> {
-  const stateSource = signal(initialState as State);
-  const signalState = toDeepSignal(stateSource.asReadonly());
+  const stateKeys = Reflect.ownKeys(initialState);
+  const stateAsRecord = initialState as Record<string | symbol, unknown>;
+
+  // define STATE_SOURCE property
```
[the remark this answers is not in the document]
rainerhahnekamp: Done

modules/signals/src/signal-state.ts
```diff
@@ -1,18 +1,48 @@
...
 export type SignalState<State extends object> = DeepSignal<State> &
   WritableStateSource<State>;
 
 export function signalState<State extends object>(
   initialState: State
 ): SignalState<State> {
-  const stateSource = signal(initialState as State);
-  const signalState = toDeepSignal(stateSource.asReadonly());
+  const stateKeys = Reflect.ownKeys(initialState);
+  const stateAsRecord = initialState as Record<string | symbol, unknown>;
```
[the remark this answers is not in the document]
rainerhahnekamp: Done for `stateAsRecord`. `stateKeys` is used three times.

modules/signals/src/signal-state.ts
```diff
@@ -1,18 +1,48 @@
...
 import { DeepSignal, toDeepSignal } from './deep-signal';
+import { SignalsDictionary } from './signal-store-models';
+import { STATE_SOURCE, WritableStateSource } from './state-source';
 
 export type SignalState<State extends object> = DeepSignal<State> &
   WritableStateSource<State>;
 
 export function signalState<State extends object>(
   initialState: State
 ): SignalState<State> {
```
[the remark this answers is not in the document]
rainerhahnekamp: @timdeschryver, @markostanimirovic. Your call. I'm happy with extending `signalState` or keeping it as it is.

Just for clarity. It should be possible to do something like this.

```typescript
const userSignal = signal({ id: 1, name: 'John' });
const userState = signalState({ user: userSignal });
```

rainerhahnekamp:
@markostanimirovic, I've updated the code or - where applicable - answered your comments. Please check, once you have time.

I've also fixed a bug in af974f9b.

I see commits from main have been merged. Can I rebase them instead or does GitHub do that automatically meanwhile?

modules/signals/src/signal-store-feature.ts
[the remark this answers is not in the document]
markostanimirovic: I'll take a closer look

modules/signals/src/state-source.ts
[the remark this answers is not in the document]
markostanimirovic: Thanks for the explanation. I'll take a look

modules/signals/src/with-state.ts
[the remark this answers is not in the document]
markostanimirovic: Oh sorry, my bad. The second assignment is for `stateSignals`, not `stateSource`. All good here 👍

modules/signals/src/state-source.ts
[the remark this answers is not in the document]
markostanimirovic: Currently, it's also not 'allowed' to have optional state slices. If a state slice is not provided to the initial state, the state signal is never created.

modules/signals/src/signal-state.ts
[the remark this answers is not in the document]
markostanimirovic: sounds good to me 👍

markostanimirovic:
> @markostanimirovic, I've updated the code or - where applicable - answered your comments. Please check, once you have time.
> 
> I've also fixed a bug in [af974f9](https://github.com/ngrx/platform/commit/af974f9b9f408192af2b467fa899d8888adce53d).
> 
> I see commits from main have been merged. Can I rebase them instead or does GitHub do that automatically meanwhile?

You can sync changes from `main` manually and push the changes, or use the "Update branch" button which is available on the PR page.

Btw, lint is also failing.

modules/signals/src/signal-state.ts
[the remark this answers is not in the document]
timdeschryver: For me as well, I will go over this PR tomorrow/wednesday.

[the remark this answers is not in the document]
rainerhahnekamp: Alright, then let me add user-defined Signals to `signalState` as well.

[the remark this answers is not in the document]
rainerhahnekamp: Done.

Implementation Detail: I moved the type `StateResult` into the `SignalState` instead of using it as return type of `signalState`. That has the advantage that the type `SignalState<State>` can be kept.

rainerhahnekamp:
@markostanimirovic, @timdeschryver 

- `signalState` now also support user-defined writable Signals
- I've updated the docs, by adding an example in `signalState` for both old and new website.

I will push the PR for `withLinkedState` later today.

projects/www/src/app/pages/guide/signals/signal-state.md
```diff
@@ -54,6 +54,35 @@ console.log(firstName()); // logs: 'Eric'
 console.log(lastName()); // logs: 'Clapton'
 ```
 
+If the root properties of a state are already of type `WritableSignal`, they will be reused, instead of creating new signals.
+This allows an integration of external `WritableSignal`s — such as `linkedSignal` or `resource.value`.
```
timdeschryver:
```suggestion
If the root properties of a state are already a `WritableSignal`, then they are reused instead of creating new signals.
This allows the integration of external `WritableSignal`s, such as `linkedSignal` or `resource.value`.
```

projects/ngrx.io/content/guide/signals/signal-state.md
```diff
@@ -54,6 +54,20 @@ console.log(firstName()); // logs: 'Eric'
...
+```ts
+const firstName = user.firstName; // type: Signal<string>
+const lastName = user.lastName; // type: Signal<string>
+
+console.log(firstName()); // logs: 'Eric'
+console.log(lastName()); // logs: 'Clapton'
+```
+
+If the root properties of a state are already of type `WritableSignal`, they will be reused, instead of creating new signals.
+This allows an integration of external `WritableSignal`s — such as `linkedSignal` or `resource.value`.
```
timdeschryver:
```suggestion
If the root properties of a state are already a `WritableSignal`, then they are reused instead of creating new signals.
This allows the integration of external `WritableSignal`s, such as `linkedSignal` or `resource.value`.
```

modules/signals/src/state-source.ts
```diff
@@ -10,16 +10,26 @@ import {
...
+    [Property in keyof State]: WritableSignal<State[Property]>;
+  };
 };
 
 export type StateSource<State extends object> = {
-  [STATE_SOURCE]: Signal<State>;
+  [STATE_SOURCE]: { [Property in keyof State]: Signal<State[Property]> };
+};
+
+export type StateResult<StateInput extends object> = {
```
timdeschryver, line 29: I wonder if this this should also be included in the barrel file as an exported type.

rainerhahnekamp: I’m not sure. What criteria do you usually follow for exposing types?
If I don’t see an immediate need, I tend to hold off and wait for a concrete use case.

timdeschryver: I was thinking about it because I noticed the test changes, and also because the other types that are exported. I'm fine with keeping it as is, we can always revisit this later if there's a need.
👍

projects/www/src/app/pages/guide/signals/signal-state.md
```diff
@@ -54,6 +54,35 @@ console.log(firstName()); // logs: 'Eric'
...
+
+```ts
+import { linkedSignal } from '@angular/core';
+import { signalState } from '@ngrx/signals';
```
michael-small:
```suggestion
import { linkedSignal, signal } from '@angular/core';
import { patchState, signalState } from '@ngrx/signals';
```

modules/signals/src/signal-store-feature.ts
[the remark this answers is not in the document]
timdeschryver: The closest I can come up with is the following signature, but this uses the `any` type , which isn't ideal.

```ts
export function signalStoreFeature(
  featureOrInput: SignalStoreFeature | Partial<SignalStoreFeatureResult>,
  ...restFeatures: SignalStoreFeature<any, SignalStoreFeatureResult>[]
): SignalStoreFeature<EmptyFeatureResult, EmptyFeatureResult> {
```

modules/signals/src/state-source.ts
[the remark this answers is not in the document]
timdeschryver: I think using `NoInfer` is fine here, otherwise the `State` type will be updated based on the updaters types. AFAIK this makes sure the `State` type remains intact.

timdeschryver (approved):

projects/www/src/app/pages/guide/signals/signal-state.md
timdeschryver: I think this one needs to be committed as well.

modules/signals/spec/with-state.spec.ts
```diff
@@ -1,18 +1,25 @@
...
+  linkedSignal,
+  resource,
+  signal,
+} from '@angular/core';
+import { TestBed } from '@angular/core/testing';
 import { withComputed, withMethods, withState } from '../src';
-import { STATE_SOURCE } from '../src/state-source';
-import { getInitialInnerStore } from '../src/signal-store';
+import { getInitialInnerStore, signalStore } from '../src/signal-store';
+import { getState, patchState } from '../src/state-source';
```
markostanimirovic: Check unused imports. Most of them them seem unused. Also, public APIs, like `getState`, `patchState`, etc. should be imported from the `../src` path.

Let's also add one or multiple unit tests (without `signalStore` usage) for `withState` to validate the changed behavior.

modules/signals/spec/state-source.spec.ts
```diff
@@ -145,6 +159,94 @@ describe('StateSource', () => {
...
+        const userState = signalState(initialState);
+
+        patchState(userState, { middleName: 'Michael' });
+
+        expect(userState()).toEqual({ id: 1, middleName: 'Michael' });
+        expect(consoleWarnSpy).not.toHaveBeenCalled();
+      });
+    });
+
+    it('should only patch affected root properties', () => {
```
markostanimirovic: nit: for consistency

```suggestion
    it('patches only affected root properties', () => {
```

modules/signals/spec/state-source.spec.ts
```diff
@@ -192,6 +294,27 @@ describe('StateSource', () => {
         expect(executionCount).toBe(2);
       });
     });
+
+    it('does not support a dynamic type as state', () => {
```
markostanimirovic:
```suggestion
    it('does not support a dynamic dictionary as state', () => {
```

modules/signals/spec/state-source.spec.ts
```diff
@@ -344,4 +467,125 @@ describe('StateSource', () => {
       });
     });
   });
+
+  describe('integration of Signals natively', () => {
```
markostanimirovic: Maybe this description would be more precise:

```suggestion
  describe('user-defined signals as state slices', () => {
```

modules/signals/spec/state-source.spec.ts
```diff
@@ -344,4 +467,125 @@ describe('StateSource', () => {
...
+
+          await new Promise((resolve) => setTimeout(resolve));
+          expect(store.user()).toEqual({ id: 1, name: 'John Doe' });
+
+          resourceTrigger.set(2);
+          await new Promise((resolve) => setTimeout(resolve));
+          expect(store.user()).toEqual({ id: 2, name: 'John Doe' });
+        });
+
+        it('allows mixed writable Signal Types', () => {
```
markostanimirovic:
```suggestion
        it('allows mixing signals and plain values', () => {
```

modules/signals/src/state-source.ts
```diff
@@ -30,40 +40,80 @@ export type StateWatcher<State extends object> = (
...
   );
 
+  const signals = stateSource[STATE_SOURCE];
+  const stateKeys = Reflect.ownKeys(stateSource[STATE_SOURCE]);
+  for (const key of Reflect.ownKeys(newState)) {
+    if (!stateKeys.includes(key)) {
+      if (typeof ngDevMode !== 'undefined' && ngDevMode) {
+        console.warn(
+          '@ngrx/signals: Skipping update for unknown property in state source.',
+          `Property: ${String(key)}`
```
markostanimirovic: "state source" may be confusing for users who are not familiar with SignalStore/State internal implementation. Message:

```suggestion
          `@ngrx/signals: patchState was called with an unknown state slice '${String(key)}'.`,
          'Ensure that all state properties are explicitly defined in the initial state.',
          'Updates to properties not present in the initial state will be ignored.',
```

modules/signals/src/state-source.ts
```diff
@@ -30,40 +40,80 @@ export type StateWatcher<State extends object> = (
...
+        );
+      }
+      continue;
+    }
+    const signalKey = key as keyof State;
+
+    if (currentState[signalKey] === newState[signalKey]) {
+      continue;
+    }
+
```
markostanimirovic: this check can be removed. It already exists in set fn implementation: https://github.com/angular/angular/blob/main/packages/core/primitives/signals/src/signal.ts#L91

modules/signals/src/state-source.ts
```diff
@@ -30,40 +40,80 @@ export type StateWatcher<State extends object> = (
   state: NoInfer<State>
 ) => void;
 
+export function isWritableSignal(
+  signal: unknown
+): signal is WritableSignal<unknown> {
```
markostanimirovic: input can be any value, not just a signal, so we can rename it to `value` or something similar:

```suggestion
  value: unknown
): value is WritableSignal<unknown> {
```

projects/ngrx.io/content/guide/signals/signal-state.md
```diff
@@ -54,6 +54,20 @@ console.log(firstName()); // logs: 'Eric'
...
+const firstName = user.firstName; // type: Signal<string>
+const lastName = user.lastName; // type: Signal<string>
+
+console.log(firstName()); // logs: 'Eric'
+console.log(lastName()); // logs: 'Clapton'
+```
+
+If the root properties of a state are already a `WritableSignal`, then they are reused instead of creating new signals.
+This allows the integration of external `WritableSignal`s, such as `linkedSignal` or `resource.value`.
+
```
markostanimirovic: The previous paragraph and example are duplicated here accidentally. The example for the new `signalState` behavior is missing.

Let's leave 'Creating a SignalState' section as it is, and after 'Custom State Updaters' (before 'Usage') add a new section. Note: the new section should be added after 'Updating State' section because the example uses `patchState`. I also suggest not mentioning implementation details, like - "they are reused instead of creating new signals".

```md
## Using User-Defined Signals

The `signalState` function supports integrating user-defined `WritableSignal` instances as part of the state.

If a root state property is a `WritableSignal`, its value becomes an integral part of the state.
The SignalState instance and the original signal remain synchronized - updating one will immediately reflect in the other.

\```ts
const name = signal('Jimi Hendrix');
const userState = signalState({
  // 👇 Providing an external signal as part of the initial state.
  name,
  isAdmin: false,
});

console.log(userState.name()); // logs: Jimi Hendrix

// Updating the external signal
name.set('Brian May');
// reflects the value in the SignalState instance.
console.log(userState.name()); // logs: Brian May

// Updating the SignalState instance
patchState(userState, { name: 'Eric Clapton' });
// reflects the value in the external signal.
console.log(name()); // logs: Eric Clapton
\```
```

We should also add an alert/paragraph with similar content for `withState`.

modules/signals/src/state-source.ts
rainerhahnekamp: Good catch Marko!

modules/signals/spec/with-state.spec.ts
rainerhahnekamp: Done

modules/signals/src/state-source.ts
rainerhahnekamp: Done

projects/ngrx.io/content/guide/signals/signal-state.md
rainerhahnekamp: Done

rainerhahnekamp:
@markostanimirovic: I think we can start a new review round. There weren't actually that many changes:
- Applied your suggestions
- Added two tests for `withState`
- Removed the "equal check" in `patchState`
- Added an alert infobox about user-defined Signals for `withState`

modules/signals/src/state-source.ts
```diff
@@ -30,40 +40,83 @@ export type StateWatcher<State extends object> = (
...
+          'Ensure that all state properties are explicitly defined in the initial state.',
+          'Updates to properties not present in the initial state will be ignored.'
+        );
+      }
+      continue;
+    }
+    const signalKey = key as keyof State;
+
+    if (currentState[signalKey] === newState[signalKey]) {
+      // continue;
```
timdeschryver: I think this can be removed

modules/signals/src/state-source.ts
```diff
@@ -30,40 +40,83 @@ export type StateWatcher<State extends object> = (
...
+    }
```
markostanimirovic:
```suggestion
```

projects/ngrx.io/content/guide/signals/signal-store/index.md
```diff
@@ -49,6 +49,27 @@ The `BookSearchStore` instance will contain the following properties:
 
 <div class="alert is-helpful">
 
+Similar to `signalState`, `withState` accepts user-defined `WritableSignal` 
+instances as part of the state.
```
markostanimirovic: Let's add a bit more details here (assuming that reader is not familiar with `signalState`):

```suggestion
In addition to plain values, the `withState` feature also accepts user-defined `WritableSignal` instances as part of the state.
In this case, the SignalStore's state and the original signal remain synchronized - updating one will immediately reflect in the other.
```

projects/ngrx.io/content/guide/signals/signal-store/index.md
```diff
@@ -49,6 +49,27 @@ The `BookSearchStore` instance will contain the following properties:
...
+import { Book } from './book';
+
+const booksSignal = signal<Book[]>([]);
+
+export const BookSearchStore = signalStore(
+  withState({
+    books: booksSignal,
+    isLoading: false
+  })
+);
```
markostanimirovic: and an example that shows how it works (note: signal suffix is removed from the variable name):

```suggestion
import { signalStore, withState } from '@ngrx/signals';
import { Book } from './book';

const books = signal<Book[]>([]);

const BookSearchStore = signalStore(
  withState({
    // 👇 Providing an external signal as part of the initial state.
    books,
    isLoading: false,
  })
);
```

modules/signals/spec/state-source.spec.ts
```diff
@@ -145,6 +159,95 @@ describe('StateSource', () => {
...
+        const userState = signalState(initialState);
+
+        patchState(userState, { middleName: 'Michael' });
+
+        expect(userState()).toEqual({ id: 1, middleName: 'Michael' });
+        expect(consoleWarnSpy).not.toHaveBeenCalled();
+      });
+    });
+
+    it("sets all root proprerites and relies on the Signal's equals function", () => {
```
markostanimirovic:
```suggestion
    it("sets all root properties and relies on the Signal's equals function", () => {
```

markostanimirovic (changes requested):

projects/www/src/app/pages/guide/signals/signal-state.md
```diff
@@ -54,6 +54,35 @@ console.log(firstName()); // logs: 'Eric'
...
+console.log(userState.user()); // logs: { id: 1, firstName: '', lastName: '' }
+patchState(userState, {
+  user: { id: 2, firstName: 'Brian', lastName: 'May' },
+});
+console.log(userState.user()); // logs: { id: 2, firstName: 'Brian', lastName: 'May' }
+
+referenceId.set(3);
+console.log(userState.user()); // logs: { id: 3, firstName: '', lastName: '' }
+```
+
```
markostanimirovic: We can revert changes from the new docs website and synchronize content when the website is ready.

```suggestion
```

markostanimirovic (approved):

PR #4798, opened 2025-05-27
test(signals): improve signalMethod injector tests

iliev-ilian:
Small improvements in signal method tests:
- Test "does not cause issues if destroyed signalMethodFn contains destroyed effectRefs" - summand2 should be set with different value so the effect is triggered for summand2 and it is not triggered for summand1, which uses the detroyed Injector. Otherwise it causes confusion that may be also the second summand is not "observed" from the effect. It is, but the set is called with the same value.
- Test "uses the provided injector (source injector) on creation" - one small additional same expect to show that after the destroy of Injector, the signal is not tracked in the effect.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[ x] No
```

### DIFF
--- a/modules/signals/spec/signal-method.spec.ts
+++ b/modules/signals/spec/signal-method.spec.ts
@@ -103,11 +103,11 @@ describe('signalMethod', () => {
       childInjector.destroy();
 
       summand1.set(2);
-      summand2.set(2);
+      summand2.set(3);
       TestBed.flushEffects();
 
       adder.destroy();
-      expect(a).toBe(4);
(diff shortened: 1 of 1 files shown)


markostanimirovic (approved):

markostanimirovic:
Thanks @iliev-ilian!

PR #4822, opened 2025-06-07
feat(signals): enhance `withComputed` to accept computation functions

### LINKED ISSUE #4782: `@ngrx/signals`: Allow returning computation functions from `withComputed` callback

signals

### Information

Provide the ability to return computation functions from `withComputed` callback in addition to signals to reduce repetitive code in the most common use case:

```ts
const CounterStore = signalStore(
  withState({ count: 0 }),
  withComputed(({ count }) => ({
    doubleCount: () => count() * 2,
  })),
);
```

If the computed signal needs to be used before it's returned or if it has custom `equal` function, it's still necessary to explicitly use `computed` API:

```ts
const CounterStore = signalStore(
  withState({ count: 0 }),
  // 1: custom equal
  withComputed(({ count }) => ({
    doubleCount: computed(() => count() * 2, { equal: /* ... */ }),
  })),
  // 2: using computed signal before returning it:
  withComputed({ count }) => ({
    const doubleCount = computed(() => count() * 2);

    return {
      doubleCountPlus1: () => doubleCount() + 1,
      doubleCount,
    };
  })),
);
```

_No response_

- [ ] Yes
- [ ] No

rainerhahnekamp:
Example:

```ts
const CounterStore = signalStore(
  withState({ count: 0 }),
  withComputed(({ count }) => ({
    doubleCount: () => count() * 2,
  })),
);
```

- Updated withComputed to automatically wrap functions in computed signals.
- Added tests to verify automatic computation and mixing user-defined computeds.
- Improved documentation to clarify the factory function's return type.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4782

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/spec/types/with-computed.types.spec.ts
+++ b/modules/signals/spec/types/with-computed.types.spec.ts
@@ -19,7 +19,7 @@ describe('withComputed', () => {
     compilerOptions()
   );
 
-  it('creates a Signal automtically', () => {
+  it('creates a Signal automatically', () => {
     const snippet = `
       const Store = signalStore(
         withComputed(() => ({
--- a/modules/signals/spec/types/with-computed.types.spec.ts
+++ b/modules/signals/spec/types/with-computed.types.spec.ts
@@ -3,7 +3,7 @@ import { compilerOptions } from './helpers';
 import { signalStore, withComputed } from 'modules/signals/src';
 import { TestBed } from '@angular/core/testing';
 
-describe('patchState', () => {
+describe('withComputed', () => {
   const expectSnippet = expecter(
     (code) => `
         import {
--- a/modules/signals/spec/types/with-computed.types.spec.ts
+++ b/modules/signals/spec/types/with-computed.types.spec.ts
@@ -1,0 +1,90 @@
+import { expecter } from 'ts-snippet';
+import { compilerOptions } from './helpers';
+import { signalStore, withComputed } from 'modules/signals/src';
+import { TestBed } from '@angular/core/testing';
+
+describe('patchState', () => {
+  const expectSnippet = expecter(
+    (code) => `
+        import {
+          deepComputed,
+          signalStore,
+          withComputed,
+        } from '@ngrx/signals';
+        import { TestBed } from '@angular/core/testing';
+        import { signal } from '@angular/core';
+        
+        ${code}
+      `,
+    compilerOptions()
+  );
+
+  it('creates a Signal automtically', () => {
+    const snippet = `
+      const Store = signalStore(
+        withComputed(() => ({
+          user: () => ({ firstName: 'John', lastName: 'Doe' })
+        }))
+      );
+
+      const store = TestBed.inject(Store);
+      const user = store.user;
+    `;
+
+    expectSnippet(snippet).toSucceed();
+    expectSnippet(snippet).toInfer(
+      'user',
+      'Signal<{ firstName: string; lastName: string; }>'
+    );
+  });
+
+  it('keeps a WritableSignal intact, if passed', () => {
+    const snippet = `
+      const user = signal({ firstName: 'John', lastName: 'Doe' });
+
+      const Store = signalStore(
+        withComputed(() => ({
+          user,
+        }))
+      );
+
+      const store = TestBed.inject(Store);
+      const userSignal = store.user;
+    `;
+
+    expectSnippet(snippet).toSucceed();
+    expectSnippet(snippet).toInfer(
+      'userSignal',
+      'WritableSignal<{ firstName: string; lastName: string; }>'
+    );
+  });
+
+  it('keeps a DeepSignal intact, if passed', () => {
+    const snippet = `
+    const user = deepComputed(
+      signal({
+        name: 'John Doe',
+        address: {
+          street: '123 Main St',
+          city: 'Anytown',
+        },
+      })
+    );
+
+    const Store = signalStore(
+      withComputed(() => ({
+        user,
+      }))
+    );
+
+    const store = TestBed.inject(Store);
+    const userSignal = store.user;
+  `;
+
+    expectSnippet(snippet).toSucceed();
+    expectSnippet(snippet).toInfer(
+      'userSignal',
+      'DeepSignal<{ name: string; address: { street: string; city: string; }; }>'
+    );
+  });
+});
--- a/modules/signals/spec/with-computed.spec.ts
+++ b/modules/signals/spec/with-computed.spec.ts
@@ -1,5 +1,10 @@
-import { signal } from '@angular/core';
-import { withComputed, withMethods, withState } from '../src';
+import { computed, signal } from '@angular/core';
+import {
+  signalStoreFeature,
+  withComputed,
+  withMethods,
+  withState,
+} from '../src';
 import { getInitialInnerStore } from '../src/signal-store';
 
 describe('withComputed', () => {
@@ -53,5 +58,35 @@ describe('withComputed', () => {
       'Trying to override:',
       'p1, s2, m1, Symbol(computed_secret)'
     );
+  });
+
+  it('adds computed automatically if the value is function', () => {
+    const initialStore = getInitialInnerStore();
+
+    const store = signalStoreFeature(
+      withState({ a: 2, b: 3 }),
+      withComputed(({ a, b }) => ({
+        sum: () => a() + b(),
+        product: () => a() * b(),
+      }))
+    )(initialStore);
+
+    expect(store.props.sum()).toBe(5);
+    expect(store.props.product()).toBe(6);
+  });
+
+  it('allows to mix user-provided computeds and automatically computed ones', () => {
+    const initialStore = getInitialInnerStore();
+
+    const store = signalStoreFeature(
+      withState({ a: 2, b: 3 }),
+      withComputed(({ a, b }) => ({
+        sum: () => a() + b(),
+        product: computed(() => a() * b()),
+      }))
+    )(initialStore);
+
+    expect(store.props.sum()).toBe(5);
+    expect(store.props.product()).toBe(6);
   });
 });
--- a/modules/signals/spec/with-computed.spec.ts
+++ b/modules/signals/spec/with-computed.spec.ts
@@ -62,7 +62,7 @@ describe('withComputed', () => {
     );
   });
 
-  it('adds computed automatically if the value is function', () => {
+  it('adds computed automatically if the value is a function', () => {
     const initialStore = getInitialInnerStore();
 
     const store = signalStoreFeature(
--- a/modules/signals/spec/with-computed.spec.ts
+++ b/modules/signals/spec/with-computed.spec.ts
@@ -1,11 +1,13 @@
 import { computed, signal } from '@angular/core';
 import {
+  deepComputed,
   signalStoreFeature,
   withComputed,
   withMethods,
   withState,
 } from '../src';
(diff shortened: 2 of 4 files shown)


projects/ngrx.io/content/guide/signals/signal-store/index.md
```diff
@@ -145,7 +145,7 @@ export class BookSearch {
 
 Computed signals can be added to the store using the `withComputed` feature.
 This feature accepts a factory function as an input argument, which is executed within the injection context.
-The factory should return a dictionary of computed signals, utilizing previously defined state signals and properties that are accessible through its input argument.
+The factory should return a dictionary containing either computed signals or functions that return values (which will automatically be wrapped in computed signals), utilizing previously defined state signals and properties that are accessible through its input argument.
```
timdeschryver:
```suggestion
The factory should return a dictionary containing either computed signals or functions that return values (which are automatically wrapped in computed signals), utilizing previously defined state signals and properties that are accessible through its input argument.
```

modules/signals/spec/with-computed.spec.ts
```diff
@@ -54,4 +59,34 @@ describe('withComputed', () => {
       'p1, s2, m1, Symbol(computed_secret)'
     );
   });
+
+  it('adds computed automatically if the value is function', () => {
```
markostanimirovic: we can add a type test to `signal-store.types.spec.ts` to validate the new `withComputed` behavior just in case

modules/signals/src/with-computed.ts
```diff
@@ -1,22 +1,49 @@
...
+> = {
+  [P in keyof ComputedSignals]: ComputedSignals[P] extends () => infer V
+    ? Signal<V>
+    : ComputedSignals[P];
+};
+
 export function withComputed<
   Input extends SignalStoreFeatureResult,
-  ComputedSignals extends SignalsDictionary
+  ComputedSignals extends Record<
```
markostanimirovic: They are not only computed signals anymore:

```suggestion
  ComputedDictionary extends Record<
```

modules/signals/src/with-computed.ts
```diff
@@ -1,22 +1,49 @@
...
-  SignalsDictionary,
   SignalStoreFeature,
   SignalStoreFeatureResult,
   StateSignals,
 } from './signal-store-models';
 import { Prettify } from './ts-helpers';
 import { withProps } from './with-props';
 
+type ComputedResult<
+  ComputedSignals extends Record<
```
markostanimirovic:
```suggestion
  ComputedInput extends Record<
```

modules/signals/src/with-computed.ts
```diff
@@ -1,22 +1,49 @@
...
+    string | symbol,
+    Signal<unknown> | (() => unknown)
+  >
+> = {
+  [P in keyof ComputedSignals]: ComputedSignals[P] extends () => infer V
+    ? Signal<V>
+    : ComputedSignals[P];
```
markostanimirovic: Provided `Signal`s are also compatible with `() => infer V` type, but we should not change their type (e.g. when `DeepSignal` is provided to `withComputed` by using `deepComputed` helper).

```suggestion
  [K in keyof ComputedInput]: ComputedInput[K] extends Signal<any>
    ? ComputedInput[K]
    : ComputedInput[K] extends () => infer V ? Signal<V> : never;
```

When adding a type test, please also add one computed signal by using `deepComputed` and one writable signal to ensure their type hasn't been changed.

modules/signals/src/with-computed.ts
```diff
@@ -1,22 +1,49 @@
...
+
 export function withComputed<
   Input extends SignalStoreFeatureResult,
-  ComputedSignals extends SignalsDictionary
+  ComputedSignals extends Record<
+    string | symbol,
+    Signal<unknown> | (() => unknown)
+  >
 >(
   signalsFactory: (
```
markostanimirovic: we can also rename this:

```suggestion
  computedFactory: (
```

modules/signals/src/with-computed.ts
```diff
@@ -1,22 +1,49 @@
...
     store: Prettify<StateSignals<Input['state']> & Input['props']>
   ) => ComputedSignals
 ): SignalStoreFeature<
   Input,
-  { state: {}; props: ComputedSignals; methods: {} }
+  { state: {}; props: ComputedResult<ComputedSignals>; methods: {} }
 > {
-  return withProps(signalsFactory);
+  return withProps((store) => {
+    const signals = signalsFactory(store);
```
markostanimirovic:
```suggestion
    const computedResult = computedFactory(store);
```

modules/signals/src/with-computed.ts
```diff
@@ -1,22 +1,49 @@
...
+    const stateKeys = Reflect.ownKeys(signals);
```
markostanimirovic:
```suggestion
    const computedResultKeys = Reflect.ownKeys(signals);
```

modules/signals/src/with-computed.ts
```diff
@@ -1,22 +1,49 @@
...
+
+    return stateKeys.reduce((prev, key) => {
+      const signalOrFunction = signals[key];
+      return {
+        ...prev,
+        [key]: isSignal(signalOrFunction)
+          ? signalOrFunction
+          : computed(signalOrFunction),
+      };
+    }, {} as ComputedResult<ComputedSignals>);
```
markostanimirovic:
```suggestion
    return computedResultKeys.reduce((prev, key) => {
      const signalOrComputation = computationResult[key];
      return {
        ...prev,
        [key]: isSignal(signalOrComputation)
          ? signalOrComputation
          : computed(signalOrComputation),
      };
    }, {} as ComputedResult<ComputedDictionary>);
```

projects/ngrx.io/content/guide/signals/signal-store/index.md
markostanimirovic:
```suggestion
The factory should return a dictionary containing either computed signals or computation functions (which are automatically converted to computed signals), utilizing previously defined state signals and properties that are accessible through its input argument.
```

markostanimirovic (changes requested):

projects/ngrx.io/content/guide/signals/signal-store/index.md
rainerhahnekamp: Done

modules/signals/spec/with-computed.spec.ts
rainerhahnekamp: Done

modules/signals/src/with-computed.ts
rainerhahnekamp: Done

modules/signals/src/with-computed.ts
rainerhahnekamp: Done, I went here also with the term `ComputedDictionary`. Also used `unknown` instead `any`. Please see me question in the comment below.

modules/signals/src/with-computed.ts
rainerhahnekamp: Done

modules/signals/src/with-computed.ts
rainerhahnekamp: Done

projects/ngrx.io/content/guide/signals/signal-store/index.md
rainerhahnekamp: @markostanimirovic @timdeschryver, do we want to use the term "computation functions"? I am afraid that this could confuse users.

rainerhahnekamp:
@markostanimirovic, @timdeschryver.

I've applied the suggested changes with the following additions:
- `ComputedResult` calls its type now also `ComputedRecord` instead `ComputedInput`.
- @markostanimirovic, you were suggesting `[P in keyof ComputedDictionary]: ComputedDictionary[P] extends Signal<any>`; I used `[P in keyof ComputedDictionary]: ComputedDictionary[P] extends Signal<unknown>`. Is there a reason why `any` should be preferred over `unkown`. 
- Type tests have been added to verify that we stick to the `Signal` type when that one is provided.

rainerhahnekamp:
As discussed, the latest commit enables access to methods within withComputed. This is a necessary step toward supporting a potential withResource feature in the 20.x release line.

When a resource is in an error state, accessing its value will throw. Therefore, users must first verify the presence of a value using the hasValue method.

In the meantime, I am going to prepare the updated docs (in a separate PR).

Since this is a significant change, I've also asked @timdeschryver for a re-review.

```typescript
const Store = signalStore(
  withResource(() =>
    // spreads members of the resource to the store, including hasValue as method
    resource({
      loader: () => Promise.resolve({ firstName: 'John', lastName: 'Doe' }),
    })
  ),
  withComputed((store) => ({
    safeUser: () =>
      store.user.hasValue()
        ? store.user.value()
        : { firstName: '', lastName: '' },
  }))
);
```

modules/signals/spec/types/with-computed.types.spec.ts
```diff
@@ -0,0 +1,90 @@
+import { expecter } from 'ts-snippet';
+import { compilerOptions } from './helpers';
+import { signalStore, withComputed } from 'modules/signals/src';
+import { TestBed } from '@angular/core/testing';
+
+describe('patchState', () => {
```
timdeschryver:
```suggestion
describe('withComputed', () => {
```

modules/signals/spec/with-computed.spec.ts
```diff
@@ -54,4 +61,87 @@ describe('withComputed', () => {
...
```
timdeschryver:
```suggestion
  it('adds computed automatically if the value is a function', () => {
```

modules/signals/spec/types/with-computed.types.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+        } from '@ngrx/signals';
+        import { TestBed } from '@angular/core/testing';
+        import { signal } from '@angular/core';
+        
+        ${code}
+      `,
+    compilerOptions()
+  );
+
+  it('creates a Signal automtically', () => {
```
michael-small:
```suggestion
  it('creates a Signal automatically', () => {
```

maxime1992:
I'd very much like to have the ability to provide a function with a parameter that returns a signal in order to create an equivalent of a factory selector in ngrx store world. I know this can be achieved in the `withMethod` block but wouldn't it make sense to be able to do that from `withComputed` when it just all comes down to dealing with signals and not just observables etc?

Just asking because of this comment which would I assume make that hard:

> Updated withComputed to automatically wrap functions in computed signals

But maybe the recommendation is to just use `withMethod` for that?

rainerhahnekamp:
@maxime1992 

> I'd very much like to have the ability to provide a function with a parameter that returns a signal in order to create an equivalent of a factory selector in ngrx store world.

Yes, factory functions for computed are actually methods. So they should go in `withMethods`. The computed itself is of course also a function in technical terms, but I hope you see the difference in a function creating a computed vs. a function  which is wrapped by a computed

projects/ngrx.io/content/guide/signals/signal-store/index.md
timdeschryver: I think this is fine. We can always revisit later.

markostanimirovic (approved):
LGTM 👍 Great work Rainer! 👏

PR #4887, opened 2025-07-16
chore: migrate eslint config to flat format

### LINKED ISSUE #4883: Request: Migrate to ESLint Flat Config Format

component-store

### Information

A flat config is now the default.  See [ES Lint blog](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/).  
Nx has a simple migration: `npx nx generate @nx/eslint:convert-to-flat-config`.  
I tested it briefly on a fresh branch and I had no issues.  

_No response_

- [ ] Yes
- [ ] No

jdegand:
## PR Checklist

## PR Type

Change to ESLint flat config.

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Outdated ESLint configuration.  Failing lint command in `package.json`.

Closes #4883

## What is the new behavior?

ESLint works with `lint` command.

```
[ ] Yes
[X] No
```

## Other information

There are quite a few lint tests failing.


eslint.config.mjs
```diff
@@ -0,0 +1,110 @@
+import { FlatCompat } from '@eslint/eslintrc';
```
timdeschryver, line 1: Can you add `@ts-check` to each file please.

```suggestion
// @ts-check
import { FlatCompat } from '@eslint/eslintrc';
```

jdegand: Adding `//@ts-check`  causes an error later on with `recommendedConfig: js.configs.recommended`.  There are multiple potential fixes.  Adding `@ts-expect-error`  above the recommendedConfig line looks to be a solution.

timdeschryver: Do you have an idea why it fails? If not, that's fine and we keep it as is.
It's something I want to revisit maybe when removing some deps, or when we fully migrate to the flat config file (without compat)

jdegand: I just pushed with changes to see if it works.  I think the failure has to do with the flat config being an object.   The error is `Object literal may only specify known properties, and 'recommendedConfig' does not exist in type '{ baseDirectory?: string | undefined; resolvePluginsRelativeTo?: any; }'.ts(2353)`.

jdegand: I tried removing `recommendedConfig` totally.   This causes all linting to fail.

timdeschryver: It seems to be a version mismatch somewhere then.
Thanks for trying, you can ignore this comment ~ I will have a look at it later but this isn't blocking for this PR.

jdegand: I added another commit but I missed some files.  I will revert new changes and try a different approach on a new branch.

jdegand: Why ts-check Fails with ESLint v9 Flat Config
The issue boils down to TypeScript not understanding ESLint’s flat config structure. Here's what’s happening:

ESLint v9 uses an array of config objects in eslint.config.js, which is not a valid TypeScript type unless you explicitly annotate it.

The flat config format lacks built-in TypeScript definitions, so ts-check can't infer types like it could with legacy configs.

Plugins like @typescript-eslint/eslint-plugin still export configs in the old format (extends, etc.), which causes type mismatches when used directly.

```
// @ts-check
/** @type {import('eslint').Linter.FlatConfig[]} */
const config = [
  {
    files: ['**/*.ts'],
    languageOptions: {
      parser: require('@typescript-eslint/parser'),
    },
    rules: {
      '@typescript-eslint/no-unused-vars': 'error',
    },
  },
];

module.exports = config;
```

This doesn't work as `Linter.FlatConfig` is deprecated and you have to use `Linter.Config`.  Plus, other issues.

Exploring converting to Typescript file with format like:
```
import type { FlatConfig } from 'eslint';
const config: FlatConfig[] = [/* your config here */];
export default config;
```

Using typescript requires changing the `tsconfig.json`'s module type.

There might be other unforeseen issues as it might be early to easily convert this to Typescript.  Type packages likely have to be installed as well.

jdegand:
```ts
import type { Linter } from 'eslint';
import { FlatCompat } from '@eslint/eslintrc';
import { dirname } from 'path';
import { fileURLToPath } from 'url';
import js from '@eslint/js';
import nxEslintPlugin from '@nx/eslint-plugin';

const compat = new FlatCompat({
  baseDirectory: dirname(fileURLToPath(import.meta.url)),
  recommendedConfig: js.configs.recommended,
});

const config: Linter.Config[] = [
  {
    ignores: ['**/dist'],
  },
  {
    plugins: {
      '@nx': nxEslintPlugin,
    },
  },
  {
    files: ['**/*.ts', '**/*.tsx', '**/*.js', '**/*.jsx'],
    rules: {
      '@nx/enforce-module-boundaries': [
        'error',
        {
          allow: [],
          depConstraints: [
            {
              sourceTag: '*',
              onlyDependOnLibsWithTags: ['*'],
            },
          ],
        },
      ],
      '@typescript-eslint/member-ordering': 'off',
      '@typescript-eslint/no-inferrable-types': 'warn',
    },
  },
  ...compat.config({ extends: ['plugin:@nx/typescript'] }).map((c) => ({
    ...c,
    files: ['**/*.ts', '**/*.tsx', '**/*.cts', '**/*.mts'],
    rules: {
      ...c.rules,
      'no-extra-semi': 'off',
    },
  })),
  ...compat.config({ extends: ['plugin:@nx/javascript'] }).map((c) => ({
    ...c,
    files: ['**/*.js', '**/*.jsx', '**/*.cjs', '**/*.mjs'],
    rules: {
      ...c.rules,
      'no-extra-semi': 'off',
    },
  })),
  ...compat
    .config({
      plugins: ['eslint-plugin-import', '@typescript-eslint'],
    })
    .map((c) => ({
      ...c,
      files: ['**/*.ts'],
      rules: {
        ...c.rules,
        '@typescript-eslint/no-explicit-any': 'off',
        '@typescript-eslint/no-unused-vars': [
          'warn',
          {
            argsIgnorePattern: '^_',
            varsIgnorePattern: '^_',
            caughtErrorsIgnorePattern: '^_',
          },
        ],
        '@typescript-eslint/naming-convention': 'off',
        '@typescript-eslint/prefer-namespace-keyword': 'error',
        '@typescript-eslint/no-empty-function': 'warn',
        '@typescript-eslint/no-empty-object-type': 'warn',
        eqeqeq: ['off', 'smart'],
        'id-blacklist': [
          'error',
          'any',
          'Number',
          'number',
          'String',
          'string',
          'Boolean',
          'boolean',
          'Undefined',
          'undefined',
        ],
        'id-match': 'error',
        'no-eval': 'off',
        'no-redeclare': 'off',
        'no-underscore-dangle': 'off',
        'no-var': 'error',
        'no-prototype-builtins': 'off',
        '@typescript-eslint/no-wrapper-object-types': 'off',
        '@typescript-eslint/no-unsafe-function-type': 'off',
      },
    })),
  {
    files: ['**/*.html'],
    rules: {},
  },
];

export default config;
```

timdeschryver:
@jdegand there seem to be some linting issues.
Are you planning to take a look at them?

jdegand:
@timdeschryver I think some issues stem from this in the config files: 
```ts
parserOptions: {
          project: ['modules/effects/testing/tsconfig.*?.json'],
},
```
There are issues resolving file paths.  I going to try tweaking this first. 

Changing the wildcard helped 3 more lints pass.

jdegand:
Using 
```
import baseConfig from '../../eslint.config.mjs';
import typescriptEslintEslintPlugin from '@typescript-eslint/eslint-plugin';
import cypressPlugin from 'eslint-plugin-cypress';

export default [
  {
    ignores: ['**/dist'],
  },
  ...baseConfig,

  {
    plugins: {
      '@typescript-eslint': typescriptEslintEslintPlugin,
      cypress: cypressPlugin,
    },
  },
  {
    rules: {},
  },
  {
    files: ['**/*.ts', '**/*.tsx', '**/*.js', '**/*.jsx'],
    rules: {},
    languageOptions: {
      parserOptions: {
        project: 'projects/example-app-e2e/tsconfig.*.json',
      },
    },
  },
  {
    files: ['src/plugins/index.js'],
    rules: {
      '@typescript-eslint/no-var-requires': 'off',
      'no-undef': 'off',
    },
  },
  {
    ignores: ['schematics-core'],
  },

  // ✅ Cypress E2E overrides
  {
    files: ['src/**/*.cy.{ts,tsx}'],
    plugins: {
      cypress: cypressPlugin,
    },
    languageOptions: {
      parserOptions: {
        project: null,
      },
    },
    rules: {},
  },
];
```
for the `standalone` and `example` app and not using FlatCompat at all works.  FlatCompat has potential incompatibilities.  I have tried extending the recommended config and ignoring `name` to no avail.  Also, there is a potential issue with a missing `@types/cypress` dependency in the `tsconfig.json`.  This causes a warning, but no real difference.  The types can be included or deleted.

jdegand:
There seem to be inconsistencies between the results of CI tests and what I get in a GitHub codespace.  I am getting more lint successes in a GitHub codespace.  If high CPU usage happens in the codespace, results seem to become unreliable.

timdeschryver:
@jdegand the cypress plugin will be removed in https://github.com/ngrx/platform/pull/4891.
It's ok to just remove it here as well if that makes it easier.

jdegand:
Now, I think the question mark after the wildcard didn't make a difference (`modules/effects/testing/tsconfig.*?.json`).  
`ngrx.io` config has the `*?` syntax and still works.  Let me know a preference and this can be changed.

Running `pnpm lint` causes high CPU usage in the codespace and then things become unreliable.  It is better to lint a few projects separately (`npx nx run store:lint`, etc) so a few cached values help performance of the codespace.  

I changed the two e2e configs so the `cypress-plugin` can be removed.

jdegand:
`projects/www/src/app/pages/guide/data/entity-metadata.md`  got added to the PR after merging main?  It is incorrectly formatted, but it still made it pass the husky lint?

modules/store/tsconfig.build.json
```diff
@@ -8,7 +8,9 @@
     "module": "ES2022",
     "moduleResolution": "node",
     "outDir": "../../dist/modules/store",
-    "paths": {},
+    "paths": {
+      "@ngrx/store": ["src"]
+    },
```
markostanimirovic: is this change required?

nx.json
```diff
@@ -92,7 +92,8 @@
       "!{projectRoot}/tsconfig.spec.json",
       "!{projectRoot}/jest.config.[jt]s",
       "!{projectRoot}/.eslintrc.json",
```
markostanimirovic: it seems this line can be removed

projects/www/src/app/pages/guide/data/entity-metadata.md
```diff
@@ -155,9 +155,10 @@ export function nameAndSayingFilter(
   entities: Villain[],
   pattern: string
 ) {
-  return PropsFilterFnFactory<Villain>[('name', 'saying')](
-    entities,
-    pattern
+  return (
+    PropsFilterFnFactory <
+    Villain >
+    [('name', 'saying')](entities, pattern)
```
markostanimirovic: can you revert this change?

jdegand: I used `no-verify` to revert the file.

nx.json
jdegand: Actually, there is a leftover `eslintrc.json` file in the `projects/www` folder.  Because it uses analog, I guess the nx script skipped converting it automatically.  I guess this can be converted as well.

modules/store/tsconfig.build.json
jdegand: This was part of my debugging (this is just helping to point to the correct files to lint)  and it's not necessary.  Looks like the only file I did this in.

nx.json
jdegand: www's `project.json` doesn't even have a lint section.

jdegand: I've had some issues getting this working perfectly, but it looks there will be extra lint errors that need to be fixed inside the `www` folder.

jdegand: The new style guide generated quite a few errors.

jdegand: I don't know if you want this included into this PR or a separate PR.  I created another branch to work on and implemented fixes there.  I can easily do whatever you prefer.  Although it does clutter the PR, I did remove `standalone:true` from a lot of components in the `www` folder (might as well do it now).

modules/component-store/eslint.config.mjs
```diff
@@ -0,0 +1,54 @@
...
+import baseConfig from '../../eslint.config.mjs';
+
+const compat = new FlatCompat({
+  baseDirectory: dirname(fileURLToPath(import.meta.url)),
+  recommendedConfig: js.configs.recommended,
+});
+
+export default [
+  {
+    ignores: ['**/dist', '**/jest.config.ts', '**/schematics-core/test-setup.ts', '**/schematics-core/utility/standalone.ts'],
```
timdeschryver, line 14: Is it possible to add this to the base config and remove it in the module specific configs?

In the base config:

```ts
    ignores: ['**/dist', '**/jest.config.ts', '**/schematics-core'],
```

I don't remember why we excluded these, this can be looked into later.

jdegand: All the exclusions were added to match previous files.  The lints fail without them.

jdegand: Potential option is to add `.eslintignore` files and add the files there. 

In base config, could add an overrides array like this: 

```ts
  overrides: [
    {
      files: ['**/jest.config.ts'],
      env: { node: true },
      rules: {
        // customize rules for config files
      },
    },
    {
      files: ['**/schematics-core/**/*.ts'],
      env: { node: true },
      rules: {
        // loosen rules or add globals as needed
      },
    },
  ],
```

Typically, files are ignored if they use "globals" or unusual patterns that could break if linted.  Code that uses TypeScript AST,  Node.js (node vs web environments), and schematics is also typically ignored.   That standalone utility looks a stop-gap or bridge file and may be something that could potentially be removed as `bootstrapApplication` can make standalone components.

nx.json
timdeschryver: Personally I think it's better to include it in this PR.
I wonder what the lint errors are if we just keep the same configuration.
If it's a lot, it's also ok to turn off some rules at this point.

modules/component-store/eslint.config.mjs
timdeschryver: It's fine, we don't need to add a eslintignore - AFAIK this isn't supported anymore.

nx.json
jdegand: No, it wasn't that bad.  With the config, I picked `ngrx` for all selectors.  So that choice generated a few errors and a few components needed `component` added to the name.   It seems like one of the app components isn't really used and potentially could be removed.

projects/www/src/app/app.component.ts
```diff
@@ -9,7 +9,7 @@ import { StackblitzComponent } from './components/docs/stackblitz.component';
 import { FooterComponent } from './components/footer.component';
 
 @Component({
-  selector: 'www-root',
+  selector: 'ngrx-root',
```
markostanimirovic, line 12: root selector needs to be changed in `index.html`

projects/www/src/app/app.component.ts
```diff
@@ -9,7 +9,7 @@ import { StackblitzComponent } from './components/docs/stackblitz.component';
 import { FooterComponent } from './components/footer.component';
 
 @Component({
-  selector: 'www-root',
+  selector: 'ngrx-www-root',
```
markostanimirovic:
```suggestion
  selector: 'ngrx-root',
```

projects/www/src/app/examples/__base/src/app.component.ts
```diff
@@ -1,15 +1,14 @@
 import { Component } from '@angular/core';
 
 @Component({
-  selector: 'app-root',
-  standalone: true,
+  selector: 'ngrx-app-root',
```
markostanimirovic:
```suggestion
  selector: 'ngrx-root',
```

`index.html` in `__base` directory also needs to be updated

projects/www/index.html
```diff
@@ -34,7 +34,7 @@
 
   </head>
   <body class="mat-typography mat-app-background">
-    <www-root></www-root>
+    <ngrx-www-root></ngrx-www-root>
```
markostanimirovic:
```suggestion
    <ngrx-root></ngrx-root>
```

projects/www/src/app/examples/ngrx-start/src/app.component.ts
```diff
@@ -1,8 +1,7 @@
 import { Component } from '@angular/core';
 
 @Component({
-  selector: 'app-root',
-  standalone: true,
+  selector: 'ngrx-app-root',
```
markostanimirovic:
```suggestion
  selector: 'ngrx-root',
```

projects/www/src/app/examples/signals-01/src/app.component.ts
```diff
@@ -14,15 +14,14 @@ const Store = signalStore(
 );
 
 @Component({
-  selector: 'app-root',
-  standalone: true,
+  selector: 'ngrx-app-root',
```
markostanimirovic:
```suggestion
  selector: 'ngrx-root',
```

projects/www/src/app/examples/store/src/app.component.ts
```diff
@@ -2,13 +2,12 @@ import { Component } from '@angular/core';
 import { MyCounterComponent } from './my-counter/my-counter.component';
 
 @Component({
-  selector: 'app-root',
-  standalone: true,
+  selector: 'ngrx-app-root',
```
markostanimirovic:
```suggestion
  selector: 'ngrx-root',
```

projects/www/src/app/examples/store/src/my-counter/my-counter.component.ts
```diff
@@ -3,8 +3,7 @@ import { Store } from '@ngrx/store';
 import { increment, decrement, reset } from '../counter.actions';
 
 @Component({
-  selector: 'app-my-counter',
-  standalone: true,
+  selector: 'ngrx-app-my-counter',
```
markostanimirovic:
```suggestion
  selector: 'ngrx-my-counter',
```

you can search for `app-my-counter` and update its usage (`store/src/app.component.ts`)

nx.json
markostanimirovic: with recent updates in the `www` project, can we now remove this line?

jdegand:
Still got to search through the markdown files to update tags there as well.

projects/www/src/app/examples/__base/src/app.component.ts
jdegand: ESLint tries to parse this `index.html` file and fails. Preventing this might not be worth the hassle.

timdeschryver (approved):
The failing Netlify docs seem to be a hiccup.
Retriggering them resolves the issue.

Thanks for this @jdegand

PR #4889, opened 2025-07-18
style(www): remove margin on code-example

### LINKED ISSUE #4876: [New Docs]: code-example styling
### Information

There seems to be too much padding/margin to code samples that use a header.

Image

### Documentation page

https://deploy-preview-4875--ngrx-site-v19.netlify.app/guide/component-store/usage

- [ ] Yes
- [ ] No

jdegand:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[X] Code style update (formatting, local variables)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Header margin issue.

Closes #4876 

## What is the new behavior?

I think it looks better without any margin between the code example and header.  Header already has enough padding. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/www/src/app/components/docs/code-example.component.ts
+++ b/projects/www/src/app/components/docs/code-example.component.ts
@@ -15,7 +15,6 @@ import { Component, Input } from '@angular/core';
         display: flex;
         flex-direction: column;
         border: 1px solid rgba(255, 255, 255, 0.12);
-        margin: 14px 0 24px;
       }
 
       .header {


jdegand:
On line 30 

```css
      .body {
        padding: 0 0px;
      }
```

There is a missing `px` in the padding rule.  I've noticed a lot of `0px` declarations in the styling.  This serves no purpose outside of `--var` and `calc` and I'd suggest a find and replace to just use `0`.  Or you can keep it.

jdegand:
Also, `standalone:true` is in this component as well.  There are likely quite a few of these that can be removed.

projects/www/src/app/components/docs/code-example.component.ts
```diff
@@ -15,7 +15,6 @@ import { Component, Input } from '@angular/core';
         display: flex;
         flex-direction: column;
         border: 1px solid rgba(255, 255, 255, 0.12);
-        margin: 14px 0 24px;
```
timdeschryver, line 18: We have to keep the margin in some cases, otherwise the code is "too attached" to the text above it.

Image

Previously:

Image

jdegand: Hard-coding margin makes the component inflexible.  Maybe, it is better to add an input for margin styling.  If you add just `margin-top: 8px` to the component, you'll likely have a situation where a double margin is created somewhere.  You could place a class on the app-code-example component and use that to handle margins.

timdeschryver (changes requested):
I noticed that this isn't related to the css.
Some code blocks are fine, while other are not.
The reason is the formatting in the docs.

If the code example component goes over multiple lines, the content is not added as the body, but as a separate tag under the component.

```
<ngrx-code-example header="src/app/app.module.ts">
```

not fine:


```
<ngrx-code-example 
   header="src/app/app.module.ts">
```

Closing this PR and issue, I will go over the examples to fix it.

PR #4904, opened 2025-07-29
fix(eslint-plugin): report on-function-explicit-return-type for call expressions

### LINKED ISSUE #4901: ESLint rule `onFunctionWithoutType` ignores arrow function body with call expression

eslint-plugin

import { createAction, createReducer, on } from '@ngrx/store';

export const testAction = createAction('[Test] Action');

export interface State {
  testNumber: number;
}

export const initialState: State = {
  testNumber: 0,
};

export const testReducer = createReducer(
  initialState,
  on(testAction, (state) => ({ ...state, unknownProperty: Number(2) })),
  // ^ missing return type is not reported as error by eslint
);

### Expected behavior

ESLint should report the following error:
`On` functions should have an explicit return type when using arrow functions: `on(action, (state): State => {}`. (https://ngrx.io/guide/eslint-plugin/rules/on-function-explicit-return-type) (@ngrx/on-function-explicit-return-type)

NgRx: 19.2.0
Angular: 19.4.0
Node: 22.14.0
eslint: 8.57.0

### Other information

I am not familiar with ESLint rules, but I could boil it down to the [rule definition](https://github.com/ngrx/platform/blob/f18111fbb0495da916eede622b1e5aa7dd0decbd/modules/eslint-plugin/src/utils/selectors/index.ts#L72) -> ArrowFunctionExpression:not([returnType.typeAnnotation], :has(CallExpression)) requiring a return type OR a call expression.

- [x] Yes
- [ ] No

timdeschryver:
…expressions

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4901

```
[ ] Yes
[ ] No
```

### DIFF
--- a/modules/eslint-plugin/spec/rules/store/on-function-explicit-return-type.spec.ts
+++ b/modules/eslint-plugin/spec/rules/store/on-function-explicit-return-type.spec.ts
@@ -29,7 +29,6 @@ const reducer = createReducer(
 const reducer = createReducer(
   initialState,
   on(increment, incrementFunc),
-  on(increment, s => incrementFunc(s)),
   on(increment, (s): State => incrementFunc(s)),
 )`,
   `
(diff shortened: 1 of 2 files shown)


modules/eslint-plugin/spec/rules/store/on-function-explicit-return-type.spec.ts
```diff
@@ -174,6 +173,57 @@ const reducer = createReducer(
...
+          },
+        ],
+      },
+    ],
+  },
+  {
+    code: `
+const reducer = createReducer(
+  initialState,
+  on(reset, s => foo(s)),
```
timdeschryver, line 186: Previously this was OK (see the removed valid test).
I can't remember why that's the case, I do think it's better to also add the return argument for these.

markostanimirovic (approved):

PR #4912, opened 2025-08-05
fix(component): use setTimeout instead of requestAnimationFrame in SSR mode

### LINKED ISSUE #4902: @ngrx/component does not work in SSR + zoneless

component

When using ngrxLet in a angular app with both SSR and zoneless, you're faced with an error:

uncaughtException ReferenceError: requestAnimationFrame is not defined.

I made a quick example:
https://github.com/JelleBruisten/ngrxlet-zoneless-ssr

It looks like due that when zoneless we use the AnimationFrameTickScheduler
```ts
@Injectable({
  providedIn: 'root',
  useFactory: () => {
    const zone = inject(NgZone);
    return isNgZone(zone)
      ? new NoopTickScheduler()
      : inject(AnimationFrameTickScheduler);
  },
})
export abstract class TickScheduler {
  abstract schedule(): void;
}
```

And within AnimationFrameTickScheduler we use this, but requestAnimationFrame does not exist in node during the server render:

```ts
  schedule(): void {
    if (!this.isScheduled) {
      this.isScheduled = true;
      requestAnimationFrame(() => {
        this.appRef.tick();
        this.isScheduled = false;
      });
    }
  }
```

We're throwing an error

### Expected behavior

To schedule using a different mechanism maybe setTimeout while on server?

ngrx: 19.2.1 ( did try it with latest 20.0.0-rc... )
angular: v20

### Other information

_No response_

- [x] Yes
- [ ] No

JelleBruisten:
…stead of requestAnimationFrame

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When using either ngrxLet or ngrxPush in ssr and zoneless, the server calls `requestAnimationFrame` which does not exist in a node environment. 

Closes #4902

## What is the new behavior?

Within a node environment ( ssr ) and zoneless we use setTimeout instead of requestAnimationFrame.

```
[ ] Yes
[X] No
```

## Other information
I did some more testing on angular v20, when using ssr + zoneless and using a `NoopTickScheduler` it does trigger changes. I opted for not removing the `NoopTickScheduler` and making big changes since `@ngrx/component` is in maintenance mode.

### DIFF
--- a/modules/component/src/core/tick-scheduler.ts
+++ b/modules/component/src/core/tick-scheduler.ts
@@ -5,8 +5,8 @@ import {
   NgZone,
   PLATFORM_ID,
 } from '@angular/core';
-import { isNgZone } from './zone-helpers';
 import { isPlatformServer } from '@angular/common';
+import { isNgZone } from './zone-helpers';
 
 @Injectable({
   providedIn: 'root',
@@ -25,13 +25,13 @@ export abstract class TickScheduler {
   providedIn: 'root',
 })
 export class ZonelessTickScheduler extends TickScheduler {
-  private isScheduled = false;
+  private readonly appRef = inject(ApplicationRef);
   private readonly platformId = inject(PLATFORM_ID);
   private readonly isServer = isPlatformServer(this.platformId);
-  private readonly appRef = inject(ApplicationRef);
   private readonly scheduleFn = this.isServer
     ? setTimeout
     : requestAnimationFrame;
+  private isScheduled = false;
 
   schedule(): void {
     if (!this.isScheduled) {
--- a/modules/component/src/core/tick-scheduler.ts
+++ b/modules/component/src/core/tick-scheduler.ts
@@ -1,5 +1,12 @@
-import { ApplicationRef, inject, Injectable, NgZone } from '@angular/core';
+import {
+  ApplicationRef,
+  inject,
+  Injectable,
+  NgZone,
+  PLATFORM_ID,
+} from '@angular/core';
 import { isNgZone } from './zone-helpers';
+import { isPlatformServer } from '@angular/common';
 
 @Injectable({
   providedIn: 'root',
@@ -7,7 +14,7 @@ import { isNgZone } from './zone-helpers';
     const zone = inject(NgZone);
     return isNgZone(zone)
       ? new NoopTickScheduler()
-      : inject(AnimationFrameTickScheduler);
+      : inject(ZonelessTickScheduler);
   },
 })
 export abstract class TickScheduler {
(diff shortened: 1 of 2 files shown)


modules/component/src/core/tick-scheduler.ts
```diff
@@ -19,6 +25,7 @@ export abstract class TickScheduler {
 })
 export class AnimationFrameTickScheduler extends TickScheduler {
   private isScheduled = false;
+  private isServer = inject(PLATFORM_ID);
```
markostanimirovic: This will not work. `isServer` will always be `true`. It's necessary to use `isPlatformServer`:

```ts
private readonly platformId = inject(PLATFORM_ID);
private readonly isServer = isPlatformServer(this.platformId);
```

modules/component/src/core/tick-scheduler.ts
```diff
@@ -19,6 +25,7 @@ export abstract class TickScheduler {
 })
 export class AnimationFrameTickScheduler extends TickScheduler {
```
markostanimirovic: This is a private API, so we can rename it properly:

```suggestion
export class ZonelessTickScheduler extends TickScheduler {
```

modules/component/src/core/tick-scheduler.ts
```diff
@@ -19,6 +25,7 @@ export abstract class TickScheduler {
...
   private isScheduled = false;
+  private isServer = inject(PLATFORM_ID);
 
   constructor(private readonly appRef: ApplicationRef) {
```
markostanimirovic: let's use either constructor-based DI or `inject` function, but not both.

modules/component/src/core/tick-scheduler.ts
```diff
@@ -27,10 +34,18 @@ export class AnimationFrameTickScheduler extends TickScheduler {
...
+        requestAnimationFrame(() => {
+          this.appRef.tick();
+          this.isScheduled = false;
+        });
+      } else {
+        setTimeout(() => {
+          this.appRef.tick();
+          this.isScheduled = false;
+        });
+      }
```
markostanimirovic: There is no need to fully duplicate logic. Also, the condition has to be reversed - `requestAnimationFrame` should be used when `isServer` is `false`, not `true`:

1. Add a class member:

```ts
private readonly scheduleFn = this.isServer ? setTimeout : requestAnimationFrame;
```

2. Then use `scheduleFn` within the `schedule` method:

```suggestion
      scheduleFn(() => {
        this.appRef.tick();
        this.isScheduled = false;
      });
```

markostanimirovic (changes requested):
Thanks for the PR @JelleBruisten! See the requested changes below.

It's also necessary to add a unit test and verify the fix.

markostanimirovic:
@JelleBruisten target branch has to be `main`, not `master`.

JelleBruisten:
@markostanimirovic I rebased the branch from the main branch.

markostanimirovic (approved):

rainerhahnekamp (approved):

modules/component/src/core/tick-scheduler.ts
```diff
@@ -17,17 +24,19 @@ export abstract class TickScheduler {
...
   providedIn: 'root',
 })
-export class AnimationFrameTickScheduler extends TickScheduler {
+export class ZonelessTickScheduler extends TickScheduler {
+  private readonly appRef = inject(ApplicationRef);
+  private readonly platformId = inject(PLATFORM_ID);
+  private readonly isServer = isPlatformServer(this.platformId);
+  private readonly scheduleFn = this.isServer
+    ? setTimeout
+    : requestAnimationFrame;
```
hakimio, line 33: @JelleBruisten this change introduced `Illegal invocation` error #5108

PR #4916, opened 2025-08-09
docs(www): enhance code blocks

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

Closes #

## What is the new behavior?

- providing a `path` to a code block now extracts the code from the given file => that's why the code is removed from the md files.
- code blocks are all follow the same theme (our ngrx theme)
- copied all examples over

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/www/eslint.config.mjs
+++ b/projects/www/eslint.config.mjs
@@ -11,7 +11,7 @@ const compat = new FlatCompat({
 
 export default [
   {
-    ignores: ['**/dist', '**/node_modules'],
+    ignores: ['**/dist', '**/node_modules', '**/examples'],
   },
   ...baseConfig,
   ...compat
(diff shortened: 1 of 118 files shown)


timdeschryver (commented):
Code blocks can't find the correct file to extract the code from.
This results in empty code blocks.

jdegand:
@timdeschryver Vite does not expose `src/` files by default. `fetch()` only works for files in the public directory or those explicitly exposed via Vite plugins or server config.

The files can be moved to `public/examples`.  Using `import.meta.glob` in example-service to replace fetch could be an alternative solution, but this would be more involved (Update: I tried a solution with `import.meta.glob`, but it appears this doesn't work with SSR).  

Using a plugin may only work with development only.  I think it would need to be an external API for production.

projects/www/src/app/components/docs/code-example.component.ts
```diff
@@ -35,4 +43,22 @@ import { Component, Input } from '@angular/core';
 })
 export class CodeExampleComponent {
   @Input() header: string = '';
+  @Input() path: string = '';
```
jdegand: You add a lot of types that can be inferred instead.  This will generate ESLint errors and causes a lint error in the build.

timdeschryver: These are resolved now.

projects/www/eslint.config.mjs
```diff
@@ -43,28 +43,7 @@ export default [
             style: 'kebab-case',
           },
         ],
-        '@typescript-eslint/prefer-namespace-keyword': 'error',
         '@nx/enforce-module-boundaries': 'off',
-        eqeqeq: ['off', 'smart'],
```
timdeschryver, line 48: Because the examples are ignored, these settings can be removed as well.

timdeschryver:
Should be good to go for a review now.
To extract the TS content I didn't find a better way than copying the content over to a txt file in order to read the content.

rainerhahnekamp (approved):

PR #4946, opened 2025-08-30
chore: update ESLint dependencies to v20

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

Closes #

## What is the new behavior?

This PR updates all linting dependencies, which also resolved unmet peer dependencies.
The major updates includes: angular-eslint and eslint.

```
[ ] Yes
[x] No
```


modules/component-store/eslint.config.mjs
```diff
@@ -30,6 +30,7 @@ export default [
         '@angular-eslint/directive-selector': 'off',
         '@angular-eslint/component-selector': 'off',
         '@angular-eslint/prefer-standalone': 'off',
+        '@angular-eslint/prefer-inject': 'off',
```
timdeschryver, line 33: I didn't update the constructors to keep this PR small.
But I think we can start using this in some/all modules, one by one.

modules/eslint-plugin/src/rule-creator.ts
```diff
@@ -1,69 +1,40 @@
...
-    create: (
-      context: Readonly<TSESLint.RuleContext<TMessageIds, TOptions>>,
-      optionsWithDefault: Readonly<TOptions>
-    ) => {
-      const {
-        meta: { ngrxModule, version },
-      } = config;
-      if (
-        version !== undefined &&
-        !ngrxVersionSatisfies(NGRX_MODULE_PATHS[ngrxModule], version)
```
timdeschryver, line 53: Everything about the version checks (to disable/enable specific  rules for a particular NgRx version) were removed in the past, but the code was not cleaned up. All rules are always applicable for all versions.
I noticed this only now, because of the TSESlint changes in the rule creator.

markostanimirovic (approved):

PR #4953, opened 2025-09-06
fix(signals): drop `assertUniqueStoreMembers` in production

arturovt:
The `assertUniqueStoreMembers` method previously used an early-return guard:

```js
  if (!ngDevMode) return;
  // dev-only code ...
```

In production builds, `ngDevMode` is replaced with `false`, so the guard compiles to `return;`. However, bundlers like ESBuild still keep the remaining statements after the return as unreachable code instead of removing them. This leaves behind unnecessary dead code in the output.

Technically, the body is unreachable. But to prove that, the bundler must be 100% certain that:

- `return` cannot be removed by some transform
- there's no later transformation that changes control flow

As thus, it's always conservative.

This commit updates the method to instead wrap the full function:

```js
  if (ngDevMode) {
    // assertUniqueStoreMembers ...
  }
```

Снимок экрана 2025-09-06 в 13 46 09

### DIFF
--- a/modules/signals/src/signal-store-assertions.ts
+++ b/modules/signals/src/signal-store-assertions.ts
@@ -1,15 +1,9 @@
 import { InnerSignalStore } from './signal-store-models';
 
-declare const ngDevMode: unknown;
-
 export function assertUniqueStoreMembers(
   store: InnerSignalStore,
   newMemberKeys: Array<string | symbol>
 ): void {
-  if (typeof ngDevMode === 'undefined' || !ngDevMode) {
-    return;
-  }
-
   const storeMembers = {
     ...store.stateSignals,
     ...store.props,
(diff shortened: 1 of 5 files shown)


rainerhahnekamp:
Thanks @arturovt. Do you see any alternatives here? The pattern with the `if`
condition reminds me of what we had in `withLinkedState`, where we duplicated
logic from `withState`. That redundancy already caused us a bug once, so I’m a
bit cautious about repeating it.

With this approach, we’d also need to ensure that all calls to
`assertUniqueStoreMember` sit safely behind the `if` condition.

If this turns out to be the best option, I’m fine with it.

markostanimirovic:
@rainerhahnekamp Yes, I don't think there is a 'more elegant' way for now.

PR #4959, opened 2025-09-19
fix(signals): expose readonly signals in slices

### LINKED ISSUE #4958: Signal Store Allows Writing to Read-Only Signals

signals

By default, the Signal Store returns signals of type Signal<...> (and not WriteableSignal<...>). However, in the current version, it's possible to write into these Signals. I'm pretty sure this was not possible in former versions.

Here is a reproduction:

```ts
import { Component, inject, signal } from '@angular/core';
import { FormsModule } from '@angular/forms';

const DemoStore = signalStore(
  {providedIn: 'root'},
  withState({
    title: 'Test!',
  }),
);

@Component({
  selector: 'app-root',
  imports: [FormsModule],
  template: `
    <h1>Reproduction</h1>
    <input [(ngModel)]="title">
    <div>{{title()}}</div>
  `,
  styleUrl: './app.css'
})
export class App {
  // Here, everything is fine:
  // protected readonly title = signal('test');

  // This option allows the two-way-binding to directly
  // write back to the signal in the store ...
  store = inject(DemoStore);
  title = this.store.title;
}
```

### Expected behavior

Writing back should not be possible. 

@angular/core@20.3.1
@ngrx/signals@20.0.1

### Other information

_No response_

- [ ] Yes
- [x] No

rainerhahnekamp:
`withState`, `signalState`, and `withLinkedState` now pass a `Signal` instead of the original `WritableSignal` to the DeepSignal.

This prevents accidental misuse where consumers (e.g. `ngModel` or other APIs with incompatible types) could assert `WritableSignal` and write directly into the state.

Closes #4958

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```typescript

const person = signalState({ title: 'Title'});
(person.title as WritableSignal<string>).set('Hallo'); <-- direct access to state
```

Closes #4958

## What is the new behavior?

Will throw an error if set is called because the Signal is readonly.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/spec/signal-state.spec.ts
+++ b/modules/signals/spec/signal-state.spec.ts
@@ -60,6 +60,13 @@ describe('signalState', () => {
     expect(isSignal(state.ngrx)).toBe(true);
   });
 
+  it('does not expose state slices as writable signals', () => {
+    const state = signalState(initialState);
+    expect(() => (state as any).foo.set('baz')).toThrow(
+      'set is not a function'
+    );
+  });
+
   it('caches previously created signals', () => {
     const state = signalState(initialState);
     const user1 = state.user;
--- a/modules/signals/spec/with-linked-state.spec.ts
+++ b/modules/signals/spec/with-linked-state.spec.ts
@@ -1,4 +1,4 @@
-import { linkedSignal, signal } from '@angular/core';
+import { linkedSignal, signal, WritableSignal } from '@angular/core';
 import {
   getState,
   patchState,
@@ -253,5 +253,16 @@ describe('withLinkedState', () => {
       'Trying to override:',
       'value'
     );
+  });
+
+  it('does not expose linked state properties as writable signals', () => {
+    const initialStore = getInitialInnerStore();
+    const userStore = withLinkedState(() => ({
+      foo: () => 'bar',
+    }))(initialStore);
+
+    expect(() =>
+      (userStore.stateSignals.foo as WritableSignal<string>).set('baz')
+    ).toThrow('set is not a function');
   });
 });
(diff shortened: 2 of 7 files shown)


markostanimirovic:
Let's merge this in v21 to prevent unintentional breaking changes.

rainerhahnekamp:
@timdeschryver just added the label "breaking changes" as suggested

markostanimirovic:
@rainerhahnekamp @timdeschryver 

After revisiting this PR, I'm not sure if we should proceed with it, because maybe in the future we will introduce deep writable signals within the store while keeping them as read-only outside of the store instance. In that case, deep writables would be protected only at the TypeScript level, and we would have to revert this (and make another breaking change).

Therefore, I'm not sure if this change is worth it. E.g., for the state source, we also have protection only at the TS level. If someone wants to avoid it (and use `as any/WritableStateSource`), I see that as a usage issue, not a bug from our side.

rainerhahnekamp:
@markostanimirovic 

the thing was that it happened via two-way binding with `ngModel`.  So users didn't apply bad practices or anything unnatural. So they didn't even receive any warnings, error messages, etc.

@manfredsteyer you reported this one. Can you elaborate on it?

rainerhahnekamp:
@markostanimirovic, do you think changes like deep-writables - even if they are only inside the store - can be done without breaking changes? As you remember, even the multi-signals approach in v20 should have been internal-only, but it turned out to be the opposite.

markostanimirovic:
> @markostanimirovic, do you think changes like deep-writables - even if they are only inside the store - can be done without breaking changes? As you remember, even the multi-signals approach in v20 should have been internal-only, but it turned out to be the opposite.

Yes, maybe there would be a breaking change for deep writables, but regardless, I'm wondering do we need to introduce this one.

markostanimirovic:
I'm going to close this PR for now. We'll revisit it for v22.

PR #4964, opened 2025-09-24
feat(signals): add ability to provide SignalStore at the platform level

### LINKED ISSUE #4963: Add support providedIn: 'platform'

signals

### Information

Please add support `providedIn: platform` within `signalStore()`.

```
signalStore(
    {
        // @ts-expect-error
        providedIn: `platform`,
    },
    ...
);
```

- [x] Yes
- [ ] No

MillerSvt:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Cannot pass any `providedIn` except `root`.

Closes #4963

## What is the new behavior?

Added support to any not deprecated `providedIn` types.

```
[ ] Yes
[x] No
```

## Other information

I was thinking of just adding the `platform` to the union. But it seems that if here we just pass the value into the `Injectable` decorator, then we should rely on the typing of this decorator. The only thing seems to be that we should exclude deprecated types from union, so as not to add support for what will be removed soon.

### DIFF
--- a/modules/store/src/feature_creator.ts
+++ b/modules/store/src/feature_creator.ts
@@ -138,10 +138,10 @@ export function createFeature<FeatureName extends string, FeatureState>(
  *   name: 'products',
  *   reducer: createReducer(
  *     initialState,
- *     on(ProductsApiActions.loadSuccess(state, { products }) => ({
+ *     on(ProductsApiActions.loadSuccess, (state, { products }) => ({
  *       ...state,
  *       products,
- *     }),
+ *     })),
  *   ),
  * });
  *
--- a/projects/ngrx.io/content/guide/store/reducers.md
+++ b/projects/ngrx.io/content/guide/store/reducers.md
@@ -271,7 +271,7 @@ Whether your feature states are loaded eagerly or lazily depends on the needs of
 
 If you have a module-based Angular application, you can still use standalone components. NgRx standalone APIs support this workflow as well.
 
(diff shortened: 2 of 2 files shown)


modules/signals/src/signal-store.ts
```diff
@@ -28,7 +32,7 @@ export function signalStore<F1 extends SignalStoreFeatureResult>(
 export function signalStore<
   F1 extends SignalStoreFeatureResult,
   F2 extends SignalStoreFeatureResult,
-  R extends SignalStoreFeatureResult = F1 & F2,
+  R extends SignalStoreFeatureResult = F1 & F2
```
markostanimirovic: Revert all changes with the removed trailing comma.

We have a pre-commit hook that runs prettier for all affected files. To register it, run `pnpm install` and it will automatically format the file on commit.

modules/signals/src/signal-store.ts
```diff
@@ -9,7 +9,11 @@ import {
 } from './signal-store-models';
 import { OmitPrivate, Prettify } from './ts-helpers';
 
-type SignalStoreConfig = { providedIn?: 'root'; protectedState?: boolean };
+type NotDeprecatedInjectable = {
+  providedIn?: Exclude<Injectable['providedIn'], 'any' | Type<any>>;
+};
+
+type SignalStoreConfig = NotDeprecatedInjectable & { protectedState?: boolean };
```
markostanimirovic: let's simplify `providedIn` type:

```suggestion
type ProvidedInConfig = { providedIn?: 'root' | 'platform' };

type SignalStoreConfig = ProvidedInConfig & { protectedState?: boolean };
```

timdeschryver:
Thanks @MillerSvt

PR #4965, opened 2025-09-25
refactor(signals): use dot-notation to access properties in withFeature

andrewseguin:
Minor refactor to use dot-notation instead of brackets for accessing the properties of the `store` object


rainerhahnekamp (approved):
Thanks

markostanimirovic (approved):

PR #4981, opened 2025-10-15
feat(signals): create a rx-promise-method to be able to use angular 2…

### LINKED ISSUE #4980: Using Angular 21 signal forms submit function with rxMethod from ngrx is impossible, please create a new function to return promise to be able to be used in angular 21 signal forms

signals

### Information

I was trying out the new feature from Angular 21 and wanted to use it with ngrx signal store. However, I realized it is impossible to use it in the signal forms submit function because it requires a promise in return, but rxMethod returns only a destroy. Therefore, we need a function that returns a promise instead.

In this article I explained the issue and created a new method called rxPromiseMethod to be used instead of rxMethod within withMethod(). You can find some examples there. 

[https://medium.com/@mohammadrezaimn/rxpromisemethod-bridging-ngrx-signal-store-with-angulars-signal-forms-f80483adcb35](url)

There is a withMutation function from the @angular-architects/ngrx-toolkit library, but the issue is that you can only send a payload to it, and the types are tightly coupled with the HTTP request DTO, which is not good.

- [x] Yes
- [ ] No

rezaimn:
…submit function

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/ngrx/platform/issues/4980

Closes #4980

```
[ ] Yes
[X] No
```

### DIFF
--- a/modules/signals/rxjs-interop/src/index.ts
+++ b/modules/signals/rxjs-interop/src/index.ts
@@ -1,1 +1,2 @@
 export { rxMethod, RxMethod } from './rx-method';
+export { rxPromiseMethod, RxPromiseMethod } from './rx-promise-method';
--- a/modules/signals/rxjs-interop/src/rx-promise-method.ts
+++ b/modules/signals/rxjs-interop/src/rx-promise-method.ts
@@ -1,0 +1,164 @@
+import {
+  assertInInjectionContext,
+  DestroyRef,
+  effect,
+  inject,
+  Injector,
+  isSignal,
+  Signal,
+  untracked,
+} from '@angular/core';
+
+declare const ngDevMode: unknown;
+
+type PromiseResult<T> =
+  | { status: 'success'; value: T }
+  | { status: 'error'; error: unknown };
+
+export type RxPromiseMethod<Input, Output> = {
+  (
+    input: Input | Signal<Input>,
+    config?: { injector?: Injector },
+  ): Promise<PromiseResult<Output>>;
+  destroy: () => void;
+};
+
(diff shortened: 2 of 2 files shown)


rainerhahnekamp:
Hey @rezaimn, I think it is a little bit too early to start implementing dedicated features for the Signal Forms. As you might know, we don't really support experimental features - otherwise `withResource` would already be available.

I do have some ideas on how a SignalForms integration could look like and I am thinking about a deeper integration. Maybe - some day - a `withForms` feature. 

BUT: Given that we are talking here about a feature which is far from even being experimental, we should wait. Looking at the project plan, I see a lot of unfinished tasks: https://github.com/orgs/angular/projects/60/views/1

rezaimn:
Hey @rainerhahnekamp, I understand, but I tried it and it was so exciting, and I feel it is a real game-changer. Anyway this could fix it, if you thinking about a withForm, then it will be even nicer.

rainerhahnekamp:
Yeah, I am also very excited and would like to have Signal Forms available today. But the past has shown that patience is quite often better for sustainable featurea.

So I'd say maybe a community extension and if v21 is out, we can revisit the idea and start to plan for sound integration.

Can we close this PR? Either way, we would require an issue with design specification first.

markostanimirovic:
Thanks for the PR. However, we handle feature requests differently. Please extend the description of the issue you opened (https://github.com/ngrx/platform/issues/4980) and share some code snippets of what you're trying to achieve. We'll take a look and let you know if you should proceed with the PR.

PR #4996, opened 2025-11-15
feat(signals): allow computation fn to be provided to `signalMethod` and `rxMethod`

### LINKED ISSUE #4986: `@ngrx/signals`: Support `() => T` in `signalMethod` and `rxMethod`

signals

## Problem

With Angular's `resource` and `linkedSignal`, we saw that the Angular team moved having a parameter of `Signal<T>` to `() => T`. This makes sense because users don't have to create an intermediate computed signal just to pass complex values, and the tracking will be done internally.

Right now, with `signalMethod` or `rxMethod`, users need to do the following:

```typescript
const printName = signalMethod<{
  firstname: string;
  lastname: string;
}>(console.log);

const firstname = signal('John');
const lastname = signal('Doe');
const name = computed(
  () => ({ firstname: ${this.firstname()}, lastname: ${this.lastname()} })
);

printName(name);
```

## Proposed Solution

```typescript
const printName = signalMethod<{
  firstname: string;
  lastname: string;
}>(console.log);

const firstname = signal('John');
const lastname = signal('Doe');

printName(() => ({ firstname: ${this.firstname()}, lastname: ${this.lastname()} }));
```

## Advantages

- **Follows Angular's pattern**: `resource` and `linkedSignal` both use `() => T`
- **Less boilerplate**: No intermediate computed signals needed
- **Backward compatible**: Signals have the same signature as `() => T`

I have to create the `computed` manually.

- [x] Yes
- [ ] No

rainerhahnekamp:
Align signalMethod and rxMethod with Angular’s resource/linkedSignal pattern so callers can pass a lazy () => T instead of creating an intermediate computed signal.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4986

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/www/src/app/components/footer.component.ts
+++ b/projects/www/src/app/components/footer.component.ts
@@ -16,7 +16,7 @@ import { RouterLink } from '@angular/router';
 
     <nav class="learn-ngrx">
       <h4>Learn NgRx</h4>
-      <a routerLink="/workshops">Workshops</a>
+      <!--      <a routerLink="/workshops">Workshops</a>-->
       <a routerLink="/api">API Reference</a>
       <a href="https://github.com/sponsors/ngrx" target="_blank">Sponsor</a>
       <a routerLink="/about">About</a>
--- a/projects/www/src/app/components/menu.component.ts
+++ b/projects/www/src/app/components/menu.component.ts
@@ -33,15 +33,15 @@ import { DOCUMENT } from '@angular/common';
         NgRx
       </a>
       <hr />
-      <a
-        routerLink="/workshops"
-        routerLinkActive="active"
-        class="menu-link"
-        (click)="closeMenu()"
-      >
-        <mat-icon>co_present</mat-icon>
-        Workshops
-      </a>
+      <!--      <a-->
+      <!--        routerLink="/workshops"-->
+      <!--        routerLinkActive="active"-->
+      <!--        class="menu-link"-->
+      <!--        (click)="closeMenu()"-->
+      <!--      >-->
+      <!--        <mat-icon>co_present</mat-icon>-->
+      <!--        Workshops-->
+      <!--      </a>-->
       <a
         routerLink="/api"
         routerLinkActive="active"
--- a/projects/www/src/app/pages/(home).page.ts
+++ b/projects/www/src/app/pages/(home).page.ts
@@ -44,16 +44,16 @@ import { MatIconModule } from '@angular/material/icon';
           >Learn SignalStore</a
         >
       </ngrx-styled-box>
-      <ngrx-styled-box>
-        <mat-icon inline>co_present</mat-icon>
-        <h3>Workshops</h3>
-        <p>
-          Attend an NgRx workshop from the creators of NgRx. Our three day
-          workshops cover the entire NgRx reactive state framework, from Store
-          to Effects to SignalStore.
-        </p>
-        <a routerLink="/workshops" mat-flat-button>Attend a Workshop</a>
-      </ngrx-styled-box>
+      <!--      <ngrx-styled-box>-->
+      <!--        <mat-icon inline>co_present</mat-icon>-->
+      <!--        <h3>Workshops</h3>-->
+      <!--        <p>-->
+      <!--          Attend an NgRx workshop from the creators of NgRx. Our three day-->
+      <!--          workshops cover the entire NgRx reactive state framework, from Store-->
+      <!--          to Effects to SignalStore.-->
+      <!--        </p>-->
+      <!--        <a routerLink="/workshops" mat-flat-button>Attend a Workshop</a>-->
+      <!--      </ngrx-styled-box>-->
       <ngrx-styled-box>
         <mat-icon inline>volunteer_activism</mat-icon>
         <h3>Support the team</h3>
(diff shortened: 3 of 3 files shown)


modules/signals/src/signal-method.ts
```diff
@@ -30,10 +28,13 @@ export function signalMethod<Input>(
   const sourceInjector = config?.injector ?? inject(Injector);
 
   const signalMethodFn = (
-    input: Input | Signal<Input>,
+    input: Input | (() => Input),
     config?: { injector?: Injector }
   ): EffectRef => {
-    if (isSignal(input)) {
+    if (typeof input !== 'function') {
```
rainerhahnekamp: Switched the if condition to eliminate the possibility of a function in the else condition.

With `Input | (() => Input)` and a `typeof input === 'function', TypeScript would not narrow done the type union.

projects/www/src/app/pages/guide/signals/rxjs-integration.md
```diff
@@ -6,7 +6,7 @@ RxJS is still a major part of NgRx and the Angular ecosystem, and the `@ngrx/sig
 
 The `rxMethod` is a standalone factory function designed for managing side effects by utilizing RxJS APIs.
 It takes a chain of RxJS operators as input and returns a reactive method.
-The reactive method can accept a static value, signal, or observable as an input argument.
+The reactive method can accept a static value, a reactive computation (like a Signal), or observable as an input argument.
```
rainerhahnekamp: I am using the term "reactive computation" instead of auto-tracking functions because that's what the Angular team calls it in `linkedSignal`.

projects/www/src/app/pages/guide/signals/signal-method.md
```diff
@@ -1,6 +1,6 @@
 # SignalMethod
 
-`signalMethod` is a standalone factory function used for managing side effects with Angular signals. It accepts a callback and returns a processor function that can handle either a static value or a signal. The input type can be specified using a generic type argument:
+`signalMethod` is a standalone factory function used for managing side effects with Angular signals. It accepts a callback and returns a processor function that can handle either a static value or a reactive computation, i.e. either a Signal or a function, which will be tracked automatically. The input type can be specified using a generic type argument:
```
markostanimirovic: To avoid explaining multiple times that the Signal is compatible with a computation function, wdyt about mentioning a computation function in addition to Signal? (the same comment for other sections where this was changed)

```suggestion
`signalMethod` is a standalone factory function used for managing side effects with Angular signals. It accepts a callback and returns a processor function that can handle either a static value, a signal, or a computation function. The input type can be specified using a generic type argument:
```

rainerhahnekamp: Do you mean I should also remove the example then? Otherwise, I don't see so many places where the docs become easier.
Should we really call it "computation function"? I used "reactive computation", because that's the offical Angular term: https://angular.dev/api/core/linkedSignal.

We could just say "reactive computation" without explanation, but I doubt that most people know what that is...

michael-small: For what it's worth, I was curious, and it turns out both names are used in varying contexts

https://github.com/search?q=repo%3Aangular%2Fangular+%22reactive+function%22&type=code

https://github.com/search?q=repo%3Aangular%2Fangular+%22reactive+computation%22&type=code

"reactive function" in form/effect/httpResource related stuff, and "reactive computation" in linkedSignal, one resource doc, and a couple primitive file/docs.

markostanimirovic: Yes, it's called differently depending on the API. The input argument of `computed` is named `computation`, and it's a function, so I think 'computation function' should be clear enough.

> We could just say "reactive computation" without explanation, but I doubt that most people know what that is...

@rainerhahnekamp Agree. I'd keep mentioning signals for clarity.

projects/www/src/app/pages/guide/signals/rxjs-integration.md
```diff
@@ -92,6 +92,37 @@ export class Numbers {
...
+      tap(console.log)
+    )
+  );
+
+  constructor() {
+    const a = signal(5);
+    const b = signal(2);
+
+    this.logDoubledNumber(() => a() + b());
+    // console output: 20
```
timdeschryver: This was a good math test 😉

```suggestion
    // console output: 14
```

projects/www/src/app/pages/guide/signals/rxjs-integration.md
```diff
@@ -92,6 +92,37 @@ export class Numbers {
...
+
+    setTimeout(() => b.set(10), 3_000);
+    // console output after 3 seconds: 40
```
timdeschryver:
```suggestion
    // console output after 3 seconds: 30
```

projects/www/src/app/pages/guide/signals/signal-method.md
```diff
@@ -49,6 +49,30 @@ export class Numbers {
 }
 ```
 
+Finally, a reactive computation example shows an automaticaly tracked computation, built from multiple Signals.
```
timdeschryver:
```suggestion
Finally, a reactive computation example shows an automatically tracked computation, built from multiple Signals.
```

projects/www/src/app/pages/guide/signals/rxjs-integration.md
```diff
@@ -63,7 +63,7 @@ export class Numbers {
 }
 ```
 
-When a reactive method is called with a signal, the reactive chain is executed every time the signal value changes.
+When a reactive method is called with a reactive computation - like a Signal - the reactive chain is executed every time the signal value changes.
```
timdeschryver:
```suggestion
When a reactive method is called with a reactive computation (like a Signal) the reactive chain is executed every time the signal value changes.
```

projects/www/src/app/pages/guide/signals/rxjs-integration.md
rainerhahnekamp: Upsi

timdeschryver (approved):

PR #4997, opened 2025-11-16
feat(signals): provide support for scoped events

### LINKED ISSUE #4776: `@ngrx/signals/events`: Provide Support for Scoped Events

signals

### Information

Currently, all events are dispatched using the globally provided `Dispatcher`. However, there are at least two scenarios where having scoped dispatchers and event handling would be beneficial:

- **Local state management:** When multiple instances of the same component are initialized, each with its own store instance, events dispatched from a specific component instance should be handled exclusively by the store instance associated with that same component.
- **Microfrontends:** Each microfrontend may require its own isolated dispatcher.

### Solution

This functionality can be introduced by supporting a tree of dispatchers. Components or microfrontends that require a scoped dispatcher can use the `provideDispatcher` function. In the following example, events dispatched from `MyComponent` can only be handled by stores provided at the same level (`MyComponent`) or in its descendants.

```ts
import { event, provideDispatcher } from '@ngrx/signals/events';

const myEvent = event('myEvent');

@Component({
  providers: [
    provideDispatcher(),
    MyStore1,
    MyStore2,
  ],
})
export class MyComponent {
  readonly dispatcher = inject(Dispatcher);

  myMethod(): void {
    this.dispatcher.dispatch(myEvent());
  }
}
```

The dispatcher tree follows the same hierarchy as Angular's injector tree: events are visible to the level at which they are dispatched and to all descendants. For example, events dispatched to the root dispatcher are available to all nested dispatchers, but events dispatched to a child dispatcher are not visible to ancestors.

If needed, events can be explicitly dispatched to a parent or root dispatcher:

```ts
  myMethod(): void {
    this.dispatcher.dispatch(myEvent(), { parent: true });
    // or
    this.dispatcher.dispatch(myEvent(), { root: true });
  }
```

The proposed approach is lightweight, highly flexible, and aligns naturally with Angular’s design principles.

While it is possible to include a unique identifier in the event payload and manually filter events within reducers and effects, this approach adds unnecessary overhead and negatively impacts the developer experience. 

- [x] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4776

## What is the new behavior?

```ts
// === providing scoped dispatcher: ===
{ providers: [provideDispatcher()] }

// === dispatching events to the parent/global scope: ===
// scope values: 'self' | 'parent' | 'global' ('self' by default)

const event1 = event('event1');
const event2 = event('event2');

// using Dispatcher:
const dispatcher = inject(Dispatcher);
dispatcher.dispatch(event1(), { scope: 'parent' });

// using injectDispatch:
const dispatch = injectDispatch({ event1 });
dispatch({ scope: 'global' }).event1();

// using toScope:
withEffects((_, events = inject(Events)) => ({
  $: events.on(event1).pipe(
    map(() => [event2(), toScope({ 'parent' })]),
  ),
}))

// using mapToScope:
withEffects((_, events = inject(Events)) => ({
  $: events.on(event1).pipe(
    map(() => event2()),
    mapToScope('global'),
  ),
}))
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/rxjs-interop/spec/rx-method.spec.ts
+++ b/modules/signals/rxjs-interop/spec/rx-method.spec.ts
@@ -65,6 +65,35 @@ describe('rxMethod', () => {
       expect(results[1]).toBe(10);
     }));
 
+  it('runs with a computation function', () => {
+    TestBed.runInInjectionContext(() => {
+      const results: number[] = [];
+      const method = rxMethod<number>(
+        pipe(tap((value) => results.push(value)))
+      );
+
+      const a = signal(1);
+      const b = signal(1);
+      const multiplier = () => a() * b();
+
+      method(multiplier);
+      expect(results.length).toBe(0);
+
+      TestBed.tick();
+      expect(results[0]).toBe(1);
+
+      a.set(5);
+      expect(results.length).toBe(1);
+
+      TestBed.tick();
+      expect(results[1]).toBe(5);
+
+      b.set(2);
+      TestBed.tick();
+      expect(results[2]).toBe(10);
+    });
+  });
+
   it('runs with void input', () => {
     const results: number[] = [];
     const subject$ = new Subject<void>();
--- a/modules/signals/rxjs-interop/src/rx-method.ts
+++ b/modules/signals/rxjs-interop/src/rx-method.ts
@@ -4,8 +4,6 @@ import {
   effect,
   inject,
   Injector,
-  isSignal,
-  Signal,
   untracked,
 } from '@angular/core';
 import { isObservable, noop, Observable, Subject } from 'rxjs';
@@ -15,7 +13,7 @@ type RxMethodRef = {
 };
 
 export type RxMethod<Input> = ((
-  input: Input | Signal<Input> | Observable<Input>,
+  input: Input | (() => Input) | Observable<Input>,
   config?: { injector?: Injector }
 ) => RxMethodRef) &
   RxMethodRef;
@@ -34,7 +32,7 @@ export function rxMethod<Input>(
   sourceInjector.get(DestroyRef).onDestroy(() => sourceSub.unsubscribe());
 
   const rxMethodFn = (
-    input: Input | Signal<Input> | Observable<Input>,
+    input: Input | (() => Input) | Observable<Input>,
     config?: { injector?: Injector }
   ): RxMethodRef => {
     if (isStatic(input)) {
@@ -62,7 +60,7 @@ export function rxMethod<Input>(
     const instanceInjector =
       config?.injector ?? callerInjector ?? sourceInjector;
 
-    if (isSignal(input)) {
+    if (typeof input === 'function') {
       const watcher = effect(
         () => {
           const value = input();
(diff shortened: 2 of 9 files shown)


markostanimirovic:
@rainerhahnekamp Yes, I'll add a new section 'Scoped Events' without changing existing content from the Events page because the previous behavior remained the same.

modules/signals/events/src/index.ts
```diff
@@ -1,9 +1,15 @@
...
 export { eventGroup } from './event-creator-group';
 export { EventInstance } from './event-instance';
-export { Events } from './events-service';
+export {
+  EventScope,
+  EventScopeConfig,
+  mapToScope,
+  toScope,
+} from './event-scope';
+export { Events, ReducerEvents } from './events-service';
```
rainerhahnekamp, line 12: That is important for the event tracker in the toolkit. 👍

modules/signals/events/src/dispatcher.ts
```diff
@@ -30,9 +31,67 @@ import { Events, EVENTS, ReducerEvents } from './events-service';
...
+ *
+ * ```ts
+ * import { Dispatcher, event } from '@ngrx/signals/events';
+ *
+ * const increment = event('[Counter Page] Increment');
+ *
+ * \@Component({
+ *   \/* ... *\/
+ *   providers: [provideDispatcher()],
+ * })
```
timdeschryver, line 72: Is there a reason to escape these?
In VSCode this gives the following result:

Image

```suggestion
 * @Component({
 *   /* ... */
 *   providers: [provideDispatcher()],
 * })
```

modules/signals/events/src/event-scope.ts
```diff
@@ -0,0 +1,103 @@
...
+ *             ],
+ *           }),
+ *         ),
+ *       ),
+ *     ),
+ *   })),
+ * );
+ * ```
+ */
+export function toScope(scope: EventScope): EventScopeConfig {
```
timdeschryver, line 55: An open question about the naming of this function, sorry I should've mentioned this in the RFC.
What about renaming it to `withScope`, I think this makes the intent easier to read.

```ts
[loadSuccess(), withScope('global')]
// or
[loadSuccess(), scopedTo('global')]
```


I'm also fine with `toScope` though.

markostanimirovic: I'd keep the `toScope` name for consistency with `mapToScope`, but also to distinguish it from SignalStore features that have the `with` prefix.

In general, this helper is not required (we can directly return `[loadedSuccess(), { scope: 'global' }]`), but I added it for a bit better DX - to avoid typos and enable autocompletion in IDEs for the scope name. Anyway, `mapToScope` will be the common way to change the default event scope in the effect, while `toScope` can be used if we want to have more granular control and have different scopes for success and error events, which would be a rare case.

modules/signals/events/src/dispatcher.ts
markostanimirovic: I removed unnecessary second backslashes in: `\/* ... \/`. The first one is required. Otherwise, it's not displayed properly:

Screenshot 2025-11-22 at 16 22 34

Regarding a backslash before the `@Component` decorator, it's needed to avoid treating `@Component` as a JSDoc annotation.

modules/signals/events/src/event-scope.ts
timdeschryver: Sounds good 👍
Thanks!

PR #5043, opened 2025-12-07
ci: fix main build

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] CI related changes
[ ] Other... Please describe:
```

The merges of the recent Vitest changes result in a failing build, due to timeouts and snapshots(?).

## What is the new behavior?

This PR does a few things:

- add `parallel=3` to address the `MaxListenersExceededWarning` warning, which happens because of the NX parallelization. I suppose this also free's up some memory for the tasks.
- add `--skip-nx-cache` for the build step, we recently already added this for the `ci` workflow
- remove `expectSnippet().toSucceed()` calls when we infer code. This will fix the time-outs, `toInfer` already verifies the compilation is succesfull.
- Update snapshots (no idea why these suddenly fail)

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/example-app/src/app/auth/containers/__snapshots__/login-page.component.spec.ts.snap
+++ b/projects/example-app/src/app/auth/containers/__snapshots__/login-page.component.spec.ts.snap
@@ -2,9 +2,9 @@
 
 exports[`Login Page should compile 1`] = `
 <bc-login-page
-  error$={[Function _Store]}
-  pending$={[Function _Store]}
-  store={[Function _MockStore]}
+  error$={[Function Store]}
+  pending$={[Function Store]}
+  store={[Function MockStore]}
 >
   <bc-login-form>
     <mat-card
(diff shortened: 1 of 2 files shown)


modules/component-store/spec/types/component-store.types.spec.ts
```diff
@@ -25,7 +25,6 @@ describe('ComponentStore types', () => {
     describe('infers Subscription', () => {
       it('when argument type is specified and a variable with corresponding type is passed', () => {
         const effectTest = `const sub = componentStore.effect((e: Observable<string>) => number$)('string');`;
-        expectSnippet(effectTest).toSucceed();
```
markostanimirovic, line 28: Why do we have to remove `.toSucceed()` checks? They're important because sometimes inference checks can succeed even if compilation fails.

timdeschryver: To speed up the test (some of these now fail).
The toInfer should fail if the compilation fails, I will double check this later this evening.

timdeschryver: For example:

image

This wasn't always the case though, and was added in v5 https://github.com/cartant/ts-snippet/blob/main/CHANGELOG.md#500-2020-10-28

PR #5074, opened 2026-01-14
docs(www): unregister legacy service worker

markostanimirovic:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, the old ngrx docs website is always displayed on browser reinitialization for all users who opened the old ngrx docs website before, because the service worker from the old app is still registered.

## What is the new behavior?

Service worker from the old ngrx website is unregistered.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/www/src/main.ts
+++ b/projects/www/src/main.ts
@@ -4,7 +4,7 @@ import { config } from './app/app.config.browser';
 
 // Deactivate service worker from old ngrx.io app
 // TODO: Remove after 6 months
-if (typeof window !== 'undefined' && window.navigator.serviceWorker) {
+if (typeof window !== 'undefined') {
   window.navigator.serviceWorker.getRegistrations().then((registrations) => {
     for (const registration of registrations) {
       registration.unregister();


projects/www/src/main.ts
```diff
@@ -2,4 +2,14 @@ import { bootstrapApplication } from '@angular/platform-browser';
 import { AppComponent } from './app/app.component';
 import { config } from './app/app.config.browser';
 
+// Deactivate service worker from old ngrx.io app
+// TODO: Remove after 6 months
+if (typeof window !== 'undefined' && window.navigator.serviceWorker) {
```
brandonroberts: Should this be

```suggestion
if (typeof window !== "undefined" && "serviceWorker" in navigator) {
```

markostanimirovic: We can actually remove this condition - `serviceWorker` (`ServiceWorkerController` instance) is always defined on `window.navigator` even if there are no registered service workers for the domain.

```suggestion
if (typeof window !== 'undefined') {
```

PR #5078, opened 2026-01-30
feat(store): enforce exact return types in on() and updater() callbacks

### LINKED ISSUE #4280: updater-explicit-return-type No longer needed with NoInfer

component-store

### Information

I'm pretty sure with `NoInfer` in TS 5.4 that we can avoid the problem of needing explicit return types for updaters. This may be true other places in other packages, too.

_No response_

- [X] Yes
- [ ] No

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

TypeScript does not perform [excess property checking](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks) on callback return values. This means that `on()` and `ComponentStore.updater()` callbacks can return objects with extra properties that don't exist on the state type, and TypeScript won't report an error:

```ts
// No compilation error — `extra` is silently accepted
on(setName, (state, { name }) => ({ ...state, name, extra: true }))

// No compilation error — `extra` is silently accepted
this.updater((state, value: string) => ({ ...state, name: value, extra: true }))
```

The `on-function-explicit-return-type` and `updater-explicit-return-type` ESLint rules exist as workarounds, requiring developers to add explicit return type annotations to trigger TypeScript's excess property checking.

Closes #4280

## What is the new behavior?

Both `on()` and `ComponentStore.updater()` now enforce exact return types at the type level. The callback return type is captured as a generic `R`, and a conditional check determines whether `R` contains keys not present in the state type. If excess keys are found, the return type is intersected with a descriptive string literal type, producing a clear compile-time error with an actionable message:

```ts
// TS error: Type '{ extra: boolean; ... }' is not assignable to type
// '"callback return type must exactly match the state type. Remove excess properties."'
on(setName, (state, { name }) => ({ ...state, name, extra: true }))

// TS error: Type '{ extra: boolean; ... }' is not assignable to type
// '"updater callback return type must exactly match the state type. Remove excess properties."'
this.updater((state, value: string) => ({ ...state, name: value, extra: true }))
```

The custom type error messages follow the same pattern used in ngrx action creators (e.g., `'action creator cannot return an array'`), providing actionable guidance instead of cryptic `never` type errors.

No explicit return type annotations are needed. The following patterns continue to work without changes:
- Spread with overrides: `(state) => ({ ...state, name: 'updated' })`
- Returning `initialState` or `state` directly
- States with optional properties or index signatures
- Multiple action creators in `on()`
- Generic reducer factories that return `state` directly

The `updater-explicit-return-type` and `on-function-explicit-return-type` ESLint rules are deprecated since the type system now handles this directly.

### Limitation with generic state types

When the state type is an unresolved generic parameter — e.g., `class MyStore<T> extends ComponentStore<T>` or `function createGenericReducer<TState>()` — TypeScript cannot fully resolve the excess property check because `keyof T` is deferred. This means callbacks that spread state and override known properties (e.g., `{...state, id: value}`) will produce a false type error in those contexts. The workaround is to return `state` directly or use a type assertion (`as T`). This is a TypeScript limitation with deferred conditional types on generic parameters, not something that can be solved at the library level. The deprecated ESLint rules do not help in this case either.

```
[x] Yes
[ ] No
```

```
BREAKING CHANGES:

The updater() and on() callback return types are now strictly checked
for excess properties. Code that previously compiled with extra
properties in these callbacks will now produce TypeScript compilation
errors with a descriptive message.

When the state type is an unresolved generic parameter (e.g.
class MyStore<T> extends ComponentStore<T>, or a generic reducer
factory function), callbacks that spread state and override known
properties may produce a false type error. Return state directly
or use a type assertion (as T) in those cases.

BEFORE:

const reducer = createReducer(
  initialState,
  on(setName, (state, { name }) => ({
    ...state,
    name,
    extra: true, // no error
  })),
);

this.updater((state, value: string) => ({
  ...state,
  name: value,
  extra: true, // no error
}))

AFTER:

// Type '{ extra: boolean; ... }' is not assignable to type
// '"callback return type must exactly match the state type.
//  Remove excess properties."'
const reducer = createReducer(
  initialState,
  on(setName, (state, { name }) => ({
    ...state,
    name,
    extra: true, // TS error with descriptive message
  })),
);

// Type '{ extra: boolean; ... }' is not assignable to type
// '"updater callback return type must exactly match the state type.
//  Remove excess properties."'
this.updater((state, value: string) => ({
  ...state,
  name: value,
  extra: true, // TS error with descriptive message
}))
```

## Other information

Validated with `ts-snippet` type tests covering: spread patterns, explicit returns, void callbacks, direct state returns, `initialState` returns, optional properties, index signatures, multiple action creators, generic reducer factories, missing properties, wrong property types, and excess property detection in both `createReducer`-wrapped and standalone `on()` usage.


david-shortman:
CI failures seem unrelated to this PR:

effects:test
Single test (should enforce an Action return value) timed out at 8s. This is a flaky ts-snippet test. Passes locally.

example-app-e2e
Cypress times out looking for bc-book-preview. This reproduces locally with the same failure, but I think it's a pre-existing issue with the example app.

modules/eslint-plugin/src/rules/component-store/updater-explicit-return-type.ts
```diff
@@ -12,8 +12,10 @@ export default createRule<Options, MessageIds>({
   name: path.parse(__filename).name,
   meta: {
     type: 'problem',
+    deprecated: true,
```
timdeschryver: Is there a reason/advantage to deprecate the rules instead of removing them?

projects/www/src/app/pages/guide/component-store/write.md
```diff
@@ -46,6 +46,34 @@ export class MoviesStore extends ComponentStore<MoviesState> {
 
 </ngrx-code-example>
 
+The `updater` method enforces that callbacks return an object matching the state type exactly. Returning an object with extra properties that don't exist on the state type will produce a TypeScript compilation error:
```
timdeschryver:
```suggestion
The `updater` method enforces that callbacks return an object matching the state type exactly. Returning an object with extra properties that don't exist on the state type produces a TypeScript compilation error:
```

projects/www/src/app/pages/guide/store/reducers.md
```diff
@@ -120,6 +120,33 @@ In the example above, the reducer is handling 4 actions: `[Scoreboard Page] Home
 
 When an action is dispatched, _all registered reducers_ receive the action. Whether they handle the action is determined by the `on` functions that associate one or more actions with a given state change.
 
+### Exact return type enforcement
+
+The `on` function enforces that callbacks return an object matching the state type exactly. Returning an object with extra properties that don't exist on the state type will produce a TypeScript compilation error:
```
timdeschryver:
```suggestion
The `on` function enforces that callbacks return an object matching the state type exactly. Returning an object with extra properties that don't exist on the state type produces a TypeScript compilation error:
```

projects/www/src/app/pages/guide/component-store/write.md
```diff
@@ -46,6 +46,34 @@ export class MoviesStore extends ComponentStore<MoviesState> {
...
+    extra: true,
+  }));
+}
+```
+
+</ngrx-code-example>
+
+<ngrx-docs-alert type="inform">
+
+**Note:** When `ComponentStore` is extended with a generic state type parameter (e.g., `class MyStore<T extends object> extends ComponentStore<T>`), TypeScript cannot fully resolve the excess property check because `keyof T` is deferred. In those cases, callbacks that spread state and override known properties may produce a false type error. Return `state` directly or use a type assertion (`as T`) as a workaround.
```
timdeschryver: I think mentioning "because `keyof T` is deferred" is not very useful here for our users. Additionally, adding a test case for this scenario can be useful.

```suggestion
**Note:** When `ComponentStore` is extended with a generic state type parameter (e.g., `class MyStore<T extends object> extends ComponentStore<T>`), TypeScript cannot fully resolve the excess property check. In those cases, callbacks that spread state and override known properties may produce a false type error. Return `state` directly or use a type assertion (`as T`) as a workaround.
```

timdeschryver (commented):
Because this is a breaking change, the earliest that this can be released is v22.
It will also require a discussion first.

modules/eslint-plugin/src/rules/component-store/updater-explicit-return-type.ts
david-shortman: I suppose they actually don't have value when the return types are no longer needed. So might as well just remove.

timdeschryver:
@david-shortman can you please take a look at the conflicts/build please?

projects/www/src/app/pages/guide/component-store/write.md
```diff
@@ -46,6 +46,34 @@ export class MoviesStore extends ComponentStore<MoviesState> {
...
+    extra: true,
+  }));
+}
+```
+
+</ngrx-code-example>
+
+<ngrx-docs-alert type="inform">
+
+**Note:** When `ComponentStore` is extended with a generic state type parameter (e.g., `class MyStore<T extends object> extends ComponentStore<T>`), TypeScript cannot fully resolve the excess property check. In those cases, callbacks that spread state and override known properties may produce a false type error. Return `state` directly or use a type assertion (`as T`) as a workaround.
```
markostanimirovic, line 73: If this fix solves the updater return type but introduces the issue with user-defined stores that use  generics for state (as it is mentioned here), I don't think we can proceed with it.

projects/www/src/app/pages/guide/store/reducers.md
```diff
@@ -120,6 +120,33 @@ In the example above, the reducer is handling 4 actions: `[Scoreboard Page] Home
...
+    extra: true,
+  }))
+);
+```
+
+</ngrx-code-example>
+
+<ngrx-docs-alert type="inform">
+
+**Note:** When `on` is used inside a generic reducer factory where the state type is an unresolved generic parameter (e.g., `function createGenericReducer<TState>()`), TypeScript cannot fully resolve the excess property check. In those cases, callbacks that spread state and override known properties may produce a false type error. Return `state` directly or use a type assertion (`as TState`) as a workaround.
```
markostanimirovic, line 146: Same here: https://github.com/ngrx/platform/pull/5078/changes#r3494912154

timdeschryver:
We've decided to not include this in the v22 release.

david-shortman:
Sounds good, seems like I need more thought on this.

PR #5079, opened 2026-01-30
fix(signals): deprecate calling rxMethod/signalMethod outside injection context

### LINKED ISSUE #4991: rxMethod warning about memory leak when none is possible

signals

`rxMethod` warns about possible memory leaks in a simple scenario:

```ts
@Component({
  template: `<button (click)=example() />`
})
export class ExampleComponent {
  someMethod = rxMethod<string>(pipe(tap(console.log)));

  value = signal('arbitrary');

  example() {
    this.someMethod(this.value);
  }
}
```

When clicking the button, you'll get a warning in dev mode that starts with "@ngrx/signals/rxjs-interop: The reactive method was called outside the injection context with a signal or observable."

### Expected behavior

There should not be a warning in this case. The source injector is used as the instance injector in the case where there is not a caller injector. It seems like the warning should be implemented like:

```ts
const instanceInjector = config?.injector ?? callerInjector ?? sourceInjector;

if (
  typeof ngDevMode !== 'undefined' &&
  ngDevMode &&
  instanceInjector === undefined
) {
  console.warn(
    '@ngrx/signals/rxjs-interop: The reactive method was called outside',
    'the injection context with a signal or observable. This may lead to',
    'a memory leak. Make sure to call it within the injection context',
    '(e.g. in a constructor or field initializer) or pass an injector',
    'explicitly via the config parameter.\n\nFor more information, see:',
    'https://ngrx.io/guide/signals/rxjs-integration#reactive-methods-and-injector-hierarchies'
  );
}
```

Latest version 20.0.1

### Other information

The warning appears to be overly cautious about RxMethods defined in scopes that are larger than the components in which they are used. I think that leads to unnecessary behavior which is always providing the injector even though it's unnecessary. I'm interested to figure out an alternative to improve the warning case (like comparing to the root injector?) or suppressing it in a tactful way.

IMO, the magic of rxMethod seamlessly handling ending subscriptions based on its declaration context is muddied by this warning that encourages always explicitly declaring an injection context when not called in an explicit injection context.

- [x] Yes
- [ ] No

david-shortman:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When `rxMethod` or `signalMethod` is called with a reactive input outside an injection context without an explicit injector, a warning about a "potential memory leak" is displayed. This warning had some false positives in common usage patterns.

Closes #4991

## What is the new behavior?

The warning is replaced with a deprecation message. The goal is to eventually align with Angular's reactive primitives by requiring an injector when reactive methods are called outside an injection context.

**Changes:**
- `rxMethod` and `signalMethod`: warning updated to deprecation notice
- Docs: alert type changed from `inform` to `warn` with deprecation language
- Tests: assertions updated to match new message, added child injection context coverage

```
[ ] Yes
[x] No
```

This is a non-breaking change to a dev-mode warning message.


markostanimirovic:
Hey @david-shortman 👋 

Thanks for raising these PRs. I'm glad to see your PRs again!

I don't think we should show the warning only for root/platform injectors. In the following situation:

```ts
@Injectable()
export class ParentService {
  readonly parentMethod = rxMethod(...);
}

@Component({
  template: `
    @if (show()) { <child /> }
  `,
  providers: [ParentService],
})
export class ParentComponent {
  readonly show = signal(false);
}

@Component({ /* ... */ })
export class ChildComponent implements OnInit {
  readonly parentService = inject(ParentService);
  
  ngOnInit() {
    this.parentService.parentMethod(interval(2_000));
  }
}
```

The memory leak will happen when `ChildComponent` is destroyed, but a warning won't be displayed.

Another problem with this implementation is using the private `injector.scope` property that can potentially change in a non-major Angular version without a breaking change notice.

david-shortman:
I'll see if I can detect this situation. Additionally, I can add testing that would catch a breaking change from Angular.

Glad to be looking at NgRx again. I was curious to look at older issues I had been involved with but with the fresh lens of researching and validating ideas with Claude Code.

rainerhahnekamp:
👋 Hey all. If we’re looking to make changes here, I’d actually propose going into the opposite direction: show a deprecation message instead of a warning.

In that sense, rxMethod would align much better with the other Angular signal-based functions like effect, linkedSignal, and the resource APIs.

david-shortman:
> 👋 Hey all. If we’re looking to make changes here, I’d actually propose going into the opposite direction: show a deprecation message instead of a warning.
> 
> 
> 
> In that sense, rxMethod would align much better with the other Angular signal-based functions like effect, linkedSignal, and the resource APIs.

You're referring to requiring supplying the injection context explicitly ~~in all cases, correct~~?

Although I like the stylishness of the context being implicitly loaded, and the slightly less boilerplate that comes with it, I would support either direction.

So you're saying that, like the other Angular signal functions, unless you call in an injection context like the constructor, then you must explicitly supply the intended injection context?

rainerhahnekamp:
> So you're saying that, like the other Angular signal functions, unless you call in an injection context like the constructor, then you must explicitly supply the intended injection context?

Exactly.

My take here is that `rxMethod`'s should follow the design of Angular's signal functions. So it is more the aspect of consistency.

markostanimirovic:
Hey @david-shortman,

Do you want to update this PR with the suggested deprecation message?

david-shortman:
> Hey @david-shortman,
> 
> Do you want to update this PR with the suggested deprecation message?

Yes; knowing that's the direction that multiple maintainers are good with, I've updated the PR to deprecate calling outside injection context.

PR #5092, opened 2026-02-12
feat(docs): add dark/light theme toggle

### LINKED ISSUE #5065: Missing dark/light theme switch
### Information

There's is no theme switch on ngrx.io anymore, and the documentation defaults to dark ignoring system preferences.

### Documentation page

_No response_

- [ ] Yes
- [ ] No

adxdits:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #5065 

## What is the new behavior?
- Added a theme toggle button in the navigation menu
- Users can switch between dark and light themes
- Theme preference is persisted in localStorage
- All components now use CSS custom properties for theming
- Light theme uses One Light color palette for syntax highlighting

- Created `ThemeService` using Angular signals for reactive state management
- Defined CSS variables in `:root` (dark) and `body.light-theme` (light)
- Updated some components to use CSS variables instead of hardcoded colors

### DIFF
--- a/projects/www/src/app/components/menu.component.ts
+++ b/projects/www/src/app/components/menu.component.ts
@@ -37,10 +37,13 @@ import { ThemeToggleComponent } from './theme-toggle.component';
       <button class="close-menu" (click)="closeMenu()">
         <mat-icon class="close-menu-icon">close</mat-icon>
       </button>
-      <a routerLink="" class="logoLink" (click)="closeMenu()">
-        <img src="/ngrx-logo-pink.svg" alt="ngrx logo" />
-        NgRx
-      </a>
+      <div class="sidebar-header">
+        <a routerLink="" class="logoLink" (click)="closeMenu()">
+          <img src="/ngrx-logo-pink.svg" alt="ngrx logo" />
+          NgRx
+        </a>
+        <ngrx-theme-toggle />
+      </div>
       <ngrx-version-navigation />
       <hr />
       <!--      <a-->
@@ -77,7 +80,6 @@ import { ThemeToggleComponent } from './theme-toggle.component';
         <mat-icon>code</mat-icon>
         GitHub
       </a>
-      <ngrx-theme-toggle />
       <hr />
       <span class="guideHeader">Guide</span>
       <ngrx-guide-section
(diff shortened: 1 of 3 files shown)


projects/www/src/app/components/theme-toggle.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  Injectable,
+  PLATFORM_ID,
+  signal,
+} from '@angular/core';
+import { isPlatformBrowser } from '@angular/common';
+import { MatButtonModule } from '@angular/material/button';
+import { MatIconModule } from '@angular/material/icon';
+
+@Injectable({ providedIn: 'root' })
+export class ThemeService {
```
timdeschryver: Please move this in a separate file

projects/www/src/app/components/theme-toggle.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  private platformId = inject(PLATFORM_ID);
+  private document = inject(DOCUMENT);
+
+  isDark = signal(true);
```
timdeschryver: Instead of a boolean, I would prefer this to be a union string.

```suggestion
  theme = signal<'light'|'dark'>();
```

projects/www/src/app/components/menu.component.ts
```diff
@@ -75,6 +77,7 @@ import { VersionNavigationComponent } from './version-navigation.component';
         <mat-icon>code</mat-icon>
         GitHub
       </a>
+      <ngrx-theme-toggle />
```
timdeschryver: It's probably me, but I find the location of the toggle a little bit off.
Maybe because it's not the same style as the items above it...

Would it make sense to move it to the top?

Image

timdeschryver (commented):
I like the light theme colors, nice!

projects/www/src/app/components/menu.component.ts
adxdits: I didn’t notice it is weird at first but now that you mention it yop, that makes sense

markostanimirovic:
Thanks @adxdits! 👑

PR #5110, opened 2026-03-04
fix(component): fix "Illegal invocation" error

### LINKED ISSUE #5108: @ngrx/component: `Illegal invocation` error

component

The error was introduce in #4912 .
The issue is that `requestAnimationFrame` is extracted and called without [`window`] context/receiver [here](https://github.com/ngrx/platform/blob/6ea9dccf1dae4b425188ba341e00c648d57e9769/modules/component/src/core/tick-scheduler.ts#L33).
The function needs to either be bound to `window` or `scheduleFn` should be converted to an arrow function:
```ts
scheduleFn = (callback: FrameRequestCallback) => this.isServer ? setTimeout(callback) : requestAnimationFrame(callback);
```

### Expected behavior

There shouldn't be any `TypeError: Illegal invocation` browser errors and the CD should be scheduled correctly.

Ngrx: 21.0.1
Angular: 21.2
Node: v22.16.0
Browser: Chrome 145
OS: Windows 11

### Other information

_No response_

- [x] Yes
- [ ] No

hakimio:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #5108

## What is the new behavior?

No more `Illegal invocation` errors.

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/component/spec/core/tick-scheduler.spec.ts
+++ b/modules/component/spec/core/tick-scheduler.spec.ts
@@ -93,6 +93,19 @@ describe('TickScheduler', () => {
       tick(300 + animationFrameDelay);
       expect(appRef.tick).toHaveBeenCalledTimes(3);
     }));
+
+    it('should ensure requestAnimationFrame is not bound to the tick scheduler', () => {
+      const rafSpy = vi.spyOn(window, 'requestAnimationFrame');
+      const { tickScheduler } = setup(noopNgZoneMock);
+
+      tickScheduler.schedule();
+
+      const invocationContext = rafSpy.mock.contexts[0];
+
+      expect(invocationContext).not.toBe(tickScheduler);
+
+      rafSpy.mockRestore();
+    });
   });
 
   describe('when NgZone is not provided and running in SSR mode', () => {
--- a/modules/component/src/core/tick-scheduler.ts
+++ b/modules/component/src/core/tick-scheduler.ts
@@ -28,9 +28,13 @@ export class ZonelessTickScheduler extends TickScheduler {
   private readonly appRef = inject(ApplicationRef);
   private readonly platformId = inject(PLATFORM_ID);
   private readonly isServer = isPlatformServer(this.platformId);
-  private readonly scheduleFn = this.isServer
-    ? setTimeout
-    : requestAnimationFrame;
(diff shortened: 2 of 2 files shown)


modules/component/spec/core/tick-scheduler.spec.ts
```diff
@@ -93,6 +93,31 @@ describe('TickScheduler', () => {
...
+      const { tickScheduler } = setup(noopNgZoneMock);
+
+      tickScheduler.schedule();
+
+      const invocationContext = rafSpy.mock.contexts[0];
+
+      expect(invocationContext === window).toBe(true);
+
+      rafSpy.mockRestore();
+    });
```
markostanimirovic: Are these tests needed? We couldn't catch the illegal invocation error through tests before, so I'm wondering if they're really a guard that ensures it cannot happen anymore.

modules/component/src/core/tick-scheduler.ts
```diff
@@ -30,7 +30,7 @@ export class ZonelessTickScheduler extends TickScheduler {
   private readonly isServer = isPlatformServer(this.platformId);
   private readonly scheduleFn = this.isServer
     ? setTimeout
-    : requestAnimationFrame;
+    : requestAnimationFrame.bind(window);
```
markostanimirovic: Let's use the approach suggested in the issue if it works and avoid direct `window` usage:

```ts
private readonly scheduleFn = (callback: () => void): void => {
  if (this.isServer) {
    setTimeout(callback);
  } else {
    requestAnimationFrame(callback);
  }
}
```

markostanimirovic (changes requested):

modules/component/spec/core/tick-scheduler.spec.ts
hakimio: I would keep at least the invocation context test to make sure it's correct. I can remove the other one which tests for the error.

modules/component/src/core/tick-scheduler.ts
hakimio: Sure.

PR #5127, opened 2026-03-25
test(signals): refactor to Vitest type tests

### LINKED ISSUE #4935: Replace ts-snippet

NA

### Information

`ts-snippet` is essentially deprecated. Running the type tests, there seems to be inconsistent results.  It is tough to debug what is the problem. It may be worth looking at using [tsd](https://github.com/tsdjs/tsd) or [tsx](https://github.com/privatenumber/tsx) and converting the tests.

_No response_

- [ ] Yes
- [ ] No

santoshyadavdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

## What is the new behavior?

uses expectTypeOf for type testing, this closed #4935  partially 

```
[ ] Yes
[ ] No
```


modules/signals/spec/types/patch-state.types.spec.ts
```diff
@@ -1,78 +1,57 @@
...
-      /Property 'numbers' is missing in type '{ count: number; foo: string; }'/
-    );
-    expectSnippet('patchState(state, increment(), addNumber(10))').toFail(
-      /Property 'numbers' is missing in type '{ count: number; foo: string; }'/
-    );
-    expectSnippet('patchState(state, addNumber(10), increment())').toFail(
-      /Property 'numbers' is missing in type '{ count: number; foo: string; }'/
-    );
+    const state = signalState({ count: 1, foo: 'bar' });
+    void (() => {
```
timdeschryver: Why is this in a callback function?

santoshyadavdev: Ohh no leftover, I will replace it, I thought I already did

rainerhahnekamp (commented):
@santoshyadavdev. Looks reasonable to me at first sight. Still have to check it thoroughly, but it seems we will settle down with a combination of `@ts-expect-error` and Vitest type assertions?

santoshyadavdev:
> @santoshyadavdev. Looks reasonable to me at first sight. Still have to check it thoroughly, but it seems we will settle down with a combination of `@ts-expect-error` and Vitest type assertions?

Yes you are right, I will fix some pending code and linter soon

modules/signals/spec/types/patch-state.types.spec.ts
santoshyadavdev: @timdeschryver i think we need arrow function, earlier these calls were string so no actual call was happening, but If i dont wrap it around arrow function, ts will execute addNumber and it will always throw exception as its not an array, let me know if there is some clean way to do this

mrazauskas:
Do I get it right that `pnpm exec nx test signals` should run the updated tests? Hm.. I tried adding obviously failing line (like `const a: number = "abc"`) to `modules/signals/spec/types/signal-state.types.spec.ts`. Running the command reports all tests passing. Or did I misunderstood something?

@rainerhahnekamp Would be be still interested to check out TSTyche? (Based on https://github.com/ngrx/platform/issues/4935#issuecomment-3214017658). I could put together a similar PR. One of the advantages, there would be no need for `@ts-expect-error` directives, that cast too wide a net (in type testing context). Because *any* errors in the following line are suppressed.

modules/signals/spec/types/patch-state.types.spec.ts
timdeschryver: That makes sense, because it's actually executing the code we don't want to run it 👍 thanks for the clarification.

rainerhahnekamp (commented):
A quick update from my side. I'd like to go through each test manually first because these are our lifesavers. I don't expect any issues, but I want to be 100% safe.

modules/signals/spec/types/signal-state.types.spec.ts
```diff
@@ -1,373 +1,355 @@
...
-    expectSnippet(`const state = signalState(true);`).toFail();
+    void (() => {
+      // @ts-expect-error - Type 'number' is not assignable to type 'object'
+      signalState(10);
+      // @ts-expect-error - Type 'string' is not assignable to type 'object'
+      signalState('');
+      // @ts-expect-error - Type 'boolean' is not assignable to type 'object'
+      signalState(null);
+      // @ts-expect-error - Type 'undefined' is not assignable to type 'object'
+      signalState(true);
```
timdeschryver: An example of a error message that isn't validated.

```suggestion
      // @ts-expect-error - Type 'null' is not assignable to type 'object'
      signalState(null);
      // @ts-expect-error - Type 'boolean' is not assignable to type 'object'
      signalState(true);
```

modules/signals/spec/types/signal-store.types.spec.ts
```diff
@@ -829,82 +699,49 @@ describe('signalStore', () => {
...
-          withBar()
-        );
-
-        const Baz2Store = signalStore(
+        signalStore(withFoo(), withState({ count: 0 }), withBar());
+        signalStore(
           withState({ foo: 'foo' }),
           withState({ count: 0 }),
           withBar()
         );
```
timdeschryver: Is there a reason why the `withBaz2` and `Baz2Store` methods are left out here?
I think we can just copy the orginal version?


```suggestion
        function withBaz() {
          return signalStoreFeature(
            withFoo(),
            withState({ count: 0 }),
            withBar()
          );
        }

        function withBaz2() {
          return signalStoreFeature(
            withState({ foo: 'foo' }),
            withState({ count: 0 }),
            withBar()
          );
        }

        const BazStore = signalStore(
          withFoo(),
          withState({ count: 0 }),
          withBar()
        );

        const Baz2Store = signalStore(
          withState({ foo: 'foo' }),
          withState({ count: 0 }),
          withBar()
        );
```

modules/signals/spec/types/signal-state.types.spec.ts
```diff
@@ -1,373 +1,355 @@
...
+    const setStateValue = setState();
+
+    const mapState = signalState(new Map<number, { bar: boolean }>());
+    const mapStateValue = mapState();
+
+    const uintArrayState = signalState(new Uint8ClampedArray());
+    const uintArrayStateValue = uintArrayState();
+
+    expectTypeOf(arrayStateValue).toEqualTypeOf<string[]>();
+    expectTypeOf<string & keyof typeof arrayState>().toBeNever();
```
timdeschryver, line 119: To keep the test cases the same, the `toBeNever` expectations need to be replaced with `unique symbol | unique symbol`. Maybe that this isn't possible anymore?

modules/signals/spec/types/signal-store.types.spec.ts
```diff
@@ -948,177 +781,123 @@ describe('signalStore', () => {
...
+        type<{
+          state: { entities: Entity[]; selectedEntity: Entity | null };
+          props: { selectedEntity2: Signal<Entity | undefined> };
+          methods: { logEntity: (entity: Entity) => void };
+        }>(),
+        withMethods(() => ({
+          loadEntities(): Promise<Entity[]> {
+            return Promise.resolve([]);
+          },
+        }))
```
timdeschryver: I could be missing something, but I think we're not testing the inner types anymore here.

```suggestion
          withMethods(({ entities, selectedEntity, selectedEntity2, logEntity }) => {
            const e: Signal<Entity[]> = entities;
            const se: Signal<Entity | null> = selectedEntity;
            const se2: Signal<Entity | undefined> = selectedEntity2;
            const le: (entity: Entity) => void = logEntity;

            return {
              loadEntities(): Promise<Entity[]> {
                return Promise.resolve([]);
              },
            };
          })
```

timdeschryver (commented):
This looks great, thanks @santoshyadavdev !
I left some comments with the differences between the old and new tests.

@rainerhahnekamp fyi, I wasn't aware of it, but the `@ts-expect-error` messages are not tested. This isn't blocking for me, but on the other hand this might give false confidence in the tests. That's why I'm fine with removing the message in these test cases.

rainerhahnekamp:
@timdeschryver  

> I wasn't aware of it, but the @ts-expect-error messages are not tested. This isn't blocking for me,

I was aware. @mrazauskas pointed that out in his PR's description https://github.com/ngrx/platform/pull/5128#issue-4151749999. It isn't a blocker for me either and I am fine if we remove the messages to avoid misunderstandings.

modules/signals/spec/types/signal-store.types.spec.ts
santoshyadavdev: they were dropped during migration added them back

modules/signals/spec/types/signal-state.types.spec.ts
santoshyadavdev: yes seems like expectTypeOf limitation

timdeschryver (approved):

modules/signals/spec/types/patch-state.types.spec.ts
```diff
@@ -1,78 +1,55 @@
...
+describe('patchState', () => {
   it('infers the state type from WritableStateSource with updater', () => {
-    expectSnippet('patchState(state, increment())').toSucceed();
+    const state = signalState({ count: 1, foo: 'bar' });
+    patchState(state, increment());
   });
 
   it('infers the state type from WritableStateSource with object', () => {
-    expectSnippet("patchState(state, { foo: 'baz' })").toSucceed();
+    const state = signalState({ count: 1, foo: 'bar' });
```
rainerhahnekamp: instead of inlining the setup for each test, please create one setup function at the beginning of `describe`, like `const setup = () => signalState({ count: 1, foo: 'bar' })`, and call that one in each state.

santoshyadavdev:
@rainerhahnekamp let me know if I need to do anything on this, if you can merge this one I can work on next issue

rainerhahnekamp:
@santoshyadavdev we do have a few issues with type tests. I was researching for some time and have a proposal on how we align our tests more to Vitest's defaults. Will be update you soon.

santoshyadavdev:
> @santoshyadavdev we do have a few issues with type tests. I was researching for some time and have a proposal on how we align our tests more to Vitest's defaults. Will be update you soon.

@rainerhahnekamp sounds good, meanwhile I will check if I can pickup something, thank you

rainerhahnekamp:
Alright @santoshyadavdev, sorry for the delay, but we are now ready. The type tests will really fail if they are not type-safe. Can you please rebase on main?

PR #5128, opened 2026-03-27
test: use TSTyche for type testing

mrazauskas:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`ts-snippet` is is currently used to test types, but there are some preseasons to replace it. See #4935

## What is the new behavior?

This PR migrates `@ngrx/signals` type tests to use `tstyche`.

Related #5127

```
[ ] Yes
[x] No
```

## Other information

Based on https://github.com/ngrx/platform/issues/4935#issuecomment-3214017658, it felt like there is interest in trying out TSTyche. I am its author.

TSTyche is a standalone type test runner. It wraps around TypeScript’s programmatic APIs. This means the types are compared programmatically and additional static analysis is possible. Therefore, TSTyche is able to cover more edge cases, check error messages suppressed by `@ts-expect-error`, or test against specific TypeScript versions.

It also looks and feels like a test runner. Check out this branch and try:

- `pnpm tstyche`, runs all test files
- `pnpm tstyche uprotected`, runs a single file
- `pnpm tstyche --target 6.0`, tests against TypeScript 6.0 (and reports problems with TSConfig)
- `pnpm tstyche --target '>=5.8'`, tests against a range of TS versions

It must be noted that TypeScript 7 is not yet supported, because their programatic API is not yet ready. There are no blockers on TSTyche side.

Few notes on this migration:

- There are two possibilities to test invalid code: `.not.toBeCallableWith()` matcher or `@ts-expect-error` directive. The matcher checks if there is an error when the expression is called with given arguments; if the directive is used, TSTyche checks if the provided and suppressed error message matches. I used the matcher for simple test cases and the directive for complex ones.
- `tsconfig.json` files were added to create isolated test projects. Otherwise, fixtures and other unrelated files get type checked. That felt unnecessary.
- To be honest, I never used `nx` and I don’t know how to set it up well. Any guidance would be very helpful.


modules/signals/spec/types/signal-state.types.spec.ts
```diff
@@ -1,373 +1,294 @@
...
+    const mapState = signalState(new Map<number, { bar: boolean }>());
+    const mapStateValue = mapState();
+
+    const uintArrayState = signalState(new Uint8ClampedArray());
+    const uintArrayStateValue = uintArrayState();
+
+    expect(arrayStateValue).type.toBe<string[]>();
+    expect<
+      Extract<keyof typeof arrayState, string | number>
+    >().type.toBe<never>();
```
mrazauskas, line 106: Here and in other similar cases I adapted the approached from #5127. Although the assertion looks strange. The meaning of it is unclear. But it might be I do not understand the context well.

Originally it was tested wherever `keyof typeof arrayState` is inferred as `unique symbol | unique symbol`. Is also puzzling. Each unique symbol is only identical to itself. It cannot be identical no other type including abstract `unique symbol` type. Feels like the union should have names of two symbols. But it does not.

For instance, in the below only `STATE_SOURCE` will work. It cannot be replaced with `unique symbol`. But in the test above, the inferred type of `keyof typeof arrayState` is `unique symbol | typeof STATE_SOURCE`. 

https://github.com/ngrx/platform/blob/7138a4324d3262de12448504e042761b606f5547/modules/signals/testing/spec/unprotected.spec.ts#L22

---

And what bothers me in the `.toBe<never>()` test is that this tests what the type *does not do* instead of *what it does*.

modules/signals/spec/types/signal-state.types.spec.ts
```diff
@@ -1,373 +1,294 @@
...
-    `;
-
-    const result = expectSnippet(snippet);
-    result.toInfer('stateValue', '() => void');
-    result.toInfer('stateKeys', 'unique symbol | unique symbol');
+    const state = signalState(() => {});
+    const stateValue = state();
+
+    expect(stateValue).type.toBe<() => void>();
+    expect<Extract<keyof typeof state, string | number>>().type.toBe<never>();
```
mrazauskas, line 157: Another thought about the `unique symbol`s. What about rewriting this and similar tests like this:

```ts
    const state = signalState(() => {});
    expect(state).type.toBeAssignableTo<StateSource<() => void>>()
    expect(state).type.toBeAssignableTo<Signal<() => void>>()

    const stateValue = state();
    expect(stateValue).type.toBe<() => void>();
```

The idea is to replace `result.toInfer('stateKeys', 'unique symbol | unique symbol')` test, with assignability checks. These proof that `state` can be used where `StateSource` or `Signal` is accepted. Feels like this could be closer to a user story. And also an indirect check that `state` has the `[STATE_SOURCE]` and `[SIGNAL]` keys.

rainerhahnekamp:
Dear @mrazauskas, I'm closing this PR.
We really don’t want to be disrespectful, and it’s great that you want to contribute, but here’s the "why":

Tooling preference: We wanted to see whether we could handle type tests by relying solely on Vitest and @ts-expect-error. TSTyche comes with additional benefits, but in that case, it doesn't outweigh the cost of adding another dependency.

Contribution policy: We're pretty strict about our [policy](https://github.com/ngrx/platform/blob/main/CONTRIBUTING.md#submitting-new-features) which requires an issue and agreement before a PR is opened. This specific task (issue #4935) was already assigned to Santosh.

Again, thanks for your contribution, but this time we can't accept it.

mrazauskas:
I understand. Thanks for taking the time to explain.

PR #5148, opened 2026-04-17
test(operators): refactor to Vitest type tests

timdeschryver:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #

```
[ ] Yes
[ ] No
```


markostanimirovic (approved):

modules/operators/spec/types/tap-response.types.spec.ts
```diff
@@ -1,71 +1,40 @@
...
-    const expectSnippet = expecter(snippetFactory, {
-      ...compilerOptions(),
-      strict: false,
-    });
-
-    testWith(expectSnippet);
+  it('infers next and error type', () => {
+    of(1).pipe(
+      tapResponse({
+        next: (value) => expectTypeOf(value).toEqualTypeOf<string>(),
```
mrazauskas: Can it be that the actual type of `value` here is `number`?

In the following example `of(true)` produces `value` of type `boolean`, it feels like here `of(1)` should produce `value` of type `number` not `string`. Opening this file in code editor shows an error, but CI is green. So it can also be something is wrong with my setup.

timdeschryver: Yes, good catch.
We're also noticed that typechecking is currently visible within your editor, but not while running the tests.

timdeschryver:
```suggestion
        next: (value) => expectTypeOf(value).toEqualTypeOf<number>(),
```

rainerhahnekamp (approved):

PR #5160, opened 2026-06-05
chore: modernize testing infrastructure

rainerhahnekamp:
This PR unifies Vitest into a root workspace, adds native type-test modes (runtime / type-only / combi), and aligns per-project configs. Modules already ran Vitest via `vite.config.mts`; this centralizes that setup and enables IDE tooling.

## What this PR does

- Introduces a shared Vitest workspace via `/vitest.config.mts` and per-project `vitest.config.mts` files (**14** config files: 12 `modules/*`, `example-app`, `www`)
- Replaces per-project `vite.config.mts` with `vitest.config.mts` extending shared `baseConfig` via `mergeConfig` (except `www`, which keeps a standalone Analog config)
- Enables `pnpm vitest --project <project-name>` and Vitest IDE tooling across the monorepo (**13** projects in the root workspace — see [Projects with customized test config](#projects-with-customized-test-config))
- Configures three test modes: runtime, type-only (`.test-d.ts`), and combi (`types/*.spec.ts` with Vitest type-check)
- Streamlines `example-app`: moves `test-setup.ts` to the project root and updates `tsconfig.spec.json` (`files` path, drops `strict: false`)
- Fixes schematic and migration tests to use `dist/modules/...` paths (relies on **existing** `dependsOn: ["build"]` when running through Nx)
- Applies common test fixes: `mock_actions` → `async`/`await`, local `strip-json-comments` for eslint-plugin schematics

The main file is `/vitest.config.mts`. It exposes shared `baseConfig` for most projects and registers workspace projects under `modules/*` and `example-app` — the repo root is not a project itself. Each standard project extends `baseConfig` via `mergeConfig`; exceptions are noted below.

## Current state

Runtime tests already run through Vitest; type suites are **not migrated yet**.

- **29** spec files under `types/` — `ts-snippet` (25), `expectTypeOf` (1), custom helpers (2), runtime regression (1)
- Naming is mixed: **16** `*.types.spec.ts`, **13** plain `*.spec.ts`; no `*.test-d.ts` files yet
- `standalone-app` is still on Jest; everything else in scope uses Vitest

The configuration below is the basis for follow-up PRs that migrate those suites to native type-only or combi tests (see [Follow-up tasks](#follow-up-tasks)).

## Projects with customized test config

- **`www`**: Standalone Analog config (not `baseConfig`). Renamed `vite.config.ts` → `vitest.config.mts`; `project.json` `configFile` updated for build/serve. Explicit `resolve.alias` for `@ngrx-io` (Vitest was resolving to `/src/app/...`). **Not** part of the root Vitest workspace. Use Nx: `nx run www:test`, `nx run www:build`, `nx serve www`.
- **`effects`**: `include: ['spec/**/*.spec.ts']` and `testTimeout: 15000` (same `include` as `main` — see [Excluded specs](#excluded-specs-dead-on-main)).
- **`entity`**: `include: ['spec/**/*.spec.ts']` and `testTimeout: 15000` (same `include` as `main`; timeout raised for slow `ts-snippet` combi tests on CI).
- **`schematics`**: `include: ['src/**/*.spec.ts']` (same as `main`).
- **`eslint-plugin`**: `testTimeout: 8000`.

## Running tests

**Nx (CI and recommended):** `nx run <project>:test` or `nx affected -t test`. Module targets use existing `dependsOn: ["build"]`, so `dist/modules/...` is populated before schematic and migration specs run.

**Native Vitest (IDE / `pnpm vitest`):** The root workspace enables `pnpm vitest` and `pnpm vitest --project <name>` for modules and `example-app`, but it does **not** trigger Nx builds. Schematic and migration tests load compiled factories from `dist/modules/...` via Node `require()` — the same requirement as on `main`. Run a build first:

```bash
nx build <project>
# or all modules:
nx run-many -t build -p component,component-store,data,effects,entity,eslint-plugin,operators,router-store,schematics,signals,store,store-devtools
```

Without that step, schematic/migration specs fail with missing `dist/modules/...` modules. Runtime-only specs (e.g. under `spec/`) are unaffected.

For **`www`**, use `nx run www:test` — it is not registered in the root workspace.

### Excluded specs (dead on `main`)

Three modules keep the narrower `include` patterns from `main`. These specs exist in the repo but are **not executed** by Vitest or `nx test` — they were never part of CI on `main` either:

| Module | `include` | Excluded specs |
|--------|-----------|----------------|
| `effects` | `spec/**/*.spec.ts` | 5× `migrations/**`, `schematics/ng-add`, `testing/spec/mock_actions` |
| `entity` | `spec/**/*.spec.ts` | `migrations/6_0_0`, `schematics/ng-add` |
| `schematics` | `src/**/*.spec.ts` | `migrations/6_0_0` |

**9** migration/schematic specs in total (plus `mock_actions` on `effects`). Enabling them is a follow-up (see [Enable excluded migration and schematic tests](#enable-excluded-migration-and-schematic-tests)).

## Folder Structure

With Vitest we have three types of tests:

- **Runtime Tests:** The default. Those verify via `expect`, etc. They are **not type-safe** — Vitest strips types and runs the code. Fast, but no type safety.
- **Type Tests:** Type-only tests using `expectTypeOf` or `assertType`. Vitest only runs the compiler on them; they are not executed at runtime.
- **Combi Tests:** Tests with `@ts-expect-error` are not precise enough to count as type tests alone. Example:

```ts
it('should fail', () => {
  const state = signalState({ foo: 'bar' });

  // @ts-expect-error foo is not a number
  patchState(state, { foo: 1 });
});
```

This is a valid type test, but `@ts-expect-error` does not differentiate between a type-only error and a syntax error. For example, the test would also succeed for:

```ts
it('should fail', () => {
  const state = signalState({ foo: 'bar' });

  // @ts-expect-error foo is not a number
  patchStore(state, { foo: 1 });
});
```

Therefore, these tests need to run both as runtime and type tests.

With the new configuration, combi tests belong in a `types/` subfolder. Files matching `**/*.test-d.ts` run as type-tests only (anywhere in the project); `**/{spec,test}.ts` files under `types/` run as both runtime and type tests.

> **Sidenote:** With `typecheck.enabled: true`, Vitest prints an experimental-feature warning at startup: _"Testing types with tsc and vue-tsc is an experimental feature. Breaking changes might not follow SemVer, please pin Vitest's version when using it."_ This is expected and harmless — type-checking works as intended. Pin the Vitest version in `package.json` until the feature stabilizes.

```
project-root/
├── modules/
│   └── mypackage/
│       ├── src/
│       ├── spec/
│       │   ├── foo.spec.ts         # Runtime Test
│       │   ├── bar.test.ts         # Runtime Test
│       │   └── types/
│       │       ├── baz.test-d.ts   # Type Test (only type-checked)
│       │       └── qux.spec.ts     # Combi Test (e.g., uses @ts-expect-error)
```

## Test migration patterns

Patterns applied when enabling schematic, migration, and legacy tests under Vitest.

### Schematic and migration paths: `__dirname` → `dist/`

`SchematicTestRunner` loads factories via Node `require()`, which expects compiled `.js` files. Tests should point at the built output, not source JSON next to `.ts` files.

```ts
// before
const collectionPath = path.join(__dirname, '../migration.json');

// after
const collectionPath = path.join(
  process.cwd(),
  'dist/modules/effects/migrations/migration.json'
);
```

The same applies to schematic collections:

```ts
// before
path.join(__dirname, '../collection.json');

// after
path.join(
  process.cwd(),
  'dist/modules/effects/schematics/collection.json'
);
```

Module `test` targets rely on existing `dependsOn: ["build"]` so `dist/` exists when tests run through Nx. That does **not** apply to native `pnpm vitest` (see [Running tests](#running-tests)). Project apps (`example-app`, `www`) have no schematic/migration specs that require a pre-test build.

### Replacing deprecated `done()` callbacks

Vitest 4 deprecates Jasmine-style `done()` in favor of `async`/`await`:

```ts
// before
it('should provide Actions from source', (done) => {
  const actions$ = TestBed.inject(Actions);
  actions$.subscribe((action) => {
    expect(action.type).toBe('foo');
    done();
  });
});

// after
it('should provide Actions from source', async () => {
  const actions$ = TestBed.inject(Actions);
  const action = await firstValueFrom(actions$);
  expect(action.type).toBe('foo');
});
```

### CJS schematic dependencies under Vitest

Schematics compiled to CommonJS cannot `require()` ESM-only packages resolved from the workspace root. For `@ngrx/eslint-plugin`, `strip-json-comments` was replaced with a local helper bundled into the schematic output:

```ts
// before
import stripJsonComments from 'strip-json-comments';

// after
import { stripJsonComments } from '../strip-json-comments';
```

### Slow type-check tests: `testTimeout`

Type tests that run `tsc` via `ts-snippet` can exceed the default timeout. Projects with heavy type suites set a higher limit in `vitest.config.mts`:

```ts
// effects
test: {
  name: 'effects',
  setupFiles: ['test-setup.ts'],
  include: ['spec/**/*.spec.ts'],
  testTimeout: 15000,
},

// entity
test: {
  name: 'entity',
  setupFiles: ['test-setup.ts'],
  include: ['spec/**/*.spec.ts'],
  testTimeout: 15000,
},

// eslint-plugin
test: {
  name: 'eslint-plugin',
  setupFiles: ['test-setup.ts'],
  testTimeout: 8000,
},
```

Combi specs under `spec/types/` can exceed 8s on CI even when the type-check pass succeeds.

## Follow-up tasks

Outstanding work after this PR (verified June 2026). Planned as separate PRs on top of the configuration introduced here.

### Enable excluded migration and schematic tests

**Planned.** Widen `include` in `effects`, `entity`, and `schematics` to match other modules (`**/*.{spec,test}.ts`), then ensure those specs pass under Vitest with `dependsOn: ["build"]` (and document the manual `nx build` step for native `pnpm vitest`). The **9** currently excluded migration/schematic specs are listed in [Excluded specs](#excluded-specs-dead-on-main).

### Migrate type suites to native Vitest type tests

**Planned.** Convert the 29 existing `types/` spec files from `ts-snippet` (and related helpers) to native `.test-d.ts` type-only tests or combi tests. Start with smaller packages, then tackle heavy suites (`signals`, `store`, `effects`). Once complete, drop the custom `typecheck.include` pattern (`**/types/**/*.{spec,test}.ts`) in `vitest.config.mts` and standardize on `*.test-d.ts` naming.

### Refactor deprecated `done()` callbacks

**Planned.** `modules/effects/testing/spec/mock_actions.spec.ts` is refactored to `async`/`await` but currently **excluded** via `effects`'s `include` pattern (dead on `main`). Vitest 4 still warns on Jasmine-style `done()` in `data`, `router-store`, `store`, and `store-devtools` (roughly 70 usages across 16 spec files). These should move to Promises or `async`/`await` (see [Replacing deprecated `done()` callbacks](#replacing-deprecated-done-callbacks)).

### Fix source type errors (`ignoreSourceErrors`)

**Planned.** Shared `baseConfig` in `vitest.config.mts` sets `typecheck.ignoreSourceErrors: true` so Vitest type-checking only reports errors in test files, not in library source. Without it, **17 type errors** in source code surface during `pnpm vitest`. Fix those errors and remove `ignoreSourceErrors` from `baseConfig` so type-check runs also guard production code. This does not apply to `www`, which uses its own standalone config and sets `ignoreSourceErrors` separately.

### Performance: long-running type tests

**Partially addressed.** `with-entities.types.spec.ts` and the rest of the `signals` type suite pass with the default timeout. `effects`, `entity` (15s), and `eslint-plugin` (8s) use raised `testTimeout` values as a stopgap (see [Slow type-check tests](#slow-type-check-tests-testtimeout)). Optimization may become less urgent as suites migrate off `ts-snippet`.

### Remove legacy test framework artifacts

**Planned.** `jasmine-marbles` remains in a few `modules/` specs (`store`, `operators`, `data`), Jasmine/Jest devDependencies remain in `package.json`, and `projects/standalone-app` is still on Jest/Karma. Migrate or remove where Vitest (or `@analogjs/vitest-angular`) equivalents exist.

### Browser runtime instead of JSDOM

**Future.** Shared config sets `environment: 'jsdom'`. Moving runtime tests to a real browser environment is a larger follow-up for closer DOM fidelity.

```
[ ] Yes
[x] No
```

Native `pnpm vitest` does not auto-build; schematic/migration tests need `nx build` first. `www` is not in the root workspace — use Nx for `www` tests and builds. No breaking changes to published packages.


modules/eslint-plugin/schematics/ng-add/index.ts
```diff
@@ -1,14 +1,16 @@
 import type { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
-import stripJsonComments from 'strip-json-comments';
```
markostanimirovic, line 2: can we remove `strip-json-comments` from `package.json`?

markostanimirovic (approved):

modules/eslint-plugin/schematics/ng-add/index.ts
rainerhahnekamp: yes, it has been removed.

modules/component-store/vitest.config.mts
```diff
@@ -0,0 +1,15 @@
+import { defineProject, mergeConfig } from 'vitest/config';
+import { baseConfig } from '../../vitest.config.mts';
+
+export default defineProject((config) =>
+  mergeConfig(baseConfig, {
+    root: __dirname,
+    test: {
+      name: 'component-store',
+      setupFiles: ['test-setup.ts'],
```
brandonroberts: Any reason setupFiles and include can't be kept in the root config?

brandonroberts (approved):

timdeschryver (approved):
Looks good to me!
(Sorry for the conflicts)

We can also add to include Analog's snapshot serializers to clean up some html snapshots and to remove the dependency to jest-preset-angular.

modules/component-store/vitest.config.mts
rainerhahnekamp: No reason 😄 , I moved all `setupFiles` to the root config.

rainerhahnekamp:
@markostanimirovic, @timdeschryver, @brandonroberts if you want you can take a look but the conflicts are gone and we are ready to go from my side.

timdeschryver (approved):
Thanks @rainerhahnekamp !

PR #5167, opened 2026-06-13
feat(signals): add resource extensions

### LINKED ISSUE #5126: RFC(`@ngrx/signals`): Resource Integration Part 1 - Introduce Extend Resource APIs
## Which @ngrx/* package(s) are relevant/related to the feature request?

signals

## Information

Introduce a set of utilities for extending Angular `Resource` behavior with composable, reusable extensions.

## Motivation

The current behavior of Angular Resource APIs is opinionated and leaves limited room for customization. Two common real-world requirements that are not configurable at the resource level:

- **Value on loading** — when a resource reloads, `value()` resets to `undefined` until the new data arrives. There is no built-in way to preserve the previous value during a reload.
- **Value on error** — when a resource enters the error state, `value()` throws. There is no built-in way to return a fallback or the previous value instead.

Angular recently introduced [Resource Snapshots](https://angular.dev/guide/signals/resource#composing-resources-with-snapshots), which allow composing resources and can help with the "preserve value on loading" case. However, the error behavior remains unaddressable, and `resourceFromSnapshot` always returns the base `Resource<T>` type, meaning any additional APIs from more specific resource types (e.g. `WritableResource`) are lost in the process.

Overriding the default error behavior is also a prerequisite for introducing Resource integration for SignalStore. With the default behavior, once a resource enters the error state, any subsequent `patchState` call will throw because it reads `value()` internally — breaking the store.

Ideally, these behaviors would be configurable directly on the resource. Since they are not supported natively, this set of utilities offers what feels like the most ergonomic way to achieve this today: a thin extension layer that wraps any resource and patches only the behavior you want to change, while fully preserving the original resource type.

## Prototype

Examples and source code of resource extensions are available here: https://github.com/markostanimirovic/extend-resource-prototype

## Quick Start

```ts
import { extendResource, withPreviousValueOnLoading, withValueOnError } from 'extend-resource';

@Component({
  /* ... */
})
export class TodoList {
  readonly todosResource = extendResource({
    resource: httpResource(() => `https://dummyjson.com/todos`),
    extensions: [withPreviousValueOnLoading(), withValueOnError(undefined)],
  });
}
```

### `extendResource`

Applies extensions to a resource and returns it with its original type preserved.

```ts
const todosResource = extendResource({
  resource: httpResource(() => `https://dummyjson.com/todos`),
  extensions: [withPreviousValueOnLoading(), withValueOnError(undefined)],
});
```

### `provideResourceExtensions`

Registers extensions for a given injector scope (application, route, component). Any resource wrapped with `extendResource` inside that scope will automatically have these extensions applied.

```ts
// main.ts
bootstrapApplication(App, {
  providers: [provideResourceExtensions(withValueOnError(undefined))],
});
```

With global extensions in place, calling `extendResource` with just a resource is enough:

```ts
@Component({
  /* ... */
})
export class TodoList {
  // Global extensions are applied automatically
  readonly todosResource = extendResource(httpResource(() => `/api/todos`));
}
```

Additional per-resource extensions are appended after the global ones:

```ts
@Component({
  /* ... */
})
export class TodoList {
  // Global extensions run first, then withPreviousValueOnLoading
  readonly todosResource = extendResource({
    resource: httpResource(() => `/api/todos`),
    extensions: [withPreviousValueOnLoading()],
  });
}
```

`provideResourceExtensions` is scope-aware and composes with parent injector extensions, so you can layer defaults at the app level and add more specific behavior at the route or component level without losing the parent configuration.

### Extensions

Extensions are plain functions with the signature `(resource: R) => void`. They receive the resource and patch its behavior in place. The built-in extensions all follow the same pattern and can be freely combined.

#### `withPreviousValueOnLoading()`

Keeps the last resolved value while the resource is reloading, instead of resetting to `undefined`.

```ts
const todosResource = extendResource({
  resource: httpResource(() => `/api/todos?page=${this.page()}`),
  extensions: [withPreviousValueOnLoading()],
});

// value() returns the previous page's data while the next page is loading,
// rather than undefined.
```

#### `withValueOnLoading(value)`

Returns a specific fallback value while the resource is loading.

```ts
const todosResource = extendResource({
  resource: httpResource(() => `/api/todos`),
  extensions: [withValueOnLoading([])],
});

// value() returns [] on initial load and on every subsequent reload.
```

#### `withPreviousValueOnError()`

Returns the last successfully resolved value when the resource enters an error state, instead of throwing.

```ts
const todosResource = extendResource({
  resource: httpResource(() => `/api/todos`),
  extensions: [withPreviousValueOnError()],
});

// If the request fails, value() returns whatever was successfully loaded before,
// rather than throwing an error.
```

#### `withValueOnError(value)`

Returns a specific fallback value when the resource enters an error state, instead of throwing.

```ts
const todosResource = extendResource({
  resource: httpResource(() => `/api/todos`),
  extensions: [withValueOnError([])],
});

// If the request fails, value() returns [] instead of throwing.
```

- [x] Yes
- [ ] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #5126

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/resource/src/extensions/with-value-on-error.ts
+++ b/modules/signals/resource/src/extensions/with-value-on-error.ts
@@ -1,0 +1,54 @@
+import { Resource, untracked } from '@angular/core';
+import { ERROR_EXTENSION_TYPE } from '../consts';
+import { ResourceExtension } from '../models';
+import { overrideHasValueOnError } from './utils';
+
+/**
+ * @experimental
+ * @description
+ *
+ * Creates a resource extension that returns a specific fallback value when the
+ * resource enters the error state, instead of throwing when `value()` is read.
+ *
+ * @usageNotes
+ *
+ * ```ts
+ * import { httpResource } from '@angular/common/http';
+ * import { extendResource, withValueOnError } from '@ngrx/signals/resource';
+ *
+ * const todosResource = extendResource(
+ *   httpResource<Todo[]>(() => `/api/todos`),
+ *   withValueOnError([])
+ * );
+ *
+ * // If the request fails, `value()` returns `[]` instead of throwing.
+ * ```
+ */
+export function withValueOnError<R extends Resource<unknown>>(
+  value: R extends Resource<infer V> ? V : never
+): ResourceExtension<R> {
+  return {
+    type: ERROR_EXTENSION_TYPE,
+    apply: (resource) => {
+      Object.defineProperty(resource, 'value', {
+        value: new Proxy(resource.value, {
+          apply(target, thisArg, args) {
+            try {
+              return Reflect.apply(target, thisArg, args);
+            } catch (error) {
+              if (untracked(resource.status) === 'error') {
+                return value;
+              }
+              throw error;
+            }
+          },
+        }),
+        configurable: true,
+        enumerable: true,
+        writable: true,
+      });
+
+      overrideHasValueOnError(resource);
+    },
+  };
+}
--- a/projects/www/src/app/pages/guide/signals/resource-extensions.md
+++ b/projects/www/src/app/pages/guide/signals/resource-extensions.md
@@ -1,0 +1,196 @@
+# Resource Extensions
+
+<ngrx-docs-alert type="inform">
+
+The APIs from `@ngrx/signals/resource` are currently marked as experimental. This means its APIs are subject to change, and modifications may occur in future versions without standard breaking change announcements until it is deemed stable.
+
+</ngrx-docs-alert>
+
+Resource extensions are a set of utilities for customizing the behavior of an Angular `Resource` in a composable, reusable way. They wrap an existing resource and patch only the parts of its behavior that should change, while fully preserving the original resource type.
+
+Two common requirements are not configurable at the resource level:
+
+- **Value on loading** - when a resource reloads, `value()` resets to `undefined` until the new data arrives.
+- **Value on error** - when a resource enters the error state, reading `value()` throws.
+
+The `extendResource` function, together with the built-in extensions, addresses both cases without losing access to the APIs of more specific resource types, such as `WritableResource`.
+
+## Extending a Resource
+
+The `extendResource` function applies extensions to a resource and returns it with its original type preserved. It accepts the resource as the first argument, followed by the sequence of extensions.
+
+<ngrx-code-example>
+
+```ts
+import { Component } from '@angular/core';
+import { httpResource } from '@angular/common/http';
+import {
+  extendResource,
+  withPreviousValueOnLoading,
+  withValueOnError,
+} from '@ngrx/signals/resource';
+
+@Component({
+  /* ... */
+})
+export class TodoList {
+  readonly todosResource = extendResource(
+    httpResource<Todo[]>(() => `/api/todos`),
+    withPreviousValueOnLoading(),
+    withValueOnError(undefined)
+  );
+}
+```
+
+</ngrx-code-example>
+
+## Built-in Extensions
+
+Extensions are plain objects with a `type` symbol and an `apply` function that receives the resource and patches its behavior in place. When multiple applied extensions share the same `type`, the last one takes precedence.
+
+### `withPreviousValueOnLoading`
+
+Keeps the last resolved value while the resource is reloading, instead of resetting to `undefined`.
+
+<ngrx-code-example>
+
+```ts
+import { signal } from '@angular/core';
+import { httpResource } from '@angular/common/http';
+import {
+  extendResource,
+  withPreviousValueOnLoading,
+} from '@ngrx/signals/resource';
+
+const page = signal(1);
+const todosResource = extendResource(
+  httpResource<Todo[]>(() => `/api/todos?page=${page()}`),
+  withPreviousValueOnLoading()
+);
+
+// `value()` returns the previous page's data while the next page is loading,
+// rather than `undefined`.
+```
+
+</ngrx-code-example>
+
+### `withValueOnLoading`
+
+Returns a specific fallback value while the resource is loading.
+
+<ngrx-code-example>
+
+```ts
+import { httpResource } from '@angular/common/http';
+import {
+  extendResource,
+  withValueOnLoading,
+} from '@ngrx/signals/resource';
+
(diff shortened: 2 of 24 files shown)


projects/www/src/app/pages/guide/signals/resource-extensions.md
```diff
@@ -0,0 +1,166 @@
...
+      withPreviousValueOnLoading(),
+      withValueOnError(undefined),
+    ],
+  });
+}
+```
+
+</ngrx-code-example>
+
+## Built-in Extensions
```
michael-small, line 47: IMO each of the built in extensions could use their respective import in each block. Not all imports in a block, but definitely 1:1 with each of the 4 functions.

markostanimirovic: what do you mean? to add e.g. `import { withPreviousValueOnError } from '@ngrx/signals/resource';` to code examples?

michael-small: Yeah, that makes them quick to try out

modules/signals/resource/src/extensions/with-value-on-error.ts
```diff
@@ -0,0 +1,53 @@
...
+ * const todosResource = extendResource({
+ *   resource: httpResource(() => `/api/todos`),
+ *   extensions: [withValueOnError([])],
+ * });
+ *
+ * // If the request fails, `value()` returns `[]` instead of throwing.
+ * ```
+ */
+export function withValueOnError<R extends Resource<unknown>>(
+  value: R extends Resource<infer V> ? V : never
```
timdeschryver, line 28: I think it's by design, but does it matter that value-on-error and value-on-loading can provide a different type compared to the resource's type?
E.g.:

```ts
   const todosResource = extendResource({
        resource: httpResource<Todo[]>(() => `/api/todos`),
        extensions: [withValueOnError('Something went wrong')],
      });
  
   const todos = todosResource.values(); 
//       ?^ Todo[]
```

markostanimirovic: Good catch. It should throw a compilation error - `string` is not assignable to `Todo[]`, but currently it doesn't throw. I'll take a closer look

markostanimirovic: The fix has been pushed. I had to change the signature to enforce type inference - from config object to standard function arguments, since it does not work with a nested array within the object literal without complicated typings.

Now, the compilation error will be thrown for this code:

```ts
const todosResource = extendResource(
  httpResource<Todo[]>(() => `/api/todos`),
  withValueOnError('Something went wrong'),
);
```

To allow a union type for `value`, it needs to be specified at the resource level:

```ts
const todosResource = extendResource(
  httpResource<Todo[] | string>(() => `/api/todos`),
  withValueOnError('Something went wrong'),
);
```

so extensions cannot modify the resource type.

timdeschryver: Does is make more sense to add the types to `extendResource`?
The reason I'm asking is because we're currently changing the type of the resource, but in this case, the http resource will never return a string value.  

```
const todosResource = extendResource<Todo[] | string>(
  httpResource(() => `/api/todos`),
  withValueOnError('Something went wrong'),
);
```

projects/www/src/app/pages/guide/signals/resource-extensions.md
```diff
@@ -0,0 +1,190 @@
+# Resource Extensions
+
+Resource extensions are a set of utilities for customizing the behavior of an Angular `Resource` in a composable, reusable way. They wrap an existing resource and patch only the parts of its behavior that should change, while fully preserving the original resource type.
+
+Two common requirements are not configurable at the resource level:
+
+- **Value on loading** - when a resource reloads, `value()` resets to `undefined` until the new data arrives.
+- **Value on error** - when a resource enters the error state, reading `value()` throws.
+
+The `extendResource` function, together with the built-in extensions, addresses both cases without losing access to the APIs of more specific resource types, such as `WritableResource`.
```
rainerhahnekamp, line 16: I would directly address here the open question, namely the difference to the native snapshots. To me the additional benefits are the ability of having 
1. A `WritableSignal`
2. The option to define them globally.

So I would rephrase it like 
```suggestion
Unlike the native `resourceFromSnapshots` function, which downcasts the output to a base `Resource`, `extendResource` can be configured globally and fully preserves the capabilities of writable types like `WritableResource`.
```

modules/signals/resource/src/extensions/with-value-on-error.ts
markostanimirovic: I'm not sure if something like this would be possible because it's important to infer the exact resource type. If the generic does not represent the resource type, but the resource value, as follows:

```ts
declare function extendResource<Value>(
  resource: Resource<Value>,
  ...extensions: ResourceExtension<Value>[]
): Resource<Value>;
```

then the original resource type won't be preserved, and we want to avoid that - when a writable resource is provided to `extendResource` for example, we don't want to 'cut' the writable part from the resource type, but return the full type instead.

This should be a rare case anyway. Even `resource` API does not allow `value` inference on multiple places, but only from the `loader` callback result:

```ts
export const demoResource = resource({
  loader: () => Promise.resolve(1),
  defaultValue: 'string', // compilation error: string is not assignable to number
});
```

By the way, with the current implementation, we can do something similar by providing the entire resource type:

```ts
const todosResource = extendResource<ResourceRef<Todo[] | string>>(
  httpResource(() => `/api/todos`),
  withValueOnError('Something went wrong'),
);
```

timdeschryver (approved):

rainerhahnekamp (approved):

projects/www/src/app/pages/guide/signals/resource-extensions.md
markostanimirovic: I'd avoid mentioning `resourceFromSnapshot` in case this limitation gets solved on the Angular side in the meantime.

PR #5184, opened 2026-07-01
fix(signals): `patchState` does not block on error

### LINKED ISSUE #5183: `@ngrx/signals`: `patchState` blocks when State signal throws an error

signals

Signals can throw an error when they are accessed. This can happen for state
slices that are backed by APIs like `resource` or `linkedSignal`.

`patchState` currently calls `getState` internally before applying an update, so
it eagerly reads every state slice. As a result, patching an unrelated state
slice can lead to a dead-end, in that sense that other state slices are not
patchable anymore. 

## Reproduction

```ts
const Store = signalStore(
  { providedIn: 'root', protectedState: false },
  withState({ name: '' }),
  withLinkedState(() => {
    const { value } = resource({
      loader: async () => {
        throw new Error('Failed to load value.');
      },
    });

    return { value };
  })
);

const store = inject(Store);

patchState(store, { name: 'John' });
```

### Expected behavior

It should not block

NgRx 21

### Other information

_No response_

- [x] Yes
- [ ] No

rainerhahnekamp:
Signals can throw an error if they are accessed. `patchState` needs to call `getState` in order to provide the current state to an updater function.

This fix catches errors in the internal `getState` call and keeps the errored signals in a proxy, where accessing them throws again.

That means `patchState` does not block unrelated updates, and the user can apply the normal error handling features like `try/catch`, `ErrorHandler`, or the upcoming error boundaries.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #5183

```
[ ] Yes
[x] No
```

### DIFF
--- a/modules/signals/src/state-source.ts
+++ b/modules/signals/src/state-source.ts
@@ -82,24 +82,24 @@ export function patchState<State extends object>(
     Partial<NoInfer<State>> | PartialStateUpdater<NoInfer<State>>
   >
 ): void {
-  const currentState = untracked(() => getState(stateSource));
-  const newState = updaters.reduce(
-    (nextState: State, updater) => ({
-      ...nextState,
-      ...(typeof updater === 'function' ? updater(nextState) : updater),
-    }),
-    currentState
-  );
-
   const signals = stateSource[STATE_SOURCE];
   const stateKeys = Reflect.ownKeys(stateSource[STATE_SOURCE]);
+  const draftState = untracked(() => getSafeState(stateSource));
+  const touchedKeys = new Set<keyof State>();
 
-  for (const key of Reflect.ownKeys(newState)) {
-    if (stateKeys.includes(key)) {
-      const signalKey = key as keyof State;
-      if (currentState[signalKey] !== newState[signalKey]) {
-        signals[signalKey].set(newState[signalKey]);
-      }
+  for (const updater of updaters) {
+    const partial =
+      typeof updater === 'function' ? updater(draftState) : updater;
+
+    for (const key of Reflect.ownKeys(partial) as Array<keyof State>) {
+      touchedKeys.add(key);
+      draftState[key] = partial[key] as State[keyof State];
+    }
+  }
+
+  for (const key of touchedKeys) {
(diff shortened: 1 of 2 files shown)


modules/signals/src/state-source.ts
```diff
@@ -82,24 +82,24 @@ export function patchState<State extends object>(
...
+  const touchedKeys = new Set<keyof State>();
+
+  for (const updater of updaters) {
+    const partial =
+      typeof updater === 'function' ? updater(draftState) : updater;
 
-  for (const key of Reflect.ownKeys(newState)) {
-    if (stateKeys.includes(key)) {
-      const signalKey = key as keyof State;
-      if (currentState[signalKey] !== newState[signalKey]) {
```
timdeschryver, line 100: Is there a reason why the equality check is removed? This will now also invoke the setter of unchanged state keys.

modules/signals/src/state-source.ts
```diff
@@ -82,24 +82,24 @@ export function patchState<State extends object>(
...
-      }
+    for (const key of Reflect.ownKeys(partial) as Array<keyof State>) {
+      touchedKeys.add(key);
+      draftState[key] = partial[key] as State[keyof State];
+    }
+  }
+
+  for (const key of touchedKeys) {
+    if (stateKeys.includes(key as string | symbol)) {
+      signals[key].set(draftState[key]);
```
timdeschryver, line 102: An AI review flagged a potential regression.
Is this intended, or should we add this to the breaking changes?

```ts
it('applies linked state updates after their source updates', () => {
  const numberStore = signalStoreFeature(
    withState({ id: 1 }),
    withLinkedState(({ id }) => ({
      level: () => id() * 2,
    }))
  )(getInitialInnerStore());

  patchState(numberStore, { level: 5, id: 2 });

  // This succeeded before, but now resolves in  { id: 2, level: 4 }
  expect(getState(numberStore)).toEqual({ id: 2, level: 5 });
});
```

It has to do with the order in which we set the values. Doing `patchState(numberStore, { id: 2, level: 5 });` is fine.

PR #5186, opened 2026-07-01
feat(signals): add `SignalStoreFeatureType`

### LINKED ISSUE #5185: `@ngrx/signals`: type representing a `SignalStoreFeature`

signals

### Information

Custom Signal Store features often expose state, props, and methods that are useful as input for other custom features. Today, that input type has to be duplicated manually with `type<{ state; props; methods }>()`, even though the producing feature already contains the canonical inferred output type.

Add a public `SignalStoreFeatureType` helper that extracts the output result type from a Signal Store feature factory.

Typing it manually.

- [x] Yes
- [ ] No

rainerhahnekamp:
This allows custom features to reuse another feature's inferred state, props, and methods as input without duplicating the feature result type manually.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #5185 

```
[ ] Yes
[x] No
```


modules/signals/src/signal-store-models.ts
```diff
@@ -46,3 +46,36 @@ export type SignalStoreFeature<
...
+ *     }))
+ *   );
+ * }
+ * ```
+ */
+export type SignalStoreFeatureType<
+  Feature extends (...params: never[]) => unknown,
+> =
+  ReturnType<Feature> extends SignalStoreFeature<infer _, infer Output>
+    ? Output
```
gabrielguerrero: @rainerhahnekamp  I did something similar to this in ngrx-traits some time ago ( I was gonna do a PR to here, nice to see you were doing the same), anyway there is a small improvement I did to mine 

```ts
export type ExtractStoreFeatureOutput<
  Result extends (...args: any[]) => SignalStoreFeature<any, any>,
> =
  ReturnType<Result> extends SignalStoreFeature<infer In, infer Out>
    ? In & Out
    : never;
```
notice I did In & Out, because if you want to extract the type of a store feature that requires an Input is it should be part of the output because is a dependency of the feature, otherwise I got some problems in cases where I use this to split the store

rainerhahnekamp: Thanks for the info. Updated ours with your suggestion 👍

timdeschryver (approved):

martin-the-geek:
We have something similar in our project except we have a second generic parameter where we can specify a subset of keys to depend on which helps limit the setup in our feature tests.

From memory it went something like this
```
type StoreFeatureResultKeys<T extends StoreFeatureResult> = keyof T['state'] | keyof T['props'] | keyof T['methods'];

type ExtractedStoreFeature<Feature extends (...params: never[]) => unknown> =
  ReturnType<Feature> extends SignalStoreFeature<infer Input, infer Output>
    ? Input & Output
    : never;

export type SignalStoreFeatureType<
  Feature extends (...params: never[]) => unknown,
  Keys extends StoreFeatureResultKeys<ExtractedStoreFeature<Feature>> = StoreFeatureResultKeys<ExtractedStoreFeature<Feature>>
> = {
  state: Pick<ExtractedStoreFeature<Feature>>['state'], Extract<Keys, keyof ExtractedStoreFeature<Feature>>['state']>>,
  props: Pick<ExtractedStoreFeature<Feature>>['props'], Extract<Keys, keyof ExtractedStoreFeature<Feature>>['props']>>,
  methods: Pick<ExtractedStoreFeature<Feature>>['methods'], Extract<Keys, keyof ExtractedStoreFeature<Feature>>['methods']>>,
};

// to depend on an entire feature
type<SignalStoreFeatureType<withSomeFeature>>()

// to depend on a subset of the feature
type<SignalStoreFeatureType<withSomeFeature, 'stateA' | 'methodB'>>()
```

rainerhahnekamp:
@martin-the-geek, but why would you want that?

I see `SignalStoreFeatureType` is a type for a specific feature. If you want a more generic one that could be implemented by multiple `signalStoreFeature` types, then I understand the motivation behind your approach. But even then, I would write a generic type manually, meaning I would not like to depend on an existing one.

I just don't know if we are overcomplicating things here.

martin-the-geek:
@rainerhahnekamp Our motivation for having a convenience type to identify a subset of a feature is to narrow the dependencies of another feature to limit the impacts to our feature tests.

If feature A depends on only a few signals from feature B but feature B has dozens, when testing feature A we only have to add the few bits to our test store that feature A depends on. Also, if in the future feature B adds some elements that are irrelevant to feature A, we don't have to update feature A's tests to include the new elements. Problem gets worse if feature B depends on feature C, etc.

Sure, I guess we just keep the key limiting type custom to our app, but it might be useful for others.

modules/signals/spec/types/signal-store-feature-type.spec.ts
markostanimirovic, line 1: file name: `signal-store-feature-type.test-d.ts`

modules/signals/src/index.ts
```diff
@@ -9,6 +9,7 @@ export {
   SignalStoreFeature,
   SignalStoreFeatureResult,
   StateSignals,
+  type SignalStoreFeatureType,
 } from './signal-store-models';
```
markostanimirovic, line 13: exports are sorted. for consistency:


```suggestion
export {
  EmptyFeatureResult,
  SignalStoreFeature,
  SignalStoreFeatureType,
  SignalStoreFeatureResult,
  StateSignals,
} from './signal-store-models';
```

projects/www/src/app/pages/guide/signals/signal-store/custom-store-features.md
```diff
@@ -184,6 +184,8 @@ This enables the utilization of input properties within the custom feature, even
 
 The expected input type should be defined as the first argument of the `signalStoreFeature` function, using the `type` helper function from the `@ngrx/signals` package.
 
+`SignalStoreFeatureType` helper can extract the complete output type of a custom feature factory and reuse it as the input type of another custom feature.
```
markostanimirovic: I'd recommend to merge this and `### Example 5: Reusing Feature Output as Input` to a separate section on this page: ## Using `SignalStoreFeatureType`, so everything related to the new API goes there.

modules/signals/src/signal-store-models.ts
```diff
@@ -46,3 +46,36 @@ export type SignalStoreFeature<
...
+ *   );
+ * }
+ * ```
+ */
+export type SignalStoreFeatureType<
+  Feature extends (...params: never[]) => unknown,
+> =
+  ReturnType<Feature> extends SignalStoreFeature<infer Input, infer Output>
+    ? Input & Output
+    : never;
```
markostanimirovic, line 82: When a factory returns one
of them bare - no `signalStoreFeature(...)` wrapper - there's nothing to infer `Input`
from, so it falls back to its constraint, and `Input & Output` drags that in:

```ts
function withLogger() {
  return withMethods(() => ({ log(): void {} }));
}

type LoggerFeature = SignalStoreFeatureType<typeof withLogger>;
// state:   object & {}
// methods: MethodsDictionary & { log: () => void }   // Record<string, Function>
```

The index signature leaks into consumers, so anything type-checks:

```ts
signalStoreFeature(
  type<LoggerFeature>(),
  withMethods((store) => {
    store.whateverTypo(); // ✅ compiles today, typed as Function
    return {};
  })
);
```

`withState` is immune - its `Input` is the concrete `EmptyFeatureResult`, not a type
parameter.

## Change

```diff
 export type SignalStoreFeatureType<
   Feature extends (...params: never[]) => unknown,
 > =
   ReturnType<Feature> extends SignalStoreFeature<infer Input, infer Output>
-    ? Input & Output
+    ? SignalStoreFeatureResult extends Input
+      ? Output
+      : Input & Output
     : never;
```

`SignalStoreFeatureResult extends Input` is true only for the constraint fallback (and
vacuous inputs like `EmptyFeatureResult`). A real requirement such as
`{ state: { count: number } }` isn't satisfied by `{ state: object }`, so it keeps the
`Input & Output` branch.

## Test to add

```ts
it('ignores unresolved input from bare feature factories', () => {
  function withLogger() {
    return withMethods(() => ({ log(): void {} }));
  }

  signalStoreFeature(
    type<SignalStoreFeatureType<typeof withLogger>>(),
    withMethods((store) => {
      expectTypeOf(store.log).toEqualTypeOf<() => void>();
      // @ts-expect-error no Function index signature
      store.anythingAtAll;

      return {};
    })
  );
});
```

modules/signals/src/signal-store-models.ts
```diff
@@ -46,3 +46,36 @@ export type SignalStoreFeature<
 > = (
   store: InnerSignalStore<Input['state'], Input['props'], Input['methods']>
 ) => InnerSignalStore<Output['state'], Output['props'], Output['methods']>;
+
+/**
+ * @description
+ *
+ * Extracts the output type of a SignalStore feature factory.
```
markostanimirovic: Documentation says it extracts the output (in `.md` as well). However, it combines `Input & Output`. This needs to be rephrased without sharing unnecessary implementation details.

projects/www/src/app/pages/guide/signals/signal-store/custom-store-features.md
```diff
@@ -307,6 +309,60 @@ export function withBaz<Foo extends string | number>() {
...
+    withComputed(() => ({
+      foo: computed(() => 10),
+    })),
+    withMethods(() => ({
+      bar(foo: number): void {
+        console.log(foo);
+      },
+    }))
+  );
+}
```
markostanimirovic: Can we provide some more realistic example? Documentation will look better.

For example:

```ts
export type RequestStatusFeature = SignalStoreFeatureType<
  typeof withRequestStatus
>;

export function withStatusMessage() {
  return signalStoreFeature(
    type<RequestStatusFeature>(),
    withComputed(({ isPending, error }) => ({
      statusMessage: () => {
        if (isPending()) {
          return 'Loading...';
        }

        return error() ?? 'Ready';
      },
    }))
  );
}
```

modules/signals/spec/types/signal-store-feature-type.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+      type<CounterFeature>(),
+      withMethods((store) => {
+        expectTypeOf(store).toMatchObjectType<{
+          count: Signal<number>;
+          increment: () => void;
+        }>();
+
+        return {};
+      })
+    );
```
markostanimirovic, line 36: Tests are missing assertions for SignalStore usage when the source feature and dependent feature are combined. Types are only tested via `signalStoreFeature`.

markostanimirovic (changes requested):

modules/signals/spec/types/signal-store-feature-type.spec.ts
rainerhahnekamp: done

modules/signals/src/signal-store-models.ts
rainerhahnekamp: Weird stuff; I've never seen a bare feature factory, but that doesn't mean it isn't possible. Good catch 👍

I've added the suggested fix and updated the test. Since tests with `@ts-expect-error` should also be executed, I've made it a full runtime test.

projects/www/src/app/pages/guide/signals/signal-store/custom-store-features.md
rainerhahnekamp: Done

projects/www/src/app/pages/guide/signals/signal-store/custom-store-features.md
rainerhahnekamp: Done

modules/signals/src/signal-store-models.ts
rainerhahnekamp: Done

modules/signals/spec/types/signal-store-feature-type.spec.ts
rainerhahnekamp: done

rainerhahnekamp:
@markostanimirovic, could please re-check?

markostanimirovic (approved):
👏

PR #5173, opened 2026-06-17
test: enable typecheck for all spec files, not only types/* suites

rainerhahnekamp:
Enable type-tests everywhere

## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngrx/platform/blob/main/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[x] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Closes #5172

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```

PR #5187, opened 2026-07-03
feat(signals): enable creation of dynamic deep signals

### LINKED ISSUE #4847: Allow reading a nested property from a DeepSignal when there's a nullable object prior to the property

signals

### Information

If you create a state containing the following:

```typescript
object: { value: number };
nullableObject: null | { value: number };
undefinableObject?: { value: number };
```

You can then do:

```typescript
store.object.value()
```

It works just fine as a `DeepSignal`.

However, both the nullable and potentially undefined ones don't work.

I'm not entirely sure what would be the best option here, either:

```typescript
store.nullableObject?.value()
```

But I'm guessing that it may break the reactivity as it may not be caught by the signal graph

OR

```typescript
store.nullableObject.value()       // returns a type `number | null`
store.undefinableObject.value() // returns a type `number | undefined`
```

Reading the top level signal, which isn't ideal ofc as I need another `computed` down the chain to get reactivity only on one property.

- [ ] Yes
- [x] No

markostanimirovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #4847

Fixes signals tests in #5173

## What is the new behavior?

Enables creation of dynamic deep signals in `signalState`, `signalStore`, and `deepComputed`. Example:

```ts
type Book = { id: number; title: string };
type Status =
  | { type: 'success'; data: string }
  | { type: 'error'; message: string };

const BookStore = signalStore(
  withState<{ book: Book | null; status: Status }>({
    book: null,
    status: { type: 'success', data: '' },
  })
);
const store = inject(BookStore);

// 👇 object literal + null: store.book is DeepSignal<Book> | Signal<null>
if ('title' in store.book) {
  const title = store.book.title; // Signal<string>
  console.log(title());
}

// 👇 union of object literals: a DeepSignal is created for each member
// store.status: DeepSignal<{ type: 'success'; data: string }> | DeepSignal<{ type: 'error'; message: string }>
if ('message' in store.status) {
  const message = store.status.message; // Signal<string>
  console.log(message());
}
```

```
[x] Yes
[ ] No
```

```
BREAKING CHANGES:

Union state slices and computed results that include an object literal now create a `DeepSignal` for each object literal member, instead of exposing the whole union as a single `Signal`.

BEFORE:

A union that included an object literal was exposed as a single `Signal` of the whole union.

signalState:

const state = signalState<{ user: { name: string } | null }>({
  user: null
});
// state.user: Signal<{ name: string } | null>

signalStore:

const Store = signalStore(
  withState<{ user: { name: string } | null }>({ user: null })
);
const store = inject(Store);
// store.user: Signal<{ name: string } | null>

deepComputed:

const source = signal<{ a: number } | { b: number }>({ a: 1 });
const result = deepComputed(() => source());
// result: Signal<{ a: number } | { b: number }>

Custom SignalStore feature with generics:

function withMyFeature<Entity extends { id: number }>() {
  return signalStoreFeature(
    type<{ state: { entity: Entity | null } }>(),
    withMethods(({ entity }) => {
      // the type of entity is Signal<Entity | null>
      const e: Signal<Entity | null> = entity;

      return {
        // ...
      };
    })
  );
}

AFTER:

Each object literal member becomes its own `DeepSignal`; the remaining members stay a regular `Signal`.

signalState:

const state = signalState<{ user: { name: string } | null }>({
  user: null
});
// state.user: DeepSignal<{ name: string }> | Signal<null>

signalStore:

const Store = signalStore(
  withState<{ user: { name: string } | null }>({ user: null })
);
const store = inject(Store);
// store.user: DeepSignal<{ name: string }> | Signal<null>

deepComputed:

const source = signal<{ a: number } | { b: number }>({ a: 1 });
const result = deepComputed(() => source());
// result: DeepSignal<{ a: number }> | DeepSignal<{ b: number }>

Custom SignalStore feature with generics:

function withMyFeature<Entity extends { id: number }>() {
  return signalStoreFeature(
    type<{ state: { entity: Entity | null } }>(),
    withMethods(({ entity }) => {
      // the type of entity is DeepSignalOf<Entity | null>
      const e: DeepSignalOf<Entity | null> = entity;

      return {
        // ...
      };
    })
  );
}
```

### DIFF
--- a/modules/signals/spec/types/signal-store.types.spec.ts
+++ b/modules/signals/spec/types/signal-store.types.spec.ts
@@ -6,6 +6,7 @@ describe('signalStore', () => {
     (code) => `
         import { computed, inject, Signal } from '@angular/core';
         import {
+          DeepSignalOf,
           getState,
           patchState,
           signalStore,
@@ -1073,8 +1074,10 @@ describe('signalStore', () => {
               logEntity: (entity: Entity) => void;
             };
           }>(),
-          withMethods(({ entities, selectedEntity2, logEntity }) => {
+          withMethods(({ entities, selectedEntity, selectedEntity2, logEntity }) => {
             const e: Signal<Entity[]> = entities;
+            const se: DeepSignalOf<Entity | null> = selectedEntity;
+            const seId: string | null = 'id' in se ? se.id() : null;
             const se2: Signal<Entity | undefined> = selectedEntity2;
             const le: (entity: Entity) => void = logEntity;
 
--- a/modules/signals/spec/types/signal-store.types.spec.ts
+++ b/modules/signals/spec/types/signal-store.types.spec.ts
@@ -240,14 +240,82 @@ describe('signalStore', () => {
     const result = expectSnippet(snippet);
     result.toInfer(
       'store',
-      '{ foo: Signal<number | { s: string; }>; bar: DeepSignal<{ baz: { b: boolean; } | null; }>; x: DeepSignal<{ y: { z: number | undefined; }; }>; } & StateSource<{ foo: number | { ...; }; bar: { ...; }; x: { ...; }; }>'
+      '{ foo: DeepSignal<{ s: string; }> | Signal<number>; bar: DeepSignal<{ baz: { b: boolean; } | null; }>; x: DeepSignal<{ y: { z: number | undefined; }; }>; } & StateSource<...>'
     );
-    result.toInfer('foo', 'Signal<number | { s: string; }>');
+    result.toInfer('foo', 'DeepSignal<{ s: string; }> | Signal<number>');
     result.toInfer('bar', 'DeepSignal<{ baz: { b: boolean; } | null; }>');
-    result.toInfer('baz', 'Signal<{ b: boolean; } | null>');
+    result.toInfer('baz', 'DeepSignal<{ b: boolean; }> | Signal<null>');
     result.toInfer('x', 'DeepSignal<{ y: { z: number | undefined; }; }>');
     result.toInfer('y', 'DeepSignal<{ z: number | undefined; }>');
     result.toInfer('z', 'Signal<number | undefined>');
+  });
+
+  it('does not split non-record union members into separate signals', () => {
+    const snippet = `
+      type FooBar = 'foo' | 'bar';
+      type State = {
+        flag: boolean;
+        status: FooBar;
+        mixed: string | number;
+        withRecord: { id: number } | boolean;
+        combo: { a: string } | FooBar | null;
+        multi: { a: number } | { b: string } | boolean;
+        withCollection: { id: number } | Set<number> | boolean;
+        nested: { foo: boolean | { deep: string } };
+      };
+
+      const Store = signalStore(
+        withState<State>({
+          flag: true,
+          status: 'foo',
+          mixed: 1,
+          withRecord: { id: 1 },
+          combo: null,
+          multi: true,
+          withCollection: { id: 1 },
+          nested: { foo: true },
+        })
+      );
+      const store = inject(Store);
+      const flag = store.flag;
+      const status = store.status;
+      const mixed = store.mixed;
+      const withRecord = store.withRecord;
+      const combo = store.combo;
+      const multi = store.multi;
+      const withCollection = store.withCollection;
+      const nested = store.nested;
+      const nestedFoo = store.nested.foo;
+    `;
+
+    const result = expectSnippet(snippet);
+    result.toInfer('flag', 'Signal<boolean>');
+    result.toInfer('status', 'Signal<FooBar>');
+    result.toInfer('mixed', 'Signal<string | number>');
+    result.toInfer(
+      'withRecord',
+      'Signal<boolean> | DeepSignal<{ id: number; }>'
+    );
+    result.toInfer(
+      'combo',
+      'DeepSignal<{ a: string; }> | Signal<FooBar | null>'
+    );
+    result.toInfer(
+      'multi',
+      'Signal<boolean> | DeepSignal<{ a: number; }> | DeepSignal<{ b: string; }>'
+    );
+    result.toInfer(
+      'withCollection',
+      'DeepSignal<{ id: number; }> | Signal<boolean | Set<number>>'
+    );
+    result.toInfer(
+      'nested',
+      'DeepSignal<{ foo: boolean | { deep: string; }; }>'
+    );
+    result.toInfer(
+      'nestedFoo',
+      'Signal<boolean> | DeepSignal<{ deep: string; }>'
+    );
   });
 
   it('succeeds when root state slices contain Function properties', () => {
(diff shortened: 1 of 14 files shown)


rainerhahnekamp (changes requested):
As always, great stuff! I just wanted to bring up two points for discussion:

### 1. Typing Issues with Deep Signals

This isn't necessarily a new issue, but it became much clearer to me during this review. The following snippet shows how our deep signal feature can lead to some tricky type safety gaps:

```ts
it('gives a wrong type guarantee', () => {
  type Status =
    | { status: 'error'; message: string }
    | { status: 'success'; value: number };
  
  const state = signal({
    status: 'error',
    message: 'did not work',
  } as Status);

  const deepState = toDeepSignal(state);
  
  if ('message' in deepState) {
    const message = deepState.message;
    state.set({ status: 'success', value: 1 });

    // returns undefined although it is typed as Signal<string> 😬
    message().toUpperCase(); 
  }
});
```

While we could argue that developers shouldn't access a `message` signal without a proper runtime check, I still see this as a problem from a strict TypeScript perspective. 

Strictly speaking, `message` should be typed as `Signal<string | undefined>`. However, I wouldn't want to make that change, as I'm really hoping the scenario above is just an edge case. Instead, I suggest we simply throw a runtime error.

---

### 2. Vitest Type Tests vs. `ts-snippet`

Regarding our testing setup, why don't we use Vitest's built-in type testing capabilities instead of `ts-snippet`? I found that they worked quite nicely for in PR #5186.

markostanimirovic:
@rainerhahnekamp 

> ### 1. Typing Issues with Deep Signals
> This isn't necessarily a new issue, but it became much clearer to me during this review. The following snippet shows how our deep signal feature can lead to some tricky type safety gaps:
> 
> ```ts
> it('gives a wrong type guarantee', () => {
>   type Status =
>     | { status: 'error'; message: string }
>     | { status: 'success'; value: number };
>   
>   const state = signal({
>     status: 'error',
>     message: 'did not work',
>   } as Status);
> 
>   const deepState = toDeepSignal(state);
>   
>   if ('message' in deepState) {
>     const message = deepState.message;
>     state.set({ status: 'success', value: 1 });
> 
>     // returns undefined although it is typed as Signal<string> 😬
>     message().toUpperCase(); 
>   }
> });
> ```
> 
> While we could argue that developers shouldn't access a `message` signal without a proper runtime check, I still see this as a problem from a strict TypeScript perspective.
> 
> Strictly speaking, `message` should be typed as `Signal<string | undefined>`. However, I wouldn't want to make that change, as I'm really hoping the scenario above is just an edge case. Instead, I suggest we simply throw a runtime error.

The same issue can be reproduced with plain TypeScript, and this is not something we can control:

```ts
    type Status = { status: 'success' } | { status: 'error'; message: string; }
    const state: Status = { status: 'error', message: 'x' };
    if (state.status === 'error') {
      Object.assign(state, { status: 'success', message: undefined });
      state.message.toUpperCase(); // typed string, undefined at runtime. Compiles clean.
    }
```

Anyway, I don't see this as any kind of blocker - in properly implemented logic, write and read blocks should be separated, and type narrowing of deep signals is needed only for the read logic. If someone wants to abuse our APIs, there is always a way (`as any` as a last resort 😅).

> ### 2. Vitest Type Tests vs. `ts-snippet`
> Regarding our testing setup, why don't we use Vitest's built-in type testing capabilities instead of `ts-snippet`? I found that they worked quite nicely for in PR #5186.

For consistency reasons, I don't think we should mix both within a single project (`@ngrx/signals` in this case still relies on `ts-snippet`). I haven't created new `spec` files, only extended the existing ones. So the reason is simple - it's out of scope for this PR. Refactoring of `@ngrx/signals` type tests from `ts-snippet` tests to Vitest should be tackled separately.

projects/www/src/app/pages/guide/signals/signal-store/index.md
```diff
@@ -49,6 +49,40 @@ The `BookSearchStore` instance will contain the following properties:
...
+  const title = store.book.title; // Signal<string>
+  console.log(title());
+}
+
+// 👇 union of object literals: a DeepSignal is created for each member
+// store.status: DeepSignal<{ type: 'success'; data: string }> | DeepSignal<{ type: 'error'; message: string }>
+if ('message' in store.status) {
+  const message = store.status.message; // Signal<string>
+  console.log(message());
+}
```
timdeschryver, line 79: Instead of having this sample duplicates, could we create a section/page for it and refer to it? This will make it easier to update later, if needed.

rainerhahnekamp (approved):
As discussed, we see my first comment as a theoretical concern and wait if those things actually happen in real life.
As for the typing tests, they would be handled in a separate issue.

modules/signals/spec/types/signal-store.types.spec.ts
```diff
@@ -999,9 +1073,8 @@ describe('signalStore', () => {
               logEntity: (entity: Entity) => void;
             };
           }>(),
-          withMethods(({ entities, selectedEntity, selectedEntity2, logEntity }) => {
```
timdeschryver, line 1002: Is this change intended?
Codex gives the following review:

modules/signals/src/deep-signal.ts:17-30, modules/signals/src/deep-computed.ts:29, modules/signals/src/signal-store-models.ts:9: DeepSignalOf<T> loses constrained generic identity and narrows to the constraint. This regresses generic consumers because the returned signal is no longer assignable to Signal<Entity>.

```ts
function select<Entity extends { id: string }>(value: Entity) {
  const sig = deepComputed(() => value);
  const entity: Entity = sig(); // TS error: returns { id: string }
}
```

The changed generic store type spec also masks this by removing selectedEntity from the destructuring at modules/signals/spec/types/signal-store.types.spec.ts:1076.

And changes the DeepSignal type to the following:

```ts
type DeepSignalMembers<T> =
  IsKnownRecord<T> extends true
    ? Readonly<{ [K in keyof T]: DeepSignalOf<T[K]> }>
    : unknown;

export type DeepSignal<T> = Signal<T> & DeepSignalMembers<T>;

export type DeepSignalOf<T> =
  HasKnownRecordMember<T> extends true
    ? DeepSignalRecordMembers<T, T> | DeepSignalNonRecordMembers<T>
    : Signal<T>;

// Preserve the original call signature when a constrained generic is narrowed
// to its constraint in the distributive branch.
type DeepSignalRecordMembers<T, Original> = T extends unknown
  ? IsKnownRecord<T> extends true
    ? T extends Original
      ? DeepSignal<T>
      : Signal<Original> & DeepSignalMembers<T>
    : never
  : never;
```

timdeschryver (commented):
LGTM, I also asked a review from codex, which gave 1 comment.

modules/signals/spec/types/signal-store.types.spec.ts
markostanimirovic: Good catch! I committed back `selectedEntity` with the updated type. I also added an additional section about custom features with generics to `BREAKING CHANGES`.

projects/www/src/app/pages/guide/signals/signal-store/index.md
markostanimirovic: Yea, it's almost a duplicate, but not fully - one example uses `signalStore`, another `signalState`. Although `signalStore` and `signalState` share the same core and have common features, I see that as an implementation detail users don't have to be aware of.

If you don't mind, I'd keep this note on both pages, because I think it's important for both APIs.

timdeschryver (approved):

Issue #5163, opened 2026-06-08
Prepare for TypeScript 7


timdeschryver:
Update the tsconfig files to fix deprecations. These are currently ignored with the  `ignoreDeprecations` option.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#preparing-for-typescript-70

robert-md-or:
planning to tackle this as: (1) fix the ES5 target + deprecated moduleResolution: 'node' first, since those are contained and low-risk, (2) then work through baseUrl/rootDir across the rest, testing carefully since those affect path resolution more broadly. Will open a PR once (1) is solid, and continue from there.

Update after digging further:
I checked all 43 tsconfig files against the 5 options TypeScript 6.0 marks as deprecated.
1. target: "ES5" — only one spot: modules/store/spec/ngc/tsconfig.ngc.json. However, checking project.json for the store package, this file isn't referenced by any Nx target (build/test/lint) — it looks like dead code, likely left over from testing the old ngc/Ivy AOT compiler. I partially fixed it, but doing so surfaced several unrelated errors (type issues, outdated lib setting), suggesting it hasn't been checked in a while. Should this file just be removed instead of fixed?
2. moduleResolution: "node" — 6 files total:
1 file (projects/www/.../__base/tsconfig.json) was simple — module was already modern (ES2022), so I switched to "bundler" and confirmed it compiles cleanly with tsc --noEmit. Done.
The remaining 5 use module: "commonjs", which is incompatible with "bundler". Per TypeScript's docs, the correct migration for Node-executed code (schematics, doc generation) is "nodenext" for both module and moduleResolution together. Haven't touched these yet — want to test each individually.
One of them (modules/eslint-plugin/spec/fixtures/tsconfig.json) is a test fixture that seems to intentionally simulate an older CommonJS-style project. Modernizing it might defeat the purpose of that test. Should it be left as-is?
3. downlevelIteration — appears in nearly all 37 package-level tsconfig files. This looks like the bulk of the actual scope here, much bigger than the target/moduleResolution piece. Haven't started on this yet.
4. baseUrl/rootDir — present everywhere; removing baseUrl can break rootDir inference (saw this directly on the ngc file above). Will need per-file attention.
Proposed order: (1) finish moduleResolution/target (small, nearly done), (2) tackle downlevelIteration (the bulk), (3) baseUrl/rootDir last (highest risk).

PR #5200, opened 2026-07-23
fix(store): remove deprecated downlevelIteration from tsconfig files (partial TS7 prep, see #5163)

robert-md-or:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Several `tsconfig.json` files across the workspace set the `downlevelIteration`
compiler option, which TypeScript has marked for removal as part of the
`target`/`moduleResolution` deprecation cleanup tracked in #5163. Left as-is,
these configs will start failing once the workspace upgrades to a TypeScript
version where the option is removed.

## What is the new behavior?

Removes `downlevelIteration` from the 32 `tsconfig.json` files where it no
longer serves a purpose:

- **10 files** already had an explicit modern `target` (`ES2015` or newer)
  set somewhere in their `extends` chain, so the flag was already redundant.
  These simply had the flag removed.
- **22 files** had no explicit `target` anywhere in their chain and relied
  only on a modern `lib` entry (e.g. `lib: ["es2018", "dom"]`). Testing
  confirmed that `lib` alone does **not** change the compiler's actual
  target — TypeScript still compiles to its ES3/ES5 default without an
  explicit `target`, which surfaced as a real compile error
  (`TS2802: ...can only be iterated through when using the
  '--downlevelIteration' flag or with a '--target' of 'es2015' or higher`)
  during local verification. These 22 files now get an explicit `target`
  added (matching their existing `lib` level), alongside the flag removal,
  so behavior no longer depends on an assumption that doesn't hold.

All affected packages (`store`, `component-store`, `effects`, `entity`,
`router-store`, `store-devtools`, `schematics`, and their respective
`schematics-core`/`build` targets) were built locally via `nx run
<project>:build` after the change, with no compilation errors.

This PR addresses the `downlevelIteration` portion of the broader scope
outlined in #5163. Remaining categories from that issue
(`moduleResolution: "node"`, the one remaining `target: "ES5"` config, and
`baseUrl`/`rootDir`) are being tracked separately and are not part of this PR.

```
[ ] Yes
[x] No
```

## Other information

Related to #5163 (partial — see scope note above; not marked as closing
since further work from that issue is still in progress).


robert-md-or:
Closing this PR — it was accidentally based on the `master` branch, which is outdated (last updated in 2023) and not this repository's default branch (`main`). That mismatch caused unrelated conflicts across dozens of files that have nothing to do with this change.

Reopened cleanly against `main` here: #5201

Same change (removing the deprecated `downlevelIteration` option from affected tsconfig files), verified with local builds across all affected packages on the correct base branch.

Issue #5218, opened 2026-08-30
@ngrx/signals: updateEntity deletes an entity when its numeric ID changes to the equivalent string


adxdits:
### Which @ngrx/* package(s) are the source of the bug?

signals

### Minimal reproduction of the bug/regression with instructions

`models.ts:3` supports `string | number` but `helpers.ts:199-205` writes the new key before deleting the old one. JavaScript treat 1 and "1" as the same object key,so the entity is deleted.

```
type Item = { id: string | number; name: string };

const ItemStore = signalStore(
  { protectedState: false },
  withEntities<Item>()
);
const store = new ItemStore();

patchState(store, addEntity({ id: 1, name: 'Ada' }));
patchState(
  store,
  updateEntity({ id: 1, changes: { id: '1' } })
);

console.log(store.ids());      // ['1']
console.log(store.entityMap()); // {}
console.log(store.entities());  // [undefined]
```

### Expected behavior

The entity should remain available under ID `"1"`:

{ '1': { id: '1', name: 'Ada' } }

### Versions of NgRx, Angular, Node, affected browser(s) and operating system(s)

NgRx: 22.0.0 / current main
Angular: 22.1.3
Browsers: all

### Other information

`updateEntitiesMutably` assigns `entityMap[newId]` and then deletes `entityMap[id]`. Numeric `1` and string `"1"` reference the same object property, so the assignment is immediately deleted.

### I would be willing to submit a PR to fix this issue

- [x] Yes
- [ ] No

PR #5219, opened 2026-08-31
fix(signals): preserve entity when id changes number to equivalent string

adxdits:
## PR Checklist

#5218 

* [x] The commit message follows our guidelines: [https://github.com/ngrx/platform/blob/main/CONTRIBUTING.md#commit](https://github.com/ngrx/platform/blob/main/CONTRIBUTING.md#commit)
* [x] Tests for the changes have been added (for bug fixes / features)
* [ ] Documentation has been added / updated (for bug fixes / features)

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Code style update (formatting, local variables)
* [ ] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation content changes
* [ ] Other... Please describe:

In `updateEntitiesMutably` (used by `updateEntity`, `updateEntities`, and `updateAllEntities`), when an entity's ID is changed to a value that is a different type but the same object key (e.g. numeric `1` to string `'1'`), the entity is deleted instead of updated. This happens because the new key is written before the old key is deleted, and JavaScript treats `1` and `'1'` as the same object property, so the delete step removes the just-written entity.

Reproduction:

```ts
const Store = signalStore({ protectedState: false }, withEntities<Entity>());
const store = new Store();

patchState(store, addEntity({ id: 1, name: 'Ada' }));
patchState(store, updateEntity({ id: 1, changes: { id: '1' } }));

store.entityMap(); // {} entity is lost
store.entities();  // [undefined]
```

## What is the new behavior?

The old entity key is deleted before the updated entity is assigned to the new key, so entities are preserved when the ID's string/number representation changes. `entityMap`, `ids`, and `entities` all stay consistent.

A regression test was added in `update-entity.spec.ts` covering the numeric-to-string ID update case.

No

### DIFF
--- a/modules/signals/entities/spec/updaters/update-entity.spec.ts
+++ b/modules/signals/entities/spec/updaters/update-entity.spec.ts
@@ -227,6 +227,26 @@ describe('updateEntity', () => {
     expect(store.ids()).toEqual([101, 2, 303]);
   });
 
+  it('updates a numeric entity id to the equivalent string id', () => {
+    type Entity = { id: string | number; name: string };
+    const Store = signalStore(
+      { protectedState: false },
+      withEntities<Entity>()
+    );
+    const store = new Store();
+
+    patchState(
+      store,
+      addEntities<Entity>([{ id: 1, name: 'Ada' }]),
+      updateEntity<Entity>({ id: 1, changes: { id: '1' } })
+    );
+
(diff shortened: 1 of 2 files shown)


markostanimirovic (approved):

timdeschryver (approved):
Thanks @adxdits

